Issue #28, opened 2019-10-17
Netaddr is missing in requirement file.


titom73:
After running `pip install -r requirements.txt` in a clean environment, we have following error message:

```
fatal: [DC1-SPINE1 -> localhost]: FAILED! => changed=false 
  msg: 'AnsibleFilterError: The ipaddr filter requires python''s netaddr be installed on the ansible controller'
```

This library is missing from [`requirements.txt`](https://github.com/aristanetworks/ansible-avd/blob/a5ada4d80322be6861f346baebd1497ee750b882/requirements.txt)

__Fix is:__

```shell
$ pip install netaddr
Collecting netaddr
  Using cached https://files.pythonhosted.org/packages/ba/97/ce14451a9fd7bdb5a397abf99b24a1a6bb7a1a440b019bebd2e9a0dbec74/netaddr-0.7.19-py2.py3-none-any.whl
Installing collected packages: netaddr
Successfully installed netaddr-0.7.19
```

PR #30, opened 2019-10-17
Fix for Issue #28 - Add missing netaddr requirement

titom73:
Add missing requirement for `netaddr==0.7.19`


carlbuchmann:
looks good!

Issue #25, opened 2019-10-17
Add Docker file


carlbuchmann:
add docker file to enable consistent use and deployment of Ansible Arista Validated Design Collection.

Provide docker file for:

- Ansible 2.8
- Ansible 2.9(devel)

kgrozis-arista:
Hi, Carl,

I’ll work on this tomorrow and plan on having a pull request for review.

Thanks
Kevin

carlbuchmann:
Cool man!

I did a big push tonight and tackled a few of the major issues, check out
the PRs!

Thanks

<http://www.arista.com/>
Software Driven Cloud Networking
<https://www.linkedin.com/company/arista-networks-inc>
<https://www.facebook.com/AristaNW/> <https://twitter.com/aristanetworks>
<https://www.youtube.com/user/AristaNetworks>

*Carl Buchmann | Systems Engineer*

carl.buchmann@arista.com

m: +1.613.222.5936
*www.arista.com* <http://www.arista.com/>

kgrozis-arista:
> add docker file to enable consistent use and deployment of Ansible Arista Validated Design Collection.
> 
> Provide docker file for:
> 
> * Ansible 2.8
> * Ansible 2.9(devel)

@carlbuchmann Single Dockerfile provided.  I made ansible version a variable that can be passed in with docker build.  By default the Dockerfile sets ansible version to 2.8.5.

Do you want me to write a docker-compose file to spin both a container with 2.8 and 2.9 at the same time?

PR #31, opened 2019-10-17
ADDED DOCKERFILE FOR CONTAINERIZATION

kgrozis-arista:
Changes to be committed:
- added Dockerfile for containerization instructions
- added Makefile to automate container build (make) and clean up (make clean)
- modified README.md with docker instructions
- added requirements-dev.txt with development pip modules
- added requirements.txt with ansible-avd pip modules


carlbuchmann:
@kgrozis-arista can we add for docker image to load /etc/hosts for name resolution to transfer from localhosts, in case DNS is not used.

requirements.txt
```diff
@@ -1,4 +1,3 @@
-ansible==2.8.3
```
titom73: Shouldn't let ansible requirement for `ansible` part of this file for every use not running Docker ?
 
In the docker file, running a task like `pip install --upgrade ansible=xxx` to override ansible version installed by `requirements.txt`

If we removed `ansible` from the file, other requirements came with ansible and their version might change with based on ansible version installed on user's system:

- `cffi`
- `cryptography`
- `enum34`
- `ipaddress`
- `Jinja2`
- `MarkupSafe`
- `pycparser`
- `PyYAML`
- `six`

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    sudo
+
+# Create the /project directory and add it as a mountpoint
+WORKDIR /ansible_avd
+VOLUME ["/ansible_avd"]
+
+# Install python modules required by the repo
+RUN pip install --upgrade pip
+ADD requirements.txt /tmp/requirements.txt
+RUN pip install --trusted-host pypi.python.org -r /tmp/requirements.txt
```
titom73: Seems to not be required since it is included in `requirements-dev.txt`

requirements.txt
kgrozis-arista: @titom73 I agree.  I wasn't thinking of the non-docker scenarios.  Not adding ansible to the requirements.txt file will definitely break these instances.

Dockerfile
kgrozis-arista: @titom73 I agree.  I'll do away with this since it adds an extra unneeded step to the dockerfile.

kgrozis-arista:
@carlbuchmann What do you have in mind for an /etc/host file?  Do you want to explicitly state IP addresses for the hosts?  This will assume it's deployed the same everywhere.  

It might be better to create a playbook to do this based on mgmt_ips in the inventory.yml file.  This way it could be used in non-docker scenarios too.

carlbuchmann:
resolves #25

Issue #19, opened 2019-10-11
Refactor BGP Templates


carlbuchmann:
Refactor to split out sub-section in BGP templates in the eos-l3ls-evpn role to allow easier extenstion.

PR #40, opened 2019-10-29
breakout bgp configuration in sub-templates

carlbuchmann:
breakout bgp templates in sub-templates, to allow better readability and allow to stack future templates


carlbuchmann:
resolve #19

PR #56, opened 2019-12-16
Feature convert to collection

carlbuchmann:
- convert to collection
- rename filter and roles to match convention
- create initial documentation


carlbuchmann:
resolved #21

PR #104, opened 2020-03-29
Fix python 3 setup for continuous testing

titom73:
Enable Py3 configuration prior to any custom job:

```yaml
- name: Setup Python 3 on runner
   uses: actions/setup-python@v1.2.0
   with: 
     python-version: '3.x'
```

Some `Makefile` targets have been update as well:

- `setup-repository` --> `install-requirements`
- `install-requirements-generic`
- `install-requirements-dev`

### DIFF
--- a/.github/workflows/ansible_test.yml
+++ b/.github/workflows/ansible_test.yml
@@ -9,10 +9,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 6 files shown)


titom73:
Need to check either we want to run linting on push and/or pull request.

- In a fork model, linting on push happens on forked repository
- In a branch model, linting on push happens within current repository

carlbuchmann (approved):
LGTM

PR #146, opened 2020-06-02
Releases/v1.1.x

ankudinov:
The initial version of upgrade role to translate 1.0 data to 1.1.
It is likely not covering all possible combinations yet. To fix that in future we need better test data That is consistent and covers entire data model.
Current version is based on netdevops-examples data.


carlbuchmann:
replaced by PR #250

PR #150, opened 2020-06-03
Fix incorrect configlet lookup path

titom73:
Fix an incorrect path feading inventory_to_container looking for wrong
place for configlets upload.

It only impact if inventories and playbooks are not in same folder.

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/main.yml
@@ -5,9 +5,9 @@
   inventory_to_container:
     inventory: '{{ inventory_file }}'
     container_root: '{{ container_root }}'
-    configlet_dir: '{{playbook_dir}}/intended/configs'
+    configlet_dir: '{{inventory_dir}}/intended/configs'
     configlet_prefix: '{{ configlets_prefix }}'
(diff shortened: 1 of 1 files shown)


ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/main.yml
```diff
@@ -5,9 +5,9 @@
   inventory_to_container:
     inventory: '{{ inventory_file }}'
     container_root: '{{ container_root }}'
-    configlet_dir: '{{playbook_dir}}/intended/configs'
+    configlet_dir: '{{inventory_dir}}/intended/configs'
     configlet_prefix: '{{ configlets_prefix }}'
-    destination: '{{inventory_dir}}/intended/structured_configs/cvp/{{inventory_hostname}}.yml'
+    destination: '{{inventory_dir}}/intended/structured_configs/cvp/{{inventory_hostname}}_2.yml'
```
carlbuchmann: @titom73 is adding "_2" intended?

carlbuchmann (approved):
LGTM

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/main.yml
carlbuchmann: from @titom73 "because we do a register, this file is not required. But it is usefull for debug.
Previous version overwritte it with next task. So I use _2 to keep a trace
I can rename it to _configlets"

Issue #180, opened 2020-07-01
JSON syntax not supported with Tenants vlans.


titom73:
## Issue Type

- Bug Report

## Summary

When using JSON as language for group_vars, template engine fails to render vars with `EOS_L3LS_EVPN`.

## Role or Module Name

Role: `eos_l3ls_evpn`

## OS / Environment

__EOS Version__

N/A

__Cloudvision version__

N/A

__OS running Ansible__

Docker from avdteam/base:3.6

## Steps to reproduce

[Convert](https://www.convertjson.com/yaml-to-json.htm) following YAML into JSON and save output under `group_vars/TENANT_NETWORKS.json`

```yaml
---
# Documentation of Tenant specific information - Vlans/VRFs
tenants:
  Tenant_test:
    l2vlans:
      111:
        name: v_test_111
        tags:
        - compute
      112:
        name: v_test_112
        tags:
        - compute
      210:
        name: v_test_210
        tags:
        - compute
    mac_vrf_vni_base: 10000
    name: Tenant_test
    vrfs:
      Tenant_test_vrf_a:
        name: Tenant_test_vrf_a
        svis:
          110:
            enabled: true
            id: 110
            ip_subnet: 10.1.110.0/24
            name: Tenant_test_vrf_a_110
            tags:
            - compute
            tenant: Tenant_test
            vrfs: Tenant_test_vrf_a
          120:
            enabled: 1
            id: 120
            ip_subnet: 10.144.100.0/24
            name: Tenant_test_vrf_a_120
            tags:
            - compute
            tenant: Tenant_test
            vrfs: Tenant_test_vrf_a
        vrf_vni: 10
        vtep_diagnostic:
          loopback: 100
          loopback_ip_range: 10.255.1.0/24
```

Remove YAML version from your group_vars

```shell
$ mv group_vars/DC1_TENANTS_NETWORKS.yml group_vars/DC1_TENANTS_NETWORKS.orig-yaml
```

### Expected results

Playbook execution should work with no error

### Actual results

Error raised at: `TASK: eos_l3ls_evpn : Generate device configuration in structured format (yaml)`

```shell
[...]

    ### Routing - OSPF ###
    router_ospf:
    {# underlay #}
    {% include 'underlay-overlay/router-ospf.j2' %}

    {# queue monitor streaming #}
    ### Queue Monitor Streaming ###
    queue_monitor_streaming:
    {% include 'base/queue-monitor-streaming.j2' %}

[...]

Issue #140, opened 2020-05-27
Add Jinja2 filters to convert type when required


ankudinov:
**Issue Type**
- Template enhancement

**Is your feature request related to a problem? Please describe.**
In some case AVD YAML files may have `<key>: "<int-value>"` instead of `<key>:<int-value>`
That can cause template parsing problems.

**Describe the solution you'd like**
Add filters to Jinja2 templates where possible.
For example, change `as: {{ spine.bgp_as }}` to `as: {{ spine.bgp_as | int }}`

carlbuchmann:
resolved by #181

PR #181, opened 2020-07-02
Implement full JSON support for group_vars

titom73:
## Description

Update AVD to support JSON as input structure as it is fully supported by Ansible.

Fix issue #180

## Changes

- Update VLANS and SVI templating engine to force
- Add a testing scenario to validate futur changes

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
+++ b/ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
@@ -103,8 +103,8 @@
 {%                 for svi_tag in tenants[tenant].vrfs[vrf].svis[svi].tags %}
 {%                     if svi_tag in leaf.filter_tags or svi_tag == leaf.group or "all" in leaf.filter_tags %}
 {%                         do leaf.vrfs.append( vrf ) %}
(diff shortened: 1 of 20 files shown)


carlbuchmann:
@titom73 & @ankudinov - I agree with the proposed changes, this is also related to issue #140 raised by @ankudinov , is there other changes required?

titom73:
PR only provides a fix by protecting group_vars value and force type for everything coming from JSON data: Keys from dictionary where `int` is expected.

I think #140 would require more work as we have to go through all templates to cast all variables to `int` when required.

ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
```diff
@@ -377,4 +377,4 @@ queue_monitor_streaming:
 {# management api http #}
 ### Management API http ###
 management_api_http:
-{% include 'base/management-api-http.j2' %}
\ No newline at end of file
+{% include 'base/management-api-http.j2' %} #}
```
carlbuchmann: @titom73 please remove "#}"

carlbuchmann (changes requested):

carlbuchmann (approved):
LGTM and tested with AVD Example

Issue #185, opened 2020-07-10
development environment build issues


ryanmerolle:
## Issue Type

- Bug Report

## Summary

Development getting started breaks

## OS / Environment

__EOS Version__

__Cloudvision version__

__OS running Ansible__

## Steps to reproduce

One liner:

```bash
rmerolle@Merolle-Desktop:~/repos/ansible-avd/development$ docker-compose up
Pulling ansible (avdteam/base:3.6)...
3.6: Pulling from avdteam/base
8559a31e96f4: Pull complete
62e60f3ef11e: Pull complete
f024dcd4fd42: Pull complete
c186acc4b86a: Pull complete
09557bd0cf27: Pull complete
d685428e547b: Pull complete
c8a9cedf5199: Pull complete
7dac0cfe1278: Pull complete
262f72c96539: Pull complete
f6f8cf90b282: Pull complete
51f7f8c643a7: Pull complete
Digest: sha256:26ebfc90a545747fcd93774f8b97be9725d2665557c3a7d0db3f538d44230185
Status: Downloaded newer image for avdteam/base:3.6
Pulling webdoc (titom73/mkdocs:latest)...
latest: Pulling from titom73/mkdocs
df20fa9351a1: Pull complete
36b3adc4ff6f: Pull complete
7031d6d6c7f1: Pull complete
81b7f5a7444b: Pull complete
0f8a54c5d7c7: Pull complete
306723926237: Pull complete
80815963f735: Pull complete
11559c1a081d: Pull complete
Digest: sha256:95dda13344968f9b85eae0cb4368ae5e5840e56a1a48c0fd5a67a5b912482f9e
Status: Downloaded newer image for titom73/mkdocs:latest
Creating development_webdoc_1  ... done
Creating development_ansible_1 ... done
Attaching to development_ansible_1, development_webdoc_1
webdoc_1   | ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'ansible-avd/ansible_collections/arista/avd/docs/requirements.txt'
development_webdoc_1 exited with code 1
^CGracefully stopping... (press Ctrl+C again to force)
Stopping development_ansible_1 ... done
```

Manually using docker-compose is expecting a requirements file in a location it does not exist.
```bash
~/repos/ansible-avd/development$ docker-compose up
Pulling ansible (avdteam/base:3.6)...
3.6: Pulling from avdteam/base
8559a31e96f4: Pull complete
62e60f3ef11e: Pull complete
[...]

PR #186, opened 2020-07-10
Fix issue where installation script failed to clone AVD

titom73:
Fix an issue (#185) where a variable was not set correctly leading to an error to clone AVD and then retrieve make file.

Also add a strep in CI to test installation file.


carlbuchmann (approved):
LGTM

PR #189, opened 2020-07-16
Change name of example repository

titom73:
Name of example repository referred to `netdevops-examples` which is confusing as it clones `ansible-avd-cloudvision-demo` and not `aristanetworks/netdevops-examples`


carlbuchmann (approved):
LGTM

PR #194, opened 2020-07-20
add next hop interfaces to static route templates

### LINKED ISSUE #192: Add Next Hop Interface for Static Routes (IPv4 & IPv6) to eos_cli_config_gen
**Issue Type**
- Template enhancement

EOS static route parameters include vrf, dest_net, next_hop (can be a combo of interface or just one of either), name, distance, and tag.

All of these options can bet set in the current template, but the data model documentation needs to be updated or a seperate key needs to be created for interface until this model.

Update the template for vlans would need to be updated as long as the following data model:
```
static_routes:
  - vrf: < vrf_name, if vrf_name = default the route will be placed in the GRT >
    destination_address_prefix: < IPv4_network/Mask >
    interface: < interface >
    gateway: < IPv4_address >
    distance: < 1-255 >
    tag: < 0-4294967295 >
    name: < description >
  - destination_address_prefix: < IPv4_network/Mask >
    gateway: < IPv4_address >
```

Document gateway not as an IPv4_address but as     `gateway: < interface &| IPv4_address >`
I think this approach would be more confusing and allow for more user error.  Also, if data validation is ever applied to these data models, it would be harder to validate a string combining 2 values.

**Additional context**
I prefer the proposed approach, but I am open to reasons to use the other.  I can see from some of the BGP default commands, a list of strings was leveraged in order to cover a host of configs without needing to model them all.  I do not think there is an advantage to do so with this model.

ryanmerolle:
closes #192

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -855,6 +855,7 @@ standard_access_lists:
 static_routes:
   - vrf: < vrf_name, if vrf_name = default the route will be placed in the GRT >
     destination_address_prefix: < IPv4_network/Mask >
+    interface: < interface >
     gateway: < IPv4_address >
     distance: < 1-255 >
     tag: < 0-4294967295 >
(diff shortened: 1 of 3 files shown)


carlbuchmann (changes requested):
when testing with following data model:

```yaml
static_routes:
  - vrf: MGMT
    destination_address_prefix: 0.0.0.0/0
    gateway: 192.168.200.1
```

result:
`ip route vrf MGMT 0.0.0.0/0`

expected result:
`ip route vrf MGMT 0.0.0.0/0 192.168.200.1`

ryanmerolle:
Corrected data model oversight due to incorrect conditional logic.

carlbuchmann (approved):
LGTM

Issue #213, opened 2020-08-07
Do not output empty variables for vlan_interfaces and port_channel_interfaces 


titom73:
## Issue Type

- Bug Report

## Summary

When we generate structured_config (YAML) for a device with no SVI like l2leaf not running in MLAG, the output contains an empty variable which has no type and leads to some issues when read by ansible:

```yaml
### Management Interfaces ###
management_interfaces:
  Management1:
    description: oob_management
    vrf: MGMT
    ip_address: 10.73.1.17/16
    gateway: 10.73.0.1

### VLAN Interfaces ###
vlan_interfaces:

### TCAM Profiles ###
tcam_profile:

```

## Role or Module Name

__arista.avd.eos_l3ls_evpn__

## Steps to reproduce

Any config where device is `type:l2leaf` and not configured in mlag

### Expected results

```yaml
```yaml
### Management Interfaces ###
management_interfaces:
  Management1:
    description: oob_management
    vrf: MGMT
    ip_address: 10.73.1.17/16
    gateway: 10.73.0.1

### TCAM Profiles ###
tcam_profile:
```

carlbuchmann:
Closing this issue, as a suitable method to handle variables that are defined but are empty with the default filter.

If you want to use the default value when variables evaluate to false or an empty string you have to set the second parameter to true -> example: {{ my_variable | default({}, true) }}

https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#defaulting-undefined-variables

PR #214, opened 2020-08-07
Fix empty vlan_interfaces and port_channel_interfaces

titom73:
Provides fixes to not generate empty variables for both port_channel_interfaces and vlan_interfaces.

## Fix

Issue: #213

## Description

- Add condition to only generate vlan_interfaces on l2leaf when they are running in MLAG since AVD does not generate SVI for tenant vlans
- Add condition to only generate port_channel_interfaces on l3leaf when they are running in MLAG mode. We assume l2leaf are only using PO to l3leafs in MLAG and do not leverage PO on a single l3leaf

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
+++ b/ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
@@ -92,6 +92,26 @@
 {%         endif %}
 {%     endfor %}
 {% endfor %}
+
+{% set leaf.has_port__channel=false %}
+{# If l3leaf hostname is in l2leaf default uplink list and 2 neighbors are defined #}
+{# We consider there is a PO to configure on L3LEAF                               #}
+{% for l3leaf in l2leaf.defaults.parent_l3leafs | arista.avd.natural_sort %}
+{%    if l3leaf == inventory_hostname and l2leaf.defaults.parent_l3leafs | length > 1%}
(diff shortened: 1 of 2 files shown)


ansible_collections/arista/avd/roles/eos_l3ls_evpn/templates/evpn-fabric-l3leaf-yml.j2
```diff
@@ -253,6 +253,7 @@ bfd_multihop:
 {% include 'base/bfd-multihop.j2' %}
 
 {# Port-Channel Interfaces #}
+{% if leaf.mlag == true %}
```
carlbuchmann: There are could some cases where `leaf.mlag == false` and there are still port-channels defined. For Example a port-channel from a server to a single-homed leaf. We should test for all use cases.

titom73: Yes. In this case, need to add an additional complex test to define if PO is expected or not.

titom73: Should be resolved with commit 2c610b9

carlbuchmann:
Closing this PR, as a suitable method to handle variables that are defined but are empty with the `default` filter.

If you want to use the default value when variables evaluate to false or an empty string you have to set the second parameter to `true` -> example: ` {{ my_variable | default({}, true) }}`

https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#defaulting-undefined-variables

PR #226, opened 2020-09-04
Critical Fix - ESI malformed by filter plugin

titom73:
Original filter plugin build ESI with one more octet than RFC.
Current commit fix behavior


carlbuchmann (approved):
LGTM

Issue #204, opened 2020-07-21
Incorrect ip-routing template conditional in eos_cli_config_gen


ryanmerolle:
## Issue Type
- EOS Configuration error
- Bug Report

## Summary

ip-routing template adds a dependency for `ip routing: true` in order to set other vrf ip routing attributes.  I tested and saw that even if a device has `no ip routing` (ie for the default vrf), you can still set other vrf routing.

I propose the template should be fixed to 

1. Add no to the begining of the command `ip routing` for the default vrf if `ip routing: false` to follow the syntax in EOS
2. Remove the VRF conditional for ip routing from the base ip routing conditional so that the default vrf being enabled for ip routing has nothing to do with any vrf's ip routing.

## Role or Module Name

eos_cli_config_gen 

## OS / Environment

__EOS Version__

4.22.6M

__Cloudvision version__

2020.1.1

__OS running Ansible__

2.9.11

## Steps to reproduce

```yaml
ip_routing: false
vrfs:
  MGMT:
    description: OOB MGMT
    ip_routing: false
  overlay:
    description: Overlay VRF
    ip_routing: true
```

### Expected results

```shell
no ip routing
no ip routing vrf MGMT
ip routing vrf overlay
```

### Actual results

```shell
no ip routing
```

### Additional context
I think it would be rare that this edge case be hit, but I forgot to set ip_routing var, and I uncovered this dependency.

aphillipps:
Thanks for the information. This will be fixed in release v1.1.0

Issue #199, opened 2020-07-21
Default VRF needs to be added to name-server config in eos_cli_config_gen 


ryanmerolle:
## Issue Type
- EOS Configuration error

## Summary

Currently, if a name-server host is set in eos_cli_config_gen, and the vrf associated is default, the config generated removes the vrf name from the generated config.  At least with EOS 4.22.6M this is not the standard behavior.  It will take the command, but EOS will automatically convert to list the vrf in the command.

## Role or Module Name

eos_cli_config_gen 

## OS / Environment

__EOS Version__

4.22.6M

__Cloudvision version__

2020.1.1

__OS running Ansible__

2.9.11

## Steps to reproduce

```yaml
name_server:
  source:
    vrf: default
  nodes:
    - 192.168.0.1
```

### Expected results

```shell
ip name-server vrf default 192.168.0.1
```

### Actual results

```shell
ip name-server 192.168.0.1
```

aphillipps:
Thanks for the information. The fix will be implemented in v1.1.0 release.

PR #228, opened 2020-09-04
Fix for issues 199 and 204

aphillipps:
- #199 : Syntax change of domain-name server declaration in default VRF in eos_cli_config_gen
- #204: Removal of VRF IP routing dependency with ip routing in GRT
- Removal of VRF IPv6 routing dependency with ipv6 routing in GRT


titom73 (approved):
LGTM

PR #230, opened 2020-09-08
eos_cli_config_gen - minor fixes

carlbuchmann:
- remove sorting from ip name-servers
- correct delimiter for IPv6 template


titom73 (approved):
LGTM

PR #231, opened 2020-09-08
ip name-server - sort to match EOS output

carlbuchmann:
After further review, noticed the ip name-server is sorted in EOS, just not naturally.

the jinja2 `| sort` filter provides the desired results.


aphillipps (approved):
Looks good.

PR #252, opened 2020-09-22
upgrade role to translate 1.0 data to 1.1

ankudinov:
The initial version of upgrade role to translate 1.0 data to 1.1.
It is likely not covering all possible combinations yet. To fix that in future we need better test data That is consistent and covers entire data model.
Current version is based on netdevops-examples data.


titom73:
As discussed with @carlbuchmann 

Updated with PR https://github.com/aristanetworks/ansible-avd/pull/255 with:

- CI
- Documetaion
- Rename role to make it generic and support all potential upgrade to do in futur

PR #258, opened 2020-10-01
initial release notes

carlbuchmann:
## Change Summary
Add release notes section, where we document:
- High-level modifications -> what roles are new or updated
- Changes to the data model
- Links to release notes (Github Release tag) for more information.

## Types of changes
- [x] Documentation content changes

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -21,7 +21,9 @@ repo_url: https://github.com/aristanetworks/ansible-avd
 use_directory_urls: true
 nav:
   - Home: README.md
-  - Release Notes: docs/release-notes.md
+  - Release Notes:
+    - 1.1.x: docs/release-notes/1.1.x.md
+    - 1.0.x: docs/release-notes/1.0.x.md
   - Getting Started: docs/getting-started.md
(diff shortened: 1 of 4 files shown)


carlbuchmann:
Expected failure in CI for mkdocs_check, as the release tag for 1.1.0 release doesn't exist yet!

mkdocs.yml
```diff
@@ -21,6 +21,7 @@ repo_url: https://github.com/aristanetworks/ansible-avd
 use_directory_urls: true
 nav:
   - Home: README.md
+  - Release Notes: docs/release-notes.md
```
titom73: Should we create a Release Note section per major versions like `1.0.x` and `1.1.x` as it might become difficult to read after few versions?

noredistribution (approved):
LGTM!

mkdocs.yml
carlbuchmann: @titom73 that is a good suggestion that makes sense to me. I will update it.
Also, any suggestions on the MKDOCs link check CI, will that impact the actual release to github.io?

PR #412, opened 2020-12-02
Adjust guidelines for vrf_vni values

### LINKED ISSUE #314: Increase VRF configuration scale
**Issue Type**
- Template enhancement

Would like to have larger scale of VRF option, it is understood that 1024 is a limitation, however if we have multiple tenants shared across the Fabric, there should be an option to have more VRFs in total spread around for different tenants.

The current 1024 is not enough, if we can either **double** the size or go a bit more beyond it that would be great.

None really. As many factors rely on this L3VNI/RD/RT

**Additional context**
Providing that there is space for scale taking into factor that not all VRFs will be in a single location.
The only limiting factor is the    

```
    # Define L3 network services organized by vrf.
    vrfs:
      # VRF name | Required
      < tenant_a_vrf_1 >:

        # VRF VNI | Required.
        # The VRF VNI range is limited.
        vrf_vni: <1-1024>
```
That many things use this to generate EVPN config.

ClausHolbechArista:
## Change Summary

We don't enforce the `vrf_vni : < 1-1024 >` limit documented in the Readme file, so this change is only for adjusting the guidelines in Readme.md to make the user aware of what to look out for, before using `vrf_vni` > 1024.

- [X] Code style update (formatting, renaming)
- [X] Documentation content changes

## Related Issue(s)

Fixes #314

## Component(s) name
eos_l3ls_evpn

## Proposed changes

* Readme updated with following guidelines:
```yaml
        # VRF VNI | Required.
        # The VRF VNI range is not limited, but it is recommended to keep vrf_vni <= 1024
        # It is necessary to keep [ vrf_vni + MLAG IBGP base_vlan ] < 4094 to support MLAG IBGP peering in VRF.
        # If vrf_vni > 1094 make sure to change mlag_ibgp_peering_vrfs: { base_vlan : < > } to a lower value (default 3000).
        # If vrf_vni > 10000 make sure to adjust mac_vrf_vni_base accordingly to avoid overlap.
        vrf_vni: < 1-1024 >
```
* Small formatting changes for `vrf_vni`

## How to test

Before changing anything I tested with vrf_vni > 1024 and it worked fine. The MLAG IBGP VLANs are calculated as `vrf_vni` + `base_vlan` so I ended up with VLAN numbers > 4096. Hence we need guidelines to make the user aware.

## Checklist:

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_l3ls_evpn/README.md
+++ b/ansible_collections/arista/avd/roles/eos_l3ls_evpn/README.md
@@ -841,8 +841,11 @@ tenants:
       < tenant_a_vrf_1 >:
 
         # VRF VNI | Required.
-        # The VRF VNI range is limited.
-        vrf_vni: <1-1024>
+        # The VRF VNI range is not limited, but it is recommended to keep vrf_vni <= 1024
+        # It is necessary to keep [ vrf_vni + MLAG IBGP base_vlan ] < 4094 to support MLAG IBGP peering in VRF.
+        # If vrf_vni > 1094 make sure to change mlag_ibgp_peering_vrfs: { base_vlan : < > } to a lower value (default 3000).
+        # If vrf_vni > 10000 make sure to adjust mac_vrf_vni_base accordingly to avoid overlap.
+        vrf_vni: < 1-1024 >
 
         # IP Helper for DHCP relay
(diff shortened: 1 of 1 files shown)


carlbuchmann (approved):
LGTM!

PR #442, opened 2020-12-08
Setting maximum paths & ecmp for eBGP in eos_l3ls_evpn role

### LINKED ISSUE #384: Maximum-paths and ECMP per platform or custom value
**Issue Type**

- Template enhancement
- Role enhancement

There should be a knob to be able to change the maximum-paths and ECMP.
With the current model we derive that based on the current topology. There is a catch with implementing it this way as when there is a change in the topology it will impact things.

Each platform has it's maximum possible paths and ECMP entries supported for any prefix. We should either enable the maximum possible which won't harm anything, just makes things usable and when there will be changes of the topology in the future we don't have to change the max paths or ECMP.
This should be per platform type.

Second way to do this would be to have a knob that allows us to configure the maximum paths either fabric wide, or per device.

```
router bgp 1
   maximum-paths 16 ecmp 16
```

**Additional context**

aphillipps:
## Change Summary
In the current AVD model, the number of maximum paths and the number of ecmp paths can not be set. It is configured automatically using the number of spines in the fabric as a reference. In some design the desired maximum paths is the target number of spines from the beginning on even though spines will be added little by little. In this case, this value can be set manually in the fabric description.

## Types of changes

## Related Issue(s)
#384 

## Component(s) name
eos_l3ls_evpn role

## Proposed changes
The new optional keys are added in the description of the fabric as follows:

```yaml
# BGP multi-path
# If not defined these values will be equal by default to the number of spines in the fabric
# If defined the values defined will both be applied to leaf and spine switches
# Note: these values are only relevant for an eBGP scenario
bgp_maximum_paths: <number_of_max_paths>
bgp_ecmp: <number_of_ecmp_paths>
```

## How to test

Add for example the following values and keys in DC1_FABRIC.yml:

```yaml
bgp_maximum_paths: 8
bgp_ecmp: 8
```

This will render the following structured config file (at the bottom of all the other declared bgp_defaults values):
```yaml
  bgp_defaults:
    - no bgp default ipv4-unicast
    - distance bgp 20 200 200
    - graceful-restart restart-time 300
    - graceful-restart
    - maximum-paths 8 ecmp 8
```
If not defined, it will simply render the line above with the number of spines instead of 8.

## Checklist:

### DIFF
--- a/ansible_collections/arista/avd/molecule/evpn_underlay_ebgp_overlay_ebgp/documentation/devices/DC1-BL1A.md
+++ b/ansible_collections/arista/avd/molecule/evpn_underlay_ebgp_overlay_ebgp/documentation/devices/DC1-BL1A.md
@@ -686,6 +686,7 @@ Router ISIS not defined
 | 192.168.255.3 | Inherited from peer group EVPN-OVERLAY-PEERS |
 | 192.168.255.4 | Inherited from peer group EVPN-OVERLAY-PEERS |
 
+
 ### Router BGP EVPN Address Family
 
 #### Router BGP EVPN MAC-VRFs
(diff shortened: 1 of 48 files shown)


titom73 (approved):
LGTM!

PR #512, opened 2020-12-19
BGP RT Membership support

### LINKED ISSUE #374: Route-Target membership - RTC support
**Issue Type**

- Role enhancement

VTEPs will advertise their route-types for different services to Route-Servers/Route-Reflectors. There is a chance that not every VRF/VLAN will be on every VTEP, and with networks that have high numbers of routes every VTEP in the network will have to learn about all these even though they do not have the services defined on them. Thus leading to different scale issues data plane or control plane.

Route target membership (RTC) will be very important addition to the AVD as it helps with scale and convergence as a result of lowering the scale.

The solution is to support new address-family (RT membership) that allows us to only reflect back to Leaves the route-types based on RT that are available on the VTEP. Thus reducing the number of routes massively.

```
router bgp {{ ASN }}
   !
   address-family rt-membership
      neighbor EVPN-OVERLAY-PEERS activate
```

In case of large data centres with hundreds of VTEPs there is a burden of implementing per neighbour (VTEP pair) policies using route-maps with extcommunity and AS path filters. This with scale will prove difficult.

**Additional context**
This is merely another address-family that will use the same peer-group config as EVPN peers and it will be enabled on EVPN speaking devices.

onurgashi:
## Change Summary

This  feature adds support for RTC or RT membership for BGP.

## Types of changes

## Related Issue(s)

Fixes #374 

## Component(s) name
eos_cli_config_gen
molecule tests

Next step create data model and add to eos_l3ls_evpn

## How to test

Was tested using molecule and hardware DUT.

## Checklist:

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
@@ -17,6 +17,15 @@ router_bgp:
       password: "q+VNViP5i4rVjW1cxFv2wA=="
       send_community: true
       maximum_routes: 0
+    EVPN-OVERLAY-RS-PEERS:
+      type: evpn
+      remote_as: 65001
+      update_source: Loopback0
+      bfd: true
+      ebgp_multihop: 3
+      password: "q+VNViP5i4rVjW1cxFv2wA=="
+      send_community: true
+      maximum_routes: 0
   neighbors:
     192.168.255.1:
       peer_group: EVPN-OVERLAY-PEERS
@@ -34,9 +43,11 @@ router_bgp:
       EVPN-OVERLAY-PEERS:
         activate: true
         default_route_target:
+      EVPN-OVERLAY-RS-PEERS:
+        activate: true
+        default_route_target:
           only: true
           encoding_origin_as_omit:
-            disabled: true
   address_family_ipv4:
     peer_groups:
       EVPN-OVERLAY-PEERS:
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
@@ -1,0 +1,104 @@
+router_bgp:
+  as: 65101
+  router_id: 192.168.255.3
+  bgp_defaults:
+    - no bgp default ipv4-unicast
+    - distance bgp 20 200 200
+    - graceful-restart restart-time 300
+    - graceful-restart
+    - maximum-paths 2 ecmp 2
+  peer_groups:
+    EVPN-OVERLAY-PEERS:
+      type: evpn
+      remote_as: 65001
+      update_source: Loopback0
+      bfd: true
+      ebgp_multihop: 3
+      password: "q+VNViP5i4rVjW1cxFv2wA=="
+      send_community: true
+      maximum_routes: 0
+  neighbors:
+    192.168.255.1:
+      peer_group: EVPN-OVERLAY-PEERS
+    192.168.255.2:
+      peer_group: EVPN-OVERLAY-PEERS
+  redistribute_routes:
+  address_family_evpn:
+    peer_groups:
+      EVPN-OVERLAY-PEERS:
+        activate: true
+      MLAG-IPv4-UNDERLAY-PEER:
+        activate: false
+  address_family_rtc:
+    peer_groups:
+      EVPN-OVERLAY-PEERS:
+        activate: true
+        default_route_target:
+          only: true
+          encoding_origin_as_omit:
+            disabled: true
+  address_family_ipv4:
+    peer_groups:
+      EVPN-OVERLAY-PEERS:
+        activate: false
+  vlan_aware_bundles:
+## Tenant_A ##
+    TENANT_A_PROJECT01:
+      rd: "192.168.255.3:11"
+      route_targets:
+        both:
+          - "11:11"
+      redistribute_routes:
+        - learned
+      vlan: 110
+    TENANT_A_PROJECT02:
+      rd: "192.168.255.3:12"
+      route_targets:
+        both:
+          - "12:12"
+      redistribute_routes:
+        - learned
+      vlan: 112
+## Tenant_B ##
+    B-ELAN-201:
+      tenant: Tenant_B
+      rd: "192.168.255.3:20201"
+      route_targets:
+        both:
+          - "20201:20201"
+      redistribute_routes:
+        - learned
+      vlan: 201
+  vlans:
+  vrfs:
+## Tenant_A ##
+    TENANT_A_PROJECT01:
+      router_id: 192.168.255.3
+      rd: "192.168.255.3:11"
+      route_targets:
+        import:
+          evpn:
+            - "11:11"
+        export:
+          evpn:
+            - "11:11"
+      neighbors:
+        10.255.251.1:
+          peer_group: MLAG-IPv4-UNDERLAY-PEER
+      redistribute_routes:
+        - connected
+    TENANT_A_PROJECT02:
+      router_id: 192.168.255.3
+      rd: "192.168.255.3:12"
+      route_targets:
+        import:
+          evpn:
+            - "12:12"
+        export:
+          evpn:
+            - "12:12"
+      neighbors:
+        10.255.251.1:
+          peer_group: MLAG-IPv4-UNDERLAY-PEER
+      redistribute_routes:
+        - connected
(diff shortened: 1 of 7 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1145,7 +1145,7 @@ router_bgp:
     - "< bgp command as string >"
   peer_groups:
     < peer_group_name_1>:
-      type: < ipv4 | evpn >
+      type: < ipv4 | evpn | rtc >
```
carlbuchmann: The type knob `rtc` doesn't appear to be used.

onurgashi: That's if someone wants to use a different peer-group other than evpn, in our case EVPN does the job as we peer to the same Route Servers or Spines. For flexibility if you want to have different peer-group used only for RTC

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -175,6 +175,34 @@ router bgp {{ router_bgp.as }}
...
+{%          for peer_group in router_bgp.address_family_rtc.peer_groups | arista.avd.natural_sort %}
+{%             if router_bgp.address_family_rtc.peer_groups[peer_group].activate == true %}
+      neighbor {{ peer_group }} activate
+{%             elif router_bgp.address_family_rtc.peer_groups[peer_group].activate == false %}
+      no neighbor {{ peer_group }} activate
+{%             endif %}
+{%             if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target is defined %}
+{%                 if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.only is defined and router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.only == true %}
+      neighbor {{ peer_group }} default-route-target only
+{%                 elif router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.disabled is defined and router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.disabled is defined == true %}
```
ClausHolbechArista: looks like you have a "is defined" too much here.

ClausHolbechArista: You have elif between "disabled" and "only". So that means we will only render one or the other. If that is on purpose, we should make it into a single variable (ex. "mode") that can be set to either "disabled" or "only".
But what is the "disabled" option anyway? I don't see it in the aid.

ClausHolbechArista: According to TOI the disabled flag is there, but I understand it as only applicable to specific neighbors as opposed to peer-groups.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -175,6 +175,34 @@ router bgp {{ router_bgp.as }}
...
+      neighbor {{ peer_group }} default-route-target disabled
+{%                 endif %}
+{%             else %}
+      neighbor {{ peer_group }} default-route-target
+{%             endif %}
+{%             if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.encoding_origin_as_omit is defined %}
+      neighbor {{ peer_group }} default-route-target encoding origin-as omit
+{%             elif router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.encoding_origin_as_omit.disabled is defined and router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.encoding_origin_as_omit.disabled is defined == true %}
+      neighbor {{ peer_group }} default-route-target encoding origin-as omit disabled
```
ClausHolbechArista: one too many "is defined" here.

ClausHolbechArista (changes requested):
The new EOS "disabled" logic makes it really hard to construct a clear data model. Since we operate on peer-group level, I don't see any purpose of disabled for each option. See comments inline.

onurgashi:
I will update the changes, appreciate your help @ClausHolbechArista

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
onurgashi: - [ ] Remove type from readme
- [ ] Update the documentation artifacts so peer-group can belong to multiple AFs

@carlbuchmann

onurgashi: If we remove the type then then there is no way to describe which AF the peer-group belongs for documentation

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
onurgashi: You are right, I changed things to reflect your comments.

onurgashi:
The cli_config_gen comments have been resolved. 

I will continue adding this as a knob to eos_l3ls_evpn:

- rt_membership: <true | false> under the Fabric, so people can enable/disable?
- Use the same peer-group for EVPN, which makes sense as you want it everywhere where EVPN is running.
- And use default only on the Spine/Super-Spine/Controller, and default on the Leaves?

@ClausHolbechArista @carlbuchmann

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
```diff
@@ -0,0 +1,115 @@
...
+  address_family_rtc:
+    peer_groups:
+      EVPN-OVERLAY-PEERS:
+        activate: true
+        default_route_target:
+      EVPN-OVERLAY-RS-PEERS:
+        activate: true
+        default_route_target:
+          only: true
+          encoding_origin_as_omit:
```
ClausHolbechArista, line 50: Your Readme says `encording_origin_as_omit : < true | false >` but your test has `none`. The code also only looks if this knob is defined, so we could just change the readme to reflect that.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1324,6 +1324,13 @@ router_bgp:
         activate: < true | false >
         route_map_in: < route_map_name >
         route_map_out: < route_map_name >
+  address_family_rtc:
+    peer_groups:
+      < peer_group_name >:
+        activate: < true | false >
+        default_route_target:
+          only: < true | false >
+          encoding_origin_as_omit: < true | false >
```
ClausHolbechArista: Your Readme says `encording_origin_as_omit : < true | false >` but your test has `none`. The code also only looks if this knob is defined, so we could just change the readme to reflect that.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -181,6 +181,30 @@ router bgp {{ router_bgp.as }}
...
+      no neighbor {{ peer_group }} activate
+{%               endif %}
+{%               if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target is defined %}
+{%                   if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.only is defined and router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.only == true %}
+      neighbor {{ peer_group }} default-route-target only
+{%                   else %}
+      neighbor {{ peer_group }} default-route-target
+{%                   endif %}
+{%               endif %}
+{%               if router_bgp.address_family_rtc.peer_groups[peer_group].default_route_target.encoding_origin_as_omit is defined %}
```
ClausHolbechArista, line 202: Your Readme says `encording_origin_as_omit : < true | false >` but your test has `none`. The code also only looks if this knob is defined, so we could just change the readme to reflect that.

ClausHolbechArista (changes requested):
Looks good ! Only a minor documentation change (or change the logic in template and test to match the documentation :)

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-rtc.yml
onurgashi: Thanks, fixed!

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
onurgashi: Thanks, fixed!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
onurgashi: Thanks, fixed!

ClausHolbechArista (approved):

ClausHolbechArista:
LGTM

PR #595, opened 2021-01-08
Fix spanning-tree documentation output

carlbuchmann:
## Change Summary

Fix Spanning tree documentation output

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Component(s) name

eos_cli_config_gen

## Checklist:


titom73 (commented):
Looks good. Just one question: do we want to render configuration when EOS spanning-tree is set to none ?

Currently it is not render but as it is not EOS default, maybe we should.

carlbuchmann:
@titom73 imho I believe the documentation should be rendered based on the intended documentation, as defaults could change in EOS over time, probably not for STP but elsewhere in the configuration.

PR #607, opened 2021-01-11
Add "defined" checks to natural_sort

ClausHolbechArista:
## Change Summary

Add a check for `None` or `not defined` to the natural_sort filter, and return empty list.

## Types of changes

- [X] Code style update (formatting, renaming)

## Component(s) name
ansible-avd collection
natural_sort

## Proposed changes

The `arista.avd.natural_sort` filter will now return an empty list if the value parsed to `natural_sort` is `none` or `undefined`. This can help simplify the templates, since we no longer need to have a test before the for-loop.

###Before
```jinja
{%     if tenants[tenant].l2vlans is defined %}
{%         for l2vlan in tenants[tenant].l2vlans | arista.avd.natural_sort %}
...
{%         endfor %}
{%     endif %}
```
###After
```jinja
{%     for l2vlan in tenants[tenant].l2vlans | arista.avd.natural_sort %}
...
{%     endfor %}
```
## How to test

I tested by running my test repo with and without this change. No difference to the produced output. Then I changed the l3leaf template like above. Without this PR change it produced an `Undefined variable` exception for some tenants. With the PR change it worked with no error exactly like before.

## Checklist:

### DIFF
--- a/ansible_collections/arista/avd/plugins/README.md
+++ b/ansible_collections/arista/avd/plugins/README.md
@@ -30,6 +30,8 @@ To use this filter:
 
 The `natural_sort` filter provides the capabilities to sort a list or a dictionary of integers and/or strings that contain alphanumeric characters naturally. When leveraged on a dictionary, only the key value will be returned.
 
+The filter will return an empty list if the value parsed to `natural_sort` is `none` or `undefined`.
+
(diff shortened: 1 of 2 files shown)


carlbuchmann (approved):
LGTM

PR #674, opened 2021-02-02
Rename eos_l3ls_evpn to eos_designs

titom73:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes)
- [ ] Documentation content changes
- [x] Other (please describe): Role renaming

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.eos_l3ls_evpn` --> `arista.avd.eos_designs`

## Proposed changes

- Rename eos_l3ls_evpn to eos_designs
- Move all the Jinja2 logic part of `templates/evpn-fabric-<device-role>.j2` to a dedicated file to make it easier to read and maintain.

### Role rename

Change playbook like:

```diff
  tasks:
    - name: generate intented variables
      import_role:
-        name: arista.avd.eos_l3ls_evpn
+        name: arista.avd.eos_designs
```

### Proposed folder structure:

```bash
├── templates
│   ├── designs # new folder to hold all design specific templates
│   │   └── l3ls-evpn # subfolder for l3ls evpn design specific templates
│   │       ├── l3ls-evpn-l2leaf.j2
│   │       ├── l3ls-evpn-l3leaf.j2
│   │       ├── l3ls-evpn-overlay-controller.j2
│   │       ├── l3ls-evpn-spine.j2
│   │       └── l3ls-evpn-super-spine.j2
```

We may also move logic files named `switch-evpn-*` in a subfolder to prepare `eos_l3ls_evpn` to become more generic with support for other designs.

Then for all `templates/evpn-fabric-<device-role>.j2`, use following include:

```jinja2
{# Set Super Spine Variables #}
{% set switch = namespace() %}
{% set leaf = namespace() %} {# When applicable #}
{% include 'designs/l3ls-evpn/l3ls-evpn-super-spine.j2' %}
```

## How to test

Refer to molecule

## TODO

- [x] rename role to eos_design
- [x] re-create eos_l3ls_evpn with deprecation warning
- [x] update documentation with new role name
- [x] update playbooks with new role name
- [x] update molecule tests with new role name
- [x] refactor template directory to support multiple designs
- [x] update images
- [x] update README content, github root, collection root and role: eos_designs.
- [x] Add Molecule test for legacy `eos_l3ls_evpn` deprecation
[...]

PR #770, opened 2021-03-05
Add platform_speed_groups to eos_designs

ClausHolbechArista:
## Change Summary
Add support for configuring speed_groups per hardware platform in eos_designs

## Types of changes

- [x] Documentation content changes

## Related Issue(s)

Branch is based on eos_designs rename branch (#674), so should _not_ be merged until that has merged.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Abstraction of hardware speed_groups. This is particular useful if placed in per-rack group_var file together with `servers`, so it is easier for the operator to maintain speed-groups together with the server connections.
Since the setting will depend on the platform, it is possible to control which devices will get speed-groups set. Ex only set on `7280R2` and not on `7010`.

Data model:
```yaml
# Set Hardware Speed Groups per Platform
platform_speed_groups:
  < platform >:
    < speed >: [ < speed_group >, < speed_group > ]
```

The template handles duplicates and reordering gracefully using a tmp dict.

Molecule has been updated with example below.

## How to test

I added the following to molecule test:
```yaml
# Set Hardware Speed Groups per platform
platform_speed_groups:
  7280R:             # Only setting speed-groups on 7280R platform, so only L3leaf should get this setting.
    25G: [ 3, 2 ]    # Unsorted order, but we should sort output correctly.
    10G: [ 1, 2, 4 ] # Duplicate speed-group 2. Since we sort on key first the result will be 25G for group 2
```
Output in molecule was as expected and configuration was correctly generated on the expected devices only.

## Checklist:


ClausHolbechArista:
I will create a new PR for this.

PR #891, opened 2021-04-23
Fix references to cvp dirs

ClausHolbechArista:
## Change Summary

Correct CI references to AVD directories instead of CVP.


ClausHolbechArista:
Not needed

PR #994, opened 2021-06-03
TerminAttr Updates

### LINKED ISSUE #950: ingestvrf default not being output for TerminAttr
## Issue Type

- [ ] Bug Report
- [X] EOS Configuration rendering error
- [ ] Documentation Issue Report

## Summary

Though its probably fine, CVP includes the `-ingestvrf=default` to the daemon TerminAttr config even when using the default vrf.  Ansible AVD's logic seems to remove that flag when it recognizes the ingestvrf value as show the default vrf.  Though technically correct, it diverges from the latest CVP deployment standard.

## Component Name

eos_cli_config_gen

## Steps to reproduce

```yaml
daemon_terminattr:
  ingestgrpcurl:
    ips:
      -  10.10.10.10
    port: 9910
  ingestauth_key: key
  ingestvrf: default
  smashexcludes: test
  ingestexclude: test
```

### Expected results

```shell
daemon TerminAttr
    exec /usr/bin/TerminAttr -ingestgrpcurl=10.10.10.10:9910 -cvcompression=gzip -ingestauth=key -smashexcludes=test -ingestexclude=test -ingestvrf default -taillogs
```

### Actual results

```shell
daemon TerminAttr
    exec /usr/bin/TerminAttr -ingestgrpcurl=10.10.10.10:9910 -cvcompression=gzip -ingestauth=key -smashexcludes=test -ingestexclude=test -taillogs
```

ryanmerolle:
## Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Other (please describe): Missing Molecule Tests

## Related Issue(s)

Fixes #950

## Component(s) name

`arista.avd.eos_cli_config_gen

## Proposed changes
Per #950 updating TerminAttr  config syntax to reflect what CVP auto deploys.

## How to test
Molecule tests added

## Checklist:

- [X] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contributing/) document.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.
- [X] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly


noredistribution:
Thanks for doing a PR! That looks good to me, that's what I would've done as well, but then I realized we will have to revamp the whole TA config to add support for streaming to multiple clusters and moving forward we'll use cvopt flags (that will be a big change and won't happen very soon so for me this is good to go) so I wasn't sure if we should do this in the interim, but if it helps why not!

Aslo wanted to point out that the `-ingestvrf=default` is only set during onboarding in CV, however none of the configlet builders are setting this, as it is a default flag (see the integrated `SYS_TelemetryBuilderV2` or `V3` or `V4`)

ryanmerolle:
Can this be merged then?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/daemon-terminattr.j2
```diff
@@ -39,11 +39,7 @@ daemon TerminAttr
 {%         else %}
 {%             set cvp_config.cli = cvp_config.cli ~ " -ingestexclude=" %}
 {%         endif %}
-{%         if daemon_terminattr.ingestvrf is arista.avd.defined %}
-{%             if daemon_terminattr.ingestvrf != 'default' %}
-{%                 set cvp_config.cli = cvp_config.cli ~ " -ingestvrf=" ~ daemon_terminattr.ingestvrf %}
-{%             endif %}
-{%         endif %}
+{%         set cvp_config.cli = cvp_config.cli ~ " -ingestvrf=" ~ daemon_terminattr.ingestvrf | default('default') %}
```
ClausHolbechArista: @ryanmerolle I agree that we can remove the suppression of vrf `default` if vrf `default` is set in the data model, but I don't think we should print out vrf default when nothing is specified. `eos_cli_config_gen` should only render what is handed to it, and not enforce any logic.
Here I suggest that we keep the check from line 42 in original code and remove the check in line 43. This would render the vrf no matter which value it has.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/daemon-terminattr.j2
```diff
@@ -68,9 +64,7 @@ daemon TerminAttr
 {%         else %}
 {%             set cvp_config.cli = cvp_config.cli ~ " -ingestexclude=" %}
 {%         endif %}
-{%         if daemon_terminattr.ingestvrf is arista.avd.defined() %}
-{%             set cvp_config.cli = cvp_config.cli ~ " -cvvrf=" ~ daemon_terminattr.ingestvrf %}
-{%         endif %}
+{%         set cvp_config.cli = cvp_config.cli ~ " -cvvrf=" ~ daemon_terminattr.ingestvrf | default('default') %}
```
ClausHolbechArista: As noted above, I think the previous code was more correct, since it would render vrf default if that is passed in the data model.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/daemon-terminattr.j2
```diff
@@ -6,7 +6,7 @@
...
 | -------------- | --------------- | ------------------------- | -------------- | -------------- | ---------- | -------- | ------ |
-| gzip | {% for cvp_ip in daemon_terminattr.ingestgrpcurl.ips %}{{ cvp_ip }}:{{ daemon_terminattr.ingestgrpcurl.port }}{% if not loop.last %},{% endif %}{% endfor %} | {{ daemon_terminattr.ingestauth_key | arista.avd.default('UNSET') }} | {{ daemon_terminattr.smashexcludes }} | {{ daemon_terminattr.ingestexclude }} | {{ daemon_terminattr.ingestvrf }} | {{ daemon_terminattr.ingestvrf }} | {{ daemon_terminattr.disable_aaa | default (false) }} |
+| gzip | {% for cvp_ip in daemon_terminattr.ingestgrpcurl.ips %}{{ cvp_ip }}:{{ daemon_terminattr.ingestgrpcurl.port }}{% if not loop.last %},{% endif %}{% endfor %} | {{ daemon_terminattr.ingestauth_key | arista.avd.default('UNSET') }} | {{ daemon_terminattr.smashexcludes }} | {{ daemon_terminattr.ingestexclude }} | {{ daemon_terminattr.ingestvrf | default('default') }} | {{ daemon_terminattr.ingestvrf | default('default') }} | {{ daemon_terminattr.disable_aaa | default (false) }} |
```
ClausHolbechArista: Should follow the logic decided in the config template.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/daemon-terminattr.j2
ryanmerolle: When I raised this in #950 there was no objection.  As I stated when CVP onboards a device, if the default vrf is used it uses the config synxtax:

```
daemon TerminAttr
    exec /usr/bin/TerminAttr -ingestgrpcurl=10.10.10.10:9910 -cvcompression=gzip -ingestauth=key -smashexcludes=test -ingestexclude=test -ingestvrf default -taillogs
```

If you want me to revert this change, there is no point for this PR.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/daemon-terminattr.j2
ryanmerolle: See my response to the other comment.  I took that CVP's default onboard config as valid, so eos_cli_config_gen should render config that matches that.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/daemon-terminattr.j2
ClausHolbechArista: CVP produces config with and without the ingestvrf depending on if you onboard existing device, zero-touch a device or use the system configletbuilder. So it is difficult to say that one is more correct than the other.
But even if ingestvrf is desired, the role eos_cli_config_gen should not be enforcing any "good practices". It is just a translator. Any abstraction and logic should go into eos_designs.
Re. #950 there is no mention of always setting default VRF if no VRF is given. So my objection is not about #950 but only this choice of always setting the VRF.
If we change default config output like this, it will affect all users of AVD.

ryanmerolle: Fair enough.  Your point, which I misunderstood is that I should allow no vrf to be set and if no vrf to be set, not to define a default value for vrf if not set.

ryanmerolle:
@ClausHolbechArista I think I resolved your concerns.

ClausHolbechArista (approved):
LGTM! Thank you for this contribution.

PR #1044, opened 2021-06-17
feat(eos_cli_config): Ensure required directories are present

sc68cal:
Someone using this role for the first time, and just loading the `eos_cli_config_gen` role shouldn't have to create the directories required by this role, beforehand.

```yaml
- hosts: all

  tasks:
    - import_role:
```

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
@@ -1,4 +1,16 @@
 ---
+- name: Create required output directories if not present
+  file:
+    path: "{{ item }}"
+    state: directory
+    mode: 0775
+  loop:
+    - "{{ structured_dir }}"
+    - "{{ eos_config_dir }}"
+    - "{{ devices_dir }}"
+  delegate_to: localhost
+  run_once: true
+
 - name: Check if structure configuration file exists
   tags: [build, provision]
   stat:


sc68cal:
I see there is a `build_output_folders` role, but this role does not include it, even though it depends on directories that this other role creates.

ClausHolbechArista:
Please set `mode`,`run_once`, `delegate_to`.
Ex:
```yaml
- name: "..."
  file:
    path: "{{ ... }}"
    state: directory
    mode: 0775
  delegate_to: localhost
  run_once: True
```
Maybe run it as a single task with items for each dir.

sc68cal:
@ClausHolbechArista good suggestions. However, would it be better to just `include_role: build_output_folders` (or `import_role`) to reduce the duplication of effort? Or is that too heavy handed? More of a style question that I'd prefer you or another maintainer answer. I'm okay with either approach.

titom73:
> @ClausHolbechArista good suggestions. However, would it be better to just `include_role: build_output_folders` (or `import_role`) to reduce the duplication of effort? Or is that too heavy handed? More of a style question that I'd prefer you or another maintainer answer. I'm okay with either approach.

Would be the best approach: single place to maintain folders creation. In this case, we should update `arista.avd.build_output_folders` to not clean and recreate folders then to avoid deleting existing files. 

Let me know if you can do it, or I will take it.

titom73:
After discussing with Claus, we are thinking to move tasks from `arista.avd.build_output_folders` to their respective roles to create only required folders. So Please follow your first idea and I will change `arista.avd.eos_designs` and all other folders accordingly

sc68cal:
Got it. I will re-spin this patch with @ClausHolbechArista 's original review suggestions, stand by.

ClausHolbechArista (approved):
LGTM

PR #1077, opened 2021-07-06
Optimize(molecule): Add ANSIBLE_OPTIONS to pass ansible options to molecule scenario

titom73:
## Types of changes

- [x] Refactoring (no functional changes)
- [x] Other (please describe): Molecule

## Related Issue(s)

N/A

## Component(s) name

molecule

## Proposed changes

Add option to limit molecule execution:

```bash
$ make converge ANSIBLE_OPTIONS='--limit router-ospf'
cd .. && \
        molecule converge --scenario-name eos_cli_config_gen -- --limit router-ospf
```

> Specifically designed for eos_cli_config_gen

## How to test

Run molecule

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contributing/) document.

### DIFF
--- a/ansible_collections/arista/avd/molecule/Makefile
+++ b/ansible_collections/arista/avd/molecule/Makefile
@@ -1,5 +1,6 @@
 
 MOLECULE ?= eos_cli_config_gen
+LIMIT ?= 'all'
 
 .PHONY: help
 help: ## Display help message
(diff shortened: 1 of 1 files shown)


ClausHolbechArista:
Instead of only passing LIMIT, could we make a knob for ansible-options or similar? So parse anything ? I often use it with tags etc.

titom73:
Good idea.
Updated code to support generic ansible options. By default it is set to `--limit all` to avoid passing a unsupported `-- ` after molecule scenario

ClausHolbechArista:
adding `--` with no option after works fine. So I think it would be cleaner to just set `""` as default.

carlbuchmann (approved):
LGMT!

PR #1000, opened 2021-06-03
Feat(eos_designs): Refactor topology data models for AVD v3.0

ClausHolbechArista:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [x] Documentation content changes
- [ ] Other (please describe):

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

### Behavioral changes
- l3leaf will no longer have spine asn hardcoded on underlay peer-group, but instead specified per uplink switch.
- raw_eos_cli defined under `tenants.vrfs.<vrf>.raw_eos_cli` is no longer rendered on l2leaf
- `structured_config` defined under `tenants.vrfs.<vrf>.raw_eos_cli` is no longer rendered on l2leaf
- Rewording in fabric documentation
  - "Uplink IPv4 Pool" instead of "P2P Summary"
  - "VTEPs only" instead of "Leafs Only"

### New features
- inheritance for all settings from `<switch_type_key>.defaults`->`switch_type_key>.node_groups.<node_group>`->`switch_type_key>.node_groups.<node_group>.nodes.<node>`
- Support for `evpn` setting per network service (`vrfs`,`svis`,`l2vlans`) to allow non-vxlan enabled services
- Update `upgrade_tools` to upgrade variables from AVD v2.x to v3.0 during the play

### Data model changes
All changes below can be handles automatically by the `upgrade_tools` role.

#### Changes to `spine` data model
- Change from spine.platform to spine.defaults.platform
- Change from spine.bgp_as to spine.defaults.bgp_as
- Change from overlay_loopback_network_summary to spine.defaults.loopback_ipv4_pool
- Change from spine.uplinks_to_super_spine_interfaces to spine.defaults.uplink_interfaces
- New spine.defaults.uplink_switches variable based on keys in super_spine.nodes
- Change from max_super_spines to spine.defaults.max_uplink_switches
- Change from max_spine_to_super_spine_links to spine.defaults.max_parallel_uplinks
- Change from super_spine_underlay_p2p_network_summary to spine.defaults.uplink_ipv4_pool
- Change from p2p_uplinks_ptp to spine.defaults.uplink_ptp
- Change from spine_bgp_defaults to spine.defaults.bgp_defaults
[...]

PR #1104, opened 2021-07-20
Feat(avd_to_containerlab): Add role to generate containerlab topology

titom73:
## Change Summary

Initial support of containerlab to build l3ls-evpn topology from
EOS-DESIGNS.

Require: avd in version 3.x & PR #1000

## Types of changes

## Component(s) name

`arista.avd.eos_designs_to_containerlab`

## How to test

```yaml
---
- name: Build containerlab topology
  hosts: DC1_FABRIC
  connection: local
  gather_facts: false
  tasks:
    - name: 'Build a containerlab topolgy'
      import_role:
        name: arista.avd.eos_designs_to_containerlab
      vars:
        mgmt_network_v4: 10.73.0.0/16
        ceos_version: arista/ceos:4.25.2F
```

## Checklist:

- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contributing/) document.


ansible_collections/arista/avd/roles/eos_designs_to_containerlab/README.md
```diff
@@ -0,0 +1,73 @@
...
+    version: 0.15.3
+     commit: 533d911
+       date: 2021-07-19T14:35:16Z
+     source: https://github.com/srl-labs/containerlab
+ rel. notes: https://containerlab.srlinux.dev/rn/0.15/#0153
+```
+
+## Example Playbook
+
+Including an example of how to use your role (for instance, with variables passed in as parameters) is always nice for users too:
```
ClausHolbechArista: I think this line should be removed?

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/README.md
```diff
@@ -0,0 +1,73 @@
...
+  hosts: DC1_FABRIC
+  connection: local
+  gather_facts: false
+  tasks:
+    - name: 'Build a containerlab topolgy'
+      import_role:
+        name: arista.avd.eos_designs_to_containerlab
+      vars:
+        mgmt_network_v4: 10.73.0.0/16
+        ceos_version: arista/ceos:4.25.2F
```
ClausHolbechArista: should this be `ceos_lab_version` ?

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+# defaults file for avd_to_containerlab
+
+# cEOS Image to use
+ceos_version: ceos:latest
+
+# Default management network IP
+mgmt_network_v4: 192.168.1.0/24
+
+# Default containerlabs configuration file generated
+containerlab_configuration: '{{ inventory_dir }}/containerlabs.yml'
```
ClausHolbechArista, line 11: should this be `containerlab_configuration_dir` ?

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/meta/main.yml
```diff
@@ -0,0 +1,52 @@
+galaxy_info:
+  author: your name
```
ClausHolbechArista: Should be updated

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/templates/topology.yml.j2
```diff
@@ -0,0 +1,73 @@
...
+
+mgmt:
+  network: 'mgmt_{{fabric_name}}'
+  ipv4_subnet: {{ mgmt_network_v4 }}
+
+{# NODES definition #}
+topology:
+  nodes:
+{% for node_type in node_types | arista.avd.natural_sort %}
+{# --------------------- NODE GROUPS LIST --------------------- #}
```
ClausHolbechArista, line 20: This approach will only work if the target for the play knows about all nodes. Since this role should run after 
`eos_designs`, I would prefer if we could read some of the generated data, instead of having to maintain the datamodel here as well.
I believe most of it should be in the structured_config including peer-switch information.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/templates/topology.yml.j2
titom73: Considering this role will be only for CI, local testing, it would be part of a specific playbook meaning the previous playbook will have to save debug files to make facts available in another play or playbook.

So user experience would be more complex than reading eos_designs inputs.

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/defaults/main.yml
titom73: Not sure as we point to a file and not a directory

ClausHolbechArista: Of course not. Sry.

ansible_collections/arista/avd/roles/eos_designs_to_containerlab/templates/topology.yml.j2
ClausHolbechArista: I meant reading structured_configuration instead. So we can change eos_designs freely without affecting this one.

titom73: Using `structured_config` files will create constraint when `eos_cli_config_gen` is similar to the current relationship with `eos_designs` and will also make rendering more complex:

- Run task for one host.
- Load `structured_config` for all hosts with a per-host approach to saving data.
- Read `structured_config` to generate file.

Where the current implementation just reads existing inventory variables.

> Note: this PR is only for the eos_designs approach

PR #1201, opened 2021-09-17
Feat(eos_designs): interface storm-control feature can be set per platform

### LINKED ISSUE #1194: ansible-avd should honor VEOS-LAB platform (not) capabilities
### Enhancement summary

Only render CLI commands (or YAML syntax in the structured configuration part) that are allowed on the target platform.

eos_designs

### Use case example

We have a 1:1 replication of our physical network in an EVE-NG environment running VEOS-LAB switches. We constantly use this lab to see hop upgrading EOS or CVP behaves before moving onto production.

While upgrading to CVP 2021.2.0 and VEOS-LAB 4.26.2F I noticed that it is no longer possible to configure the storm-control settings on an interface (it's useless in a lab anyway but we wan't to check that our generated configuration is correct).

What happened? I can no longer deploy our configuration to VEOS-LAB as storm-control is no longer supported (this was also the case with ACL and queue monitor in 4.25.4M).

With the platform type beeing available in ansible-avd l3ls design this key/value pair should be honored when rendering the structured configuration.

If a vEOS lab system is detected - the keys should not be embedded into the configuration when processing https://github.com/aristanetworks/ansible-avd/blob/790d69febb52b10ffe32968f413aa7f682ba8338/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2#L102-L111 - an additional if-expression could check for vEOS-LAB devices

Splitting the port profiles into multiple files for QA and PROD seems to be the wrong approach as we managed to single-source 98% of our codebase.

### Additional context

_No response_

### Contributing Guide

c-po:
## Change Summary
vEOS-LAB platform does not support storm-control configuration as of EOS 4.26 - remove the feature from the rendered configuration.

## Related Issue(s)

Fixes #1194

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Use same approach as for `queue_monitor_length_notify` to disable the feature on the "default" platform
```
platform_settings:
  - platforms: [ < Arista Platform Family >, < Arista Platform Family > ]
    feature_support:
      interface_storm_control: < true | false | default -> true >
```

## How to test

Use a port profile that enables storm-control:

```yaml
port_profiles:
  ##### parent port-profiles to be inherited by device related port-profiles #####
  port_profile_DOT1Q_EDGE:
    mode: trunk
    spanning_tree_bpdufilter: false
    spanning_tree_bpduguard: true
    spanning_tree_portfast: edge
    storm_control:
      broadcast:
        level: 1
        unit: percent
      multicast:
        level: 1
        unit: percent
      unknown-unicast:
        level: 1
        unit: percent
```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -88,7 +88,6 @@ platform_settings:
       non_mlag: 330
     feature_support:
       queue_monitor_length_notify: false
-      interface_storm_control: false
   - platforms: [ 7280R, 7280R2, 7500R, 7500R2 ]
     tcam_profile: vxlan-routing
     lag_hardware_only: true
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -88,6 +88,7 @@ platform_settings:
       non_mlag: 330
     feature_support:
       queue_monitor_length_notify: false
+      interface_storm_control: false
   - platforms: [ 7280R, 7280R2, 7500R, 7500R2 ]
     tcam_profile: vxlan-routing
     lag_hardware_only: true
--- a/ansible_collections/arista/avd/roles/eos_designs/doc/l3ls-evpn/common-settings.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/doc/l3ls-evpn/common-settings.md
@@ -213,6 +213,7 @@ platform_settings:
     lag_hardware_only: < true | false >
     feature_support:
       queue_monitor_length_notify: < true | false | default -> true >
+      interface_storm_control: < true | false | default -> true >
     reload_delay:
       mlag: < seconds >
       non_mlag: < seconds >
@@ -233,6 +234,7 @@ platform_settings:
       non_mlag: 330
     feature_support:
       queue_monitor_length_notify: false
+      interface_storm_control: false
   - platforms: [ 7280R, 7280R2, 7500R, 7500R2 ]
     tcam_profile: vxlan-routing
     lag_hardware_only: true
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
@@ -1,7 +1,5 @@
 {# Leaf connected_endpoint interfaces #}
 ethernet_interfaces:
-{# This must be assigned outside the loop as switch is overridden in the loop sinner scope and can no longer be referenced #}
-{% set support_storm_control = switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false) %}
 {% for endpoint_key in connected_endpoints | arista.avd.natural_sort %}
 {%     for connected_endpoint in connected_endpoints[endpoint_key] | arista.avd.natural_sort %}
 {%         for adapter in connected_endpoints[endpoint_key][connected_endpoint].adapters | arista.avd.natural_sort %}
@@ -101,7 +99,7 @@ ethernet_interfaces:
 {%                     if adapter_spanning_tree_bpduguard is arista.avd.defined %}
     spanning_tree_bpduguard: {{ adapter_spanning_tree_bpduguard }}
 {%                     endif %}
-{%                     if adapter_storm_control is arista.avd.defined and support_storm_control is arista.avd.defined(true) %}
+{%                     if adapter_storm_control is arista.avd.defined and switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false) %}
     storm_control:
 {%                         for section in adapter_storm_control %}
       {{ section }}:
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
@@ -1,5 +1,7 @@
 {# Leaf connected_endpoint interfaces #}
 ethernet_interfaces:
+{# This must be assigned outside the loop as switch is overridden in the loop sinner scope and can no longer be referenced #}
+{% set support_storm_control = switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false) %}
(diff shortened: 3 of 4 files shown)


c-po:
Hi folks (@ClausHolbechArista @titom73 @carlbuchmann ),

any implementation feedback from your side on this? As this is a blocker in our QA environment.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -88,6 +88,7 @@ platform_settings:
       non_mlag: 330
     feature_support:
       queue_monitor_length_notify: false
+      interface_storm_control: false
```
titom73: I would vote for a new platform type to support VEOS-LAB as storm-control is valid for all HW as far as I know

ansible_collections/arista/avd/roles/eos_designs/doc/l3ls-evpn/common-settings.md
```diff
@@ -213,6 +213,7 @@ platform_settings:
     lag_hardware_only: < true | false >
     feature_support:
       queue_monitor_length_notify: < true | false | default -> true >
+      interface_storm_control: < true | false | default -> true >
```
titom73, line 216: I would vote for a new platform type to support VEOS-LAB as storm-control is valid for all HW as far as I know

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -1,5 +1,7 @@
 {# Leaf connected_endpoint interfaces #}
 ethernet_interfaces:
+{# This must be assigned outside the loop as switch is overridden in the loop sinner scope and can no longer be referenced #}
+{% set support_storm_control = switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false) %}
```
titom73: I would use `true` as default to not break existing deployment unless we explicitly activate it for default platform type.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
```diff
@@ -1,5 +1,7 @@
 {# Leaf server port-channels #}
 port_channel_interfaces:
+{# This must be assigned outside the loop as switch is overridden in the loop sinner scope and can no longer be referenced #}
+{% set support_storm_control = switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false) %}
```
titom73: I would use true as default to not break existing deployment unless we explicitly activate it for default platform type.

titom73 (commented):
Just few comment about default value and behavior

ansible_collections/arista/avd/roles/eos_designs/doc/l3ls-evpn/common-settings.md
ClausHolbechArista: Well for documentation this is still correct and default is `true` so supported.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
ClausHolbechArista: I believe this logic is correct, but instead of setting a boolean up here like this, I think we should just change the logic in line 104 to:
`if adapter_storm_control is arista.avd.defined and switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false)`

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
ClausHolbechArista: I believe this logic is correct, but instead of setting a boolean up here like this, I think we should just change the logic in line 96 to:
`if adapter_storm_control is arista.avd.defined and switch.platform_settings.feature_support.interface_storm_control is not arista.avd.defined(false)`

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
ClausHolbechArista: Let us keep the [ default ] as is.
EDIT: `queue_monitor_length_notify` should stay in `default`. Only supported on R-series.
~~The default setting for queue_monitor_length_notify should also be removed with AVD 3.0. We kept it there to avoid breaking existing deployments. Let us file a separate PR for that and document the change.~~

ClausHolbechArista: And +1 that we could add a default platform for veos, disabling all these knobs.

c-po:
So I would proceed and rather add the vEOS-LAB platform as additional platform?

ClausHolbechArista:
> So I would proceed and rather add the vEOS-LAB platform as additional platform?

Just leave the defaults file untouched and we will add that extra platform in a separate PR. (already on it's way)

c-po:
@ClausHolbechArista ah okay - so it looks like I just need to stay put? I can do that!

ClausHolbechArista:
> @ClausHolbechArista ah okay - so it looks like I just need to stay put? I can do that!

Please fix the other comments. Thanks

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
ClausHolbechArista: Solved by #1216

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
ClausHolbechArista: fixed

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
ClausHolbechArista: fixed

carlbuchmann (approved):
LGTM!

c-po:
Thank you guys for renaming `switch` -> `adapter_switch` in 9c0bb3e562aa86310cd766791952daff5209ecb0 - that makey it much easier to access the `switch` key and my hack with `set ..` is obsolete.

emilarista (approved):
LGTM!

xaviramon (approved):
LGTM

Issue #1293, opened 2021-10-28
"TypeError: 'NoneType' object is not iterable" thrown when device_filter doesn't match devices


noredistribution:
### Issue Summary

Seems like if `device_filter` is set in the playbook, even when using cvp v3 modules it is not ignored so if the filter doesn't match any devices the playbook will fail with something like below:

```
TypeError: 'NoneType' object is not iterable\n", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error",
```

found some refs here: https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py

we should either completely ignore this with v3 and/or improve the error handling and print that the filter didn't match on any devices

### Which component(s) of AVD impacted

cvp_configlet_upload, eos_config_deploy_cvp

### How do you run AVD ?

Ansible CLI (with virtual-env or native python)

### Steps to reproduce

```shell
to repro simply add a `device_filter` that wouldn't match on any device, like in this playbook: https://github.com/arista-netdevops-community/ipspace-webinar-september15-2020/blob/master/playbooks/dc1-fabric-deploy-cvp.yml


- name: Configuration deployment with CVP
  hosts: AVD-CVP-1
  connection: local
  gather_facts: no
  tasks:
    - name: run CVP provisioning
      import_role:
         name: arista.avd.eos_config_deploy_cvp
      vars:
        container_root: 'TP_FABRIC'
        configlets_prefix: 'tp-avd'
        cv_collection: v3
        device_filter: 'tP-avd'
        state: present
```
```


### Relevant log output

```shell
TASK [arista.avd.eos_config_deploy_cvp : Configure devices on AVD-CVP-1] ******************************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: 'NoneType' object is not iterable
[...]

PR #1300, opened 2021-11-01
Fix(eos_config_deploy_cvp): avoid making CV_DEVICES becoming None

noredistribution:
… devices

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

## Proposed changes
in https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
if the device_filter does not match any devices inside the for loop `CVP_DEVICES` will be empty, so it'll become `None` and then on the ansible-cvp side we'll get that TypeError traceback mentioned in the #1293 

This will only set `CVP_DEVICES` to an empty list, and then validation should be done in the cv_devices_v3 module on ansible-cvp side

## How to test
To check if the list will become empty simply  print the `CVP_DEVICES` var in https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/v3/present.yml after L41
like this:
```
- name: "Print CVP_DEVICES_RESULTS"
  ansible.builtin.debug:
    var: CVP_DEVICES
```

which should result in

```
TASK [arista.avd.eos_config_deploy_cvp : Configure devices on AVD-CVP-1] **********************************************************
ok: [AVD-CVP-1]

TASK [arista.avd.eos_config_deploy_cvp : Print CVP_DEVICES_RESULTS] ***************************************************************
ok: [AVD-CVP-1] => {
    "CVP_DEVICES": []
}
```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -39,5 +44,5 @@ CVP_CONTAINERS:
 {%         for configlet in hostvars[groups[container_root][0]]['cv_configlets']['containers'][container_name] | arista.avd.natural_sort %}
       - {{ configlet }}
 {%         endfor %}
-{%      endif %}
+{%     endif %}
 {% endfor %}
```
tgodaA: New line required at the end of file.

noredistribution: done thx!

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -5,15 +5,17 @@
 {% else %}
 {%    set local_var.device_filter = device_filter %}
 {% endif %}
+{% set dev_cntr = namespace(value=0) %}
 ---
 CVP_DEVICES:
 {% for device in groups[container_root] | arista.avd.natural_sort %}
 {%     if device | arista.avd.is_in_filter(local_var.device_filter) %}
+{%         set dev_cntr.value = dev_cntr.value + 1 %}
```
tgodaA: I think it's better to move this line of code to under below if condition
```{%         if hostvars[device]['is_deployed'] is arista.avd.defined(true) or hostvars[device]['is_deployed'] is not arista.avd.defined %}```

If this condition fails then also CVP_DEVICES will be returned None.

Comment (noredistribution) [reply ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2]: yea, that's how I did it initially, not sure why I changed my mind, but that's a good point, let me move it there, thanks!

File: ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2 [hunk 3]
```diff
@@ -5,15 +5,18 @@
...
```
tgodaA: I guess you forgot to delete this line.

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -5,15 +5,17 @@
 {% else %}
 {%    set local_var.device_filter = device_filter %}
 {% endif %}
+{% set dev_cntr = namespace(value=0) %}
```
tgodaA, line 8: Space required around ```=```. value = 0

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -5,15 +5,17 @@
 {% else %}
 {%    set local_var.device_filter = device_filter %}
```
ClausHolbechArista: missing indentation (1 space). Same above.

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
ClausHolbechArista: No, according to PEP-8 this is correct. Adding spaces would be wrong.

ClausHolbechArista (dismissed):
LGTM

tgodaA (approved):
LGTM

ClausHolbechArista (approved):
LGTM - we don't need an error message, since it is not necessarily an error that no devices where ready to be deployed. So we will just parse the empty list along to the cvp modules and they will return without any action being done.

PR #1301, opened 2021-11-01
Feat(eos_cli_config_gen): Support LACP global configuration parameters

### LINKED ISSUE #1074: EVPN multihoming lacp port-id range support
**Issue Type**

- [x] Template enhancement

Add support to configure global command "lacp port-id range", which is part of the recommended configuration for EVPN multihoming:
https://www.arista.com/custom_data/downloads/?f=/support/download/DesignGuides/EVPN_Multihoming_Design_Guide.pdf

"A global configuration command must be used to configure LACP port-id range for all PEs connected to multihomed CE(s).
When the port-id range command is configured, it limits the range of allocated port-ids to the one specified.

If the CE is receiving the packet from two different ports in the same LAG (one each to a MH PE) with the same port-id, the CE device interprets (wrongly) that the port has moved from one LAG to a different LAG. Hence, configure the range for the port-id on the PE router to ensure there are no overlapping ranges for the PEs in the same ES."

CLi configuration:
```
lacp port-id range <min_value> <max_value> (1-64K)
```
Similar to RFE: #1043 

Ideally when multihoming is configured in the SERVER.yml file, also ensure separate port-id ranges are allocated to the PEs in that same ES.
 
For example, multihoming peer A (assuming two peers, and up to 127 port-channels):
```
lacp port-id range 1 127
```
Multihoming peer B:
```
lacp port-id range 128 255
```
These range could overlap in other vteps, same range applied to MH pair Leaf1/2,Leaf3/4, etc.

thanks,
David

### LINKED ISSUE #1295: Support lacp global parameters
### Enhancement summary

**Issue Type**

- [x] Template enhancement

Add support to configure lacp parameters.
- **port-id**: This range is required for recommended EVPN Multihoming configuration. Different PE(Provider Edge Router or L3Leafs) will have different port-id range.
-  **rate-limit**: To set LACPDU rate-limit options
- **system-priority**: To set local system LACP priority

CLi configuration:

```
lacp port-id range <min_value> <max_value> (1-64K)

lacp system-priority  <0-65535>
```

Similar to RFE #1074

eos_cli_config_gen

### Use case example

lacp port-id range must be defined different in EVPN MH environment.

For example

LEAF1A: 
``` lacp port-id range 1 127 ```

LEAF1B: 
``` lacp port-id range 128 255 ```

To set LACP system priority: Lower value means higher priority. The lower priority port will be in active state.

Proposed model:

```yaml
lacp:
  port_id_min: < min_port >
  port_id_max: < max_port >
  system_priority: < integer >
```

```eos_cli```

### Additional context

_No response_

### Contributing Guide

### PR DESCRIPTION
**Pull request by tgodaA**
## Change Summary

LACP global parameters like port-id and system-priority can be set with this enhancement.

## Related Issue(s)

Fixes #1295 

## Component(s) name

## Proposed changes
Added the below data-model:

```yaml
lacp:
  port_id:
    range:
      begin: < min_port >
      end: < max_port >
  rate_limit:
    default: < true | false >
  system_priority: < 0-65535 >
```

## How to test
molecule test ```lacp```

### User Checklist

- N/A

### Repository Checklist

- [ x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [ x] My change requires a change to the documentation and documentation have been updated accordingly.
- [ x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -61,6 +61,7 @@
     - [Internal VLAN Order](#internal-vlan-order)
     - [IP DHCP Relay](#ip-dhcp-relay)
     - [IP ICMP Redirect](#ip-icmp-redirect)
+    - [LACP](#lacp)
     - [LLDP](#lldp)
     - [MACsec](#macsec)
     - [Maintenance Mode](#maintenance-mode)
@@ -1197,6 +1198,19 @@ ip_dhcp_relay:
 ```yaml
 ip_icmp_redirect: < true | false >
 ipv6_icmp_redirect: < true | false >
+```
+
+### LACP
+
+```yaml
+lacp:
+  port_id:
+    range:
+      begin: < min_port >
+      end: < max_port >
+  rate_limit:
+    default: < true | false >
+  system_priority: < 0-65535 >
 ```
 
 ### LLDP
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/lacp-configuration.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/lacp-configuration.j2
@@ -1,0 +1,21 @@
+{% if lacp is arista.avd.defined %}
+
+# LACP
+
+## LACP Summary
+
+| Port-id range | Rate-limit default | System-priority |
+| ------------- | ------------------ | --------------- |
+{%     if lacp.port_id.range.begin is arista.avd.defined and lacp.port_id.range.end is arista.avd.defined %}
+{%         set row_range = lacp.port_id.range.begin ~ ' - ' ~ lacp.port_id.range.end %}
+{%     else %}
+{%         set row_range = '-' %}
+{%     endif %}
+| {{ row_range }} | {{ lacp.rate_limit.default | arista.avd.default('-') }} | {{ lacp.system_priority | arista.avd.default('-') }} |
+
+## LACP Device Configuration
+
+```eos
+{%     include 'eos/lacp.j2' %}
+```
+{% endif %}
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/lacp.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/lacp.j2
@@ -1,0 +1,15 @@
+{# eos - lacp #}
+{% if lacp is arista.avd.defined %}
+!
+{%     if lacp.port_id.range.begin is arista.avd.defined and lacp.port_id.range.end is arista.avd.defined %}
+lacp port-id range {{ lacp.port_id.range.begin }} {{ lacp.port_id.range.end }}
+{%     endif %}
+{%     if lacp.rate_limit.default is arista.avd.defined(true) %}
+lacp rate-limit default
+{%     elif lacp.rate_limit.default is arista.avd.defined(false) %}
+no lacp rate-limit default
+{%     endif %}
+{%     if lacp.system_priority is arista.avd.defined %}
+lacp system-priority {{ lacp.system_priority }}
+{%     endif %}
+{% endif %}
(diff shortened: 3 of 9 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1192,6 +1193,15 @@ ip_icmp_redirect: < true | false >
 ipv6_icmp_redirect: < true | false >
 ```
 
+### LACP
+
+```yaml
+lacp:
+  port_id_min: < min_port >
+  port_id_max: < max_port >
```
xaviramon: We should keep structure similar to EOS configuration:
```POD1_LEAF01A(config)#lacp ?
  port-id          LACP port-id range configuration
  rate-limit       Set LACPDU Rate Limit Options
  system-priority  Set local system LACP priority

POD1_LEAF01A(config)#lacp port-id ?
  range  LACP port-id range configuration

POD1_LEAF01A(config)#lacp port-id range ?
  <1-65535>  LACP port-ID range```

I suggest we ident port-id variables into port_id and then into range to allow for future extensibility if required.

Also, do we want to make rate-limit configurable?

**Review by xaviramon — CHANGES_REQUESTED** [review]
Structured should be reviewed. See comments. Compare with VLAN internal allocation policy

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md (hunk 1 above)
Comment (tgodaA) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md]: rate-limit doesn't make any difference even if I configure because it is set to default only. I don't know the reason. But I will include it.

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/lacp.j2 [hunk 2]
```diff
@@ -0,0 +1,15 @@
+{# eos - lacp #}
+{% if lacp is arista.avd.defined %}
+!
+{%     if lacp.port_id.range.begin is arista.avd.defined and lacp.port_id.range.end is arista.avd.defined %}
+lacp port-id range {{ lacp.port_id.range.begin }} {{ lacp.port_id.range.end }}
+{%     endif %}
+{%     if lacp.rate_limit.default is arista.avd.defined(false) %}
+no lacp rate-limit default
+{%     else %}
```
ClausHolbechArista: please change this into `{% elif lacp.rate_limit.default is arista.avd.defined(true) %}`
If someone just wants to set the lacp system-priority, we should not generate extra config.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1192,6 +1193,19 @@ ip_icmp_redirect: < true | false >
...
+### LACP
+
+```yaml
+lacp:
+  port_id:
+    range:
+      begin: < min_port >
+      end: < max_port >
+  rate_limit:
+    default: < true | false | default -> true >
```
ClausHolbechArista: please remove the default

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/lacp-configuration.j2
```diff
@@ -0,0 +1,25 @@
+{% if lacp is arista.avd.defined %}
+
+# LACP
+
+## LACP Summary
+
+| Port-id range | Rate-limit | System-priority |
```
ClausHolbechArista: If you change the heading to `Rate-limit default` you can just use the boolean value directly with default to `-`

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/lacp-configuration.j2
tgodaA: Rate-limit is default. We can only set it to no default.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/lacp.j2
tgodaA: https://eos.arista.com/eos-4-18-0f/incoming-lacpdu-rate-limit/, lacp rate-limit is set by default. I will only keep the no lacp rate-limit command and remove the else condition.

ClausHolbechArista (approved):
LGTM

xaviramon (approved):
LGTM

PR #1310, opened 2021-11-04
Fix(eos_cli_config_gen): Add missing case

ccsnw:
## Change Summary

add missing case for the if/else for the case that "ip_routing" is undefined. Else we miss a "!" between the blocks in the rendered output

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.<eos_cli_config_gen>`

### User Checklist

- N/A

### Repository Checklist

- [x ] My code has been rebased from devel before I start
- [x ] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.


ccsnw:
Actually I think it would make more sense to move the undefined test to the previous clause like this:

`{% elif ip_routing is arista.avd.defined(false) or ip_routing is not arista.avd.defined  %}`

and therefore set "no ip routing" in case it is not explicitly enabled.

What do you think?

ClausHolbechArista:
> Actually I think it would make more sense to move the undefined test to the previous clause like this:
> 
> `{% elif ip_routing is arista.avd.defined(false) or ip_routing is not arista.avd.defined %}`
> 
> and therefore set "no ip routing" in case it is not explicitly enabled.
> 
> What do you think?

Normally we don't go to these lengths to match the "show running-config". Sometimes `!` is added for readability, but between keeping templates simpler, readable and maintainable vs matching `show running-config` perfectly, I would go with simplicity.

ccsnw:
Thanks Clause, let me close the PR in this case. What you said makes sense as this would only be a cosmetical "fix" anyway.

PR #1315, opened 2021-11-09
Bump: Support for Ansible 2.12.0

ClausHolbechArista:
## Change Summary

Various fixes to support Ansible version 2.12.0

## Component(s) name

`arista.avd`

## Proposed changes

- [x] Update `yaml_templates_to_fact` to use `lookup_loader` for `template` plugin instead of hardcoded reference.
- [x] Update `to_nice_yaml` calls with max width to avoid changes to molecule artifacts between 2.11 and 2.12
- [x] Update CI artifacts

The issue with max width is that ansible 2.11 is hardcoded to use `pyyaml` for `to_yaml` and `to_nice_yaml` filters. In Ansible 2.12 this has been changed to use `pyyaml` only if `libyaml` is not available. One difference between the two is that `pyyaml` has a default width of 80 characters in the `dump` routines, where `libyaml` has no width set by default.
In some of our molecule scenarios we have `eos_cli` multi-line strings that are longer than 80 characters, so they wrap around when using Ansible 2.11, but not in 2.12.
To avoid molecule diffs between the two versions, we now set width statically to max value (2147483647).

## How to test

Molecule
Small test scenario to compare the different versions. See [mail to Ansible-project group](https://groups.google.com/g/ansible-project/c/1iFQx53RBtk) for details.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/templates/debug/generate-debug-vars.j2
```diff
@@ -1 +1 @@
-{{ hostvars[inventory_hostname] | to_nice_yaml(indent=2, sort_keys=False) }}
\ No newline at end of file
+{{ hostvars[inventory_hostname] | to_nice_yaml(indent=2, sort_keys=False, width=2147483647) }}
```
tgodaA: Missing new line

ansible_collections/arista/avd/molecule/eos_designs-twodc-5stage-clos/intended/structured_configs/DC1-POD1-L2LEAF2A.yml
```diff
@@ -31,9 +31,7 @@ management_api_http:
   enable_vrfs:
     MGMT: {}
   enable_https: true
-eos_cli: "interface Loopback1002\n  description Loopback created from raw_eos_cli\
-  \ under l2leaf node-group RACK2_MLAG\n\ninterface Loopback1111\n  description Loopback\
-  \ created from raw_eos_cli under platform_settings vEOS-LAB\n"
+eos_cli: "interface Loopback1002\n  description Loopback created from raw_eos_cli under l2leaf node-group RACK2_MLAG\n\ninterface Loopback1111\n  description Loopback created from raw_eos_cli under platform_settings vEOS-LAB\n"
```
tgodaA, line 34: Would it be ok to use literal block scalars for such commands as below:

```yaml
eos_cli: |
         interface Loopback1002
           description Loopback created from raw_eos_cli under l2leaf node-group RACK2_MLAG
         
         interface Loopback1111
           description Loopback created from raw_eos_cli under platform_settings vEOS-LAB
```

ClausHolbechArista: We do use block scalars in the input data, but in the structured_config this is converted into a regular string. We tried many different approaches and this was the most robust one. Also this is the standard output of the `to_nice_yaml` filter.

carlbuchmann (approved):
LGTM!

PR #1335, opened 2021-11-16
Fix(eos_cli_config_gen): Port-channel can use LACP fallback configuration on physical interface

### LINKED ISSUE #1279: Bug L2 interfaces & Channel_Groups set wtih lacp_fallback
### Issue Summary

Link #542 given the relation.

When channel_groups are set, no other L2 config is applied which is useful to eliminate L2 redudant config.  The problem is when you set LACP fallback individual on the bond, you need to be able to set config parameters on the port.

Why do people need to set fallback?  PIXE Boot...

eos_cli_config_gen

### How do you run AVD ?

Ansible AWX

### Steps to reproduce

```shell
set channel_group on an interface & lacp_fallback on its associated Port-Channel.
```

### Relevant log output

_No response_

### Contributing Guide

xaviramon:
## Change Summary

In order to use LACP fallback feature, it iw requested that physical interface of a Port-channel has specific and different configuration than the PC itself.
Under this PR, a new validation is set under the ethernet interfaces where a flag is used to decide if configuration under the physical interface must be rendered or not. 

## Related Issue(s)

Fixes #1279 

## Component(s) name

## Proposed changes
The flag checks if the interface is a Port-channel. If not, flag is activated and configuration is rendered. 
If the interface is a Port-channel, it is verified if LACP individual mode is defined. In this case, flag is activated and configuration will be rendered. For normal PC with no LACP fallback individual, flag is not activated and configuration is not rendered.
This is the way that introduces less changes while keeping the previous behaviour.

## How to test
Molecule test included

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -32,7 +32,15 @@ interface {{ ethernet_interface }}
 {%         if ethernet_interfaces[ethernet_interface].lacp_timer.multiplier is arista.avd.defined %}
    lacp timer multiplier {{ ethernet_interfaces[ethernet_interface].lacp_timer.multiplier }}
 {%         endif %}
+{%     endif %}
+{%     if ethernet_interfaces[ethernet_interface].channel_group.id is arista.avd.defined and ethernet_interfaces[ethernet_interface].channel_group.mode is arista.avd.defined %}
+{%         if port_channel_interfaces['Port-Channel' ~ ethernet_interfaces[ethernet_interface].channel_group.id].lacp_fallback_mode is arista.avd.defined("individual") %}
+{%             set print_ethernet = true %}
```
carlbuchmann, line 36: @xaviramon - I really like this approach, but I believe it is still too much abstraction for`eos_cli_config_gen`.
IMO, we should remove the conditional for port-channel, and do this type of logic in `eos_designs` instead.

Let's discuss with @ClausHolbechArista before we do any changes.

xaviramon: In my opinion, roles should be independent, unless we need to combine them to make the EVPN AA functionality work together.
For this, i believe, we should support this in eos_config_cli

ClausHolbechArista: I have tested a few other workarounds, but because of all the logic we already have inside `ethernet_interfaces.j2` they all break one or the other way. I would have preferred a logic-free template, so what you set is what you get. But getting there would be a breaking change.
So if we accept to have this logic in here - to avoid breaking changes - we are down to deciding how the logic should be implemented.
Line 35 and 36 can be removed since the `if` block ended by line 35 contains the same logic as line 36. The rest is fine for me.

ClausHolbechArista (approved):
LGTM

tgodaA (approved):
LGTM!

PR #1341, opened 2021-11-17
Feat(eos_cli_config_gen): Refactor STP template and support STP root primary/secondary

### LINKED ISSUE #1329: Support for stp root primary/secondary commands
### Enhancement summary

Asking to add support for the following STP CLI commands:

```
spanning-tree root primary/secondary
spanning-tree mst <0-4094> root primary/seconday
```
Today it is possible to set the priority manually to a value but it would be useful to add support also for these commands:
https://avd.sh/en/releases-v3.x.x/roles/eos_cli_config_gen/index.html#spanning-tree

eos_cli_config_gen

### Use case example

I would like to set these commands for an active-standby scenario and align the active PE with the primary STP root without the need of setting the priority explicitly.

Something along these lines:
```
spanning_tree:
  root: < primary | secondary >
  mst:
     root: < primary | secondary >
```

Setting the priority value is an alternative and it is supported.

### Additional context

_No response_

### Contributing Guide

tgodaA:
## Change Summary

Support STP root primary or secondary commands and refactor stp template according to cli.

## Related Issue(s)

Fixes #1329 

## Component(s) name

## Proposed changes

Refactor template output according to cli

Old output:

```
spanning-tree root super
spanning-tree mode mstp
no spanning-tree vlan-id 4094
```

New output:

```
spanning-tree mode mstp
no spanning-tree vlan-id 4094
spanning-tree root super
```

Added new knobs for spanning tree to support root primary/secondary:

Old model:

```
spanning_tree:
  mst_instances:
    # Use '-' to specify range of instance ids, like 200-210
    "< instance_id(s) >":
      priority: < priority >
```

```
spanning_tree:
  root: < primary | secondary >
  mst_instances:
    # Use '-' to specify range of instance ids, like 200-210
    "< instance_id(s) >":
      priority: < priority >
      root: < primary | secondary >
```

## How to test
Molecule

## Checklist

- [x] Update artifacts for the refactored template
- [x] Verify the STP output configuration over cli
- [x] Update documentation  

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/spanning-tree.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/spanning-tree.j2
@@ -10,22 +10,40 @@ STP mode: **{{ spanning_tree.mode | arista.avd.default("mstp") }}**
 
 STP Root Super: **{{ spanning_tree.root_super }}**
 {%     endif %}
+{%     if spanning_tree.root is arista.avd.defined %}
+
+STP Root: **{{ spanning_tree.root }}**
+{%     endif %}
 {%     if spanning_tree.mst_instances is arista.avd.defined %}
 
 ### MSTP Instance and Priority
 
 | Instance(s) | Priority |
-| -------- | -------- |
+| ----------- | -------- |
+{%         set root_instances = [] %}
 {%         for mst_instance in spanning_tree.mst_instances | arista.avd.natural_sort %}
 | {{ mst_instance }} | {{ spanning_tree.mst_instances[mst_instance].priority | arista.avd.default("-") }} |
+{%             if spanning_tree.mst_instances[mst_instance].root is arista.avd.defined %}
+{%                 do root_instances.append(mst_instance) %}
+{%             endif %}
 {%         endfor %}
+{%         if root_instances | length > 0 %}
+
+### MSTP Instance and Root
+
+| Instance(s) | Root |
+| ----------- | ---- |
+{%             for mst_instance in root_instances %}
+| {{ mst_instance }} | {{ spanning_tree.mst_instances[mst_instance].root }} |
+{%             endfor %}
+{%         endif %}
 {%     endif %}
 {%     if spanning_tree.mst.configuration is arista.avd.defined %}
 
 ### MST Configuration
 
 | Variable | Value |
-| -------- | -------- |
+| -------- | ----- |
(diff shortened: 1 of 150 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/spanning-tree.j2
```diff
@@ -10,22 +10,40 @@ STP mode: **{{ spanning_tree.mode | arista.avd.default("mstp") }}**
 
 STP Root Super: **{{ spanning_tree.root_super }}**
 {%     endif %}
+{%     if spanning_tree.root is arista.avd.defined %}
+
+STP Root: **{{ spanning_tree.root }}**
+{%     endif %}
 {%     if spanning_tree.mst_instances is arista.avd.defined %}
 
 ### MSTP Instance and Priority
```
ClausHolbechArista, line 19: I would suggest changing this heading to `MSTP Instances` and then combine the two tables below, so it shows priority if set, and check for root primary/secondary if priority is not set. (Like done in config template)

ClausHolbechArista: Maybe add the instance `vlans` to the table while we are at it. (move from below in another table)

tgodaA: Done.

ClausHolbechArista (commented):
I don't think we should support these knobs, since they are actually aliases to the priority commands in EOS:
```cli
R3(config)#show run | sec spanning-tree
spanning-tree mode mstp
R3(config)#spanning-tree root primary
R3(config)#show run | sec spanning-tree
spanning-tree mode mstp
spanning-tree mst 0 priority 8192
R3(config)#
```

tgodaA:
Yeah, these are aliases to priority. But adding primary or secondary commands to EOS makes it easy for the user. In background they convert to priority. Primary - 8192, Secondary 16384. Also I didn't add the converted priority to the documentation but added only if it's primary or secondary because the default value might change anytime.

xaviramon:
I don't like the idea of root primary or secondary. Do we have a use case or special need for it? I would prefer to set the STP priority manually, also because different manufacturers could use a different alias with a different priority.

tgodaA:
I am waiting for suggestion from Claus. If this doesn't look good, please close it or comment to close it.

tgodaA:
The same feat can achieved by using priority. So not adding this feature.

PR #1346, opened 2021-11-18
Feat(eos_cli_config_gen): Enhance vlan_interfaces commands (arp_*)

ccsnw:
## Change Summary

adding

```    
vlan_interfaces:
  <snip>
  arp_cache_dynamic_capacity: < 0-4294967295 >
  arp_gratuitous_accept: < true | false >
  arp_monitor_mac_address: < true | false >
 <snip>
```

## Related Issue(s)

N/A

Just noticed that arp_gratuitous_accept is not explicitly defined and though of adding it, because its not so uncommon to use.

## Component(s) name

## Proposed changes
Added several arp related commands under vlan_interfaces.

## How to test
molecule run

### User Checklist

- N/A

### Repository Checklist


ccsnw:
Does a section for the arp related commands  per interface make sense in the documentation, or would it be overhead?

Other option would be to add some key information, e.g. if gratuitious arp is accepted to the "VLAN Interfaces Summary" table?

ClausHolbechArista:
> Does a section for the arp related commands per interface make sense in the documentation, or would it be overhead?
> 
> Other option would be to add some key information, e.g. if gratuitious arp is accepted to the "VLAN Interfaces Summary" table?

I think these knobs are so "minor" that we should exempt them from documentation. They will still show up in the config snippets, but not be in any tables. @carlbuchmann @titom73 agree?

xaviramon:
I would like to have a molecule configuration generated with all 3 variables you are creating: arp_cache_dynamic_capacity, arp_gratuitous_accept and arp_monitor_mac_address, so we can monitor any change on any of them with molecule automatically.

ccsnw:
added! @xaviramon 
also fixed a typo in one of the commands

carlbuchmann:
> 

Yes I agree with @ClausHolbechArista, all the various minor knobs don't need to be in the documentation table. That is why we have the config snippets!

xaviramon:
Just a last comment before approving. We render the configuration the same way EOS will do it. Please, change the order of configuration render to:
```
arp gratuitous accept
arp monitor mac-address
arp cache dynamic capacity <number>
```
Thank you!

ccsnw:
order changed @xaviramon

xaviramon (approved):
LGTM!

carlbuchmann (approved):
LGTM!

PR #1350, opened 2021-11-22
Feat(eos_designs): Add support for network_services_keys

### LINKED ISSUE #1021: Allow tenants to be defined similar to connected_endpoints
**Issue Type**

- [ ] Template enhancement
- [X] Role enhancement
- [ ] Documentation enhancement (avd.sh)

Currently all tenants are defined under a single tenants variable which makes splitting up tenants into multiple files impossible due to ansible not allowing duplicate vars. 

This could be solved similar to how connected_endpoints is done, by allowing us to define list of tenants under a tenant variable and then define them as a top level variable.

Don't think there are any other than actually making it.

**Additional context**

melkypie:
## Change Summary

Add support for network_services_keys

Switches from using the static `tenants` key for defining network services to a user-defined key name. We keep `tenants` as the default key name.

This makes the behavior of tenants similar to how connected_endpoints can be defined and also allows splitting each tenant into a separate file rather than keeping them all in one.

## Related Issue(s)

Fixes #1021

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

```yaml
network_services_keys:
  - name: <key_1>
  - name: <key_2>
# default
network_services_keys:
  - name: "tenants"
```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -131,6 +131,9 @@ platform_settings:
 
 connected_endpoints_keys: "{{ default_connected_endpoints_keys[design.type] }}"
 
+network_services_keys:
+  - name: tenants
+
 #custom_structured_configuration_prefix allows the user to customize the prefix for Custom Structured Configuration variables.
 custom_structured_configuration_prefix: custom_structured_configuration_
 custom_structured_configuration_list_merge: 'replace'
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -138,7 +138,6 @@ network_services_keys:
 custom_structured_configuration_prefix: custom_structured_configuration_
 custom_structured_configuration_list_merge: 'replace'
 
-
 node_type_keys: "{{ default_node_type_keys[design.type] }}"
 
 templates: "{{ default_templates[design.type] }}"
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
@@ -446,61 +446,64 @@ switch:
 {% if switch.network_services_l2 is arista.avd.defined(true) or
       switch.network_services_l3 is arista.avd.defined(true) or
       switch.network_services_l1 is arista.avd.defined(true) %}
-{%     set tenants = dict() %}
-{%     include 'network_services/logic.j2' %}
 {%     set leaf = namespace() %}
 {%     set leaf.vlans = []  %}
 {%     set leaf.tenants = {} %}
 {%     set leaf.pseudowires = [] %}
-{%     for tenant in tenants | arista.avd.natural_sort if tenant in switch_filter_tenants or "all" in switch_filter_tenants %}
-{%         set add_vrfs = {} %}
-{%         for vrf in tenants[tenant].vrfs | arista.avd.natural_sort %}
-{%             set add_svis = [] %}
-{%             for svi in tenants[tenant].vrfs[vrf].svis | arista.avd.natural_sort %}
-{%                 for svi_tag in tenants[tenant].vrfs[vrf].svis[svi].tags | arista.avd.natural_sort %}
-{%                     if svi_tag in switch_filter_tags or switch_data.group is arista.avd.defined(svi_tag) or "all" in switch_filter_tags %}
-{%                         do add_svis.append(svi) %}
-{%                         do leaf.vlans.append(svi | int) %}
-{%                         break %}
-{%                     endif %}
-{%                 endfor %}
-{%             endfor %}
+{%     set tmp_hostvars = hostvars[inventory_hostname] %}
+{%     for network_services_key in network_services_keys | arista.avd.natural_sort('name') %}
+{%         if network_services_key.name is arista.avd.defined and tmp_hostvars[network_services_key.name] is arista.avd.defined %}
+{%             for tenant in tmp_hostvars[network_services_key.name] | arista.avd.convert_dicts | arista.avd.natural_sort('name') if tenant.name in switch_filter_tenants or "all" in switch_filter_tenants %}
+{%                 set add_vrfs = {} %}
+{%                 for vrf in tenant.vrfs | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
+{%                     set add_svis = [] %}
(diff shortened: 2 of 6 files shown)


ClausHolbechArista (commented):
This change would make all existing deployments fail. I don't think we should define the tenant names twice.
What we could do is to allow the user to change the key name instead of `tenants`. So the outer dict containing other tenants. We could even allow them to supply multiple keys containing multiple tenants.
Ex:
```yaml
tenants_keys:
  - name: dc1_tenants
  - name: dc2_tenants
  - name: dc_tenants
dc1_tenants:
  tenant_a:
    <...>
  tenant_b:
    <...>
dc2_tenants:
  tenant_c:
    <...>
  tenant_d:
    <...>
dc_tenants:
  tenant_e:
    <...>
  tenant_f:
    <...>
```

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -121,6 +121,9 @@ connected_endpoints_keys:
   storage_arrays:
     type: storage_array
 
+tenants_keys:
```
ClausHolbechArista: sorry I didn't catch this before. Can you make it `network_services_keys` ? Thanks.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
```diff
@@ -0,0 +1,7 @@
+{% for network_services_key in network_services_keys | arista.avd.natural_sort %}
+{%     if hostvars[inventory_hostname][network_services_key.name] is arista.avd.defined %}
```
ClausHolbechArista: from tests with large scale inventories, we have learned that using `hostvars` should be avoided, since jinja copies the entire `hostvars` dataset between each call to the underlying code. In our tests it improved the speed a lot if we copy the part we need into a temporary variable.
Ex.
`{% set tmp_network_services = hostvars[inventory_hostname][network_services_key.name] | arista.avd.default({}) %}`
And then reference the temporary variable below.
I think it should be possible to do a single update with `{% do tenants.update(tmp_network_services) %}`

ClausHolbechArista (commented):
I am worried about how this change will affect the execution time with scaled inventories (10000+ services). It could be "expensive" to run the logic twice. Maybe we could run the logic as part of facts, and output only the relevant subset into a new `switch.network_services` fact. We know which are "relevant" when we add to the simplified `switch.tenants` in line 374 of `facts/switch/switch.j2`.
But I am not sure what is most time consuming. Adding all this extra data to facts or parsing the logic twice.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
melkypie: This should probably be done for connected_endpoints_keys also then. Since this implementation is pretty close to what connected_endpoints_keys does. Although the improvement might be not that significant as it does not have this second loop.

ClausHolbechArista: Correct, this was discovered recently, and we will need to refactor all templates where we use potential large data sets.
We plan to change the data models to lists of dicts instead of dicts of dicts. This will give us improvements on speed, since we don't need to address the complete parent dict every time we need a single child key. So all new models will be built with this in mind, and for 4.0 (subject to change :) ) we will try to do the conversion.
The conversion is also part of the work with schema-based input validation and documentation.

ClausHolbechArista: @melkypie I have implemented the changes to handle future list-of-dicts as well as `network_serivces_keys`. It took a while, but I hope you can test this out, so we can get it merged before our planned release on Friday.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/main.j2
```diff
@@ -12,6 +12,13 @@
 
 {% endif %}
 
+{% if switch.vtep is arista.avd.defined(true) or
```
emilarista: Does this add anything that the stuff at 56/63 doesn't already cover except the "set tenants = dict()"?

ClausHolbechArista: it was a miss from rebase. I will remove this. Thank you.

ClausHolbechArista: Fixed

emilarista (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -450,55 +450,60 @@ switch:
...
+{%     set tmp_hostvars = hostvars[inventory_hostname] %}
+{%     for network_services_key in network_services_keys | arista.avd.natural_sort('name') %}
+{%         if network_services_key.name is arista.avd.defined and tmp_hostvars[network_services_key.name] is arista.avd.defined %}
+{%             for tenant in tmp_hostvars[network_services_key.name] | arista.avd.convert_dicts | arista.avd.natural_sort('name') if tenant.name in switch_filter_tenants or "all" in switch_filter_tenants %}
+{%                 set add_vrfs = {} %}
+{%                 for vrf in tenant.vrfs | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
+{%                     set add_svis = [] %}
+{%                     for svi in vrf.svis | arista.avd.convert_dicts('id') | arista.avd.natural_sort('id') %}
+{%                         for svi_tag in svi.tags | arista.avd.natural_sort %}
+{%                             if svi_tag in switch_filter_tags or switch_data.group is arista.avd.defined(svi_tag) or "all" in switch_filter_tags %}
```
tgodaA, line 462: Looks good to me. Just curious to know if we can use ```switch_filter_tags is arista.avd.contains([svi_tag, 'all'])``` instead of ```svi_tag in switch_filter_tags``` and ```"all" in switch_filter_tags```

tgodaA: Same for line 456, ```if tenant.name in switch_filter_tenants or "all" in switch_filter_tenants``` and line 487.

tgodaA (approved):
LGTM

tgodaA:
I have a comment but still approved because they are minor suggestions, that could improve code readability.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
ClausHolbechArista: Those are part of another PR, so this will come later

PR #1149, opened 2021-08-16
Feat(eos_designs): Add support for setting VXLAN Source interface

lermilov:
## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes)
- [ ] Documentation content changes
- [ ] Other (please describe):

## Related Issue(s)

Fixes #704

## Component(s) name

`arista.avd.eos-designs`

## Proposed changes
Introducing new variable "vtep_loopback" with default value "Loopback1".
If "vtep_loopback" is not set then there will be no change to existing configurations, i.e. Loopback1.

## Checklist:

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly
- [ ] All new and existing tests passed ([`pre-commit`](https://www.avd.sh/docs/installation/development/#python-virtual-environment), `make linting` and `make sanity-lint`).

PR #1360, opened 2021-11-25
Fix(eos_validate_state): Abstraction of VXLAN source interface

### LINKED ISSUE #1284: eos_validate_state not accounting for vtep_loopback in eos_designs
### Issue Summary

Since the introduction of #1149 that allows you to set custom VTEP loopback id, the `eos_validate_state` role still uses hardcoded Loopback1 as the VTEP loopback while it should use the vtep_loopback id defined in eos_designs.

eos_validate_state

### How do you run AVD ?

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Contributing Guide

xaviramon:
## Change Summary

Under PR#1149 (https://github.com/aristanetworks/ansible-avd/pull/1149) support for user defined Loopback interface was added. It was forgotten to add the support on the eos_validate_state, where old hardcoded interface (loopback1) was left. 
This PR solves this issue by using the vxlan1 interface source interface as the VTEP IP.

## Related Issue(s)

Fixes #1284

## Component(s) name

## Proposed changes
Tests using VTEP IPs are not hardcoded to use Loopback1 interface anymore. The test is changed and the IP interface used for the test is change to take the IP of the loopback which is described in the VXLAN tunnel as source interface.

## How to test
Difficult, as your own environment needs to be used to run the validation role.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_validate_state/tasks/router_id_reachability.yml
```diff
@@ -0,0 +1,56 @@
+---
+
+- name: Gather ip reachability state between devices (Router ID <-> Router ID)
```
ClausHolbechArista: IMO we should call this `loopback_reachability`, since router-id is not strictly an IP, but just a 32-bit number.
We could pick up the first loopback interface in `loopback_interfaces` as long as it is _not_ the VTEP.
If we need to add another variable to structured_config, we can do this - we just need to document it and comment that it is only used for validation role. (and maybe documentation?)

xaviramon: Since this test is very specific for EVPN, what if we take the ``update_source`` variable from ``EVPN-OVERLAY-PEERS``? This makes the test non-reusable for L2LS or L3LS but as said, I think it is not needed.

ansible_collections/arista/avd/roles/eos_validate_state/tasks/loopback_reachability.yml
```diff
@@ -0,0 +1,56 @@
+---
+
+- name: Gather ip reachability state between devices (Router ID <-> Router ID)
+  eos_command:
+    commands: "ping {{ loopback_ip_address | ansible.netcommon.ipaddr('address') }} source {{ loopback_interfaces[router_bgp.peer_groups[\"EVPN-OVERLAY-PEERS\"].update_source].ip_address | ansible.netcommon.ipaddr('address') }} repeat 1"
```
ClausHolbechArista: Points for creativity, but this will not work. The peer group name is configurable.
I really think we should just add a variable for this

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_vars_for_testing.j2
```diff
@@ -1,23 +1,19 @@
-{# loopback0 reachability test #}
+{# Loopback reachability test #}
 {% set address = [] %}
 {% for node in groups[fabric_name] | arista.avd.natural_sort %}
-{%     if hostvars[node].type == 'l3leaf' %}
-{%         if hostvars[node].loopback_interfaces.Loopback0.ip_address is defined and hostvars[node].loopback_interfaces.Loopback0.ip_address is not none %}
-{%             do address.append(hostvars[node].loopback_interfaces.Loopback0.ip_address | ansible.netcommon.ipaddr('address')) %}
-{%         endif %}
+{%     if hostvars[node].router_bgp.peer_groups["EVPN-OVERLAY-PEERS"].update_source is arista.avd.defined %}
```
tgodaA: Looks good to me.

Because you are accessing ```hostvars[node].router_bgp.peer_groups["EVPN-OVERLAY-PEERS"].update_source``` more than once, please set it to a temporary variable. Because accessing hostvars multiple times will increase the run-time of the code.

Or alternatively set node_vars = hostvars[node] and then access different values using the temporary variable ```node_vars```. We used the similar technique in the recent refactor.

Same for below variables.

tgodaA:
Single commit semantic error!

ansible_collections/arista/avd/roles/eos_validate_state/tasks/routing_table.yml
```diff
@@ -1,31 +1,30 @@
...
   eos_command:
-    commands: "show ip route {{ loopback1_address }}"
-  loop: "{{ loopback1_reachability.loopback1_range }}"
+    commands: "show ip route {{ vtep_ip }}"
+  loop: "{{ vtep_reachability.vtep_ip_reachability }}"
   loop_control:
-    loop_var: loopback1_address
+    loop_var: vtep_ip
   ignore_errors: "{{ eos_validate_state_validation_mode_loose }}"
   when: |
```
tgodaA, line 10: Since there is only one logic check, can we write this as:

``` when: vtep_reachability.vtep_ip_reachability is arista.avd.defined```

tgodaA (approved):
LGTM!

ClausHolbechArista (approved):
LGTM

PR #1366, opened 2021-11-27
Feat(eos_cli_config_gen): Add additional router_bgp options

### LINKED ISSUE #1289: Feature request: router_bgp allowas-in missing features in eos_cli_config_gen
### Enhancement summary

allowas-in options are missing for router_bgp neighbors, peer-groups, etc

eos_cli_config_gen

### Use case example

```
router bgp 65000
   neighbor test peer group
   neighbor test allowas-in 1
   no neighbor 10.10.10.10 allowas-in 10
```

```yaml
router_bgp:
  peer_groups:
    test:
         allowas_in:  < enabled | disabled | 1-10 | default (enabled) >
  neighbors:
    10.10.10.10:
         allowas_in:  < enabled | disabled | 1-10 | default (enabled) >
```

Do not enable EOS feature parity

### Additional context

_No response_

### Contributing Guide

ccsnw:
## Change Summary

Adding additional options for router_bgp

## Related Issue(s)

Fixes #1289

## Component(s) name

## Proposed changes
Adding support for:
- Explicitly defining: bgp wait-for-convergence
- Explicitly defining: wait-update
- Explicitly defining: bgp distance
- Explicitly defining: bgp maximum-paths
- Extend Maximum Routes with Limit and Warning Only
- Add support for "allowas-in"
- Add support for "no redistribute" in vlans/vlan aware bundles

```
router_bgp:
<snip>
  distance:
    external_routes: < 1-255 >
    internal_routes: < 1-255 >
    local_routes: < 1-255 >
  maximum_paths:
    paths: < 1-600 >
    ecmp: < 1-600 >
  updates:
    wait_for_convergence: < true | false >
    wait_install: < true | false >
<snip>
  peer_groups/neighbors/vrf:
      maximum_routes: < integer >
      maximum_routes_warning_limit: < "<integer>" | "<0-100> percent" >
      maximum_routes_warning_only: < true | false >
      allowas_in:
          enabled: < true | false >
          times: < 1-10 >
<snip>
  vlans/vlanawarebundles:
      no_redistribute_routes:
        - < host-route >
 <snip>
```

## How to test
test using molecule once working again

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -90,8 +113,8 @@
 
 ### BGP Neighbors
 
-| Neighbor | Remote AS | VRF | Send-community | Maximum-routes |
-| -------- | --------- | --- | -------------- | -------------- |
+| Neighbor | Remote AS | VRF | Send-community | Maximum-routes | Maximum-routes warning only |
```
ClausHolbechArista: Instead of adding this column everywhere, for a config that is not too common, I would suggest to add it to the Maximum-routes column.
Example:
```md
| Maximum-routes |
| 1000 (warning-only, warning-limit 900) |
| 1000000 (warning-limit 90%) |
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -219,31 +250,33 @@
 
 ##### VLAN aware bundles
 
-| VLAN Aware Bundle | Route-Distinguisher | Both Route-Target | Import Route Target | Export Route-Target | Redistribute | VLANs |
-| ----------------- | ------------------- | ----------------- | ------------------- | ------------------- | ------------ | ----- |
+| VLAN Aware Bundle | Route-Distinguisher | Both Route-Target | Import Route Target | Export Route-Target | Redistribute | No Redistribute | VLANs |
```
ClausHolbechArista: Instead of adding another column for a not too common setting, I would suggest to reuse the `Redistribute` column.
Example:
```md
| Redistribute |
| static<br>learned<br>no host-routes |
```
@carlbuchmann what do you think?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2152,6 +2152,16 @@ router_general:
...
   as: < bgp_as >
   router_id: < IPv4_address >
+  distance:
+    external_routes: < 1-255 >
+    internal_routes: < 1-255 >
+    local_routes: < 1-255 >
+  maximum_paths:
+    ec_paths: < 1-600 >
+    ecmp_routes: < 1-600 >
+  updates:
```
ClausHolbechArista, line 2226: I think we should use the singular form like it is in CLI.

ClausHolbechArista (commented):
Please start by taking a look at the extra columns.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: Hi Claus, unfortunately "update" is a reserved keyword.. so can't be used as key (if there is no workaround for that). 

For the paths it is possible to switch to singular.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-base.yml
```diff
@@ -1,12 +1,20 @@
 router_bgp:
   as: 65101
   router_id: 192.168.255.3
+  update:
```
tgodaA: This keyword is not consistent with your jinja template.

ccsnw: Thanks and fixed

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: Claus, I've just doublechecked. EOS CLI also says "maximum_path**s**". How about changing to this:

```
maximum_paths:
  ec: < 1-600 >
  ecmp: < 1-600 >
updates:
```

Regarding the key "updates", I'm still unsure. Can it remain plural, as "update" is reserved keyword, or should we rather pick a completely different term?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
ccsnw: changed accordingly

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
tgodaA: I have a hack. We can use ```update``` knob but while retrieving data we should not use ```.``` method (like router_bgp.update) but retrieve data by router_bgp['update']. I know this against our rules but this works and we avoid plural form.

tgodaA:
Hi Nico, please check your PR, Repository Checklist. You have uploaded all of the artifacts but didn't check the Molecule artifacts checkbox.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2226,6 +2236,12 @@ router_bgp:
       password: "< encrypted_password >"
       send_community: < standard | extended | large | all >
       maximum_routes: < integer >
+      maximum_routes_warning_limit: < integer >
```
tgodaA: This model is too static. What we can do was:

maximum_routes_warning_limit: < < integer > | < percent > "percent" >
Let the user specify this as either 10 or 10 percent. This way we can eliminate the knob ```maximum_routes_warning_limit_percent```. Also if there is any future update to specify different format that can also be supported.

Same for the below config.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2226,6 +2236,12 @@ router_bgp:
...
+      maximum_routes_warning_limit_percent: < true | false >
+      maximum_routes_warning_only: < true | false >
+      allowas_in:
```
tgodaA, line 2242: Make this knob ```allow_as_in``` for easy to understand.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2226,6 +2236,12 @@ router_bgp:
...
+          enabled: < true | false >
+          allow_local_asn_n_times: < 1-10 >
```
tgodaA: Can you make this ```allowed_local_asns``` or please wait for others to comment.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: Since this will cause all kinds of issues in templates or inline jinja, we should just keep the "updates" as it is now.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: We already had many goes at this key name. I'm not sure your's is more precise than the one in the PR. I also tried a few in the issue discussion, but never found a good one.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: So the proposal is to change to a string: `< "<integer>" | "<0-100> percent" >`.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-base.cfg
```diff
@@ -13,11 +13,14 @@ interface Management1
    ip address 10.73.255.122/24
 !
 router bgp 65101
-   no bgp default ipv4-unicast
```
ClausHolbechArista, line 16: Why is this line getting removed?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: It's tricky, if the goal is to make the key shorter, we could to something like

```
allowas_in:
  times: < 1-10 >
```

reads like: 

```
allow(local)as in < 1-10> times
```

Somehow I don't like to have "local asn**s**"(with plural s), because there is only one local ASN from the logical point of view.. it's just repeated n-times.

What do you think?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: My goal was to follow the CLI as much as possible,  so either allow the value in percent or integer. In order to not mix types in the value (integer | string) I've introduced the key 

```
maximum_routes_warning_limit_percent: < true | false >
```

would it make sense to convert this into something like 

```
maximum_routes_warning_limit_unit: < percent | integer >
```

This would also allow for different units if they get introduced in the future. Or is it better to just take an arbitrary string like proposed?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: @tgodaA yup, using bracket notation was mentioned in the Ansible doc as well. Thinks its better to adjust the key though, otherwise there is bracket notation just for this case and nobody will remember why in like 3 months :-D

ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-base.cfg
ccsnw: Hi Claus,

I don't think it's removed, just the order changes slightly:

```
router bgp 65101
   router-id 192.168.255.3
   distance bgp 20 200 200
   maximum-paths 32 ecmp 32
   update wait-for-convergence
   update wait-install
   **no bgp default ipv4-unicast**
   graceful-restart restart-time 300
   graceful-restart
```

due to how I arranged it.

In the current code, most of these commands are kept under a single key:

```
  bgp_defaults:
    - no bgp default ipv4-unicast
    - distance bgp 20 200 200
    - graceful-restart restart-time 300
    - graceful-restart
    - maximum-paths 2 ecmp 2
 ```

and I've removed a couple of them and defined them explicitly instead, so they appear on top now..

Comment (ccsnw) [reply ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-base.cfg:16]: can we keep it like that, or do I need to adjust the order?

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md (hunk 7 above)
Comment (tgodaA) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md]: Yeah, looks good to me. Either ```times``` or may be ```repeat``` works for me.

**Comment by ccsnw** [comment]
> Hi Nico, please check your PR, Repository Checklist. You have uploaded all of the artifacts but didn't check the Molecule artifacts checkbox.

Done Tony, had this PR as draft without the artifcats and forgot to check the box after changing the PRs state

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md (hunk 7 above)
Comment (ccsnw) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md]: changed to:

```
      allowas_in:
          enabled: < true | false >
          times: < 1-10 >
```

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md (hunk 5 above)
Comment (ccsnw) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md]: Changed model to:

```
      maximum_routes: < integer >
      maximum_routes_warning_limit: < "<integer>" | "<0-100> percent" >
      maximum_routes_warning_only: < true | false >
```

If new formats get added, we'll just have to update the documentation accordingly, like Tony mentioned.

File: ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-base.cfg (hunk 8 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-base.cfg:16]: My bad, I missed it.

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md [hunk 9]
```diff
@@ -2205,6 +2205,16 @@ router_general:
 router_bgp:
   as: < bgp_as >
   router_id: < IPv4_address >
+  distance:
+    external_routes: < 1-255 >
+    internal_routes: < 1-255 >
+    local_routes: < 1-255 >
+  maximum_paths:
+    ec: < 1-600 >
```
ClausHolbechArista: This key should be named `paths`. Sorry for not catching this earlier.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2227,6 +2237,11 @@ router_bgp:
       password: "< encrypted_password >"
       send_community: < standard | extended | large | all >
       maximum_routes: < integer >
+      maximum_routes_warning_limit: < "<integer>" | "<0-100> percent" >
+      maximum_routes_warning_only: < true | false >
+      allowas_in:
+          enabled: < true | false >
```
ClausHolbechArista: We should only do two spaces indentation here.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2471,6 +2496,12 @@ router_bgp:
           timers: < keepalive_hold_timer_values >
           send_community: < standard | extended | large | all >
           maximum_routes: < integer >
+          maximum_routes_warning_limit: < integer >
+          maximum_routes_warning_limit_percent: < true | false >
+          maximum_routes_warning_only: < true | false >
+          allowas_in:
+            enabled: < true | false >
+            allow_local_asn_n_times: < 1-10 >
```
ClausHolbechArista: Change to `times` like the others.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2252,6 +2267,12 @@ router_bgp:
       route_map_out: < outbound route-map >
       send_community: < all | extended | large | standard >
       maximum_routes: < integer >
+      maximum_routes_warning_limit: < integer >
```
ClausHolbechArista: Missing the last update with string value and remove the percent key.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2471,6 +2496,12 @@ router_bgp:
           timers: < keepalive_hold_timer_values >
           send_community: < standard | extended | large | all >
           maximum_routes: < integer >
+          maximum_routes_warning_limit: < integer >
```
ClausHolbechArista: Missing the last update with string value and remove the percent key.

ClausHolbechArista (changes requested):
A few more requests. The downside of putting this many knobs into the same PR.

ccsnw:
> A few more requests. The downside of putting this many knobs into the same PR.

don't worry. I'm the dumbass for not paying more attention on the update.. -.- sometimes you gotta learn the hard way.

ClausHolbechArista (dismissed):
LGTM

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -131,7 +165,21 @@ router bgp {{ router_bgp.as }}
...
+{%                 set maximum_routes_cli = maximum_routes_cli ~ " warning-only" %}
+{%             endif %}
+   {{ maximum_routes_cli }}
+{%         endif %}
+{%         if router_bgp.neighbors[neighbor].allowas_in.enabled is arista.avd.defined(true) %}
+{%             set allowas_in_cli = "neighbor " ~ neighbor ~ " allowas-in" %}
+{%             if router_bgp.neighbors[neighbor].allowas_in.times is arista.avd.defined %}
+{%                 set allowas_in_cli = allowas_in_cli ~ " " ~ router_bgp.neighbors[neighbor].allowas_in.times %}
+{%             endif %}
+   {{ allowas_in_cli }}
```
xaviramon, line 182: There is no documentation for the allowas-in knob. I think we should include it in the section "BGP tunning" or the peer-group.

ccsnw: Hmm don't think the tuning table would be the proper place. 
Under neighbors or the Peer group would make more sense to me.
Question is if the setting is that common, you'd in any case see it via the embedded config snippet..

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -72,6 +92,20 @@
...
+{%                     if router_bgp.peer_groups[peer_group].maximum_routes_warning_limit is arista.avd.defined %}
+{%                         set value = value ~ "warning-limit " ~ router_bgp.peer_groups[peer_group].maximum_routes_warning_limit %}
+{%                         if router_bgp.peer_groups[peer_group].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                             set value = value ~ ", " %}
+{%                         else %}
+{%                             set value = value ~ ")" %}
+{%                         endif %}
+{%                     endif %}
+{%                     if router_bgp.peer_groups[peer_group].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                         set value = value ~ "warning-only)"%}
```
tgodaA: Missing space before ```%}```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -112,7 +146,28 @@
...
+{%                     if router_bgp.neighbors[neighbor].maximum_routes_warning_limit is arista.avd.defined %}
+{%                         set value = value ~ "warning-limit " ~ router_bgp.neighbors[neighbor].maximum_routes_warning_limit %}
+{%                         if router_bgp.neighbors[neighbor].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                             set value = value ~ ", " %}
+{%                         else %}
+{%                             set value = value ~ ")" %}
+{%                         endif %}
+{%                     endif %}
+{%                     if router_bgp.neighbors[neighbor].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                         set value = value ~ "warning-only)"%}
```
tgodaA: Missing space before ```%}```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -133,7 +188,28 @@
...
+{%                             if router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes_warning_limit is arista.avd.defined %}
+{%                                 set value = value ~ "warning-limit " ~ router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes_warning_limit %}
+{%                                 if router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                                     set value = value ~ ", " %}
+{%                                 else %}
+{%                                     set value = value ~ ")" %}
+{%                                 endif %}
+{%                             endif %}
+{%                             if router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes_warning_only is arista.avd.defined(true) %}
+{%                                 set value = value ~ "warning-only)"%}
```
tgodaA: Missing space before ```%}```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -614,7 +668,21 @@ router bgp {{ router_bgp.as }}
       neighbor {{ neighbor }} send-community {{ router_bgp.vrfs[vrf].neighbors[neighbor].send_community }}
 {%             endif %}
 {%             if router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes is arista.avd.defined %}
-      neighbor {{ neighbor }} maximum-routes {{ router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes }}
+{%             set maximum_routes_cli = "neighbor " ~ neighbor ~ " maximum-routes " ~ router_bgp.vrfs[vrf].neighbors[neighbor].maximum_routes %}
```
tgodaA: Bad indentation lines 670 to 679. Missing 4 spaces (lines 671-677)

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
```diff
@@ -311,8 +389,10 @@
...
 {%                 endfor %}
 {%             endif %}
-{%             set redistribute = router_bgp.vlans[vlan].redistribute_routes | arista.avd.default("-") %}
-| {{ vlan }} | {{ route_distinguisher }} | {{ both_route_target | arista.avd.default(["-"]) | join("<br>") }} | {{ import_route_target | arista.avd.default(["-"]) | join("<br>") }} | {{ export_route_target | arista.avd.default(["-"]) |  join("<br>") }} | {{ redistribute | join("<br>") }} |
+{%             set redistribute_route = router_bgp.vlans[vlan].redistribute_routes | arista.avd.default("") | list %}
+{%             set no_redistribute_route = router_bgp.vlans[vlan].no_redistribute_routes | arista.avd.default("") | map("replace","","no ",1) | list %}
+{%             set redistribution = redistribute_route + no_redistribute_route %}
+| {{ vlan }} | {{ route_distinguisher }} | {{ both_route_target | arista.avd.default(["-"]) | join("<br>") }} | {{ import_route_target | arista.avd.default(["-"]) | join("<br>") }} | {{ export_route_target | arista.avd.default(["-"]) |  join("<br>") }} | {{ redistribution | join("<br>") | arista.avd.default("-") }} |
```
tgodaA: One minor correction:

Remove double space before join ```{{ export_route_target | arista.avd.default(["-"]) |  join("<br>") }}```

This might affect all other artifact scenarios.

tgodaA:
Rest of it LGTM! Will approve once you solve those jinja lint errors.

ccsnw:
> Rest of it LGTM! Will approve once you solve those jinja lint errors.

Can I setup a linter to catch these style errors automatically? If so do you have a config for one thats compliant with the rules here? Might make sense to share this via the webpage, or?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-bgp.j2
ccsnw: will do, but think this came in from the previous PR

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
xaviramon: I agree we don't need to document everything in the tables and the configuration is in the snipped. However, in this case, as we are modifying the BGP best path selection algorithm, I believe it is important to add it specifically in the documentation so it is clearly stated.

ccsnw: Xavi, adding it to the peer groups like this would make sense I think:

```
| Settings | Value |
| Allowas-in | n |
```

It would also be possible in the neighbors table, as a new column:

```
| Neighbor | Remote AS | VRF | Send-community | Maximum-routes | Allowas-in |
```

I dont see it fit in any of the existing columns..

@ClausHolbechArista @tgodaA whats your view on this?

tgodaA:
Yeah, we do have linter available for VS code. But I am checking it manually though. This is the page for coding style: https://avd.sh/en/releases-v3.x.x/docs/contribution/style-guide.html

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
```diff
@@ -88,16 +88,18 @@ interface Management1
 
 | BGP AS | Router ID |
 | ------ | --------- |
-| 65101|  - |
+| 65101|  192.168.255.3 |
```
tgodaA, line 91: You fixed this extra space in the last commit. But might have forgot to update the artifacts.

ccsnw: Tony, this is not coming from my code. Its from here:

```
| BGP AS | Router ID |
| ------ | --------- |
| {{ router_bgp.as }}|  {{ router_bgp.router_id | arista.avd.default('-') }} |
{%     if router_bgp.bgp_cluster_id is arista.avd.defined %}
```

tgodaA: Got you. Thank you! My bad I was confused, how did the CI test pass.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
tgodaA: You can add this to neighbors table. Allowas-in with values like Allowed, Allowed 5 times, '-'

I don't have much idea about this but neighbor table looks good to me.

ccsnw: @tgodaA Will add it for peer-groups and neighbors

tgodaA (approved):
LGTM!

xaviramon (approved):
LGTM!
Nice one! A lot of different cases to consider!

ClausHolbechArista (approved):
LGTM

PR #1373, opened 2021-11-30
Feat(eos_validate_state): Allowing configurable xcvr manufacturers

### LINKED ISSUE #997: eos_validate_state role whitelist transceiver vendors
**Issue Type**

- [ ] Template enhancement
- [x] Role enhancement
- [ ] Documentation enhancement (avd.sh)

We are using 3rd party optics with a verified vendor from our point of view. eos_validate_state role  reports "transceivers manufacturers" as FAIL.

 Make it possible to whitelist vendors via a variable.

**Additional context**

xaviramon:
## Change Summary

Currently, if XCVR manufacturer is not in the following list, the test is failing, while the interface may be working as expected: 
- 'Not Present'
- 'Arastra, Inc.'
- 'Arista Networks'

This PR allows the user to configure a list of accepted XCVR manufacturers to avoid failing these tests.

## Related Issue(s)

Fixes #997 

## Component(s) name

## Proposed changes
Task in hardware.yml file is modified to validate output of XCVR manufacturer agains a list defined in eos_validate_state/defaults/main.yml. 
This list is either the user-defined list under validation_role.xcvr_own_manufacturers variable OR the defined default (Arastra, Arista Networks). The variable is defined at fabric level: 
```
validation_role:
  xcvr_own_manufacturers:
    - Manufacturer 1
    - Manufacturer 2
```
The ansible task is only validating the manufacturer of the XCVR when it is present, so no need of defining the manufacturer "Not Present". This also eliminates noise in the results file, since we do not consider "Not present" as a successful test, we simply do not consider it and this can be catch in different tests (ie P2P ping test) if it should be installed.

This leads to a validation of the result.failed variable when building the CSV file, as the ansible task is executed but is does not fail.
Also, the report in CSV is piped with `| arista.avd.natural_sort` to print all interfaces in order, making it more "human readable".

## How to test
Generate a playbook with the testing devices included and run the eos_validate_state role against it.

Device should have a XCVR different from the above list.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
@@ -23,4 +23,4 @@ structured_dir_name: 'structured_configs'
 structured_dir: '{{ output_dir }}/{{ structured_dir_name }}'
 
 # Allow different manufacturers
-accepted_xcvr_manufacturers: "{{ validation_role.xcvr_own_manufacturers + ['Not Present', 'Arastra, Inc.', 'Arista Networks'] }}"
+accepted_xcvr_manufacturers: "{{ validation_role.xcvr_own_manufacturers | arista.avd.default(['Arastra, Inc.', 'Arista Networks']) }}"
--- a/ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
@@ -13,6 +13,9 @@ eos_validate_state_csv_report_path: '{{ eos_validate_state_dir }}/{{ fabric_name
 # Only support default and github
 validate_state_markdown_flavor: "default"
 
+# Allow different manufacturers
+accepted_xcvr_manufacturers: "{{ validation_role.xcvr_own_manufacturers | arista.avd.default(['Arastra, Inc.', 'Arista Networks']) }}"
+
 # AVD configurations output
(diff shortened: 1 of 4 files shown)


tgodaA (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
```diff
@@ -21,3 +21,6 @@ output_dir: '{{ root_dir }}/{{ output_dir_name }}'
 # Output for structured YAML files:
 structured_dir_name: 'structured_configs'
 structured_dir: '{{ output_dir }}/{{ structured_dir_name }}'
+
+# Allow different manufacturers
+accepted_xcvr_manufacturers: "{{ validation_role.xcvr_own_manufacturers + ['Not Present', 'Arastra, Inc.', 'Arista Networks'] }}"
```
ClausHolbechArista: missing documentation update on this new knob.
Also I think it is unnecessary complex to take one knob as input and have hidden builtin extras. Would be simpler to just set the right knob with the default value and leave it to the user to set `accepted_xcvr_manufacturers` directly and include the defaults when they add more. Maybe they want to _not_ allow one of the defaults.

xaviramon:
Implemented changes and changed the behaviour.

onurgashi (approved):
LGTM

tgodaA (approved):
LGTM!

PR #1381, opened 2021-12-01
Refactor(eos_designs): Network-services templates to use lists internally

ClausHolbechArista:
## Change Summary

Refactor eos_designs Network Services to improve performance by converting dictionaries to lists of dictionaries

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Introduce new filter plugin to convert_dict_to_list
- refactor network_services to use this method on for loops to address performance issues.

## How to test
- No change reported in Molecule
- Run test scenario with large data set

### User Checklist

- [x] Create new filter plugin - convert_dict_to_list
- [x] Upldate filter plugin documentation 
- [x] Refactor tenant network services template
- [x] Test with large data set for performance

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
+++ b/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
@@ -1,5 +1,5 @@
 #
-# def arista.avd.migrate_dicts
+# def arista.avd.convert_dicts
 #
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
@@ -7,23 +7,27 @@ __metaclass__ = type
 from jinja2.runtime import Undefined
 
 
-def migrate_dicts(dictionary, primary_key="name"):
+def convert_dicts(dictionary, primary_key="name"):
     """
-    migrate_dicts will convert dictionary to list.
+    convert_dicts will convert dictionary to list.
 
-    Arista.avd.migrate_dicts will convert nested dictionaries to list of dictionaries
+    The `arista.avd.convert_dicts` filter will convert nested dictionaries to lists of dictionaries
     and insert the outer dictionary keys into each list item using the primary_key name.
 
-    This filter is intended for seemless data model migration.
-    If variable is already converted to list, it will pass through untouched
+    This filter is intended for:
+
+    - Seemless data model migration from dictionaries to lists.
+    - Improve Ansible's processing performance when dealing with large dictionaries by converting them to lists of dictionaries.
+
+    Note: If the variable is already a list, it will pass through untouched
 
     Example
     -------
     {# Migrate access_lists data model from dict to list #}
-    {% set access_lists = access_lists | arista.avd.migrate_dicts %}
+    {% set access_lists = access_lists | arista.avd.convert_dicts %}
     {% for access_list in access_lists %}
     {#     Migrate access-lists.sequence_numbers data model from dict to list #}
-    {%     do access_lists[loop.index0].update({'sequence_numbers': access_list.sequence_numbers | arista.avd.migrate_dicts('sequence')}) %}
+    {%     do access_lists[loop.index0].update({'sequence_numbers': access_list.sequence_numbers | arista.avd.convert_dicts('sequence')}) %}
     {% endfor %}
     access_lists: {{ access_lists }}
 
@@ -58,5 +62,5 @@ def migrate_dicts(dictionary, primary_key="name"):
 class FilterModule(object):
     def filters(self):
         return {
-            'migrate_dicts': migrate_dicts,
+            'convert_dicts': convert_dicts,
         }
--- a/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
+++ b/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
@@ -9,10 +9,8 @@ from jinja2.runtime import Undefined
 
 def convert_dicts(dictionary, primary_key="name"):
     """
-    convert_dicts will convert dictionary to list.
-
-    The `arista.avd.convert_dicts` filter will convert nested dictionaries to lists of dictionaries
-    and insert the outer dictionary keys into each list item using the primary_key `name`.
+    The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of dictionaries
+    and insert the outer dictionary keys into each list item using the primary_key `name` (key name is configurable).
 
     This filter is intended for:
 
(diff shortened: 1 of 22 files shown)


carlbuchmann (approved):
LGTM!

carlbuchmann (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
```diff
@@ -0,0 +1,34 @@
+
+{% set tmp_tenants = [] %}
+{# tenants #}
+{% for tenant in tenants | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
+{%     if tenant.name in switch.tenants | arista.avd.default([]) %}
+{%         set switch_tenant = switch.tenants[tenant.name] %}
+{#         vrfs #}
+{%         set tmp_vrfs = [] %}
+{%         for vrf in tenant.vrfs | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
```
tgodaA, line 9: 1) I don't know the reason, why we sort the data in eos_designs, because we sort them anyway in eos_cli_config_gen.
2) If you think sorting is required then try not to sort them here. Just append them to the tmp_list now and sort that tmp_list when you are using it. Because tmp_list mostly will have few items compared to the actual parent list (this one). Also sometimes the if check might fail for all entries leading to an empty tmp_list. So in that case we are wasting resources to sort the list unnecessarily. 

Sorting here might be efficient sometimes but I am still not sure, which one is more efficient as it depends on how we are writing the code.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/route-maps.j2
```diff
@@ -1,19 +1,19 @@
 {# Route-maps for tenant bgp peers set_ipv4_next_hop parameter #}
 route_maps:
-{% for tenant in switch.tenants | arista.avd.natural_sort %}
-{%     for vrf in switch.tenants[tenant].vrfs | arista.avd.natural_sort %}
-{%         for bgp_peer in tenants[tenant].vrfs[vrf].bgp_peers | arista.avd.natural_sort %}
-{%             if inventory_hostname in tenants[tenant].vrfs[vrf].bgp_peers[bgp_peer].nodes | arista.avd.natural_sort %}
-{%                 if tenants[tenant].vrfs[vrf].bgp_peers[bgp_peer].set_ipv4_next_hop is arista.avd.defined or tenants[tenant].vrfs[vrf].bgp_peers[bgp_peer].set_ipv6_next_hop is arista.avd.defined %}
-  RM-{{ vrf }}-{{ bgp_peer }}-SET-NEXT-HOP-OUT:
+{% for tenant in network_services_data.tenants %}
```
tgodaA, line 3: Here it is better to sort than in the logic.j2. Similar for other knobs. I mean, we sort only once, first time when we wanted to use that respective list.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -0,0 +1,70 @@
...
+    -------
+    any
+        Returns list of dictionaries or input variable untouched if not a nested dictionary
+    """
+    if not isinstance(dictionary, dict):
+        # Not a dictionary, return the original
+        return dictionary
+    else:
+        output = []
+        for key in dictionary.keys():
```
tgodaA, line 55: I saw this article online: https://stackoverflow.com/questions/44244281/how-come-key-in-dictionary-keys-slower-than-key-in-dictionary, which says to avoid using .keys(). 

Also I tried this python3 code my self:

```
import string, time

d = dict.fromkeys(string.ascii_lowercase, 0)

def f_nokeys(d):
    start_time = time.time()
    print("t" in d)
    print("Without .keys method")
    print("--- %s seconds ---" % (time.time() - start_time))

def f_keys(d):
    start_time = time.time()
    print("t" in d.keys())
    print("With .keys method")
    print("--- %s seconds ---" % (time.time() - start_time))

f_nokeys(d)
print('\n\n')
f_keys(d)
```

Most of the time the .keys() method is taking much more time.
```
docker % python test.py
True
Without .keys method
--- 2.4080276489257812e-05 seconds ---



True
With .keys method
--- 5.9604644775390625e-06 seconds ---
```

tgodaA: I tried your function (convert to list) and found no major difference with .keys() method and without it. I could not find a major difference in time.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
```diff
@@ -0,0 +1,34 @@
+
+{% set tmp_tenants = [] %}
+{# tenants #}
+{% for tenant in tenants | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
```
tgodaA, line 4: Why not loop through switch.tenants, so that we eliminate the next if condition. ``` for tenant in switch.tenants``` and then to access the tenant data we can use tenants[tenant] directly. This will save a lot of time. Similarly for other data like, vrfs and svis. Because we did this similar things already in switch facts.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
ClausHolbechArista: Just tested on my end too, and even if I don't see any measurable difference, I will add this fix (remove `.keys()`) together with the next refactor. Thank you for the deep review!

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
ClausHolbechArista: We will reference these variables in many places, so sorting them every time is ineffective. We _could_ add a final sort before writing the variable, instead of sorting on the input, but I did not see any measurable difference. Also the protection builtin to the `natural_sort` filter saves us another `default([])`

We need to sort the data in eos_designs primarily to avoid any changes. The documentation and the csv files changed when I didn't sort. I wanted to avoid _any_ changes.

ClausHolbechArista: In general the "structured config" is also a "product" which is used by many other areas, so having it sorted and nice makes it easier for the next role.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/route-maps.j2
ClausHolbechArista: you would have to add the sort on every level, so in this example on tenants and on vrfs. Many of the others also use SVIs and l2vlans, so it becomes more expensive.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
tgodaA: Got it, thanks.

PR #1386, opened 2021-12-02
Feat(eos_designs): Platform management interface

### LINKED ISSUE #1056: Choosing a modular platform should update the default management port
## Issue Type
- [ ] Bug Report

## Summary
When a platform such as 7500R is chosen, the default management port is still "Management1" which does not exists in modular switches. It would be nice that is automatically set to "Management0" (active supervisor).

## Steps to reproduce
1. Chose platform 7500
2. Management port for all platforms are still using default port "Management1".

### Expected results
Modular platforms to use Management0 instead.

xaviramon:
## Change Summary

Management interface is changed for modular platforms

## Related Issue(s)

Fixes #1056

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -81,7 +81,7 @@ mlag_ibgp_peering_vrfs:
 # Recommended default values for Jericho based platform, VEOS and all other platforms default tag.
 
 platform_settings:
-  - platforms: [default]
+  - platforms: [ default ]
     reload_delay:
       mlag: 300
       non_mlag: 330
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
@@ -26,7 +26,6 @@ design:
 
 mgmt_vrf_routing: false
 
-mgmt_interface: Management1
 mgmt_interface_vrf: MGMT
 
 # eAPI
@@ -89,13 +88,25 @@ platform_settings:
     feature_support:
       # "queue-monitor length notify" is only valid for R-Series so should be disabled on default platform.
       queue_monitor_length_notify: false
-  - platforms: [ 7280R, 7280R2, 7500R, 7500R2 ]
+  - platforms: [ 7280R, 7280R2 ]
     tcam_profile: vxlan-routing
     lag_hardware_only: true
     reload_delay:
       mlag: 900
       non_mlag: 1020
-  - platforms: [ 7280R3, 7500R3, 7800R3 ]
+  - platforms: [ 7280R3 ]
+    reload_delay:
+      mlag: 900
+      non_mlag: 1020
+  - platforms: [ 7500R, 7500R2 ]
+    tcam_profile: vxlan-routing
+    lag_hardware_only: true
+    management_interface: Management0
+    reload_delay:
+      mlag: 900
+      non_mlag: 1020
+  - platforms: [ 7500R3, 7800R3 ]
+    management_interface: Management0
     reload_delay:
       mlag: 900
       non_mlag: 1020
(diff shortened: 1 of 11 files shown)


ansible_collections/arista/avd/roles/eos_designs/templates/base/management-interfaces.j2
```diff
@@ -1,10 +1,9 @@
 {# Management interface #}
-{% if mgmt_interface is arista.avd.defined and
-      mgmt_interface_vrf is arista.avd.defined and
+{% if mgmt_interface_vrf is arista.avd.defined and
       mgmt_gateway is arista.avd.defined and
       switch.mgmt_ip is arista.avd.defined %}
 management_interfaces:
-  {{ mgmt_interface }}:
```
ClausHolbechArista, line 7: please resolve the correct mgmt_interface in switch facts template, so we set a `switch.management_interface`. It makes it simpler to find the defaults, also to use the management interface in other templates or in user-configured inline jinja.

ansible_collections/arista/avd/roles/eos_designs/templates/upgrade/upgrade-2.x-to-3.0/ntp.j2
```diff
@@ -8,7 +8,7 @@
 {%     endif %}
 {{ prefix }}ntp:
   local_interface:
-    name: {{ mgmt_interface }}
+    name: {{ switch.platform_settings.management_interface | arista.avd.default('Management1') }}
```
ClausHolbechArista: I don't think you can use the switch.* variables in the upgrade templates, since we don't run the facts templates first.
Since this upgrade is about 2.x to 3.0 it is safe to asume that mgmt_interface is set. Maybe just add it as a condition to the `if` in line 1

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -82,20 +81,32 @@ mlag_ibgp_peering_vrfs:
...
       # "queue-monitor length notify" is only valid for R-Series so should be disabled on default platform.
       queue_monitor_length_notify: false
-  - platforms: [ 7280R, 7280R2, 7500R, 7500R2 ]
+  - platforms: [ 7280R, 7280R2 ]
     tcam_profile: vxlan-routing
     lag_hardware_only: true
     reload_delay:
       mlag: 900
       non_mlag: 1020
-  - platforms: [ 7280R3, 7500R3, 7800R3 ]
```
ClausHolbechArista, line 98: Before merging this, we need to consider if any of these changes to the default platforms might affect existing deployments negatively. IMO this will only fix issues and not affect them badly. They would most likely already have different fixes in place to address these issues.

ansible_collections/arista/avd/roles/eos_designs/templates/upgrade/upgrade-2.x-to-3.0/ntp.j2
ClausHolbechArista: Ah no, we would loose the old default, so you should just do `mgmt_interface | arista.avd.default('Management1')` since this is the upgrade role.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
xaviramon: IF user has all platforms correctly selected, that should be ok. I will improve documentation, and maybe it is a good idea to emphasise it un RN.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -26,7 +26,6 @@ design:
 
 mgmt_vrf_routing: false
 
-mgmt_interface: Management1
```
tgodaA, line 29: CI error caused because you deleted this, as this knob is used in upgrade_2.x_to_3.x role. But please comment above this knob of it's depreciation if you are not deleting this line. So we can remove in later versions. Else we have to update in the 2.x_to_3.x upgrade role. Whatever is the best option!

tgodaA (approved):
LGTM!

carlbuchmann (approved):
LGMT!

ClausHolbechArista (approved):
LGTM

PR #1393, opened 2021-12-07
Doc(eos_designs): Minor documentation fixes

PhirePhly:
## Change Summary

Fix two typos in the eos_designs documentation:
* interface_descriptions variable is spelled incorrectly in node_types.md
* endpoint_ports is required to be defined in connected_endpoints.md

## Component(s) name

`arista.avd.eos_designs`

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
@@ -96,13 +96,13 @@ port_profiles:
         # Adapter speed - if not specified will be auto.
       - speed: < interface_speed | forced interface_speed | auto interface_speed >
 
-        # Local endpoint port(s)
+        # Local endpoint port(s) | required
         endpoint_ports: [ < interface_name > ]
 
-        # List of port(s) connected to switches
+        # List of port(s) connected to switches | required
         switch_ports: [ < switchport_interface > ]
 
-        # List of switche(s)
+        # List of switche(s) | required
         switches: [ < device > ]
 
         # Port-profile name, to inherit configuration.
--- a/ansible_collections/arista/avd/roles/eos_designs/doc/node-types.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/doc/node-types.md
@@ -53,7 +53,7 @@ node_type_keys:
 
     # Optional | Override interface_descriptions templates
     # If description templates use Jinja2, they have to strip whitespaces using {%- -%} on any code blocks
-    interface descriptions:
+    interface_descriptions:
       underlay_ethernet_interfaces: <path to J2 template - default inherited from templates.interface_descriptions.underlay_ethernet_interfaces >
       underlay_port_channel_interfaces: <path to J2 template - default inherited from templates.interface_descriptions.underlay_port_channel_interfaces >
       connected_endpoints_ethernet_interfaces: <path to J2 template - default inherited from templates.interface_descriptions.connected_endpoints_ethernet_interfaces >


PhirePhly:
Does the bot not like that I included an optional scope?

PhirePhly:
Goodness. Sorry about the title edit spam on the Title

ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
```diff
@@ -96,7 +96,7 @@ port_profiles:
         # Adapter speed - if not specified will be auto.
       - speed: < interface_speed | forced interface_speed | auto interface_speed >
 
-        # Local endpoint port(s)
+        # Local endpoint port(s) | required
         endpoint_ports: [ < interface_name > ]
 
         # List of port(s) connected to switches
```
xaviramon: Could you please add the "| required" tag here and also in next variable, "switches"?

xaviramon (changes requested):
Thank you for your PR, Kenneth!
That's a good catch, not all required variables are really marked with the required tag. I believe it was somehow assumed that defined a new server will require switches and ports, but it is much better if we explicitly say it.
Please, add the tag in the variables I mentioned. Otherwise, looks good to me and I will approve it after these changes.

PhirePhly:
So interestingly, while the switch and switch port are conceptually required, I couldn't find any syntactic requirement for them. The lack of the endpoint_ports caused a build failure, but you can define a batch of servers and not allocate them to a switch right away. I'll make your requested changes and fix both commit messages now that I better understand the title semantics.

Where I tripped on this is because I'm trying to use AVD to build an L2 EVPN fabric for an Internet Exchange Point. The endpoints are Internet edge peering and outside our autonomous system, so I have no idea what port on their device they're using. So ideally I think we should make the endpoint ports optional, but I'm trying to go one step at a time here.

xaviramon:
I took another look. If you don't have ``switches`` and ``switch_ports`` documentation is generated correctly. If ``switch_ports`` is missing you will face an error, so it is validated:
```
TASK [arista.avd.eos_designs : Generate device configuration in structured format] ***********************************************************************************************
Friday 10 December 2021  17:20:03 +0100 (0:00:00.768)       0:00:07.330 ******* 
Friday 10 December 2021  17:20:03 +0100 (0:00:00.768)       0:00:07.329 ******* 
ok: [POD1_SPINE01 -> localhost]
ok: [POD1_SPINE02 -> localhost]
fatal: [POD1_LEAF01A -> localhost]: FAILED! => {"msg": "'dict object' has no attribute 'switch_ports'"}
```

For your specific case, I might suggest they use as ``endpoint_ports`` whatever value helps them to identify the connection on the other peer, as this value is only used in the description field:
```
my_cool_server:
  cool_1:
    rack: RACK_TEST
    adapters:
      - speed: "10g auto"
        endpoint_ports: [ blah ]
        switch_ports: [ eth7 ]
        switches: [ POD1_LEAF01A ]
        mode: access
        vlans: "3"
```
Generated documentation:
```
#### L2

| Interface | Description | Mode | VLANs | Native VLAN | Trunk Group | Channel-Group |
| --------- | ----------- | ---- | ----- | ----------- | ----------- | ------------- |
| eth7 |  cool_1_blah | access | 3 | - | - | - |
```
Generated config:
```
interface eth7
   description cool_1_blah
   no shutdown
   speed 10g auto
   switchport
   switchport access vlan 3
   switchport mode access
!
```

PhirePhly:
Good point. Once the switch is assigned and the endpoint gets pulling into the switches state, the lack of the port assignment is a problem.

I've force pushed over my previous branch with your requested changes.

tgodaA:
To add up to your discussion. I just found that either ```server_ports``` or ```endpoint_ports``` is required. Because if we don't define ```endpoint_ports``` then the default value is taken from ```server_ports```. But this looks good to me to set **endpoint_ports** to required.

tgodaA:
Also, change your PR title to Fix documentation rather than saying two or three suggestions.

I found one more typo in the same connected-endpoints. 
image
I don't think we need:
```yaml
< port_profile_2 >:
  mode: < access | dot1q-tunnel | trunk >
  vlans: < vlans as string >
```

This is a port_profile_2 which should be defined under port_profiles knob but accidentally sitting under adapters settings.

PhirePhly:
I've changed the pull request title. 

`port_profile_2` is outside the scope of what I'm working on.

ClausHolbechArista (approved):
LGTM!
Thank you for this valuable contribution! It will save others the trouble you might have had.

carlbuchmann (approved):
LGTM!

PR #1395, opened 2021-12-09
Refactor(eos_designs,eos_cli_config_gen): Refactor management-api-http and match the documentation

tgodaA:
## Change Summary

- management_eapi in eos_designs is using defined test instead of arista.avd.defined, and MGMT vrf is created by default under this config. To use eapi with the devices directly rather than through CVP, we don't need MGMT vrf, where as it should be set to vrf ```default```. If we use arista.avd.defined tests then, setting management_eapi: null will not generate the config that is creating issues and then define the intended management-api-http config as per requirement. Overall, this refactor will **allow us to overwrite the default behavior** if required.
- management_api_http in eos_cli_config_gen: The documentation template is using **lower** filter over a knob that is boolean. So I guess we don't need that. Also refactored whitespaces according to the logic. Added a missing feature in README

## Related Issue(s)

Fixes #

## Component(s) name

`arista.avd.eos_designs`, `arista.avd.eos_cli_config_gen`

## How to test
Updated molecule artifacts (make refresh-facts)

### User Checklist

- N/A

### Repository Checklist

- [ x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [ x] My change requires a change to the documentation and documentation have been updated accordingly.
- [ x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/base.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/base.md
@@ -126,15 +126,14 @@ management console
 ### Management API HTTP Summary
 
 | HTTP | HTTPS |
-| ---------- | ---------- |
-| true | true |
+| ---- | ----- |
+| True | True |
 
 ### Management API VRF Access
 
 | VRF Name | IPv4 ACL | IPv6 ACL |
 | -------- | -------- | -------- |
 | mgt | ACL-API | - |
-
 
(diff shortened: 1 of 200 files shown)


ClausHolbechArista (approved):
LGTM

carlbuchmann (approved):
LGMT!

PR #1414, opened 2021-12-21
Fix(eos_designs): Management interface renders incorrectly when set other than default

xaviramon:
## Change Summary

- Fixes issue when setting `mgmt_interface` to Management0, it won't take effect and always gets set to the default `Management1`
- Add defaults for cEOS platform.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

fix logic in templates/facts/switch.j2  accordingly.

## How to test
See Molecule tests.

### User Checklist

- N/A

### Repository Checklist


ankudinov:
@xaviramon Don't think that `arista.avd.contains` is case sensitive. Please add `cEOS` and `ceos` or make it case insensitive.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -15,11 +15,12 @@ switch:
 {# switch.mgmt_ip #}
   mgmt_ip: {{ switch_data.node.mgmt_ip | arista.avd.default() }}
 
-{# switch.mgmt_interface#}
-  mgmt_interface: {{ platform_settings.management_interface | arista.avd.default('Management1') }}
+{# switch.mgmt_interface #}
+{% set switch_platform = switch_data.combined.platform | arista.avd.default() %}
+{% set platform_mgmt_int = (platform_settings | selectattr("platforms", "arista.avd.contains", switch_platform) | first).management_interface %}
+  mgmt_interface: {{ switch_data.node.mgmt_interface | arista.avd.default(platform_mgmt_int,mgmt_interface) }}
```
tgodaA: space after the ```,``` in arista.avd.default filter

tgodaA: Also, a small suggestion, use ```switch_data.combined.mgmt_interface``` as the management interface can be common for multiple switches in the group (like leaf1a and leaf2a might have same interface, Management0). switch_data.combined will pull the node config eventually if we specify the management_interface on a node.

```mgmt_interface: {{ switch_data.combined.mgmt_interface | ```

Comment (carlbuchmann) [reply ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2]: agreed and updated!

**Review by carlbuchmann — APPROVED** [review]
LGMT and tested with avd-quickstart-containerlab.

**Comment by tgodaA** [comment]
Missing fabric-topology.md documentation. Add the knob ```mgmt_interface``` under node_type defaults or somewhere explaining how to use it.


Except this everything LGTM!

File: ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml [hunk 2]
```diff
@@ -81,36 +82,44 @@ mlag_ibgp_peering_vrfs:
...
       mlag: 900
       non_mlag: 1020
-  - platforms: [ 7500R3, 7800R3 ]
+  - platforms: [ '7500R3', '7800R3' ]
     management_interface: Management0
     reload_delay:
       mlag: 900
       non_mlag: 1020
-  - platforms: ["VEOS", "VEOS-LAB"]
+  - platforms: [ 'VEOS', 'VEOS-LAB', 'vEOS', 'vEOS-lab' ]
```
tgodaA, line 114: Thank you for these changes!

tgodaA (approved):
LGTM!

carlbuchmann (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -145,6 +145,9 @@ defaults <- node_group <- node_group.node <- node
     # Arista platform family | Required.
     platform: < Arista Platform Family >
 
+    # Management interface configuration
+    mgmt_interface: < mgmt_interface | default -> platform_management_interface -> Management1 >
```
ClausHolbechArista: 1. This is already documented in `management-settings.md`. It was never removed from there.
2. This implies that the platform setting has less preference than the mgmt_interface setting. I would think it should be the other way around, so a general setting can be overwritten per platform.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -161,6 +161,27 @@ l3leaf:
           mgmt_ip: 192.168.200.118/24
           mac_address: '0c:1d:c0:1d:62:01'
           uplink_switch_interfaces: [ Ethernet9, Ethernet9, Ethernet9, Ethernet9 ]
+    DC1_BL3:
```
ClausHolbechArista: - These changes are not a fix as PR title says.
- Do we need 2 more devices with full configs - adding to the runtime - in unit test scenario to be able to test the CEOS platform? Or could we add a smaller definition like the `evpn_services_l2_only_true` example, where we set only the relevant parameters and name it according to the feature we are testing.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -15,11 +15,12 @@ switch:
 {# switch.mgmt_ip #}
   mgmt_ip: {{ switch_data.node.mgmt_ip | arista.avd.default() }}
 
-{# switch.mgmt_interface#}
-  mgmt_interface: {{ platform_settings.management_interface | arista.avd.default('Management1') }}
+{# switch.mgmt_interface #}
+{% set switch_platform = switch_data.combined.platform | arista.avd.default() %}
```
ClausHolbechArista: to keep this file consistent, please move this setting back down below the switch.platform. You can then move the while `switch.mgmt_interface` section further down if it relies on the `switch_platform` variable.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -15,11 +15,12 @@ switch:
...
+{% set platform_mgmt_int = (platform_settings | selectattr("platforms", "arista.avd.contains", switch_platform) | first).management_interface %}
```
ClausHolbechArista: It would make this more readable and easier to maintain, if you move this down below the `switch.platform_settings`, set a `switch_platform_settings` variable in that section, and then use that variable here. This logic is also missing coverage for the default platform, so another reason to reuse the cluttered logic for platform_settings.

ClausHolbechArista (changes requested):
I think there are some things we need to improve before releasing this.

ankudinov:
LGTM once comments from Claus resolved. Tested with avd-quickstart-containerlab

ankudinov (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
tgodaA: This setting is per < node-type-key >. But in **management-settings.md** the old setting was a global parameter.

tgodaA: Because we support mgmt_interface at both < node-type-key > and globally, I guess we need to document both.

carlbuchmann: @ClausHolbechArista , I actually tried your suggestion (point number 2) in my testing and would introduce a potential breaking change.

ClausHolbechArista: All good. But then we need a 3rd inheritance above - so `< mgmt_interface | default -> platform_management_interface -> mgmt_interface -> "Management1" >

ClausHolbechArista (approved):
LGTM

PR #1422, opened 2021-12-28
Feat(eos_cli_config_gen): Add MAC ACL support to Ethernet and Port-channels

### LINKED ISSUE #1397: Add MAC ACL support
### Enhancement summary

Add support for MAC ACLs in eos_cli_config_gen.

eos_cli_config_gen

### Use case example

MAC ACLs are needed for the IXP L2 peering network I'm working on porting to AVD. This feature will allow defining an ingress MAC ACL, which gets applied to the connected end points as `port_profiles: structured_config: mac_access_group_in`.

Since this is essentially going to be the same as IPv6, I think the AVD structure syntax is already pretty set. The MAC ACLs need to be globally defined, then also assigned to interfaces, so the steps are:
1. `mac_access_lists:` with children for each MAC ACL name, and their sequence numbers with raw action strings. (I'm assuming we don't want to get in the business of building individual  ACL lines as structured config)
2. `ethernet_interfaces: Ethernet_interface_1: mac_access_group_in` & `mac_address_group_out` - The same for port-channels and SVIs? I don't know how much commonality there is in the code base between different interface types, so I'll need to dig into this.
3. Documentation updates.
4. Unit tests? I'll need help once I get to this point.

This won't be integrated into eos_designs since it seems like the existing ACLs aren't integrated into that either. Are we eventually going to want eos_designs endpoint adapters / port profiles to include support for IPv4/IPv6/MAC ACLs, or leave those to custom structured config?

It seems like the current solution would be to dump the MAC ACL entirely as raw EOS config lines into the ports. 

### Additional context

_No response_

### Contributing Guide

PhirePhly:
## Change Summary

Add ingress and egress MAC ACLs to front panel ports.

## Related Issue(s)

Fixes #1397

## Component(s) name

## Proposed changes

Adds MAC ACL support with essential identical data models as for IPv6 ACLs. 

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -253,6 +254,23 @@ ipv6_access_lists:
         action: "< action as string >"
 ```
 
+#### MAC Access-Lists
```
ClausHolbechArista, line 317: Since this is a new model, we should move to list of dicts now. We will convert all the existing ones in 4.0.
So I suggest:
```yaml
mac_access_lists:
  - name: < mac_access_list_name_1 >
    counters_per_entry: < true | false >
    entries:
      - sequence: < sequence_id_1 >
        action: "< action as string >"
```
And BTW then make sequence optional in the template.

PhirePhly:
I think I need a little help with the new data model. I've got a new branch https://github.com/PhirePhly/ansible-avd/tree/mac-acl-v2

Do we want the j2 template to still sort entries by number? What's the best way to do that? https://github.com/PhirePhly/ansible-avd/blob/20ad2740762b5421a46ee9729e162d5705c85f76/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/mac-access-lists.j2

(ignore the documentation templates; they're still a mess.)

ClausHolbechArista:
> I think I need a little help with the new data model. I've got a new branch https://github.com/PhirePhly/ansible-avd/tree/mac-acl-v2
> 
> Do we want the j2 template to still sort entries by number? What's the best way to do that? https://github.com/PhirePhly/ansible-avd/blob/20ad2740762b5421a46ee9729e162d5705c85f76/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/mac-access-lists.j2
> 
> (ignore the documentation templates; they're still a mess.)

I have not looked at your branch. I can take a look sometime next week, but simplest would be if you update PR with every small step (Just commit to the same branch as PR is using). Then I can easily review/help.

Re sort, our `arista.avd.natural_sort` supports an argument being the key to sort on if it is a list of dicts. (Yeah!! ;) 
Check doc for natural_sort filter and other plugins here: https://avd.sh/en/latest/plugins/index.html

PhirePhly:
> Re sort, our `arista.avd.natural_sort` supports an argument being the key to sort on if it is a list of dicts. (Yeah!! ;) Check doc for natural_sort filter and other plugins here: https://avd.sh/en/latest/plugins/index.html

I tried using `{%     for acl_entry in mac_access_list.entries | arista.avd.natural_sort('sequence') %}` but for the entries where sequence isn't defined, it doesn't seem to leave them in the provided order, but tries to sort them based on the string.

So this unit test:
```
mac_access_lists:
  - name: TEST4
    entries:
      - action: "permit any 01:00:0c:cc:cc:cd 00:00:00:00:00:00"
      - action: "deny any 01:80:c2:00:00:00 00:00:00:00:00:00"
```
Emits an ACL that is sorted on deny before permit:
```
!
mac access-list TEST4
   deny any 01:80:c2:00:00:00 00:00:00:00:00:00
   permit any 01:00:0c:cc:cc:cd 00:00:00:00:00:00
!
```

So is there a way to invoke `natural_sort` as a stable sort?

I assume we're ok with it not handling an ACL definition that mixes sequence numbers and not numbered? Or should I be looping over the dicts as a preprocessing stage to assign missing sequence numbers?

ClausHolbechArista:
The more I think of this, I think we should just _not_ sort at all. Keep the list as provided. Emit sequence if provided.

tgodaA (approved):
LGTM!

ClausHolbechArista (approved):
LGTM

PR #1424, opened 2021-12-30
Feat(eos_cli_config_gen): Improve data model for extended IP ACLs #1411

### LINKED ISSUE #1411: Improve data model for extended IP ACLs
### Enhancement summary

Currently AVD supports basic extended ACL data model as the list of strings under `access_lists` key.
The data model must be improved to support more advance use cases. For example, integration with 3rd party policy controllers.

As part of the change AVD must allow to deploy port ACLs without producing the rest of the configuration (3rd party policy controller use case).

PS: the existing [ arista.eos.eos_acls ]( https://docs.ansible.com/ansible/latest/collections/arista/eos/eos_acls_module.html ) data model will not be considered, as it looks a bit suboptimal.

### Use case example

Integration with 3rd party policy controllers and other advanced use cases.

Proposed data model brainstormed with @ClausHolbechArista 

```yaml
ip_access_lists:
  - name: "< access_list_name >"
    counters_per_entry: < true | false >
    entries:
      - sequence: < sequence_id >
        remark: "<Comment, up to 100 characters >"
        action: < "permit" | "deny" | "remark" >
        protocol: < "ip" | "tcp" | "udp" | "<protocol name or number>" >
        source: < "IPv4Address/Mask" | "host <IPv4Address>" | "any" >
        source_ports: [ "< port name or number >", ]
        source_ports_match: < "eq"(default) | "gt" | "lt" | "neq" | "range" >
        destination: < "IPv4Address/Mask" | "host <IPv4Address>" | "any" >
        destination_ports: [ "< port name or number >", ]
        destination_ports_match: < "eq"(default) | "gt" | "lt" | "neq" | "range" >
        tcp_flags: [ "< tcp flags >", ]
        fragments: < true | false >
        log: < true | false >
        ttl: < 0-255 >
        ttl_match: < "eq"(default) | "gt" | "lt" | "neq" >
```

_No response_

### Additional context

This feature request focuses on IPv4 and extended ACLs only.

### Contributing Guide

ankudinov:
## Change Summary

Improve data model for extended IP ACLs

## Related Issue(s)

Fixes #1411 

## Component(s) name

## Proposed changes
Improve data model for extended IP ACLs

Data model:
```yaml
ip_access_lists:
  - name: "< access list name as string >"
    counters_per_entry: < true | false >
    entries:
      # remark entry
      - sequence: < acl entry sequence number >  # optional
        # NOTE: if remark is defined, other keys in acl entry will be ignored
        remark: "< Comment, up to 100 characters >"
      # normal entry
      - sequence: < acl entry sequence number >  # optional
        action: "< permit | deny >"  # required
        protocol: "< ip | tcp | udp | icmp | other protocol name or number >"  # required
        # NOTE: A.B.C.D without a mask means host
        source: "< any | A.B.C.D/E | A.B.C.D >"  # required
        source_ports_match: "< eq | gt | lt | neq | range >"  # eq is default
        source_ports: ["< tcp/udp port name or number >",]  # optional
        # NOTE: A.B.C.D without a mask means host
        destination: "< any | A.B.C.D/E | A.B.C.D >"  # required
        destination_ports_match: "< eq | gt | lt | neq | range >"  # eq is default
        destination_ports: ["< tcp/udp port name or number >",]  # optional
        tcp_flags: ["< tcp flag name >",]  # optional
        fragments: < true | false >  # optional, match non-head fragment packets
        log: < true | false >  # optional, log matches against this rule
        ttl: < <0-254> TTL value >  # optional
        ttl_match: "< eq | gt | lt | neq >"  # optional
        icmp_type: "< Message type name/number for ICMP packets >"  # optional
        icmp_code: "< Message code for ICMP packets >"  # optional
        nexthop_group: "< nexthop-group name >"  # optional
        tracked: < true | false > # optional, match packets in existing ICMP/UDP/TCP connections
        dscp: "< DSCP value or name >"  # optional
        vlan_number: < vlan number >  # optional
        vlan_inner: < true | false >  # optional, default - false
        vlan_mask: "< 0x000-0xFFF  Vlan mask >"  # optional
```

### Repository Checklist


ankudinov:
Implemented initial draft of the data model, but still work in progress. OK to review the data model, but do not merge.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -202,6 +213,52 @@ access_lists:
...
+        vlan_inner: < true | false >  # optional, default - false
+        vlan_mask: "< 0x000-0xFFF  Vlan mask >"  # optional
+```
+
+The improved data model allows to define the maximum number of ACL entries that AVD is allowed to generate.
+If the number is above the limit, the playbook will fail. That provides a simplified control over hardware utilisation.
+The numbers must be based on the hardware tests and AVD does not provide any guidance.
+
+```yaml
+ip_access_lists_max_acle: <maximum number of ACL entries allowed per switch>  # default is 10000
```
ClausHolbechArista: I suggest to rename to `ip_access_lists_max_entries`
please don't set a default here.
The first sentence in the text is not clear. Suggestion: "`ip_access_lists_max_entries` can be used to limit the number of ACL entries that AVD is allowed to generate. Only entries under `ip_access_lists` will be counted. If the number is above the limit, the playbook will fail. This provides a simplified control over hardware utilization.
The numbers must be based on the hardware tests and AVD does not provide any guidance. Note that other EOS features may use the same hardware resources and affect the supported scale"

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{#             * add dscp #}
+{%             if acle.dscp is arista.avd.defined %}
+{%                 set acl_entry.string = acl_entry.string ~ 'dscp ' ~ acle.dscp | lower ~ ' ' %}
+{%             endif %}
+{#             * add log #}
+{%             if acle.log is arista.avd.defined(true) %}
+{%                 set acl_entry.string = acl_entry.string ~ 'log' %}
+{%             endif %}
+{%         endif %}
+{#     - always remove trailing spaces at the end of the string #}
```
ClausHolbechArista: In all other templates we try to implement spacing by adding the space _before_ the keyword you add. This way we ensure that the `acl_entry.string` is always clean and ready to be printed. Please adjust here and remove the rstrip

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{#     check if the number of ACL entries is below the limit #}
+{%     set counter.acle_number = counter.acle_number + 1 %}
+{%     if counter.acle_number > counter.acle_max %}
+   {{ ('OK' if false) | mandatory('The number of ACL entries is above defined maximum!') }}
+{%     endif %}
+{#     END ACL_ENTRY CYCLE: walk over the list of ACL entries #}
+{%     endfor %}
+{# END ACL CYCLE: end walk over the list of defined ACLs #}
+{%     endfor %}
+{% endif %}
```
ClausHolbechArista: missing newline at the end of the file

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{%             if acle.log is arista.avd.defined(true) %}
+{%                 set acl_entry.string = acl_entry.string ~ 'log' %}
+{%             endif %}
+{%         endif %}
+{#     - always remove trailing spaces at the end of the string #}
+{%         set acl_entry.string = acl_entry.string.rstrip() %}
+   {{ acl_entry.string }}
+{#     check if the number of ACL entries is below the limit #}
+{%     set counter.acle_number = counter.acle_number + 1 %}
+{%     if counter.acle_number > counter.acle_max %}
```
ClausHolbechArista: indentation is wrong for this `if` block

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{%     set counter.acle_number = 0 %}
+{%     set counter.acle_max = ip_access_lists_max_acle | arista.avd.default(10000) %}
+{# START ACL CYCLE: walk over the list of defined ACLs #}
+{%     for acl in ip_access_lists %}
+ip access-list {{ acl.name }}
+{%         if acl.counters_per_entry is arista.avd.defined(true) %}
+   counters per-entry
+{%         endif %}
+{#     START ACL_ENTRY CYCLE: walk over the list of ACL entries #}
+{%     for acle in acl.entries %}
```
ClausHolbechArista: indentation is wrong for this `for` block.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{%             endif %}
+{#             * add log #}
+{%             if acle.log is arista.avd.defined(true) %}
+{%                 set acl_entry.string = acl_entry.string ~ 'log' %}
+{%             endif %}
+{%         endif %}
+{#     - always remove trailing spaces at the end of the string #}
+{%         set acl_entry.string = acl_entry.string.rstrip() %}
+   {{ acl_entry.string }}
+{#     check if the number of ACL entries is below the limit #}
```
ClausHolbechArista: only perform this check if the limit is set. Also ignore remark entries. Otherwise it would be simpler to just take length of the `entries` list and add to the counter.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{%     for acl in ip_access_lists %}
+ip access-list {{ acl.name }}
+{%         if acl.counters_per_entry is arista.avd.defined(true) %}
+   counters per-entry
+{%         endif %}
+{#     START ACL_ENTRY CYCLE: walk over the list of ACL entries #}
+{%     for acle in acl.entries %}
+{#     Build ACL entry: #}
+{#     - init ACL entry string #}
+{%             set acl_entry = namespace(string='') %}
```
ClausHolbechArista, line 21: there is really no need for a namespace here. You stay within the same `for`, so you can just use a regular variable. Ex `acl_entry_cli` like elsewhere.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
+{# eos - extended access-lists - improved data model #}
+{% if ip_access_lists is arista.avd.defined %}
+{# set maximum number of ACL entries allowed, the playbook will fail if number is over the limit #}
+{%     set counter = namespace() %}
+{%     set counter.acle_number = 0 %}
+{%     set counter.acle_max = ip_access_lists_max_acle | arista.avd.default(10000) %}
+{# START ACL CYCLE: walk over the list of defined ACLs #}
+{%     for acl in ip_access_lists %}
+ip access-list {{ acl.name }}
```
ClausHolbechArista, line 13: missing variable tests for `name` and `entries`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,123 @@
...
+{%                 set acl_entry.string = acl_entry.string ~ 'log' %}
+{%             endif %}
+{%         endif %}
+{#     - always remove trailing spaces at the end of the string #}
+{%         set acl_entry.string = acl_entry.string.rstrip() %}
+   {{ acl_entry.string }}
+{#     check if the number of ACL entries is below the limit #}
+{%     set counter.acle_number = counter.acle_number + 1 %}
+{%     if counter.acle_number > counter.acle_max %}
+   {{ ('OK' if false) | mandatory('The number of ACL entries is above defined maximum!') }}
```
ClausHolbechArista: This is not very clear. Please find a cleaner way to call `mandatory`. Ex. `dummy_undefined_variable | mandatory(...)`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-access-lists.j2
```diff
@@ -0,0 +1,10 @@
+{% if ip_access_lists is arista.avd.defined %}
+
+## Extended Access-lists Device Configuration (Improved Data Model)
```
ClausHolbechArista: This is the lazy documentation :) I understand that the value of a table will be limited.
Please change the heading to `### IP Access-lists Summary` and then add another heading above the configuration with `### IP Access-lists Configuration` and a `## IP Access-lists` at the top. Don't include the `improved data model` in the device documentation. We have that covered in the readme.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ankudinov: Resolved. Inventing the correct default was an impossible task anyway, removing it an making the var optional making the logic a lot more straightforward.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: fixed at the cost of a few extra grey hairs =)

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: fixed

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,124 @@
...
+{#                 * add dscp #}
+{%                 if acle.dscp is arista.avd.defined %}
+{%                     set acl_entry.string = acl_entry.string ~ 'dscp ' ~ acle.dscp | lower ~ ' ' %}
+{%                 endif %}
+{#                 * add log #}
+{%                 if acle.log is arista.avd.defined(true) %}
+{%                     set acl_entry.string = acl_entry.string ~ 'log' %}
+{%                 endif %}
+{%             endif %}
+{#             - always remove trailing spaces at the end of the string #}
```
ClausHolbechArista: In all other templates we try to implement spacing by adding the space _before_ the keyword you add. This way we ensure that the `acl_entry.string` is always clean and ready to be printed. Please adjust here and remove the rstrip

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: The reason for namespace is inner for loops, for ex. tcp ports. With a regular variable the ports will not be added to the string. Namespace is a simple way to guarantee consistency in all cases. @ClausHolbechArista

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: resolved

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: resolved

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: @ClausHolbechArista The reason for space after the keyword is acle.sequence, that can be present or not. If space will be added before the keyword, that will come at the cost of additional if conditions before acle.remark and acle.entry. While it's not a huge cost, IMO current logic is easier and is not breaking AVD standards heavily.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: closing, as that is discussed in the rstrip related comment above

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-access-lists.j2
ankudinov: resolved

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -186,6 +187,16 @@ Requirements are located here: [avd-requirements](../../README.md#Requirements)
 
 #### IP Extended Access-Lists
 
+AVD currently supports 2 different data models for extended ACLs:
+
+- Legacy (for compatibility with existing designs, to avoid breaking changes)
```
tgodaA, line 192: compatible with existing designs

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -186,6 +187,16 @@ Requirements are located here: [avd-requirements](../../README.md#Requirements)
...
+- Improved extended IP ACL data model
+
+Both data models can coexists without conflicts, as differet keys are used: `access_lists` vs `ip_access_lists`.
+Access list names must be unique.
+
+Legacy data model is supports simplified ACL definition with `sequence_number` -> `action_string`:
```
tgodaA: Please rephrase this sentence. It is a bit hard to understand. Typo: ```is supports```.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -202,6 +213,53 @@ access_lists:
         action: "< action as string >"
 ```
 
+Improved data model has a more sophisticated design documented below:
```
tgodaA, line 216: Improved data model has a more sophisticated design, documented below:

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -202,6 +213,53 @@ access_lists:
...
+      # remark entry
+      - sequence: < acl entry sequence number >  # optional
+        # NOTE: if remark is defined, other keys in acl entry will be ignored
+        remark: "< Comment, up to 100 characters >"
+      # normal entry
+      - sequence: < acl entry sequence number >  # optional
+        action: "< permit | deny >"  # required
+        protocol: "< ip | tcp | udp | icmp | other protocol name or number >"  # required
+        # NOTE: A.B.C.D without a mask means host
+        source: "< any | A.B.C.D/E | A.B.C.D >"  # required
```
tgodaA, line 232: If we support ```source: < A.B.C.D >``` then we need another knob to provide the ```Destination Wildcard bits```. Same for the destination. Either we remove this option or add another knob to support the mask.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-access-lists.j2
```diff
@@ -0,0 +1,19 @@
+{% if ip_access_lists is arista.avd.defined %}
+
+## IP Access-lists
+
+### IP Access-lists Summary
+
```
tgodaA, line 7: Please move this blank line (6) below if (between line 7 and 8) and else (between line 9 and 10) conditions. So if anyone adds their snippet to this template in future, the consistency will be maintained.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,130 @@
+{# eos - extended access-lists - improved data model #}
+{% if ip_access_lists is arista.avd.defined %}
+{#     set maximum number of ACL entries allowed, the playbook will fail if number is over the limit #}
+{%     set counter = namespace() %}
+{%     set counter.acle_number = 0 %}
+{#     START ACL CYCLE: walk over the list of defined ACLs #}
+{%     for acl in ip_access_lists %}
+{%         if not (acl.name is arista.avd.defined(fail_action='warning', var_name='ip_access_lists[*].name')) %}{# break cycle if not defined #}{% continue %}{% endif %}
```
ClausHolbechArista: Please avoid multiple jinja statements on the same line. Break this into multiple lines. Same elsewhere.
Also as we talked about, we don't normally do all these warnings. I will let @carlbuchmann and @tgodaA chip in on that.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
```diff
@@ -0,0 +1,130 @@
...
+{%                 set acl_entry.string = acl_entry.string ~ acle.sequence ~ ' ' %}
+{%             endif %}
+{#             - add action or remark #}
+{#             if remark is defined, add remark + text only #}
+{%             if acle.remark is arista.avd.defined %}
+{%                 set acl_entry.string = acl_entry.string ~ 'remark ' ~ acle.remark %}
+{#             if no remark defined, build normal rule #}
+{%             else %}
+{#                 * add action #}
+{%                 if not (acle.action is arista.avd.defined(fail_action='warning', var_name='ip_access_lists[*].entries[*].action')) %}{# break cycle if not defined #}{% continue %}{% endif %}
```
ClausHolbechArista: All this error handling would be better handled in the beginning of each cycle for ACL entries, so I suggest that you gather all the required variables for each entry and test for those. It will make it simpler to follow the code and more in line with all our other templates. You can split it over multiple lines like this:
```jinja
{% if acle.action is arista.avd.defined(fail_action='warning', var_name='ip_access_lists[*].entries[*].action') and
      acle.protocol is arista.avd.defined(fail_action='warning', var_name='ip_access_lists[*].entries[*].protocol')) %}
```
and then enclose all the code into the `if` block

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: The code automatically inserts `host` before the IP if only `A.B.C.D` is given.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ClausHolbechArista: It is fine to do this for `sequence` since it is a prefix, but for everything else if should be done before. Please adjust.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ankudinov: @tgodaA It's also documented in the doc and IMO makes the data model easier, at least for humans. Suggest closing this.

tgodaA: Sorry, I didn't see the option ```host A.B.C.D``` assuming mask is a must. My bad!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: ack. part of elif block for non-remark rules. resolved

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: Sure. It's definitely not a problem to drop the warning and leave just arista.avd.defined(), but IMO it's better than failing silently. I know that it will be covered by schema, but I don't think it hearts to have the warning log right now. @carlbuchmann and @tgodaA let me know what do you think?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: resolved

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
tgodaA: I like same warnings in log files rather than too many warnings popped up when the playbook is running.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-access-lists.j2
```diff
@@ -0,0 +1,20 @@
...
+## IP Access-lists
+
+### IP Access-lists Summary
+{%     if ip_access_lists_max_entries is arista.avd.defined %}
+
+> The maximum number of ACL entries allowed to be provisioned per switch: {{ ip_access_lists_max_entries }}
+{%     else %}
+
+> The maximum number of ACL entries allowed to be provisioned per switch is not limited.
+> Set `ip_access_lists_max_entries` to configure a guard if required.
```
ClausHolbechArista: This should not be in the device documentation. I suggest to remove lines 9-12 and move the `if` in line 6 up to line 4 (above the blank line)

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: The improved

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -186,6 +187,16 @@ Requirements are located here: [avd-requirements](../../README.md#Requirements)
...
 
+AVD currently supports 2 different data models for extended ACLs:
+
+- Legacy (for compatibility with existing designs, to avoid breaking changes)
+- Improved extended IP ACL data model
+
+Both data models can coexists without conflicts, as differet keys are used: `access_lists` vs `ip_access_lists`.
+Access list names must be unique.
+
+Legacy data model supports simplified ACL definition with `sequence_number` to `action_string` mapping:
```
ClausHolbechArista: The legacy

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: I suggest changing to:

- The legacy `access_lists` data model, for compatibility with existing deployments
- The improved `ip_access_lists` data model, for access to more EOS features

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -186,6 +187,16 @@ Requirements are located here: [avd-requirements](../../README.md#Requirements)
 
 #### IP Extended Access-Lists
 
+AVD currently supports 2 different data models for extended ACLs:
+
+- Legacy (for compatibility with existing designs, to avoid breaking changes)
+- Improved extended IP ACL data model
+
+Both data models can coexists without conflicts, as differet keys are used: `access_lists` vs `ip_access_lists`.
```
ClausHolbechArista: Typo: different

ClausHolbechArista (commented):
almost there

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ip-access-lists.j2
ankudinov: ok. It will fail silently and we'll wait for schema.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-access-lists.j2
```diff
@@ -0,0 +1,16 @@
+{% if ip_access_lists is arista.avd.defined %}
+
+## IP Access-lists
+{%     if ip_access_lists_max_entries is arista.avd.defined %}
+
+### IP Access-lists Summary
+
+> The maximum number of ACL entries allowed to be provisioned per switch: {{ ip_access_lists_max_entries }}
```
ClausHolbechArista: Sorry for not catching this earlier. This should be a `-` in the beginning instead of `>`

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -202,6 +213,53 @@ access_lists:
...
+        source_ports: ["< tcp/udp port name or number >",]  # optional
+        # NOTE: A.B.C.D without a mask means host
+        destination: "< any | A.B.C.D/E | A.B.C.D >"  # required
+        destination_ports_match: "< eq | gt | lt | neq | range >"  # eq is default
+        destination_ports: ["< tcp/udp port name or number >",]  # optional
+        tcp_flags: ["< tcp flag name >",]  # optional
+        fragments: < true | false >  # optional, match non-head fragment packets
+        log: < true | false >  # optional, log matches against this rule
+        ttl: < <0-254> TTL value >  # optional
+        ttl_match: "< eq | gt | lt | neq >"  # optional
```
tgodaA: eq is default

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -202,6 +213,53 @@ access_lists:
...
+      - sequence: < acl entry sequence number >  # optional
+        # NOTE: if remark is defined, other keys in acl entry will be ignored
+        remark: "< Comment, up to 100 characters >"
+      # normal entry
+      - sequence: < acl entry sequence number >  # optional
+        action: "< permit | deny >"  # required
+        protocol: "< ip | tcp | udp | icmp | other protocol name or number >"  # required
+        # NOTE: A.B.C.D without a mask means host
+        source: "< any | A.B.C.D/E | A.B.C.D >"  # required
+        source_ports_match: "< eq | gt | lt | neq | range >"  # eq is default
```
tgodaA: LGTM! just a minor suggestion. Please use our common way to set default value, ```default -> eq```. same for destination_ports_match, ttl_match, vlan_inner.

ClausHolbechArista: agree

ankudinov: IMO, it can not be counted as common way until schema is releases and comment is more readable. But as discussed, it's minor. Agree and resolving.

tgodaA: Reasonable!

tgodaA (approved):
LGTM!

PR #1429, opened 2022-01-04
Feat(eos_cli_config_gen): Add support for as-path filter

ccsnw:
## Change Summary

- Add support for as-path filters

## Related Issue(s)

Fixes N/A

## Component(s) name

## Proposed changes
- Allows to filter based on as-path
- Add documentation (using backticks to escape possible regex, otherwise might interfere with markdown)

```
as_path:
  regex_mode: < asn | string >
  access_lists:
    - name: < access_list_name_1 >
      entries:
        - type: < permit | deny >
          match: "< regex to match >"
          origin: < "any" | "egp" | "igp" | "incomplete" | default -> "any" >
```

## How to test
- Added molecule test / artifcats

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -645,6 +646,18 @@ match_list_input:
           match_regex: < match string >
 ```
 
+#### AS Paths
+
+```yaml
+as_path:
+  regex_mode: < asn | string >
+  access_lists:
```
ClausHolbechArista, line 649: Since this is a new data model, we should use list of dicts instead. 
Also you can have multiple entries in the same list.
```eos
ip as-path access-list test permit 123 any
ip as-path access-list test permit 321 any
```
My suggestion:
```yaml
as_path:
  regex_mode: < asn | string >
  access_lists:
    - name: < access_list_name_1 >
      entries:
        - type: < permit | deny >
          match: "< regex to match >"
          origin: < "any" | "egp" | "igp" | "incomplete" | default -> "any" >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/as-path.j2
```diff
@@ -0,0 +1,19 @@
+{# eos - as path #}
+{% if as_path is arista.avd.defined %}
+!
+{%     if as_path.regex_mode is arista.avd.defined %}
+ip as-path regex-mode {{ as_path.regex_mode }}
+{%     endif %}
+{%     if as_path.access_lists is arista.avd.defined %}
+{%         for as_path_access_list in as_path.access_lists %}
```
tgodaA: Use ```natural_sort``` filter. Same for all for loops in this template.

You can modify the line 7-10 as below (correct the indentation as required):

```
{% for as_path_access_list in as_path.access_lists | arista.avd.naural_sort if as_path_access_list.name is arista.avd.defined %}
{%     for as_path_access_list_entry in as_path_access_list.entries | arista.avd.naural_sort %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/as-path.j2
```diff
@@ -0,0 +1,26 @@
...
+
+### AS Path Lists Summary
+{%     if as_path.regex_mode is defined %}
+
+AS Path Regex Mode is **{{ as_path.regex_mode }}**.
+{%     endif %}
+
+| List Name | Type | Match | Origin |
+| --------- | ---- | ----- | ------ |
+{%     for as_path_access_list in as_path.access_lists %}
```
tgodaA: Use natural_sort filter. Also we can remove additional if check of that list as the filter will take care of that.
https://avd.sh/en/releases-v3.x.x/plugins/index.html#natural_sort-filter

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/as-path.j2
ccsnw: Thanks Tony, let me add the filter. I think I'll make use of it by default in the future.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/as-path.j2
```diff
@@ -0,0 +1,24 @@
...
+{%     if as_path.regex_mode is defined %}
+
+AS Path Regex Mode is **{{ as_path.regex_mode }}**.
+{%     endif %}
+
+| List Name | Type | Match | Origin |
+| --------- | ---- | ----- | ------ |
+{%     for as_path_access_list in as_path.access_lists | arista.avd.natural_sort %}
+{%         for as_path_access_list_entry in as_path_access_list.entries | arista.avd.natural_sort %}
+| {{ as_path_access_list.name }} | {{ as_path_access_list_entry.type }} | {{ as_path_access_list_entry.match | replace('|', '\|') }} | {{ as_path_access_list_entry.origin | arista.avd.default('any') }} |
```
tgodaA, line 17: Line 13 -> natural_sort will only take care if ```as_path.access_lists``` is not none.
Line 14 -> natural_sort will only take care if ```as_path_access_list.entries``` is not none. If entries has only ```-type: <> then it's not none, then we need handle other knobs, so add an if condition above this line (15)to not add entry if both of these values doesn't exist```

So we still need to take care of ```as_path_access_list.name```, ```as_path_access_list_entry.type and match```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/as-path.j2
```diff
@@ -0,0 +1,24 @@
...
+
+### AS Path Lists Summary
+{%     if as_path.regex_mode is defined %}
+
+AS Path Regex Mode is **{{ as_path.regex_mode }}**.
+{%     endif %}
+
+| List Name | Type | Match | Origin |
+| --------- | ---- | ----- | ------ |
+{%     for as_path_access_list in as_path.access_lists | arista.avd.natural_sort %}
```
tgodaA: after natural_sort add ``` if as_path_access_list.name is arista.avd.defined %}```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/as-path.j2
ClausHolbechArista: We should _not_ use `natural_sort` on `entries` on ACLs. The order is important, and we should let the user decide on which order the entries will be configured. So please revert.

ccsnw: got it. let me revert this but add a condition to make sure that at least type and match are defined.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/as-path.j2
ClausHolbechArista: You should keep the changes to the outer for-loop. so add the `natural_sort` and remove the check in line 7. Same for documentation. It was only the `entries` which should _not_ use the sorting, since it will change the order.

There is another trick that you can use instead of natural_sort, if you don't want to sort, but still want to include the variable check on the `for` loop.
```jinja
{% for entry in entries | arista.avd.default([]) %}
```

But what you have now is fine, since you also need to check for name.

ccsnw: just to make sure I understand 100% correctly: the whole point is to include the check if as_path_access_list.name is defined in the for loop, right? so basically safe an explicit if-condition for it?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/as-path.j2
```diff
@@ -0,0 +1,17 @@
+{# eos - as path #}
+{% if as_path is arista.avd.defined %}
+!
+{%     if as_path.regex_mode is arista.avd.defined %}
+ip as-path regex-mode {{ as_path.regex_mode }}
+{%     endif %}
+{%     for as_path_access_list in as_path.access_lists | arista.avd.naural_sort if as_path_access_list.name is arista.avd.defined %}
```
tgodaA: natural_sort typo!!

ClausHolbechArista (approved):
LGTM

tgodaA (approved):
LGTM!

PR #1285, opened 2021-10-26
Feat(eos_cli_config_gen): Link tracking in EVPN environments

tgodaA:
## Change Summary

- Support for Link Tracking Groups to blackhole the traffic when the links are down
- Add upstream and downstream interfaces
- Documentation 

## Related Issue(s)

Fixes #1043 

## Component(s) name

`arista.avd.eos_cli_config_gen`

## Proposed changes

Added the below data model
#### link-tracking-groups
```yaml
link_tracking_groups:
  - name: < group_name >
    links_minimum: < 1-100000 >
    recovery_delay: < 0-3600 >
```

#### ethernet-interfaces
```yaml
ethernet_interfaces:
  <Ethernet_interface_1 >:
    link_tracking_groups:
      - name: < group_name >
        direction: < upstream | downstream >
```

#### portchannel-interfaces
```yaml
port_channel_interfaces:
  < Port-Channel_interface_1 >:
    link_tracking_groups:
      - name: < group_name >
        direction: < upstream | downstream >
```

Should we add this to eos_designs to automatically enable this feature by default for EVPN MH?

## How to test
Molecule examples

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #1441, opened 2022-01-18
Feat(eos_designs): Support EVPN multihoming features link-tracking and lacp port-id range

### LINKED ISSUE #1043: Support for link tracking in EVPN environments.
## Issue Type

- [X] New Feature

## Summary

In most environments that run EVPN-MH or sometimes EVPN-MLAG we need the capability that if upstream spine links will go down that the MLAG port-channel to downstream interfaces (servers) also goes down and does not forward traffic where it would then be black holed. 

A common CLI would look like the following.

link tracking group FOO
   recovery delay 300
!

! upstream interface
interface Ethernet3
   link tracking group FOO upstream
!

! downstream interface
interface Port-Channel1
   switchport access vlan 30
   !
   evpn ethernet-segment
      identifier 0000:0000:0000:0000:3030
      route-target import 30:30:30:30:30:30
   lacp system-id 3030.3030.3030
   link tracking group FOO downstream
!

Where the Ethernet 3 interface would be the spine uplink and the port-channel1 would be the downstream connection to the server.

I would imagine this would have a dictionary containing the following..

## Component Name

eos_cli_config_gen/templates/eos/port-channel-interfaces.j2
eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
+ eos_cli_config_gen/templates/eos/link-tracking.j2
eos_cli_config_gen/templates/documentation/port-channel-interfaces.j2
eos_cli_config_gen/templates/documentation/ethernet-interfaces.j2
+ eos_cli_config_gen/templates/documentation/link-tracking.j2

I would imagine the model would look like the following.

#### link-tracking
```yaml
link-tracking:
    - group_name: string
      delay: int32 
```

#### ethernet-interfaces
```yaml
ethernet_interfaces:
  <Ethernet_interface_1 >:
    link_tracking: <true | false>
      - name: string 
        direction: upstream | downstream
```

#### portchannel-interfaces
```yaml
port_channel_interfaces:
  < Port-Channel_interface_1 >:
    link_tracking: <true | false>
      - name: string 
        direction: upstream | downstream
```

This is very common within EVPN-MH environments if agreed upon please suggest this model or we can put in a PR to add to the ethernet/port-channel interfaces and potentially add a new j2 template for link-tracking.

tgodaA:
## Change Summary

The PR #1285 and #1301 were related to the eos_cli_config_gen role and the same features were supported in eos_designs now. These features are related to the EVPN multihoming, 1) Link-Tracking Group 2) LACP Port-id

[Design Guide](https://www.arista.com/custom_data/downloads/?f=/support/download/DesignGuides/EVPN_Multihoming_Design_Guide.pdf) Page 16 and 17

## Related Issue(s)

Fixes #1043 #1074 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Need to support: 1) Link tracking groups 2) Different Lacp Port-id in different vtep groups

A) Fabric variables:
To create Link tracking groups under any type of node as well as adds all the switch uplink interfaces as upstream interfaces in that group:
```
<node_type_keys>:
  default:
    link_tracking:
      enabled: < true | false >
      # Optional default group name -> "LT_GROUP1", recovery_delay -> 500
      groups:
        - name: < tracking_group_name >
          recovery_delay: < 0-3600 | default -> 500 >
          links_minimum: < 1-100000 >
```

Lacp Port-id range must be set in Vteps only (l3leaf) to make sure the end-point can differentiate the different switches(l3leafs), it is connected to in the ES-LAG group.
To support different lacp port-id in different Vtep groups:
```
<node_type_keys>:
  default:
    lacp_port_id_range:
      enabled: < true | false | default -> false >
      # Recommended size > = number of ports in the switch.
      size: < 1-65535 | default -> 128 >
      # Offset is used to avoid overlapping port-id ranges of different switches | Optional
      # Useful when a "connected-endpoint" is connected to switches in different "node_groups".
      offset: < offset_for_lacp_port_id_range | default -> 0 >
```

B) On connected-endpoints we set up the link tracking group downstream interfaces:
```
link_tracking:
  enabled: < true | false >
  # The default group name is taken from fabric variable of the switch, link_tracking.groups[0].name with default value being "LT_GROUP1".
  # Optional if default link_tracking settings are configured on the node.
  name: < tracking_group_name >
```

## How to test
Molecule

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -344,3 +344,27 @@ defaults <- node_group <- node_group.node <- node
...
+```yaml
+< node_type_key >:
+
+  defaults:
+    # This configures the Link Tracking Group on a switch as well as adds the p2p-uplinks of the switch as the upstream interfaces.
+    link_tracking:
+      enabled: < true | false | default -> false >
+      name: < name_of_the_tracking_group | default -> "LT_GROUP1" >
+      recovery_delay: < 0-3600 Seconds | default -> 500 >
+      links_minimum: < 1-100000 | Optional >
```
ClausHolbechArista: hard to read the `| Optional` in here. I would just do `links_minimum: < 1-100000 >`

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -344,3 +344,27 @@ defaults <- node_group <- node_group.node <- node
...
+
+```yaml
+< node_type_key >:
+
+  defaults:
+    # This configures the Link Tracking Group on a switch as well as adds the p2p-uplinks of the switch as the upstream interfaces.
+    link_tracking:
+      enabled: < true | false | default -> false >
+      name: < name_of_the_tracking_group | default -> "LT_GROUP1" >
+      recovery_delay: < 0-3600 Seconds | default -> 500 >
```
ClausHolbechArista: avoid the word "seconds" in here unless that is supposed to go into the value.

tgodaA:
I have added new molecule scenario. Need some help with the review. If everything is good, let me know how to add it to the workflow. Thank you!

ClausHolbechArista:
> I have added new molecule scenario. Need some help with the review. If everything is good, let me know how to add it to the workflow. Thank you!

Please do not add extra molecule scenarios. We are trying to get fewer, since the CI runtime is getting very long (and expensive). Best would be to just add the data you need in a hostvars file and add the device(s) under the unit_test scenario. Avoid using any existing groups. Just add a new `UNIT_TESTS` group under `all` and supply all the required vars in hostvars. Ping me if we should work this out together.

tgodaA:
@ClausHolbechArista , I saw an example of UNIT_TESTS in eos_designs_unit_tests. But for the multihoming example, I need the whole spine-leaf-end_point architecture to test link_tracking_groups and just a l3leaf to test lacp_port_id_range. But I am working to move this to eos_designs_unit_tests.

tgodaA:
@ClausHolbechArista let me know if these artifacts looks good to you. I will try to remove few more knobs in the example to make it a unit test only for those new knobs.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -397,3 +397,37 @@ switch:
...
+    - name: LT_GROUP1
+      recovery_delay: 500
+{%     endif %}
+{% endif %}
+
+{% if switch.vtep is arista.avd.defined(true) %}
+{#     switch.lacp_port_id #}
+{%     if switch_data.combined.lacp_port_id_range.enabled is arista.avd.defined(true) %}
+{%         set group_length = switch_data.node_group.nodes | arista.avd.default([inventory_hostname]) | length %}
+{%         set port_range = switch_data.combined.lacp_port_id_range.size | arista.avd.default(128) %}
```
xaviramon: Any specific reason to choose 128 as a default? In my opinion, 100 is much more human-friendly, specially for troubleshooting. I also think 100 AA ports per device cover majority of scenarios

tgodaA: These ports are not the switch ports. These are the lacp port-id. Recommended practice is to keep the lacp port-id range = # ports on the switch(not mandatory but considering the worst case). This doesn't allocate the AA ports on switch but allocates the port-id from the range(1-128), to those actual switch ports which are currently on LAG.

Page 16 of the design guide: 
image

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -344,3 +344,31 @@ defaults <- node_group <- node_group.node <- node
     # VLAN number assigned to Inband Management SVI on l2leafs in default VRF.
     inband_management_vlan: < vlan-id | Default -> 4092 >
 ```
+
+### EVPN Multihoming
+
+```yaml
```
ClausHolbechArista: These new options are not specific to EVPN Multihoming, so it should be moved or renamed to a more generic heading

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -67,6 +67,11 @@ ethernet_interfaces:
     channel_group:
       id: {{ channel_group_id }}
       mode: "{{ adapter_settings.port_channel.mode }}"
+{%                     elif adapter_settings.link_tracking.enabled is arista.avd.defined(true) and switch.link_tracking_groups is arista.avd.defined %}
```
ClausHolbechArista: Even though this logic works, I would prefer to have it in a separate `if` block to make it more clear. Then we don't need the comment either.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -397,3 +397,37 @@ switch:
...
+{# EVPN multihoming configuration on VTEPs #}
+{# switch.link_tracking #}
+{% if switch_data.combined.link_tracking.enabled is arista.avd.defined(true) %}
+{%     if switch_data.combined.link_tracking.groups is arista.avd.defined %}
+  link_tracking_groups:
+{#         Do not sort the group as the first group will be the default group for the end-points #}
+{%         for lt_group in switch_data.combined.link_tracking.groups %}
+{%             if lt_group.name is arista.avd.defined %}
+    - name: {{ lt_group.name }}
+      recovery_delay: {{ lt_group.recovery_delay | arista.avd.default(500) }}
```
ClausHolbechArista: I think we should default to 0 / no delay. If this is seconds, it will have severe impacts on the network.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -397,3 +397,37 @@ switch:
...
+{%             endif %}
+{%         endfor %}
+{%     else %}
+  link_tracking_groups:
+    - name: LT_GROUP1
+      recovery_delay: 500
+{%     endif %}
+{% endif %}
+
+{% if switch.vtep is arista.avd.defined(true) %}
```
ClausHolbechArista: This feature should not rely on vtep. It should also work on l2leaf etc.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/topology/p2p-uplinks.j2
```diff
@@ -33,6 +33,14 @@ topology:
 
       peer_ip_address: {% include switch.ip_addressing.p2p_uplinks_peer_ip %}
 
+{%             endif %}
+{%             if switch.link_tracking_groups is arista.avd.defined %}
```
ClausHolbechArista, line 37: This should also be added to port-channel uplink type

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/ethernet-interfaces.j2
```diff
@@ -37,6 +37,9 @@ ethernet_interfaces:
     isis_metric: 50
     isis_network_point_to_point: true
 {%         endif %}
+{%         if link.link_tracking_groups is arista.avd.defined %}
```
ClausHolbechArista, line 40: This should also be added for `underlay_l2` type below

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
tgodaA: I can move them to two groups. Link-tracking and Lacp port-id and mention there that these knobs are useful for multihoming.

Alternatively, we can have a generic heading like (### Other Supported Configuration) and put any generic knob supported in eos_designs in here and then comment to explain their essence.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
tgodaA: Then I need to add if check for port-channel doesn't exist. ```if adapter_settings.port_channel is not arista.avd.defined``` and then add other logic with ```and```

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
tgodaA: No, recovery_delay is required as far as I understand. 500 was the value used in document. We are allowing the BGP to come up and then LAGs to work to avoid the traffic arrival before the network is actually up. I understand you are presenting the generic config, but I think mostly this knob is going to be used for multihoming as far as I guess.


image

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
tgodaA: How about SPINE ?

tgodaA: I think this is an l2 feature. Will change it accordingly.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/topology/p2p-uplinks.j2
tgodaA: will do.

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/ethernet-interfaces.j2
tgodaA: This is for complete MLAG system right?

But anyway, I will add there.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
tgodaA: Even for a regular use case, I think there should be a delay to allow for route convergence unless if static routing is used.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
tgodaA: Used the heading ```Other Supported Configuration```.

tgodaA:
Tested link_tracking on a l2leaf. It works now. Couldn't test on SPINE though!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/documentation/devices/MHGRP1_LEAF1.md
```diff
@@ -0,0 +1,720 @@
+# MHGRP1_LEAF1
```
ClausHolbechArista: Leftover file. Please remove. Same for 2 others.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -344,3 +344,33 @@ defaults <- node_group <- node_group.node <- node
...
+      # Link Tracking Groups | Optional
+      # By default a single group named "LT_GROUP1" is defined with default values. Any groups defined under "groups" will replace the default.
+      groups:
+        - name: < tracking_group_name >
+          recovery_delay: < 0-3600 | default -> 500 >
+          # Optional
+          links_minimum: < 1-100000 >
+
+    # This will generate the "lacp port-id range", "begin" and "end" values based on node "id" and the number of nodes in the "node_group".
+    # It is recommended to use this knob if there are any downlink lag interfaces on a VTEP in a Mutihomed network.
```
ClausHolbechArista: The second line is redundant. The third is more to the point.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
ClausHolbechArista: I suggest to move it up under the genereric section.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
ClausHolbechArista: For l3 devices (check switch.underlay_router) I think the default should be 300 to match the default mlag reload delay. For pure l2 devices where we just have a port-channel uplink, IMO the delay should be minimum. Maybe 10 seconds. Just for spanning-tree to settle.

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/ethernet-interfaces.j2
```diff
@@ -59,6 +62,9 @@ ethernet_interfaces:
 {%             if p2p_uplinks_qos_profile is arista.avd.defined %}
     service_profile: {{ p2p_uplinks_qos_profile }}
 {%             endif %}
+{%             if link.link_tracking_groups is arista.avd.defined %}
```
ClausHolbechArista, line 65: I was wrong to have you add it here. We should only set the port-channel interface as the uplink. Not the individual members.

tgodaA: This will only add the config if channel_group_id is not defined (that means, port-channle doesn't exist).

image

ClausHolbechArista: got it. Keep it as is then. Thanks

tgodaA:
Mlag reload delay:

Here is the info I found on https://www.arista.com/en/um-eos/eos-multi-chassis-link-aggregation?searchword=mlag%20timer

image

-----------

image

------------
image

ClausHolbechArista (approved):
LGTM! Thank you for this great contribution.

carlbuchmann (approved):
LGTM!

PR #1451, opened 2022-01-24
Fix(eos_cli_config_gen): Fix name-servers jinja templates to allow non-vrf nameservers

ccsnw:
## Change Summary

name-servers.j2 is missing a condition to handle non-vrf (vrf default) name servers.

In previous versions the code existed (inline) and I think it was accidently removed during refactoring.

## Related Issue(s)

Fixes #N/A

## Component(s) name

## Proposed changes

add condition to also cover output rendering of no vrf is defined

model unchanged

## How to test
molecule test only covers scenario with vrf, so maybe this is why we didn't catch the issue?

### User Checklist

- N/A

### Repository Checklist

- [O] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/name-servers.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/name-servers.j2
@@ -7,7 +7,7 @@
 | Name Server | Source VRF |
 | ----------- | ---------- |
 {%     for node in name_server.nodes %}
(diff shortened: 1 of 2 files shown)


ClausHolbechArista (approved):
LGTM

tgodaA (approved):
LGTM!

That's a good catch. Thank you for the contribution

PR #1457, opened 2022-01-26
Feat(plugins): Add debug and  file options to yaml_template_to_facts

ClausHolbechArista:
## Change Summary

Add debug and  file options to yaml_template_to_facts

## Related Issue(s)

This is a prerequisite for future PRs changing the task layout of AVD to improve speed.

## Component(s) name

`arista.avd.yaml_templates_to_facts`

## Proposed changes

Adding new optional arguments to `yaml_templates_to_facts`:
- dest (Write output to this file as well as returning facts) - autodetecting yml or json from file suffix
- template_output (Run templater over the output before returning - allowing for inline jinja)
- debug (Collect timestamps of all actions performed, for troubleshooting performance issues)

## How to test

- No changes to molecule or existing unit tests expected.
- New unit tests will be added to cover the new features

### User Checklist

- [x] Add enhancements to action module
- [ ] Update base module argspec
- [ ] Update documentation
- [ ] Create new unit tests for the new features

### Repository Checklist


ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -8,6 +8,8 @@
 from ansible.plugins.filter.core import combine
 from ansible.plugins.loader import lookup_loader
 from ansible_collections.arista.avd.plugins.module_utils.strip_empties import strip_null_from_data
+from datetime import datetime
+import yaml
 
 
 class ActionModule(ActionBase):
```
tgodaA, line 15: In line 16, is it ok to set ```task_vars = {}``` instead of line 17,18

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -37,6 +39,12 @@ def run(self, tmp=None, task_vars=None):
                     raise AnsibleActionFail("The argument 'templates' is not a list")
             else:
                 raise AnsibleActionFail("The argument 'templates' must be set")
+
+            output_filename = self._task.args.get("output_filename",False)
```
tgodaA: Where is the file going to be saved? I think we need to specify that in the output_filename. When I tested this feature, the file(s) is in the playbook directory.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -37,6 +39,12 @@ def run(self, tmp=None, task_vars=None):
...
+            output_format = self._task.args.get("output_format","json")
+            template_output = self._task.args.get("template_output",False)
```
tgodaA, line 44: Unable to understand the essence of this. So couldn't think of a ansible-test scenario for this.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
ClausHolbechArista: No, because if the module is called with the argument `task_vars=None` it would not take the default value. So this is a way of ensuring that both no value and the `None` value will lead to an empty dict.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
ClausHolbechArista: I should change this variable name from `output_filename` to `dest` to match the regular ansible naming for file desitnations. It can be a relative or absolute path, since this is just parsed on to the `dest` argument of the `copy` module.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
ClausHolbechArista: This becomes a bit "inception"-like. The point is if you have an input variable with a jinja2 expression, which cannot be resolved until after the data rendered in the template has been set.
Example:
Input variables:
```yaml
cvsourceip: "{{ management_interfaces.Vlan4092.ip_address | ipaddr('address') if type == 'l2leaf' else router_bgp.router_id }}"
```
Since `management_interfaces`and `router_bgp` are not set when the playbook is called, this jinja is just passed along raw. Leading to the output configuration containing the jinja2 expression instead of the intended value.

Today we work around this by writing the structured_config to a file using a template. This extra rendering step resolves the jinja expressions. Next task is importing the resulting file again, overwriting the facts set on the first run.

This new knob is a way of removing those two extra tasks from the role and running the extra templater inside the `yaml_templates_to_facts`. This makes it simpler to detect changes and setting the correct `changed` flag.

tgodaA: I got it, you told this previously.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -120,22 +130,25 @@ def run(self, tmp=None, task_vars=None):
...
                     'timestamps': {
                         'write_file': datetime.now()
                     }
                 }
 
-            if output_format in ["yml", "yaml"] :
-                write_file_result = self.write_file(yaml.dump(output, indent=2, sort_keys=False, width=130), output_filename, task_vars)
+            # Depending on the file suffix of 'dest' (default: 'json') we will format the data to yaml or just write the output data directly.
+            # The Copy module used in 'write_file' will convert the output data to json automatically.
+            if dest.split('.')[-1] or '' in ["yml", "yaml"] :
```
tgodaA: Why do we need ```or ''```? If someone doesn't provide any file extension, then the file name is the output for [-1]. If you are considering this scenario then we can remove this check.

FYI
image

ClausHolbechArista: correct. I will remove it.

KonikaChaurasiya-GSLab (commented):
LGTM...
 Added review comments

ansible_collections/arista/avd/tests/integration/targets/yaml_templates_to_facts/tests/test_with_template_output_false.yml
```diff
@@ -0,0 +1,21 @@
...
+      - result.changed == true
+      - result.ansible_facts is defined
+      - result.ansible_facts.key_1 is defined
+      - result.ansible_facts.key_1 == "value_1"
+      - result.ansible_facts.key_2 is defined
+      - result.ansible_facts.key_2[0] == '{'
+      - result.ansible_facts.key_2[1] == '{'
+      - result.ansible_facts.key_2[2:9] == ' key_1 '
+      - result.ansible_facts.key_2[9] == '}'
+      - result.ansible_facts.key_2[10] == '}'
```
KonikaChaurasiya-GSLab, line 21: Is there any specific reason to add assert for characters instead of for whole string?. Can we add assert like ansible_facts.key_2 == '{{  key_1 }}'

ansible_collections/arista/avd/tests/integration/targets/yaml_templates_to_facts/tests/test_with_debug.yml
```diff
@@ -0,0 +1,29 @@
...
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].template is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].template == "templates/template.j2"
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.starting is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.run_template is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.load_yaml is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.strip_empty_keys is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.combine_data is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.done is defined
+      - result.ansible_facts.avd_yaml_templates_to_facts_debug[-1].timestamps.run_template is defined
```
KonikaChaurasiya-GSLab: Looks like duplicate assert statement added by mistakenly. I think there is no need of last assert statement.

ClausHolbechArista: I will fix. Thanks

ansible_collections/arista/avd/tests/integration/targets/yaml_templates_to_facts/tests/test_with_template_output_false.yml
ClausHolbechArista: Ansible automatically tried to render any value that looks like jinja, so the {{ will trigger an error. Here I specifically want to test if the content is the string `{{ key_1 }}` and _not_ the value of key_1. I tried escaping etc, but couldn't find a way.

KonikaChaurasiya-GSLab: Got it. Thanks

KonikaChaurasiya-GSLab:
LGTM...

KonikaChaurasiya-GSLab (approved):

manuwelakanade (approved):
LGTM

PR #1470, opened 2022-01-31
Fix(eos_designs): Fixing template errors and invalid config when configuring cloudvision as a service

emilarista:
## Change Summary

Fixed bug causing cvp_instance_ip variable value containing "arista.io" to error out the template.

## Component(s) name

`arista.avd.eos_designs`

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
@@ -35,7 +35,7 @@ daemon_terminattr:
 {%         endfor %}
   cvauth:
     method: "token-secure"
-    key: "{{ cvp_ingestauth_key | arista.avd.default('/tmp/cv-onboarding-token') }}"
+    token_file: "{{ cvp_token_file | arista.avd.default('/tmp/cv-onboarding-token') }}"
 {%     endif %}
   cvvrf: {{ mgmt_interface_vrf }}
   smashexcludes: "{{ terminattr_smashexcludes }}"
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
@@ -8,8 +8,8 @@
 {%     if cvp_instance_ip is arista.avd.defined %}
 {%         if "arista.io" not in cvp_instance_ip %}
 {%             do cvp_on_prem.ips.append(cvp_instance_ip) %}
-{%         elif "arista.io" in cvp_ip %}
-{%             do cvaas.ips.append(cvp_ip) %}
+{%         elif "arista.io" in cvp_instance_ip%}
+{%             do cvaas.ips.append(cvp_instance_ip) %}
 {%         endif %}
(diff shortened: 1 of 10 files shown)


ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
```diff
@@ -8,8 +8,8 @@
 {%     if cvp_instance_ip is arista.avd.defined %}
 {%         if "arista.io" not in cvp_instance_ip %}
 {%             do cvp_on_prem.ips.append(cvp_instance_ip) %}
-{%         elif "arista.io" in cvp_ip %}
-{%             do cvaas.ips.append(cvp_ip) %}
+{%         elif "arista.io" in cvp_instance_ip %}
```
tgodaA: I understand this is already existing code. But my suggestion was:

As we already checked in line 9, if "arista.io" exists in cvp_instance_ip or not. Simply an ```else``` should work without any issue.

or

IMO, we can modify the line 9-12 to (Prefer to avoid not(logic) )

```
{%     if cvp_instance_ip is arista.avd.defined %}
{%         if "arista.io" in cvp_instance_ip %}
{%             do cvaas.ips.append(cvp_instance_ip) %}
{%         else %}
{%             do cvp_on_prem.ips.append(cvp_instance_ip) %}
```

tgodaA: If it's ok for everyone, we can modify the below logic as well.

ClausHolbechArista: If we change that, I would swap to `if "arista.io" in cvp_instance_ip` and `else` for the onprem.

emilarista: Implemented the suggestion, thanks!

ClausHolbechArista: Thanks. I took the liberty of also fixing auth for cvaas. @noredistribution can you take a look at the last commit here and see if it looks right?

ansible_collections/arista/avd/roles/eos_designs/templates/base/daemon-terminattr.j2
```diff
@@ -26,14 +26,17 @@ daemon_terminattr:
...
 {%     elif cvaas.ips | length > 0 %}
 {%         for cvp_instance_ip in cvaas.ips %}
     - {{ cvp_instance_ip }}:443
 {%         endfor %}
-{%     endif %}
   cvauth:
-    method: "key"
-    key: "{{ cvp_ingestauth_key }}"
+    method: "token-secure"
+    key: "{{ cvp_ingestauth_key | arista.avd.default('/tmp/cv-onboarding-token') }}"
```
noredistribution: replace `key` with `token_file` as with CVaaS there's no key based auth, otherwise the playbook will not generate `cvauth` flag, just tested this and it works after replacing it to `token_file`

tgodaA (approved):
LGTM

ClausHolbechArista (approved):
LGTM

PR #1477, opened 2022-02-01
Optimize(plugins): Optimize the esi_management filter with join

tgodaA:
## Change Summary

The generate_route_target method in esi_management filter is using a for loop which is replaced with a join method.
## Component(s) name

`arista.avd.<role-name>`

## How to test

Requires the latest devel to test
```
cd ansible-avd/ansible_collections/arista/avd

ansible-test units --python <version>

cd molecule

make refresh-facts
```

- All the unit tests should pass.
- Avoid the make refresh-facts because it takes a lot of time. This is tested only to make sure there are no changes in the artifacts.

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista (approved):
LGTM. I ran the unit and integration tests and no issues.

PR #1483, opened 2022-02-04
Fix(eos_designs): Pull endpoint interface information from combined adapter+port profile settings

PhirePhly:
## Change Summary

endpoint_ports was incorrectly being pulled from the original raw adapter
configuration, which did not allow the endpoint port to be configured in
a port profile and inherited, which would be useful when stamping out a large number of
connected endpoints with the same port configuraiton.

Also makes it much easier to stub out endpoint_ports when they're not
applicable for the user since they can set endpoint_ports: [ 'N/A' ] in
just a parent port profile template.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Pull endpoint_ports from the merged dataset instead of the original top level dict

## How to test

Moved one server_ports definition up into a profile.

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
@@ -15,11 +15,11 @@ ethernet_interfaces:
 {%             set endpoint_ports = adapter_settings.endpoint_ports | arista.avd.default(adapter_settings.server_ports) %}
 {%             set peer_type = connected_endpoints_keys[endpoint_key].type %}
 {%             set peer = connected_endpoint %}
-{%             for adapter_switch in adapter.switches %}
+{%             for adapter_switch in adapter_settings.switches %}
 {%                 if adapter_switch == inventory_hostname %}
-{%                     set channel_group_id = (adapter.switch_ports[0] | regex_findall("\d") | join) %}
+{%                     set channel_group_id = (adapter_settings.switch_ports[0] | regex_findall("\d") | join) %}
 {%                     set peer_interface = endpoint_ports[loop.index0] %}
-  {{ adapter.switch_ports[loop.index0] }}:
+  {{ adapter_settings.switch_ports[loop.index0] }}:
     peer: {{ peer }}
     peer_interface: {{ peer_interface }}
     peer_type: {{ peer_type }}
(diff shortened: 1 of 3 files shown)


ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -12,7 +12,7 @@ ethernet_interfaces:
 {%             set adapter_settings = {} | combine(parent_profile | arista.avd.default({}),
                                                    adapter_profile | arista.avd.default({}),
                                                    adapter, recursive=true, list_merge='replace') %}
-{%             set endpoint_ports = adapter.endpoint_ports | arista.avd.default(adapter.server_ports) %}
+{%             set endpoint_ports = adapter_settings.endpoint_ports | arista.avd.default(adapter_settings.server_ports) %}
 {%             set peer_type = connected_endpoints_keys[endpoint_key].type %}
 {%             set peer = connected_endpoint %}
 {%             for adapter_switch in adapter.switches %}
```
tgodaA: @ClausHolbechArista, I think we can modify this parameter, ```adapter.switches``` to ```adapter_settings.switches```. This gives the user a lot of flexibility. #e.g. 20 end_points connected to the same group of switches. So if we add this to profile settings, the user can use the profile settings. Anyway we have the local settings to override the default ones.

PhirePhly: I added a third commit to fix the switch side as well, in case someone wanted to have port profile strides in that dimension instead of in the server interface dimension like I did.

ClausHolbechArista (commented):
`switch_ports`is also referenced in the `port-channel-interfaces.j2`, to calculate the `channel_group_id`. This should also be updated to take it from adapter_settings.

tgodaA:
But ```switch_ports``` are unique per adapter. So this can be referenced from ```adapter```.

PhirePhly:
It depends on what dimension you're using in the template.

1. Every endpoint uses the same NIC: define `endpoint_ports` in the port profile. i.e. Every server uses their `enp130s2` NIC
2. Every endpoint is connected to the same TOR switch: define `switches` in the port profile. i.e. Every load balancer is connected to the service leaf.
3. Every endpoint is connected to the same port on their TOR switches: define `switch_ports` in the profile. i.e. Every provisioning server is connected to Eth1 on the TORs. 

I only care about scenario 1, but I can believe someone hitting scenario 2. Scenario 3 seems possible but the least likely. 

I'm really just trying to avoid having to define `endpoint_ports: ['N/A']` 140 times for inter-AS peering ports.

ClausHolbechArista (approved):
LGTM

tgodaA (approved):
LGTM

carlbuchmann (approved):
LGTM!

PR #1499, opened 2022-02-11
Feat(eos_designs): Support default vrf in Network Services

### LINKED ISSUE #1203: SVI in default routing table
### Enhancement summary

In current implementation SVIs have to be configured in a specific VRF.
Is it possible to make AVD able to add SVI interface in default routing table?

eos_designs

### Use case example

```
interface Vlan110
   description Tenant_A_OP_Zone
   ip address 172.16.100.2/24
   ip virtual-router address 172.16.100.1
```

In inventory/group_vars/DC1_TENANTS_NETWORKS.yaml

```yaml
tenants:
  # Tenant A Specific Information - VRFs / VLANs
 TenantA:
    mac_vrf_vni_base: 10000
    vrfs:
     Tenant_A_OP_Zone:  <----- HERE IT SHOULD BE POSSIBLE TO USE "default" FOR DEFAULT ROUTING INSTANCE
        vrf_vni: 123
        svis:
          # Service One
          100:
            name: Service_100
            tags: [pub, servers]
            enabled: true
            ip_virtual_router_address: 172.16.100.1
            nodes:
              DC1_LEAF1:
                ip_address: 172.16.100.2/24
              DC1_LEAF2:
                ip_address: 172.16.100.3/24
              DC1_LEAF3:
                ip_address: 172.16.100.4/24
              DC1_LEAF4:
                ip_address: 172.16.100.5/24
```

_No response_

### Additional context

_No response_

### Contributing Guide

tgodaA:
## Change Summary

Support default vrf in Network services. This shouldn't generate default vrf in vlan-aware-bundles, router_bgp, vlan-interfaces. 

## Related Issue(s)

Fixes #1203

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


tgodaA:
Let me know if this is the correct way.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vlan-aware-bundles.j2
```diff
@@ -4,7 +4,7 @@
 {% if evpn_vlan_aware_bundles == true %}
 {%     for tenant in network_services_data.tenants %}
 {%         set tenant_mac_vrf_id_base = tenant.mac_vrf_id_base | arista.avd.default(tenant.mac_vrf_vni_base) %}
-{%         for vrf in tenant.vrfs %}
+{%         for vrf in tenant.vrfs if vrf.name != 'default' %}
```
ClausHolbechArista: vlan aware bundles are still needed, even if the vlan is in default VRF. You will just have to add some route-maps on the vrf import/export for it to work with regular IRB

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vrfs.j2
```diff
@@ -4,7 +4,7 @@
 {%     set bgp_vrf = namespace() %}
   vrfs:
 {%     for tenant in network_services_data.tenants %}
-{%         for vrf in tenant.vrfs %}
+{%         for vrf in tenant.vrfs if vrf.name != 'default' %}
```
ClausHolbechArista: I think we should just adjust the config according to this:
https://www.arista.com/en/support/toi/eos-4-22-0f/14304-evpn-s-irb-in-default-vrf-type-2-type-5-routes

tgodaA: Thank you for sharing this doc.

tgodaA: I remember now. This is Asymmetric IRB but with slight difference.

ClausHolbechArista: No, not really. It is S-IRB, but we need to strictly control which prefixes we send into the pure "underlay" routing, and which we use EVPN for.

tgodaA: I modified the logic to produce the similar config as explained in the document. I tested it locally.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vrfs.j2
```diff
@@ -24,84 +24,93 @@
...
 {%             endfor %}
     {{ vrf.name }}:
+{%             if vrf.name == 'default' %}
+{%                 set default_vrf.enabled = true %}
+{%                 do route_targets.export.evpn.append('route-map RM-TENANT-NETWORK-DEFAULT-VRF') %}
+      rd: "{{ network_services_data.rd_admin_subfield }}:{{ bgp_vrf_id }}"
+      route_targets:
+        import: {{ route_targets.import | to_json }}
+        export: {{ route_targets.export | to_json }}
+{%             else %}
```
ClausHolbechArista, line 35: if we exclude all the following knobs from the default vrf case, we should note that in the documentation. We can always add them one by one later.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/main.j2
```diff
@@ -47,6 +48,12 @@
       switch.network_services_l3 is arista.avd.defined(true)) %}
 {%     include 'network_services/vtep-logic.j2' %}
 
+{%     if default_vrf.svi_subnets %}
```
ClausHolbechArista, line 51: here you check for svi_subnets. Somewhere else you check for enabled. I don't think you need the enabled flag.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp.j2
```diff
@@ -7,4 +7,10 @@ router_bgp:
 
 {%     include 'network_services/router-bgp-vlan-aware-bundles.j2' %}
 
+{%     if default_vrf.enabled %}
+
+{%         include 'network_services/default-vrf.j2' %}
```
ClausHolbechArista: rename the template to `router-bgp-default-vrf.j2`

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf-route-map.j2
```diff
@@ -0,0 +1,17 @@
+{# router_bgp.prefix_lists #}
```
ClausHolbechArista: not router_bgp.prefix_lists. Just prefix_lists. Same below for route-maps.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf.j2
```diff
@@ -0,0 +1,6 @@
+  redistribute_routes:
+    connected: true
```
ClausHolbechArista: This will affect the existing fabric behavior. Do we need to redistribute all connected without route-map ? Maybe we could add our new subnets (we have it in the prefix list) to the existing route_map and then exclude those specifically on the peer-group. My point is that we might have other contributors to underlay routing, that we don't know about.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vrfs.j2
tgodaA: I tested other knobs as well. When I configured the neighbors (under vrf default) similar to vrf non-default configuration, I found that those commands were placed under router_bgp.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/main.j2
tgodaA: Yeah, I made multiple changes. But I left that for testing. Will remove the enabled flag. I will modify the logic inside router_bgp. Instead of checking if vrf == default only I wil also check if there are any vlans under default vrf now. This will make sure the S-IRB with default vrf is configured only when there are vlans under default vrf.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp.j2
tgodaA: Sure.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf-route-map.j2
tgodaA: My bad. Will change the comment.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf.j2
tgodaA: Actually, I moved the RM-CONN-2-BGP to IPv4-UNDERLAY-PEER out.

image

And redistributed all connected routes here as per the manual.

tgodaA: The new connected subnets were added to vrf default route-target export 

image

tgodaA: I used this [doc](https://docs.google.com/document/d/1VeXddQNdm-k9AtIHWyhL2aB9OZwNuBTpSGbZDjsJpRo/edit#heading=h.4dj21gsbghfz) and the one provided by you for reference.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vrfs.j2
tgodaA: I will document the changes.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf.j2
tgodaA: I have tested the changes you mentioned: 

image

image

image

tgodaA: I am not sure about route-map being used in redistribute connected. I will check with others. The reason why I left it with out route-map was due to the way of implementation discussed in the doc (uses mix of both default an non-default vrfs) https://www.arista.com/en/support/toi/eos-4-22-0f/14304-evpn-s-irb-in-default-vrf-type-2-type-5-routes

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf-route-map.j2
```diff
@@ -0,0 +1,17 @@
+{# prefix_lists #}
+prefix_lists:
+  TENANT-CONN-DEFAULT-VRF:
```
xaviramon: Naming convention, use "PL-" as prefix

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf.j2
xaviramon: I tested that with customer. It seems like we have a "double filter" for the redistribution. 
Basically, if you use the "redistribute connected" command with a route-map, no other networks can be considered later in the VRF to be announced out of these allowed in the route-map.
This is the reason you filter only to spines direction, and you can't add a route-map to the redistribute connected.

May a general redistribute connected with a general route-map and then a more specific one towards the spines to filter underlay? This is not documented, I am not sure if we should go down this path.

xaviramon:
My concern here is that this implementation is limiting the networks in default VRF to only SVI defined. There is no possibility to import/export other sources into the EVPN table. Is this sufficient for the issue opened?

tgodaA:
Could be @xaviramon . I don't have the core knowledge of using default vrf(I just read the docs on how to use it). Please share the doc/resource if you have any. So I can implement them in this PR or may be in another one.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf-route-map.j2
tgodaA: Sure, will change it

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/default-vrf.j2
tgodaA: Thank you for the clarification.

ClausHolbechArista:
@xaviramon @tgodaA 
I don't think this feature should change our existing policy of only redistributing the loopbacks into default table and advertise to neighbors. Of course we also need to add the SVIs defined, but it should be easy to just add the SVI prefix-list to the existing redistribution route-map, so we permit the current loopbacks + the SVIs.
Since the feature will require _not_ to advertise the SVIs, we could add an outbound route-map to underlay_peers specifically _denying_ these, but allowing everything else.

xaviramon:
My 5 euro-cents. I tested it in EVE-NG and it works.

ClausHolbechArista:
> My 5 euro-cents. I tested it in EVE-NG and it works.

Great. I will go through it, but at first glance, I think we need to find some better RM names. IMO the RM should be about where it is applied, and the PL about what it matches. So IMO the RM for underlay should be RM-BGP-UNDERLAY-PEERS-OUT or similar.

xaviramon:
While pushing with AVD to testing environment I discovered a duplicate in iBGP for default VRF that was causing duplicated IPs. Since we are in default VRF, we will use default L3 peering and there is no need to create a new one.
Names changed, I hope it makes more sense now.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp.j2
```diff
@@ -7,4 +7,10 @@ router_bgp:
 
 {%     include 'network_services/router-bgp-vlan-aware-bundles.j2' %}
 
+{%     if default_vrf.svi_subnets %}
+
```
emilarista: Nitpicking here but I think we only need extra blank lines just after an include statement. So not needed on 11 and 15.

emilarista (approved):
LGTM aside from some unneeded blank lines.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/router-bgp-vrfs.j2
```diff
@@ -24,84 +24,94 @@
 {%                 endif %}
 {%             endfor %}
     {{ vrf.name }}:
+{%             if vrf.name == 'default' %}
+{%                 if default_vrf.svi_subnets %}
+{%                     do route_targets.export.evpn.append('route-map RM-EVPN-OVERLAY-PEERS-DEFAULT-VRF-OUT') %}
```
ClausHolbechArista: Since this RM is applied on evpn export IMO it should be named something like `RM-EVPN-EXPORT-VRF-DEFAULT`

emilarista (approved):

ClausHolbechArista (approved):
LGTM

PR #1522, opened 2022-02-19
Fix(eos_cli_config_gen): Remove policy sort, support index and drop command in policy-maps PBR

tgodaA:
## Change Summary

The old PBR rules were sorted based on the class name, which is not desired. The drop command needs to be added to the PBR rules. This drop policy might end up anywhere in the list based on it's class name. So the sorting over policies is removed and a new index field is added to support more flexibility.

## Component(s) name

## Proposed changes
The **drop**  knob is added.

```
policy_maps:
  pbr:
    < policy-map name >:
      classes:
        < class name >:
          index: < integer >
          # Set only one of the below actions per class
          drop: < true | false >
          set:
            nexthop:
              ip_address: < IPv4_address | IPv6_address >
              recursive: < true | false >
```

## How to test

molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/policy-maps-pbr.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/policy-maps-pbr.j2
@@ -2,17 +2,15 @@
 {% for policy_map in policy_maps.pbr | arista.avd.natural_sort %}
 !
 policy-map type pbr {{ policy_map }}
-{%     set delimiter = namespace(is_set = false) %}
 {%     for class in policy_maps.pbr[policy_map].classes | arista.avd.default([]) %}
 {%         set class_cli = 'class ' ~ class %}
 {%         if policy_maps.pbr[policy_map].classes[class].index is arista.avd.defined %}
 {%             set class_cli = policy_maps.pbr[policy_map].classes[class].index ~ ' ' ~ class_cli %}
 {%         endif %}
-{%         if delimiter.is_set %}
+{%         if not loop.first %}
    !
 {%         endif %}
    {{ class_cli }}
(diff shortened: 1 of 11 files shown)


ClausHolbechArista:
Adding a new model is a bit drastic. I suggest that we just remove the sorts and do `arista.avd.default([])` instead. I think it is a bug that we sort.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/policy-maps-pbr.j2
```diff
@@ -0,0 +1,28 @@
...
+{% for policy_map in policy_maps.pbr | arista.avd.natural_sort %}
+!
+policy-map type pbr {{ policy_map }}
+{%     set delimiter = namespace(is_set = false) %}
+{%     for class in policy_maps.pbr[policy_map].classes | arista.avd.default([]) %}
+{%         set class_cli = 'class ' ~ class %}
+{%         if policy_maps.pbr[policy_map].classes[class].index is arista.avd.defined %}
+{%             set class_cli = policy_maps.pbr[policy_map].classes[class].index ~ ' ' ~ class_cli %}
+{%         endif %}
+{%         if delimiter.is_set %}
```
ClausHolbechArista: Normally we don't go to this extend to output delimiters. A simpler implementation would be to just use `if not loop.first`

tgodaA: Thanks for the hint.

ClausHolbechArista (approved):
LGTM

xaviramon (approved):
LGTM

PR #1533, opened 2022-02-22
Feat(eos_cli_config_gen): Add support for changing the notification timestamp mode for gnmi (Issues/1532)

### LINKED ISSUE #1532: Feat(eos_cli_config_gen): Add support for changing the notification timestamp mode for gnmi
### Enhancement summary

In 4.27.0F we've added support for changing the notification timestamp from the last-change-time (GNMI spec) to send-time (outside of spec) to support (legacy) systems that don't keep track of the last known state of a counter

eos_cli_config_gen

### Use case example

folks who setup streaming via telegraf or any gnmi client to influx or other dbs and plot graphs with grafana won't see states outside of the selected range if that state has not changed for a long time, even if sampling mode is configured, as other systems are not like CloudVision which keeps track of the last known state of each counter...

we have a new cli "notification timestamp send-time" that can override the ts of the notifications everytime we stream it even if they did not change
ref: https://aristanetworks.github.io/openmgmt/configuration/openconfig/#telemetry-timestamps

_No response_

### Additional context

_No response_

### Contributing Guide

noredistribution:
## Change Summary

added support for `notification timestamp send-time` under `management api gnmi` that was introduced in EOS 4.27.0F (ref: https://aristanetworks.github.io/openmgmt/configuration/openconfig/#telemetry-timestamps)

## Related Issue(s)

Fixes #1532 

## Component(s) name

```arista.avd.eos_cli_config_gen```

## Proposed changes
didn't modify the existing data model, just added another key for `set_send_time`

## How to test

add `set_send_time: true` under `management_api_gnmi` in the fabric group_vars

```yaml
management_api_gnmi:
  set_send_time: true
  enable_vrfs:
    MGMT:
    #  access_group: test1
  octa: true
```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -1701,16 +1701,28 @@ ip_http_client_source_interfaces:
 
 ```yaml
 management_api_gnmi:
+  provider: < "eos-native" >
+  transport:
+     grpc:
+     - name: < transport_name >
+       # vrf is optional
+       vrf: < vrf_name >
+      # Per the GNMI specification, the default timestamp field of a notification message is set to be
+      # the time at which the value of the underlying data source changes or when the reported event takes place.
+      # In order to facilitate integration in legacy environments oriented around polling style operations,
+      # an option to support overriding the timestamp field to the send-time is available from EOS 4.27.0F.
+      notification_timestamp: < "send-time" | "last-change-time" >
+      ip_access_group: < acl_name >
+
+  # Below keys will be deprecated in AVD v4.0 - These should not be mixed with the new keys above.
   enable_vrfs:
     < vrf_name_1 >:
       access_group: < Standard IPv4 ACL name >
     < vrf_name_2 >:
       access_group: < Standard IPv4 ACL name >
+  # Enable provider 'eos-native' to stream both OpenConfig and EOS native paths.
   octa:
 ```
-
-!!! info "gNMI provider"
-    Octa activates `eos-native` provider and it is the only provider currently supported by EOS.
 
 #### Management Console
 
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -1703,16 +1703,16 @@ ip_http_client_source_interfaces:
 management_api_gnmi:
   provider: < "eos-native" >
   transport:
-     grpc:
-     - name: < transport_name >
-       # vrf is optional
-       vrf: < vrf_name >
-      # Per the GNMI specification, the default timestamp field of a notification message is set to be
-      # the time at which the value of the underlying data source changes or when the reported event takes place.
-      # In order to facilitate integration in legacy environments oriented around polling style operations,
-      # an option to support overriding the timestamp field to the send-time is available from EOS 4.27.0F.
-      notification_timestamp: < "send-time" | "last-change-time" >
-      ip_access_group: < acl_name >
+    grpc:
+      - name: < transport_name >
+        # vrf is optional
+        vrf: < vrf_name >
+        # Per the GNMI specification, the default timestamp field of a notification message is set to be
+        # the time at which the value of the underlying data source changes or when the reported event takes place.
+        # In order to facilitate integration in legacy environments oriented around polling style operations,
+        # an option to support overriding the timestamp field to the send-time is available from EOS 4.27.0F.
+        notification_timestamp: < "send-time" | "last-change-time" >
+        ip_access_group: < acl_name >
 
   # Below keys will be deprecated in AVD v4.0 - These should not be mixed with the new keys above.
   enable_vrfs:
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
@@ -4,20 +4,40 @@
 ## Management API GNMI
 
 ### Management API GNMI Summary
-
+{# legacy table view that will be deprecated in future releases#}
 {%     if management_api_gnmi.enable_vrfs is defined %}
+{%         if management_api_gnmi.octa is defined  %}
+{%             set octa = 'enabled' %}
+{%         else %}
+{%             set octa = 'disabled' %}
+{%         endif %}
+
 | VRF with GNMI | OCTA |
-| ---------- | ---------- |
+| ------------- | ---- |
 {%         for vrf in management_api_gnmi.enable_vrfs | arista.avd.natural_sort %}
-{%             if management_api_gnmi.octa is defined  %}
-{%                 set octa= 'enabled' %}
-{%             else %}
-{%                 set octa= 'disabled' %}
-{%             endif %}
 | {{ vrf }} | {{ octa }} |
 {%         endfor %}
-{%     else %}
-GNMI is not configured on any VRF.
+{%     endif %}
+{# new table view using the new flags #}
+{%     if management_api_gnmi.transport is arista.avd.defined %}
+
+| Transport | VRF | Notification Timestamp | ACL |
+| --------- | --- | ---------------------- | --- |
+{%         if management_api_gnmi.transport.grpc is arista.avd.defined %}
+{%             for transport in management_api_gnmi.transport.grpc %}
+{%                 if transport.name is arista.avd.defined %}
+{%                     set transport_name = transport.name %}
+{%                     set vrf = transport.vrf | arista.avd.default('-') %}
+{%                     set notif = transport.notification_timestamp | arista.avd.default('last-change-time') %}
+{%                     set acl = transport.ip_access_group | arista.avd.default('-') %}
+| {{ transport_name }} | {{ vrf }} | {{ notif }} | {{ acl }} |
+{%                 endif %}
+{%             endfor %}
+{%         endif %}
+{%     endif %}
+{%     if management_api_gnmi.provider is arista.avd.defined %}
+
+Provider {{ management_api_gnmi.provider }} is configured.
 {%     endif %}
 
 ### Management API gnmi configuration
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
@@ -4,7 +4,7 @@
 ## Management API GNMI
 
 ### Management API GNMI Summary
-{# legacy table view that will be deprecated in future releases#}
+{# legacy table view that will be deprecated in future releases #}
 {%     if management_api_gnmi.enable_vrfs is defined %}
 {%         if management_api_gnmi.octa is defined  %}
 {%             set octa = 'enabled' %}
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
@@ -13,7 +13,7 @@ management api gnmi
       vrf {{ vrf }}
 {%         endif %}
 {%     endfor %}
-{%     if management_api_gnmi.octa is arista.avd.defined %}
+{%     if management_api_gnmi.octa is defined %}
    provider eos-native
 {%     endif %}
 {%     if management_api_gnmi.transport is arista.avd.defined %}
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
@@ -16,4 +16,25 @@ management api gnmi
 {%     if management_api_gnmi.octa is arista.avd.defined %}
    provider eos-native
 {%     endif %}
+{%     if management_api_gnmi.transport is arista.avd.defined %}
+{%         if management_api_gnmi.transport.grpc is arista.avd.defined %}
+{%             for transport in management_api_gnmi.transport.grpc %}
+{%                 if transport.name is arista.avd.defined %}
+   transport grpc {{ transport.name }}
+{%                     if transport.vrf is arista.avd.defined %}
+      vrf {{ transport.vrf }}
+{%                     endif %}
+{%                     if transport.ip_access_group is arista.avd.defined %}
+      ip access-group {{ transport.ip_access_group }}
+{%                     endif %}
+{%                     if transport.notification_timestamp is arista.avd.defined %}
+      notification timestamp {{ transport.notification_timestamp }}
+{%                     endif %}
+{%                 endif %}
+{%             endfor %}
+{%         endif %}
+{%     endif %}
+{%     if management_api_gnmi.provider is arista.avd.defined %}
+   provider {{ management_api_gnmi.provider }}
+{%     endif %}
 {% endif %}
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi-new-flags.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi-new-flags.md
@@ -1,4 +1,4 @@
-# management-gnmi
+# management-gnmi-new-flags
 # Table of Contents
 
 - [Management](#management)
@@ -49,10 +49,12 @@ interface Management1
 
 ### Management API GNMI Summary
 
-| VRF with GNMI | OCTA |
-| ------------- | ---- |
-| MGMT | enabled |
-| MONITORING | enabled |
+| Transport | VRF | Notification Timestamp | ACL |
+| --------- | --- | ---------------------- | --- |
+| MGMT | MGMT | send-time | acl1 |
+| mytransport | - | send-time | acl1 |
+
+Provider eos-native is configured.
 
 ### Management API gnmi configuration
 
@@ -60,10 +62,13 @@ interface Management1
 !
 management api gnmi
    transport grpc MGMT
-      ip access-group ACL-GNMI
       vrf MGMT
-   transport grpc MONITORING
-      vrf MONITORING
+      ip access-group acl1
+      notification timestamp send-time
+   transport grpc mytransport
+      ip access-group acl1
+      notification timestamp send-time
+   provider eos-native
 ```
 
 # Authentication
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi.md
@@ -64,6 +64,7 @@ management api gnmi
       vrf MGMT
    transport grpc MONITORING
       vrf MONITORING
+   provider eos-native
 ```
 
 # Authentication
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/management-gnmi-new-flags.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/management-gnmi-new-flags.cfg
@@ -1,0 +1,25 @@
+!RANCID-CONTENT-TYPE: arista
+!
+transceiver qsfp default-mode 4x10G
+!
+hostname management-gnmi-new-flags
+!
+no enable password
+no aaa root
+!
+interface Management1
+   description oob_management
+   vrf MGMT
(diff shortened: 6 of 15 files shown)


ClausHolbechArista:
I have changed to draft. Please fix all test failures first. Then click "ready for review"

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,25 @@
 
 ### Management API GNMI Summary
 
-{%     if management_api_gnmi.enable_vrfs is defined %}
-| VRF with GNMI | OCTA |
-| ---------- | ---------- |
+| VRF | OCTA | Notification |
+| --- | ---- | ------------ |
+{%     if management_api_gnmi.octa is arista.avd.defined %}
```
tgodaA: This has to be defined true to set it. Please fix this.

For example: 
```
management_api_gnmi:
  octa: false                              <---- This will enable the octa even though logically, we are trying to set it off.
```

tgodaA: Needed change in both templates.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,25 @@
```
tgodaA, line 5: Is it ok for you to change line 2 logic to use arista.avd.defined ?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,25 @@
 
 ### Management API GNMI Summary
 
-{%     if management_api_gnmi.enable_vrfs is defined %}
```
tgodaA, line 8: We still need this check to avoid printing empty table (without values) when enable_vrfs is not defined.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -9,6 +9,9 @@ management api gnmi
    transport grpc {{ vrf }}
 {%             if management_api_gnmi.enable_vrfs[vrf].access_group is arista.avd.defined %}
       ip access-group {{ management_api_gnmi.enable_vrfs[vrf].access_group }}
+{%             endif %}
+{%             if management_api_gnmi.set_send_time is arista.avd.defined %}
```
tgodaA: Has to defined **true**. Also the data-model is not in compliance with the cli_config. Please use the knob (under management_api_gnmi)
```
notification_timestamp:
  send_time: < true | false >```

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2 (hunk 3 above)
Comment (noredistribution) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2:8]: I printed the values further down now when vrf is not defined or if it's set to default

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2 (hunk 4 above)
Comment (noredistribution) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2]: fair point, done

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2 (hunk 1 above)
Comment (noredistribution) [reply ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2]: I think I did it now, tested both `octa: false` or when `octa` flag is not set and works now

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2 [hunk 5]
```diff
@@ -1,18 +1,30 @@
 {# Management API GNMI #}
-{% if management_api_gnmi is defined and management_api_gnmi is not none %}
+{% if management_api_gnmi is arista.avd.defined and management_api_gnmi is not none %}
```
tgodaA: We can remove the last check **is not none**. The arista.avd.defined does two checks ``` is defined``` and ```is not none```. So because we are using arista.avd.defined we can remove the last one.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
tgodaA: Ok

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -2,17 +2,30 @@
...
+{%         if management_api_gnmi.notification_timestamp.send_time is arista.avd.defined(true) %}
+      notification timestamp send-time
+{%         endif %}
+{%         if management_api_gnmi.access_group is arista.avd.defined %}
+      ip access-group {{ management_api_gnmi.access_group }}
+{%         endif %}
+{%     else %}
+{%         for vrf in management_api_gnmi.enable_vrfs | arista.avd.natural_sort %}
+{%             if vrf == 'default' %}
+   transport grpc default
```
tgodaA: Why are we defining this once again for default vrf?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -2,17 +2,30 @@
...
+{%             else %}
    transport grpc {{ vrf }}
+{%             endif %}
 {%             if management_api_gnmi.enable_vrfs[vrf].access_group is arista.avd.defined %}
       ip access-group {{ management_api_gnmi.enable_vrfs[vrf].access_group }}
+{%             endif %}
+{%             if management_api_gnmi.notification_timestamp.send_time is arista.avd.defined(true) %}
+      notification timestamp send-time
 {%             endif %}
       vrf {{ vrf }}
```
tgodaA, line 13: This will print ```vrf default```. Shouldn't we need to do a check here if vrf != default then only print this command

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -1,18 +1,30 @@
...
+{%         set octa = 'enabled' %}
+{%     else %}
+{%         set octa = 'disabled' %}
+{%     endif %}
+{%     if management_api_gnmi.notification_timestamp.send_time is arista.avd.defined(true) %}
+{%         set send_time = 'send-time' %}
+{%     else %}
+{%         set send_time = 'last-change-time' %}
+{%     endif %}
+{%     if management_api_gnmi.enable_vrfs is not arista.avd.defined %}
```
tgodaA: I understand that you were going with the default behavior of eos_cli. But in this AVD role, we print only the input given by user. So if enable_vrfs is defined then only we print the values.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-gnmi.yml
```diff
@@ -1,6 +1,8 @@
 management_api_gnmi:
+  notification_timestamp:
+    send_time: true
   enable_vrfs:
     MGMT:
       access_group: ACL-GNMI
     MONITORING:
-  octa:
+  octa: true
```
ClausHolbechArista: This is a data model change, and cannot be accepted in a minor release.
Unfortunately the `management_api_gnmi.octa` key is also valid with a `none` value. It was done like that originally, but should be fixed in the next major release.
Please revert this and all associated changes in the templates. Ex. you have to use `is defined` since it also passes on `none` value.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
noredistribution: I don't think I understand this, the `default` vrf is a vrf too, so imo we should document if you have GNMI enabled in default vrf, plus the other things like what's the send-time or if OCTA is enabled, I don't think we should only render a table if someone uses a non-default vrf, what's the use case?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
noredistribution: good catch, fixed this now, let me push a change

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-gnmi.yml
tgodaA: Yes, it is a breaking change. I though this was a new feature.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -2,17 +2,30 @@
...
+      notification timestamp send-time
+{%             endif %}
+{%         endfor %}
+{%     else %}
+   transport grpc default
+{%         if management_api_gnmi.notification_timestamp.send_time is arista.avd.defined(true) %}
+      notification timestamp send-time
 {%         endif %}
-{%     endfor %}
+{%         if management_api_gnmi.access_group is arista.avd.defined %}
```
ClausHolbechArista: This variable is not documented.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -2,17 +2,30 @@
...
-{%         if vrf == 'default' %}
-   transport grpc default
-{%         else %}
+{%     if management_api_gnmi.enable_vrfs is arista.avd.defined %}
+{%         for vrf in management_api_gnmi.enable_vrfs | arista.avd.natural_sort %}
+{%             if vrf != 'default' %}
    transport grpc {{ vrf }}
+      vrf {{ vrf }}
+{%             else %}
+   transport grpc default
```
ClausHolbechArista: No need to have this condition, since you produce the same output (if vrf == "default" you would output `transport grpc default` any way.)
It is not clear to me if EOS cli `transport grpc default` means settings for the VRF `default` or if it means "default settings". to me it seems a bit mixed in this template.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-gnmi-new-flags-novrf.yml
```diff
@@ -0,0 +1,7 @@
+management_api_gnmi:
```
ClausHolbechArista: instead of running multiple test cases, you can just add another list item to the `grpc` list.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -17,3 +17,24 @@ management api gnmi
    provider eos-native
 {%     endif %}
 {% endif %}
+{% if management_api_gnmi.transport is arista.avd.defined %}
+{%     if management_api_gnmi.transport.grpc is arista.avd.defined %}
+{%         for transport in management_api_gnmi.transport.grpc %}
+{%             if transport.name is arista.avd.defined %}
```
ClausHolbechArista: this `if` should enclose all the following config, since it is the entire config section

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,38 @@
...
+| ------------- | ---- |
 {%         for vrf in management_api_gnmi.enable_vrfs | arista.avd.natural_sort %}
-| {{ vrf }} | {% if management_api_gnmi.octa is defined  %} enabled {% else %} disabled {% endif %} |
+| {{ vrf }} | {{ octa }} |
 {%         endfor %}
-{%     else %}
-GNMI is not configured on any VRF.
+{%     endif %}
+{# new table view using the new flags #}
+{%     if management_api_gnmi.transport is arista.avd.defined %}
```
ClausHolbechArista, line 22: add one blank line after the `if`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,38 @@
 
 ### Management API GNMI Summary
```
ClausHolbechArista: Remove this blank line

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -5,14 +5,38 @@
...
 
+{# legacy table view that will be deprecated in future releases#}
+{%     if management_api_gnmi.octa is defined  %}
+{%         set octa = 'enabled' %}
+{%     else %}
+{%         set octa = 'disabled' %}
+{%     endif %}
 {%     if management_api_gnmi.enable_vrfs is defined %}
```
ClausHolbechArista, line 8: add one blank line after the `if`

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi-new-flags-novrf.md
```diff
@@ -0,0 +1,124 @@
+# management-gnmi
```
ClausHolbechArista: you see deletions in the CI because you have not added these new "hosts" to the "hosts.ini" under molecule. So when the CI rerun it is not creating the config and documentation files. I am not sure you managed to get these files created without adding them to "hosts.ini"

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -17,3 +17,24 @@ management api gnmi
    provider eos-native
 {%     endif %}
 {% endif %}
+{% if management_api_gnmi.transport is arista.avd.defined %}
```
ClausHolbechArista: please enclose all the new config inside the `if` block started in line 2 and ending in line 19. So you only evaluate everything if the outer `if` is true.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -17,3 +17,24 @@ management api gnmi
...
+{%     if management_api_gnmi.transport.grpc is arista.avd.defined %}
+{%         for transport in management_api_gnmi.transport.grpc %}
+{%             if transport.name is arista.avd.defined %}
+   transport grpc {{ transport.name }}
+{%             endif %}
+{%             if transport.vrf is arista.avd.defined %}
+      vrf {{ transport.vrf }}
+{%             endif %}
+{%             if transport.notification_timestamp is arista.avd.defined %}
```
ClausHolbechArista: EOS output shows `ip access-group` first, and `notification timestamp` after. Please reorder.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
noredistribution: done

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
noredistribution: done

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
ClausHolbechArista: Still needed. If you check the molecule MD files, you can see the missing blank line between the heading and the table.

ClausHolbechArista: I will add this as approve.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
ClausHolbechArista: I have updated this

ClausHolbechArista (approved):
LGTM @noredistribution please check it again, since I had to rebase etc.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-api-gnmi.j2
```diff
@@ -4,20 +4,40 @@
 ## Management API GNMI
 
 ### Management API GNMI Summary
-
+{# legacy table view that will be deprecated in future releases#}
```
tgodaA: Add space between release and #

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-gnmi.yml
tgodaA: There is a problem with this. The eos template is using the ```arista.avd.defined``` test while the documentation is using ```defined``` test. So this is a bug which doesn't match with the README as well. @ClausHolbechArista

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -16,4 +16,25 @@ management api gnmi
 {%     if management_api_gnmi.octa is arista.avd.defined %}
```
tgodaA: Need to fix this. This should be using defined test instead of arista.avd.defined. At least this needs to be fixed in 4.0

tgodaA: Or change the logic to use arista.avd.defined(true)

ClausHolbechArista: this is not a boolean in the old data model. It was defined in a vague (wrong) way, where simply setting the key to null, would enable the feature. We need to keep this behavior for now. I have updated to `defined` filter.

tgodaA: Thank you. That should generate a new artifact, prints the eos-native command.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1701,17 +1701,29 @@ ip_http_client_source_interfaces:
...
+     grpc:
+     - name: < transport_name >
+       # vrf is optional
+       vrf: < vrf_name >
+      # Per the GNMI specification, the default timestamp field of a notification message is set to be
+      # the time at which the value of the underlying data source changes or when the reported event takes place.
+      # In order to facilitate integration in legacy environments oriented around polling style operations,
+      # an option to support overriding the timestamp field to the send-time is available from EOS 4.27.0F.
+      notification_timestamp: < "send-time" | "last-change-time" >
+      ip_access_group: < acl_name >
```
tgodaA: Indentation mismatch in lines 1714 and 1715 with respect to 1707 and 1709.

tgodaA:
Need to run molecule for eos_cli_config_gen. Because the octa is tested with a regular ```defined``` test the artifacts are generated now for the old examples.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: fixed

tgodaA (approved):
LGTM

PR #1472, opened 2022-01-31
Feat(eos_designs): Add Monitor Session support

ccsnw:
## Change Summary

Add support to create monitor sessions via eos_designs

## Related Issue(s)

Fixes #1412

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Adding option to create monitoring sessions via eos_designs
- Allows multiple sources per session
- Allows multiple destinations per session
- Will merge "global" session settings (in case different values are configured [improper config])
- Only creates structured YAML for sessions that have at least one SRC and one DST defined

```
monitor_session:
  name: < session_name >
  role: < source | destination >
  < interfaces: taken from adapters.switch_ports > 
  source_settings:
    < name: taken from adapters.switch_ports >
    - direction: < rx | tx | both >
      access_group:
        type: < ip | ipv6 | mac >
        name: < acl_name >
        priority: < priority >
  session_settings:
    < destinations:
      - taken from adapters.switch_ports >
    encapsulation_gre_metadata_tx: < true | false >
    header_remove_size: < bytes >
    access_group:
      type: < ip | ipv6 | mac >
      name: < acl_name >
    rate_limit_per_ingress_chip: < "<int> bps" | "<int> kbps" | "<int> mbps" >
    rate_limit_per_egress_chip: < "<int> bps" | "<int> kbps" | "<int> mbps" >
    sample: < int >
    truncate:
      enabled: < true | false >
      size: < bytes >
```

## How to test
- Created multiple Molecule tests
- Test with SRC and DST + merge replacements for some global settings -> test if combine works
- Test with only SRC defined -> not rendered

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #1535, opened 2022-02-22
Doc(eos_designs): Fix typos in Monitor Session Documentation

ccsnw:
## Change Summary

Pasted old datamodel in documentation. Fixing by adding the actual datamodel that was used

## Related Issue(s)

Fixes #1472 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Correct data model is:

```
monitor_sessions:
  - name: < session_name >
    role: < source | destination >
    source_settings:
      direction: < rx | tx | both >
      access_group:
        type: < ip | ipv6 | mac >
        name: < acl_name >
        priority: < priority >
    session_settings:
      encapsulation_gre_metadata_tx: < true | false >
      header_remove_size: < bytes >
      access_group:
        type: < ip | ipv6 | mac >
        name: < acl_name >
      rate_limit_per_ingress_chip: < "<int> bps" | "<int> kbps" | "<int> mbps" >
      rate_limit_per_egress_chip: < "<int> bps" | "<int> kbps" | "<int> mbps" >
      sample: < int >
      truncate:
        enabled: < true | false >
        size: < bytes >
```

## How to test
Molecule tests

### User Checklist

- N/A

### Repository Checklist

- [o] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
@@ -171,28 +171,28 @@ port_profiles:
             unit: < percent | pps > | Optional var and is hardware dependant - default is percent)
 
         # Monitor Session configuration: use defined switchports as source or destination for monitoring sessions | Optional
-        monitor_session:
-          name: < session_name >
-          role: < source | destination >
-          source_settings:
-            - direction: < rx | tx | both >
+        monitor_sessions:
+          - name: < session_name >
+            role: < source | destination >
+            source_settings:
+              - direction: < rx | tx | both >
+                access_group:
+                  type: < ip | ipv6 | mac >
+                  name: < acl_name >
+                  priority: < priority >
+            # Session settings are defined per session name. Different session_settings with for same session name will be combined/merged
+            session_settings:
+              encapsulation_gre_metadata_tx: < true | false >
+              header_remove_size: < bytes >
               access_group:
                 type: < ip | ipv6 | mac >
                 name: < acl_name >
-                priority: < priority >
-          # Session settings are defined per session name. Different session_settings with for same session name will be combined/merged
-          session_settings:
-            encapsulation_gre_metadata_tx: < true | false >
-            header_remove_size: < bytes >
-            access_group:
-              type: < ip | ipv6 | mac >
-              name: < acl_name >
-            rate_limit_per_ingress_chip: < "<int> bps" | "<int> kbps" | "<int> mbps" >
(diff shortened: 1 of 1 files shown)


ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
```diff
@@ -171,28 +171,28 @@ port_profiles:
...
-        monitor_session:
-          name: < session_name >
-          role: < source | destination >
-          source_settings:
-            - direction: < rx | tx | both >
+        monitor_sessions:
+          - name: < session_name >
+            role: < source | destination >
+            source_settings:
+              - direction: < rx | tx | both >
```
ClausHolbechArista: this should not be a list either I guess.

ccsnw: I should stop using computers, obviously. 

:-))

tgodaA (approved):
LGTM

ClausHolbechArista (commented):
I can see that with this data model we will have a failure with the `combine`, if we have multiple sources on the same session. It will overwrite the `source_settings`. So maybe we should take a half step back and parse all sources first, and _then_ do the `combine` and parse the `session_settings`.
Either we could expand this fix, or add another PR with that fix.

ccsnw:
> I can see that with this data model we will have a failure with the `combine`, if we have multiple sources on the same session. It will overwrite the `source_settings`. So maybe we should take a half step back and parse all sources first, and _then_ do the `combine` and parse the `session_settings`. Either we could expand this fix, or add another PR with that fix.

@ClausHolbechArista Are you sure? I'm not using the combined/merged settings to render the sources, only for the dests. In the molecule tests I have the session "MonitoringSessionServer18WithDest" which has multiple sources and different settings on them.

ClausHolbechArista:
> > I can see that with this data model we will have a failure with the `combine`, if we have multiple sources on the same session. It will overwrite the `source_settings`. So maybe we should take a half step back and parse all sources first, and _then_ do the `combine` and parse the `session_settings`. Either we could expand this fix, or add another PR with that fix.
> 
> @ClausHolbechArista Are you sure? I'm not using the combined/merged settings to render the sources, only for the dests. In the molecule tests I have the session "MonitoringSessionServer18WithDest" which has multiple sources and different settings on them.

You are right. It works as expected.

ClausHolbechArista (approved):
LGTM

PR #1556, opened 2022-03-02
Fix(eos_designs): Feature interaction issues with evpn_services_l2_only

### LINKED ISSUE #1554: Consider "evpn_services_l2_only" only for overlay services but not on for inband_management
### Issue Summary

When the  "evpn_services_l2_only" is set to "true" it will set the fact "switch.network_services_l3" to "false" which has the effect that the virtual router MAC address template is not considered. This is a problem when we use "inband_management" at the same time for l2leafs attached to l3leafs. 

The option "evpn_services_l2_only" should only have an effect on overlay networks but not on inband management.

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
---
type: l2leaf

l2leaf:
  defaults:
    ...
    ## INBAND MANAGEMENT VLAN ##
    inband_management_subnet: 192.168.200.0/24
    inband_management_vlan: 900



---
type: l3leaf

l3leaf:
  defaults:
    ...
    
    # Possibility to prevent configuration of Tenant VRFs and SVIs. This allows support for centralized routing.
    evpn_services_l2_only: true


If needed I can provide access to a gitlab repository.
```

### Relevant log output

_No response_

### Contributing Guide

ClausHolbechArista:
## Change Summary

Feature interaction issues with evpn_services_l2_only

## Related Issue(s)

Fixes #1554 
Fixes another issue where source nat for VRFs was configures even with `evpn_services_l2_only: true`.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Configure virtual-router MAC under inband management, if condition used in network-services is false.
- Only configure VRF source-nat if VTEP and L2 services and L3 services are true.

## How to test

Added molecule coverage for the combination of inband management and `evpn_services_l2_only`.

### Repository Checklist


bjmeuer (approved):

tgodaA (approved):
LGTM

PR #1642, opened 2022-03-28
Fix(plugins): Issue in convert_dicts with empty dictionary values

ClausHolbechArista:
## Change Summary

Fix issue in convert_dicts with empty dictionary values

## Related Issue(s)

Some of the existing data models have optional keys inside nested dictionaries, so if none of the optional keys are set, the inner dictionary will not always be a dictionary, but instead the value `None`.

Example:
```yaml
ip_domain_lookup:
  source_interfaces:
    Management0:
      vrf: mgt
    Loopback0:
```

## Component(s) name

`arista.avd.convert_dicts`

## Proposed changes

Catch if the inner value is `Null` and insert a dict with the primary key, instead of just returning the original dict as today.

## How to test

The update to `dns-ntp.yml` in this PR will fail with the old code.

```shell
TASK [arista.avd.eos_cli_config_gen : Generate eos intended configuration] *****
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'ansible.parsing.yaml.objects.AnsibleUnicode object' has no attribute 'name'
fatal: [dns-ntp -> localhost]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.parsing.yaml.objects.AnsibleUnicode object' has no attribute 'name'"}
```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/dns-ntp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/dns-ntp.md
@@ -81,11 +81,13 @@ ip name-server vrf mgt 10.10.129.10
 
 | Source interface | vrf |
 | ---------------- | --- |
+| Loopback0 | - |
 | Management0 | mgt |
 
 ### DNS Domain Lookup Device Configuration
 
 ```eos
+ip domain lookup source-interface Loopback0
 ip domain lookup vrf mgt source-interface Management0
 ```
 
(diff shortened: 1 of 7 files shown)


carlbuchmann (approved):
LGTM!

Tested by replicating the issue on my setup by reverting the fix: 

```bash
TASK [arista.avd.eos_cli_config_gen : Generate eos intended configuration] *****
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'ansible.parsing.yaml.objects.AnsibleUnicode object' has no attribute 'name'
fatal: [dns-ntp -> localhost]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.parsing.yaml.objects.AnsibleUnicode object' has no attribute 'name'"}
```

PR #1653, opened 2022-03-31
Feat(eos_cli_config_gen): Support mlag primary-priority command

tgodaA:
## Change Summary

Add a knob to support mlag primary-priority command. This is not visible in the eos command line when ```?``` is typed in to command line but the switch takes the command when the full command is entered (tested it locally on a DUT). This command is needed to change the mlag priority on a primary mlag switch.

## Component(s) name

## Proposed changes

```
mlag_configuration:
  primary_priority: < integer >
```

## How to test

Molecule.

### User Checklist

- N/A

### Repository Checklist


tgodaA:
For reference adding the screenshot of the command tested: 
image

ClausHolbechArista:
I believe this command is deprecated in EOS - which also explains why it is hidden. We should not "revive it" here. I suggest to close this PR and circle back on the design.

carlbuchmann:
> I believe this command is deprecated in EOS - which also explains why it is hidden. We should not "revive it" here. I suggest to close this PR and circle back on the design.

On EOS there's a hidden command to set the priority value - "primary-priority <value>" It is hidden to reinforce the concept that it doesn't matter which peer is primary. The use case is to make the behavior predictable and not depend on the MAC system for primary election.

So my suggestion is to add this knob, but remove it from the README.md documentation.

tgodaA:
Removed it from README.

PR #1666, opened 2022-04-04
Feat(eos_cli_config_gen): Support remove-private-as in router_bgp

### LINKED ISSUE #1641: Ability to configure bgp feature remove-private-as per neighbor
### Enhancement summary

Support to enable bgp knob "remove-private-as" at the neighbor level.

eos_cli_config_gen

### Use case example

Customer is going to use AVD to implement VXLAN evpn fabric with multiple external routers(MPLS routers,SDHWAN routers) hanging off of the VTEP and enabling bgp peering with VTEPs. Border leaf (acting as VTEP) will be connecting to the MPLS DCI where customer would like to remove the private ASN that will be used within the VXLAN fabric before routes are advertised into the core . This use case will require configuring remove-private-as bgp knob for the neighbors.

CLI:
neighbor <ip/ipv6/peer-group> remove-private-as <cr/all/ingress> <replace-as>

replace-as extension is not required as of today, but may be required as part of this feature by the customer in future. By implementing replace-as this knob will be fully supported by AVD.

CLI reference guide:
https://docs.google.com/document/d/1XPfwnzSWX_a-KmBpsdAztsgeYqJeuwZ_twwkUlX04GQ/edit

```yaml
router_bgp:
 neighbors:
   remove_private_as: 
      enabled: <true | false>
      all: <true | false>
      ingress: <true | false >
      replace-as: <true | false>
 peer_groups:
   remove_private_as: 
      enabled: <true | false>
      all: <true | false>
      ingress: <true | false >
      replace-as: <true | false>
```

_No response_

### Additional context

_No response_

### Contributing Guide

tgodaA:
## Change Summary

Add support for remove-private-as in router_bgp. This feature was added to neighbors (vrf neighbors as well), peer_groups

## Related Issues(s)

Fixes: #1641 

## Component(s) name

## Proposed changes

```yaml
router_bgp:
  peer_groups:
    < peer_group_name >:
      remove_private_as:
        enabled: < true | false >
        all: < true | false >
        replace_as: < true | false >
      remove_private_as_ingress:
        enabled: < true | false >
        replace_as: < true | false >
  neighbors:
    < neighbor_ip_address >:
      remove_private_as:
        enabled: < true | false >
        all: < true | false >
        replace_as: < true | false >
      remove_private_as_ingress:
        enabled: < true | false >
        replace_as: < true | false >
  vrfs:
    < vrf_name >:
      neighbors:
        < neighbor_ip_address >:
          remove_private_as:
            enabled: < true | false >
            all: < true | false >
            replace_as: < true | false >
          remove_private_as_ingress:
            enabled: < true | false >
            replace_as: < true | false >
```

## How to test

Molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
@@ -104,15 +104,15 @@ interface Management1
 
 ### BGP Neighbors
 
-| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain | Remove Private AS |
-| -------- | --------- | --- | -------- | -------------- | -------------- | ---------- | --- | --------------------- | ----------------- |
-| 192.0.3.1 | 65432 | default | - | all | - | - | - | True | - |
-| 192.0.3.2 | 65433 | default | - | extended | 10000 | - | - | True (All) | - |
-| 192.0.3.3 | 65434 | default | - | standard | - | - | - | True | - |
-| 192.0.3.4 | 65435 | default | - | large | - | - | - | False | - |
-| 192.0.3.5 | 65436 | default | - | standard | 12000 | - | - | - | - |
-| 192.0.3.6 | 65437 | default | - | - | - | - | - | - | True |
-| 192.0.3.7 | 65438 | default | - | - | - | - | - | - | True (All) |
+| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain |
+| -------- | --------- | --- | -------- | -------------- | -------------- | ---------- | --- | --------------------- |
+| 192.0.3.1 | 65432 | default | - | all | - | - | - | True |
+| 192.0.3.2 | 65433 | default | - | extended | 10000 | - | - | True (All) |
+| 192.0.3.3 | 65434 | default | - | standard | - | - | - | True |
+| 192.0.3.4 | 65435 | default | - | large | - | - | - | False |
+| 192.0.3.5 | 65436 | default | - | standard | 12000 | - | - | - |
+| 192.0.3.6 | 65437 | default | - | - | - | - | - | - |
+| 192.0.3.7 | 65438 | default | - | - | - | - | - | - |
 
 ### BGP Neighbor Interfaces
 
@@ -170,7 +170,7 @@ router bgp 65101
    neighbor 192.0.3.6 remove-private-as ingress
    neighbor 192.0.3.6 description test_remove_private_as
    neighbor 192.0.3.7 remote-as 65438
-   neighbor 192.0.3.7 remove-private-as all
+   neighbor 192.0.3.7 remove-private-as all replace-as
    neighbor 192.0.3.7 remove-private-as ingress replace-as
    neighbor 192.0.3.7 description test_remove_private_as_all
    aggregate-address 1.1.1.0/24 advertise-only
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
@@ -104,13 +104,15 @@ interface Management1
 
 ### BGP Neighbors
 
-| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain |
-| -------- | --------- | --- | -------- | -------------- | -------------- | ---------- | --- | -------------- |
-| 192.0.3.1 | 65432 | default | - | all | - | - | - | True |
-| 192.0.3.2 | 65433 | default | - | extended | 10000 | - | - | True (All) |
-| 192.0.3.3 | 65434 | default | - | standard | - | - | - | True |
-| 192.0.3.4 | 65435 | default | - | large | - | - | - | False |
-| 192.0.3.5 | 65436 | default | - | standard | 12000 | - | - | - |
+| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain | Remove Private AS |
+| -------- | --------- | --- | -------- | -------------- | -------------- | ---------- | --- | --------------------- | ----------------- |
+| 192.0.3.1 | 65432 | default | - | all | - | - | - | True | - |
+| 192.0.3.2 | 65433 | default | - | extended | 10000 | - | - | True (All) | - |
+| 192.0.3.3 | 65434 | default | - | standard | - | - | - | True | - |
+| 192.0.3.4 | 65435 | default | - | large | - | - | - | False | - |
+| 192.0.3.5 | 65436 | default | - | standard | 12000 | - | - | - | - |
+| 192.0.3.6 | 65437 | default | - | - | - | - | - | - | True |
+| 192.0.3.7 | 65438 | default | - | - | - | - | - | - | True (All) |
 
 ### BGP Neighbor Interfaces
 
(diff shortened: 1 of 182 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2633,6 +2633,11 @@ router_bgp:
       local_as: < bgp_as >
       description: "< description as string >"
       shutdown: < true | false >
+      remove_private_as:
```
ankudinov, line 2765: Just a suggestion to consider. But feel free to decline:

```yaml
remove_private_as:
  egress: < enabled | disabled | all >
  ingress: < enabled | disabled | replace_as >
```

Apart from being more compact and readable, it prevents cases where all is defined without enabled true.
As an alternative, I'd suggest to treat `all: true` as `enabled: true` and `replace_as: true` as `ingress: true`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -56,6 +56,22 @@ router bgp {{ router_bgp.as }}
 {%         if router_bgp.peer_groups[peer_group].next_hop_unchanged is arista.avd.defined(true) %}
    neighbor {{ peer_group }} next-hop-unchanged
 {%         endif %}
+{%         if router_bgp.peer_groups[peer_group].remove_private_as.enabled is arista.avd.defined(true) %}
```
ankudinov, line 59: In my opinion, if `all` is true, we should still build the CLI command. Consider or condition if data model remains as is. It's still not obvious what should have more priority when `all: true` and `enabled: false`. Although check the doc template.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -56,6 +56,22 @@ router bgp {{ router_bgp.as }}
...
+{%             set remove_private_as_cli = "neighbor " ~ peer_group ~ " remove-private-as" %}
+{%             if router_bgp.peer_groups[peer_group].remove_private_as.all is arista.avd.defined(true) %}
+{%                 set remove_private_as_cli = remove_private_as_cli ~ " all" %}
+{%             endif %}
+   {{ remove_private_as_cli }}
+{%             if router_bgp.peer_groups[peer_group].remove_private_as.ingress is arista.avd.defined(true) %}
```
ankudinov: Same as above. If only `replace_as` is true, we can still build the Cli. Same for docs. And need some clarification on what will have priority if `enabled: false` and `replace_as: true`. Otherwise just consider an alternative data model

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
tgodaA: Thank you for the suggestion. I like this model. Will implement this.

tgodaA: The problem was with egress keyword. Because I think we need to implement something close to the cli match. But as I don't see any better option I will go with this.

ClausHolbechArista, line 2636: The EOS CLI mentions `outbound` for the regular mode and inbound for the `ingress` keyword.
Remember that `neighbor 1.2.3.4 remove-private-as all replace-as` is also valid CLI - not covered in Petr's example.

ClausHolbechArista: The EOS CLI mentions `outbound` for the regular mode and inbound for the `ingress` keyword, but ingress/egress is fine for me as well.
Remember that `neighbor 1.2.3.4 remove-private-as all replace-as` is also valid CLI - not covered in Petr's example.

ClausHolbechArista: Alternative could be:
```yaml
remove_private_as:
  enabled:
  all:
  replace_as:
remove_private_as_ingress:
  enabled:
  replace_as:
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
ClausHolbechArista: This would not match other areas in AVD. it is fine to require `enabled: true` and just ignore the rest if not set correctly.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
ClausHolbechArista: This would not match other areas in AVD. it is fine to require `enabled: true` and just ignore the rest if not set correctly.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
ankudinov: ack. Let's focus on the data model anyway, as if we change it the problem may be fixed automatically

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
ankudinov: ack. Let's focus on the data model anyway, as if we change it the problem may be fixed automatically

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ankudinov: OK, if we want to keep enbled true | false approach here is the suggestion:

```yaml
remove_private_as:
  in:
    enabled:
    replace_as:
  out:
    enabled:
    all:
    replace_as:
```

But I'm fine with the option suggested by Claus as well. The in/out, ingress/egress, inbound/outbound makes it more explicit. Which is good in my opinion. If CLI isn't explicit enough, we can still add some clarity in the data model if it's the right thing to do.

ryanmerolle: I think it should state ingress in the name to reflect the config key being applied.  Claus model is my preference because it would be closer to the actual config being modeled.

ryanmerolle:
Let me know if I can help today.  This is a good / low hanging fruit PR to get into 3.4.0

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: I don't think this will make it to 3.4.0. I am preparing release-notes etc now, so train has left the station :)

tgodaA: Even I can't push it now.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-base.md
```diff
@@ -104,13 +104,15 @@ interface Management1
...
 ### BGP Neighbors
 
-| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain |
-| -------- | --------- | --- | -------- | -------------- | -------------- | ---------- | --- | -------------- |
-| 192.0.3.1 | 65432 | default | - | all | - | - | - | True |
-| 192.0.3.2 | 65433 | default | - | extended | 10000 | - | - | True (All) |
-| 192.0.3.3 | 65434 | default | - | standard | - | - | - | True |
-| 192.0.3.4 | 65435 | default | - | large | - | - | - | False |
-| 192.0.3.5 | 65436 | default | - | standard | 12000 | - | - | - |
+| Neighbor | Remote AS | VRF | Shutdown | Send-community | Maximum-routes | Allowas-in | BFD | RIB Pre-Policy Retain | Remove Private AS Outbound | Remove Private AS Inbound |
```
ClausHolbechArista: I don't think these two are used often enough to warrant two new columns on all documentation. We need to find a better way to document all this, but until then, I think we should just leave out these knobs from the tables. It is still shown in the config below. You can keep it under peer-groups, since they are only rendered when set

tgodaA: Removed.

ClausHolbechArista (approved):
LGTM

gmuloc (approved):
LGTM

PR #1721, opened 2022-04-13
Feat(eos_cli_config_gen): add SNMPv3 hashed user passphrases support

### LINKED ISSUE #1720: (eos_cli_config_gen): Add support for SNMPv3 user localized syntax
### Enhancement summary

As of now the eos_cli_config_gen module only support the following syntax:

```
snmp-server user <USER> <GROUP> v3 [auth <AUTH_ALGORITHM> <CLEARTEXT_AUTH_PASSPHRASE> [priv <PRIV_ALGORITHM> <CLEARTEXT_PRIV_PASSPHRASE>]]
```

(though the documentation indicates "encrypted" passphrases for auth and priv, this is not accurate as the passphrases will be rehashed with the local engineID when configured on the switch.)

This feature request intends to allow the configuration of:

```
snmp-server user <USER> <GROUP> v3 localized <ENGINE-ID> [auth <AUTH_ALGORITHM> <HASHED_AUTH_PASSPHRASE> [priv <PRIV_ALGORITHM> <HASHED_PRIV_PASSPHRASE>]]
```

in addition to the existing options.

Note that it requires a local engine ID to be provided as well (by default the engine ID is derived from the switch system MAC). 

eos_cli_config_gen

### Use case example

As it is now, it is not possible to have the auth and priv passphrase NOT in cleartext inside the AVD config as it is not possible to send the command using the localized keyword which is dependent on the engineID (which default value on the switch is unknown to AVD).

The intention is to allow the configuration to be done at eos_cli_config_gen level and to later on leverage this capability from eos_designs role to auto generate the hashed passphrases when required, using a potentially auto-generate engine ID.

```
snmp_server:
  engineid:
    local: <ENGINE-ID>
[...]
     - name: < username >
       group: < group_name >
       version: < v1 | v2c | v3 >
       localized: < true | false >
       auth: < hash_algorithm >
       auth_passphrase: < hashed_auth_passphrase if localized is true else cleartext auth_passphrase >
       priv: < encryption_algorithm >
     priv_passphrase: < hashed_priv_passphrase if localized is true else cleartext priv_passphrase >
```

Using vault to hide the password in the variables YAML file but it will still be present in the end configuration.

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

As of now the eos_cli_config_gen module only support the following syntax:

```
snmp-server user <USER> <GROUP> v3 [auth <AUTH_ALGORITHM> <CLEARTEXT_AUTH_PASSPHRASE> [priv <PRIV_ALGORITHM> <CLEARTEXT_PRIV_PASSPHRASE>]]
```

(though the documentation indicates "encrypted" passphrases for auth and priv, this is not accurate as the passphrases will be rehashed with the local engineID when configured on the switch.)

This feature request intends to allow the configuration of:

```
snmp-server user <USER> <GROUP> v3 localized <ENGINE-ID> [auth <AUTH_ALGORITHM> <HASHED_AUTH_PASSPHRASE> [priv <PRIV_ALGORITHM> <HASHED_PRIV_PASSPHRASE>]]
```

## Related Issue(s)

Fixes #1720 

## Component(s) name

## Proposed changes

Modification of the `snmp-settings.j2` template to:
* Add the support for local engineID (remote can be added in the future as it is not needed for this feature)
* Add the capability to detect the presence of the `localized` keyword that indicate that the engineID is 

Documentation template has been updated to indicate the configured Local Engine ID.

README.md updated

_CAVEATS_:
* if the `snmp_server.engineid.local` is not configured the current behavior (which can be changed) is that the template will revert to non localized syntax (the one is use currently in the repo). However passing hashed passphrases to the non localized syntax will eventually fail as the passphrases will be rehashed with the device engineID (the default one if none is configured) at configuration time. This faulty behavior is already allowed in the current AVD version - and the passphrases in this version MUST be passed in cleartext (contrary to what the documentation indicates).

## How to test

Run molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
@@ -53,9 +53,9 @@ interface Management1
 
 ### SNMP Configuration Summary
 
-| Contact | Location | SNMP Traps | State |
-| ------- | -------- | ---------- | ----- |
-| DC1_OPS | DC1 | All | Disabled |
+| Local Engine ID | Contact | Location | SNMP Traps | State |
+| --------------- | ------- | -------- | ---------- | ----- |
+| Switch System Mac | DC1_OPS | DC1 | All | Disabled |
 | DC1_OPS | DC1 | bgp, bridge, lldp, mpls, msdp backward-transition, msdp established, snmp link-down, snmpConfigManEvent | Enabled |
 | DC1_OPS | DC1 | bgp arista-backward-transition, bridge arista-mac-age | Disabled |
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
@@ -55,7 +55,7 @@ interface Management1
 
 | Local Engine ID | Contact | Location | SNMP Traps | State |
 | --------------- | ------- | -------- | ---------- | ----- |
-| Switch System Mac | DC1_OPS | DC1 | All | Disabled |
+| - | DC1_OPS | DC1 | All | Disabled |
 | DC1_OPS | DC1 | bgp, bridge, lldp, mpls, msdp backward-transition, msdp established, snmp link-down, snmpConfigManEvent | Enabled |
 | DC1_OPS | DC1 | bgp arista-backward-transition, bridge arista-mac-age | Disabled |
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
@@ -53,9 +53,9 @@ interface Management1
 
 ### SNMP Configuration Summary
 
-| Local Engine ID | Contact | Location | SNMP Traps | State |
-| --------------- | ------- | -------- | ---------- | ----- |
-| - | DC1_OPS | DC1 | All | Disabled |
+| Contact | Location | SNMP Traps | State |
+| ------- | -------- | ---------- | ----- |
+| DC1_OPS | DC1 | All | Disabled |
 | DC1_OPS | DC1 | bgp, bridge, lldp, mpls, msdp backward-transition, msdp established, snmp link-down, snmpConfigManEvent | Enabled |
 | DC1_OPS | DC1 | bgp arista-backward-transition, bridge arista-mac-age | Disabled |
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -53,9 +53,9 @@ interface Management1
 
 ### SNMP Configuration Summary
 
-| Contact | Location | SNMP Traps | State |
-| ------- | -------- | ---------- | ----- |
-| DC1_OPS | DC1 | All | Enabled |
+| Local Engine ID | Contact | Location | SNMP Traps | State |
+| --------------- | ------- | -------- | ---------- | ----- |
+| Switch System Mac | DC1_OPS | DC1 | All | Enabled |
 
 ### SNMP ACLs
 | IP | ACL | VRF |
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -116,10 +116,10 @@ interface Management1
 
 ### SNMP Users Configuration
 
-| User | Group | Version | Authentication | Privacy |
-| ---- | ----- | ------- | -------------- | ------- |
-| USER-READ | GRP-READ-ONLY | v3 | sha | aes |
-| USER-WRITE | GRP-READ-WRITE | v3 | sha | aes |
+| User | Group | Version | Authentication | Privacy | Remote IP | UDP Port | Engine ID |
+| ---- | ----- | ------- | -------------- | ------- | --------- | -------- | --------- |
+| USER-READ | GRP-READ-ONLY | v3 | sha | aes | - | - | - |
+| USER-WRITE | GRP-READ-WRITE | v3 | sha | aes | - | - | - |
 
 ### SNMP Device Configuration
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -116,8 +116,8 @@ interface Management1
 
 ### SNMP Users Configuration
 
-| User | Group | Version | Authentication | Privacy | Remote IP | UDP Port | Engine ID |
-| ---- | ----- | ------- | -------------- | ------- | --------- | -------- | --------- |
+| User | Group | Version | Authentication | Privacy | Remote Address | Remote Port | Engine ID |
+| ---- | ----- | ------- | -------------- | ------- | -------------- | ----------- | --------- |
 | USER-READ | GRP-READ-ONLY | v3 | sha | aes | - | - | - |
 | USER-WRITE | GRP-READ-WRITE | v3 | sha | aes | - | - | - |
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -57,7 +57,6 @@ interface Management1
 | ------- | -------- | ---------- | ----- |
 | DC1_OPS | DC1 | All | Enabled |
 
-
 ### SNMP ACLs
 | IP | ACL | VRF |
 | -- | --- | --- |
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -53,9 +53,10 @@ interface Management1
 
 ### SNMP Configuration Summary
 
-| Local Engine ID | Contact | Location | SNMP Traps | State |
-| --------------- | ------- | -------- | ---------- | ----- |
-| - | DC1_OPS | DC1 | All | Enabled |
+| Contact | Location | SNMP Traps | State |
+| ------- | -------- | ---------- | ----- |
+| DC1_OPS | DC1 | All | Enabled |
+
 
 ### SNMP ACLs
 | IP | ACL | VRF |
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
@@ -55,7 +55,7 @@ interface Management1
 
 | Local Engine ID | Contact | Location | SNMP Traps | State |
 | --------------- | ------- | -------- | ---------- | ----- |
-| Switch System Mac | DC1_OPS | DC1 | All | Enabled |
+| - | DC1_OPS | DC1 | All | Enabled |
 
 ### SNMP ACLs
 | IP | ACL | VRF |
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2318,8 +2318,12 @@ sflow:
 
 ```yaml
 snmp_server:
-  engineid:
-    local: <local_engine_id>
+  engine_ids:
+    local: < engine_name in hex >
+    remote:
+      - name: < engine_name in hex >
+        ip: < hostname | ip of remote engine >
+        port: < udp-port of remote engine >
   contact: < contact_name >
   location: < location >
   communities:
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2320,10 +2320,10 @@ sflow:
 snmp_server:
   engine_ids:
     local: < engine_name in hex >
-    remote:
-      - name: < engine_name in hex >
-        ip: < hostname | ip of remote engine >
-        port: < udp-port of remote engine >
+    remotes:
+      - id: < engine_name in hex >
+        address: <  hostname_or_ip_of_remote_engine >
+        port: < udp_port >
   contact: < contact_name >
   location: < location >
   communities:
@@ -2376,13 +2376,13 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
-      # remote_ip and udp_port are used for remote users
-      remote_ip: < ip_address >
-      # udp_port will not be used if no remote_ip is configured
+      # remote_address and udp_port are used for remote users
+      remote_address: < hostname_or_ip_of_remote_engine >
+      # udp_port will not be used if no remote_address is configured
       udp_port: < udp_port >
       version: < v1 | v2c | v3 >
-      # for a local user (i.e. no remote_ip) - the local engine ID should be
-      # used
+      # For a local user (i.e. no remote_ip), use the local engine_id
+      # For a remote user, use the remote engine_id
       localized: < engine_name in hex >
       auth: < hash_algorithm >
       auth_passphrase: < hashed_auth_passphrase if localized is used else cleartext auth_passphrase >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2318,6 +2318,8 @@ sflow:
 
 ```yaml
 snmp_server:
+  engineid:
+    local: <local_engine_id>
   contact: < contact_name >
   location: < location >
   communities:
@@ -2371,10 +2373,11 @@ snmp_server:
     - name: < username >
       group: < group_name >
       version: < v1 | v2c | v3 >
+      localized: < true | false > 
       auth: < hash_algorithm >
-      auth_passphrase: < encrypted_auth_passphrase >
+      auth_passphrase: < hashed_auth_passphrase if localized is true else cleartext auth_passphrase >
       priv: < encryption_algorithm >
-      priv_passphrase: < encrypted_priv_passphrase >
+      priv_passphrase: < hashed_priv_passphrase if localized is true else cleartext priv_passphrase >
     - name: < username >
       group: < group_name >
       version: < v1 | v2c | v3 >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2373,7 +2373,7 @@ snmp_server:
     - name: < username >
       group: < group_name >
       version: < v1 | v2c | v3 >
-      localized: < true | false > 
+      localized: < true | false >
       auth: < hash_algorithm >
       auth_passphrase: < hashed_auth_passphrase if localized is true else cleartext auth_passphrase >
       priv: < encryption_algorithm >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2319,11 +2319,11 @@ sflow:
 ```yaml
 snmp_server:
   engine_ids:
-    local: < engine_name in hex >
+    local: < engine_id in hex >
     remotes:
-      - id: < engine_name in hex >
+      - id: < engine_id in hex >
         address: <  hostname_or_ip_of_remote_engine >
-        port: < udp_port >
+        udp_port: < udp_port >
   contact: < contact_name >
   location: < location >
   communities:
@@ -2383,7 +2383,7 @@ snmp_server:
       version: < v1 | v2c | v3 >
       # For a local user (i.e. no remote_ip), use the local engine_id
       # For a remote user, use the remote engine_id
-      localized: < engine_name in hex >
+      localized: < engine_id in hex >
       auth: < hash_algorithm >
       auth_passphrase: < hashed_auth_passphrase if localized is used else cleartext auth_passphrase >
       priv: < encryption_algorithm >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2376,12 +2376,18 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
+      # remote_ip and udp_port are used for remote users
+      remote_ip: < ip_address >
+      # udp_port will not be used if no remote_ip is configured
+      udp_port: < udp_port >
       version: < v1 | v2c | v3 >
-      localized: < true | false >
+      # for a local user (i.e. no remote_ip) - the local engine ID should be
+      # used
+      localized: < engine_name in hex >
       auth: < hash_algorithm >
-      auth_passphrase: < hashed_auth_passphrase if localized is true else cleartext auth_passphrase >
+      auth_passphrase: < hashed_auth_passphrase if localized is used else cleartext auth_passphrase >
       priv: < encryption_algorithm >
-      priv_passphrase: < hashed_priv_passphrase if localized is true else cleartext priv_passphrase >
+      priv_passphrase: < hashed_priv_passphrase if localized is used else cleartext priv_passphrase >
     - name: < username >
       group: < group_name >
       version: < v1 | v2c | v3 >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
@@ -2319,9 +2319,9 @@ sflow:
 ```yaml
 snmp_server:
   engine_ids:
-    local: < engine_id in hex >
+    local: < engine_id_in_hex >
     remotes:
-      - id: < engine_id in hex >
+      - id: < engine_id_in_hex >
         address: <  hostname_or_ip_of_remote_engine >
         udp_port: < udp_port >
   contact: < contact_name >
@@ -2383,7 +2383,7 @@ snmp_server:
       version: < v1 | v2c | v3 >
       # For a local user (i.e. no remote_ip), use the local engine_id
       # For a remote user, use the remote engine_id
-      localized: < engine_id in hex >
+      localized: < engine_id_in_hex >
       auth: < hash_algorithm >
       auth_passphrase: < hashed_auth_passphrase if localized is used else cleartext auth_passphrase >
       priv: < encryption_algorithm >
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
@@ -26,7 +26,6 @@
 {%     endif %}
 {%     if snmp_server.engine_ids is arista.avd.defined %}
 
-
 ### SNMP EngineID Configuration
 
 | Type | Name (Hex) | IP | Port |
@@ -44,7 +43,6 @@
 {%         endfor %}
 {%     endif %}
 {%     if snmp_server.ipv4_acls is arista.avd.defined %}
-
 
 ### SNMP ACLs
 | IP | ACL | VRF |
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
@@ -4,8 +4,9 @@
 
 ### SNMP Configuration Summary
 
-| Contact | Location | SNMP Traps | State |
-| ------- | -------- | ---------- | ----- |
+| Local Engine ID | Contact | Location | SNMP Traps | State |
+| --------------- | ------- | -------- | ---------- | ----- |
+{%     set row_engineid = snmp_server.engineid.local | arista.avd.default('Switch System Mac') %}
 {%     set row_contact = snmp_server.contact | arista.avd.default('-') %}
 {%     set row_location = snmp_server.location | arista.avd.default('-') %}
 {%     if snmp_server.traps.enable is arista.avd.defined(true) %}
@@ -13,7 +14,7 @@
 {%     else %}
 {%         set row_state = 'Disabled' %}
 {%     endif %}
-| {{ row_contact }} | {{ row_location }} | All | {{ row_state }} |
+| {{row_engineid}} | {{ row_contact }} | {{ row_location }} | All | {{ row_state }} |
 {%     if snmp_server.traps.snmp_traps is arista.avd.defined %}
 {%         set row_traps_disabled = snmp_server.traps.snmp_traps | selectattr('enabled','false') | map(attribute = 'name') | arista.avd.natural_sort | join(', ') %}
 {%         set row_traps_enabled = snmp_server.traps.snmp_traps | rejectattr('enabled','false') | map(attribute = 'name') | arista.avd.natural_sort | join(', ') %}
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
@@ -6,7 +6,7 @@
 
 | Local Engine ID | Contact | Location | SNMP Traps | State |
 | --------------- | ------- | -------- | ---------- | ----- |
-{%     set row_engineid = snmp_server.engineid.local | arista.avd.default('Switch System Mac') %}
+{%     set row_engineid = snmp_server.engineid.local | arista.avd.default('-') %}
 {%     set row_contact = snmp_server.contact | arista.avd.default('-') %}
 {%     set row_location = snmp_server.location | arista.avd.default('-') %}
 {%     if snmp_server.traps.enable is arista.avd.defined(true) %}
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
@@ -4,9 +4,8 @@
 
 ### SNMP Configuration Summary
 
-| Local Engine ID | Contact | Location | SNMP Traps | State |
-| --------------- | ------- | -------- | ---------- | ----- |
-{%     set row_engineid = snmp_server.engineid.local | arista.avd.default('-') %}
+| Contact | Location | SNMP Traps | State |
+| ------- | -------- | ---------- | ----- |
 {%     set row_contact = snmp_server.contact | arista.avd.default('-') %}
 {%     set row_location = snmp_server.location | arista.avd.default('-') %}
 {%     if snmp_server.traps.enable is arista.avd.defined(true) %}
@@ -14,7 +13,7 @@
 {%     else %}
 {%         set row_state = 'Disabled' %}
 {%     endif %}
-| {{ row_engineid }} | {{ row_contact }} | {{ row_location }} | All | {{ row_state }} |
+| {{ row_contact }} | {{ row_location }} | All | {{ row_state }} |
 {%     if snmp_server.traps.snmp_traps is arista.avd.defined %}
 {%         set row_traps_disabled = snmp_server.traps.snmp_traps | selectattr('enabled','false') | map(attribute = 'name') | arista.avd.natural_sort | join(', ') %}
 {%         set row_traps_enabled = snmp_server.traps.snmp_traps | rejectattr('enabled','false') | map(attribute = 'name') | arista.avd.natural_sort | join(', ') %}
@@ -25,7 +24,27 @@
 | {{ row_contact }} | {{ row_location }} | {{ row_traps_disabled }} | Disabled |
 {%         endif %}
 {%     endif %}
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+
+
+### SNMP EngineID Configuration
+
+| Type | Name (Hex) | IP | Port |
+| ---- | ---------- | -- | ---- |
+{%         if snmp_server.engine_ids.local is arista.avd.defined %}
+| local |  {{ snmp_server.engine_ids.local }} | - | - |
+{%         endif %}
+{%         for engine_id in snmp_server.engine_ids.remote %}
+{%             if engine_id.name is arista.avd.defined and engine_id.ip is arista.avd.defined %}
+{%                 set row_name = engine_id.name | arista.avd.default('-') %}
+{%                 set row_ip = engine_id.ip | arista.avd.default('-') %}
+{%                 set row_port = engine_id.port | arista.avd.default('-') %}
+| remote | {{ row_name }} | {{ row_ip }} | {{ row_port }} |
+{%             endif %}
+{%         endfor %}
+{%     endif %}
 {%     if snmp_server.ipv4_acls is arista.avd.defined %}
+
 
 ### SNMP ACLs
 | IP | ACL | VRF |
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
@@ -159,15 +159,18 @@
 
 ### SNMP Users Configuration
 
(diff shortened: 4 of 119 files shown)


gmuloc:
So the changes are changing some outputs in some molecule:
* eos_cli_config_gen_v4.0
* evpn_underlay_ebgp_overlay_ebgp

this is because of the change in documentation - is the process to regenerate all these molecules as well?

tgodaA:
Yes, we need to update molecule on all scenarios.

gmuloc:
> Yes, we need to update molecule on all scenarios.

this has already been done in the latest commit

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp-server-traps.md
```diff
@@ -53,9 +53,9 @@ interface Management1
 
 ### SNMP Configuration Summary
 
-| Contact | Location | SNMP Traps | State |
-| ------- | -------- | ---------- | ----- |
-| DC1_OPS | DC1 | All | Disabled |
+| Local Engine ID | Contact | Location | SNMP Traps | State |
+| --------------- | ------- | -------- | ---------- | ----- |
+| Switch System Mac | DC1_OPS | DC1 | All | Disabled |
```
ClausHolbechArista: The default value of the new field should just be `-`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -105,15 +108,27 @@ snmp-server location {{ snmp_server.location }}
 {%             if user.version is arista.avd.defined %}
 {%                 set user_cli = user_cli ~ " " ~ user.version %}
 {%             endif %}
-{%             if user.auth is arista.avd.defined
+{%             if user.localized is arista.avd.defined
```
ClausHolbechArista: Missing check for `true`.

Instead of repeating the code, the idea with the `user_cli` var, is that you can just build the parts you need. I would go back to the old code, and then just insert another `if user.localized is arista.avd.defined(true) and user.version is arista.avd.defined('v3') and snmp_server.engineid.local is arista.avd.defined`. The CLI accepts just the `localized <engineid>` CLI, without auth following. So it should not be combined with the checks for auth. You _could_ move the check for v3 out to a wider if block covering both.

gmuloc: The CLI does accept it but then it is not part of the show running config if you don't have auth:
```
Spine1(config)#snmp-server user blah blahgroup v3 localized 424242424242
Spine1(config)#show run | se blah
snmp-server user blah blahgroup v3
```

so there is no point in configuring it righ?

gmuloc: that being said happy to reorganise the code with a bunch of if statement to insert localized or not but given the behavior I would end up adding the lines:

`if user.localized is arista.avd.defined(true) and user.version is arista.avd.defined('v3') and snmp_server.engineid.local is arista.avd.defined and user.auth is arista.avd.defined and user.auth_passphrase is arista.avd.defined `

Also the previous code could lead to configure priv without auth (even without the `localized` keyword) which is not impossible:
```
Leaf1(config)#snmp-server user test test-group v3 ?
  auth       User authentication settings
  localized  Specify localized auth and privacy passphrases
  <cr>
```

ClausHolbechArista: If you don't want to output localized alone, you can still just add a `if user.localized is arista.avd.defined(true)` inside the existing `if user.auth ...`. It would make it simpler to add other knobs later.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,8 @@ sflow:
 
 ```yaml
 snmp_server:
+  engineid:
+    local: <local_engine_id>
```
ccsnw: I think it would make more sense to use the plural for the keyname: "engine_ids" (because we can have one local and multiple remote ids) and add the parameters as a list:

```
snmp_server:
<snip>
  engine_ids:
    - type: < local | remote >
      name: < engine_name in hex >
      ip: < hostname | ip of remote engine >
      port: < udp-port of remote engine >
</snip>
```

Drawback: This allows to create multiple local ID's even only one is possible

Advantage: You can easily include the remote ID's

==

Other alternative (only allows 1 local ID and multiple remote IDs)

```
engine_ids:
  local: < engine_name in hex >
  remote:
      -name: < engine_name in hex >
       ip: < hostname | ip of remote engine >
       port: < udp-port of remote engine >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: indeed - that would work - I will modify the code

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: I was not planning to add the remote engine IDs feature in this PR as I was really targeting adding the localized part (and then from eos_designs adding the possibility to hash some passwords protected in a vault to generate the eos_cli_config_gen part) but I can see you opened another one so lets maybe do everything in one go.

I guess I would rather go with your second option to make sure there is always one local id to keep things simple. What are your thoughts?

ccsnw: makes sense to me! 

for the remote's the logic would be quite easy, as you don't have to "cross" check if localized is true or not, I think you can easily add it it with this PR. Otherwise I can add it later once your PR is merged

gmuloc: I have updated my PR with the discussed structure - I have not added the remote support for SNMPv3 users in it as there are some points I am not sure about - typically I can configure this and it will show up in the confg:

```
snmp-server user blah blahg remote 1.1.1.1 udp-port 1337 v3 localized 121212121212 auth sha 1234567890123456789012345678901234567890
```

but I am not sure how this is supposed to relate to a configured remote engine ID on the same device

gmuloc: I have also adopted your documentation style (which make the PR not affect all the other testbeds and makes it easier to review) - thanks!

ccsnw: cool! skip the remote user for now (keep the remote engine config)

ccsnw: @gmuloc the documentation is a bit confusing about the remote users. Let me try to sum it up:

first you will need to configure the remote engines name on the local system:
```
snmp-server engineID remote <REMOTE-IP> <REMOTENEGINEID>
```

We already have this in the DM (the list under remote: key)

==

Next you'll need to add the user of the remote agent to the local system:
```
snmp-server user <USERNAME> <GROUPNAME> remote <REMOTEIP> v3 auth sha <CLEARTEXT AUTHPW> priv aes <CLEARTEXT PRIV PW>
```

if you do it like this, EOS will use the REMOTENEGINEID from the previous command to encrypt the cleartexts.

you can also add the remote users by explicitly outting the REMOTEENGINEID in localization:
```
snmp-server user <USERNAME> <GROUPNAME> remote <REMOTEIP> v3 localized <REMOTENEGINEID> auth sha <CLEARTEXT AUTHPW> priv aes <CLEARTEXT PRIV PW>
```

==

the remote users are needed to send snmp informs (instead of traps) using snmpv3.

@ClausHolbechArista et al. Please correct me if I'm wrong about this.

gmuloc: Thanks @ccsnw  - so essentially this is not too difficult to add for the `remote <REMOTEIP> udp-port <UDP-PORT>` part in the user declaration - the only thing is that it then requires also to change the `localized` variable as it is now possible to give any value to it - I initially used a boolean to be able to pick up the local engine ID and automatically regenerate the correct user CLI (well the keys still need to be changed) when changing the engine ID:
```
    users:
      - name: < username >
        group: < group_name >
        version: < v1 | v2c | v3 >
        localized: < true | false >
```

So now I am thinking that it should look like:
```
    users:
      - name: < username >
        group: < group_name >
        # remote_ip and udp_port are used for remote users
        remote_ip: < ip_address >
        # udp_port will not be used if no remote_ip is configured
        udp_port: < udp_port >
        version: < v1 | v2c | v3 >
        localized: < engine_name in hex >
```
and just like for the regular CLI the user is in charge of giving the correct value there.

ccsnw: @gmuloc yup, will work if you turn localized into a hex string (also did it like that in a PR that I had prepared before seeing yours)

if the user doesn't provide the hex for the localized key, they pw's are expected in cleartext. Think you can add this to the readme if its not in there yet 

"and just like for the regular CLI the user is in charge of giving the correct value there."
>> Yes, but the Arista guys are working on input validation for v4 of AVD

gmuloc: Ack - changes pushed in the latest commit

> if the user doesn't provide the hex for the localized key, they pw's are expected in cleartext. Think you can add this to the readme if its not in there yet

Yeah that's in the README

> Yes, but the Arista guys are working on input validation for v4 of AVD

hehe - I guess I should take a look at what's going on there as well

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2370,11 +2376,18 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
+      # remote_ip and udp_port are used for remote users
+      remote_ip: < ip_address >
+      # udp_port will not be used if no remote_ip is configured
+      udp_port: < udp_port >
       version: < v1 | v2c | v3 >
+      # for a local user (i.e. no remote_ip) - the local engine ID should be
+      # used
```
ClausHolbechArista:
```suggestion
      # For a local user (i.e. no remote_ip), use the local engine_id
      # For a remote user, use the remote engine_id
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2370,11 +2376,18 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
+      # remote_ip and udp_port are used for remote users
```
ClausHolbechArista: Maybe we could match the model similar to the engine_ids:
```yaml
...
      remote:
        ip: < hostname_or_ip_of_remote_engine >
        port: < udp_port >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,12 @@ sflow:
 
 ```yaml
 snmp_server:
+  engine_ids:
+    local: < engine_name in hex >
+    remote:
+      - name: < engine_name in hex >
+        ip: < hostname | ip of remote engine >
+        port: < udp-port of remote engine >
```
ClausHolbechArista: Remotes should be in plural since it is a list.
```suggestion
    local: < engine_id in hex >
    remotes:
      - id: < engine_id in hex >
        ip: < hostname_or_ip_of_remote_engine >
        port: < udp_port >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -139,15 +159,18 @@
 
 ### SNMP Users Configuration
 
-| User | Group | Version | Authentication | Privacy |
-| ---- | ----- | ------- | -------------- | ------- |
+| User | Group | Version | Authentication | Privacy | Remote IP | UDP Port | Engine ID |
```
ClausHolbechArista:
```suggestion
| User | Group | Version | Authentication | Privacy | Remote Address | Remote Port | Engine ID |
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -102,18 +117,27 @@ snmp-server location {{ snmp_server.location }}
...
+{%                 if user.udp_port is arista.avd.defined %}
+{%                     set user_cli = user_cli ~ " udp-port " ~ user.udp_port %}
+{%                 endif %}
+{%             endif %}
 {%             if user.version is arista.avd.defined %}
 {%                 set user_cli = user_cli ~ " " ~ user.version %}
 {%             endif %}
 {%             if user.auth is arista.avd.defined
                   and user.version is arista.avd.defined('v3')
                   and user.auth_passphrase is arista.avd.defined %}
```
ccsnw, line 131: Maybe we should rethink this condition (the 2nd and which checks if passphrase is set). The CLI allows the following:

just setting "v3" without any auth or priv pass

it also allows just setting "v3 and localized BLABLA" (it will accept this, but omit the localized in the config)

=======

To me it would make sense to build the CLI step by step, so "plain" v3 only would be allowed, "v3 localized xxx" would be allowed, just auth allowed (no priv specified) and the same with localized....

most easiest would be to simply construct the cli from whatever the user inputs.. and keep the conditions minimal.

What do you think?

ClausHolbechArista: Will the v3 alone work?
We already discussed the localized option alone, but since it is removed by EOS, we decided to ignore that option.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: we could also consider using `name` or `address` instead of `ip`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
ccsnw: yup, just v3 works.. also appears like that in the running config:

```
snmp-server engineID local 617269737461
snmp-server group bla v3 priv
snmp-server user bla test v3
```

you can use unauthenticated snmpv3. In general you just need a username to use snmpv3.. kinda defeats the purpose of it, but it is possible.

What troubles me more (and might be worth addressing to EOS developers) is that EOS currently accepts auth passwords < 8 characters (cleartext). 

Problem occurs if you want to query these agents with netsnmp for example:

```
Error: passphrase chosen is below the length requirements of the USM (min=8).
snmpwalk:  (The supplied password length is too short.)
Error generating a key (Ku) from the supplied authentication pass phrase.
```

At least SHA seems to require at least 8 character password. Might make sense to enforce 8char min on OS level, or?

>> Ignore. this Issue explained. Just lacking the documentation

ccsnw: with snmpv3 they tried to make up for everything that snmpv2 was missing while also still allowing what snmpv2 allowed.. (my opinion).

thats why you have: unauthenticated, authenicated but no priv, and authpriv and so on and so forth :-D

and then came telemetry streaming which will hopefully be the end of snmp in general at some point.

ClausHolbechArista: I think we are getting sidetracked here. Support for v3 alone could be nice. @gmuloc

gmuloc: It works already and is part of the tests: https://github.com/aristanetworks/ansible-avd/pull/1721/files#diff-d65839360352a741a4aaf63924e24a87127c419b98ff93f2758a748f3aaf744cR164

gmuloc: line 167 in the same file is also there to check that when localized is used but no auth or priv are given only v3 remains.

ClausHolbechArista: sorry I missed that.

gmuloc: @ccsnw - regarding removing the localized keywords to match the actual running-config rather than what is configurable - I just verified two extra points

1. Behavior is the same for local and remote users
2. I check what happens in CVP as if CVP were to push the localized keywords it should be seeing the device out of sync.

I Created this Configlet:

```
snmp-server user CVP-TEST CVP-GROUP v3 localized 42424242424242
```

and this is the configuration generated (I noticed that it adds the System Mac as local engine ID by itself which is not supported in this PR though I guess it _could_ be considered as a future optimization):
Screenshot 2022-04-21 at 09 28 38

so as discussed with Claus earlier I feel confident this can be removed in this situation ( and there is a molecule test to check the behavior)

ccsnw: @gmuloc 

I'd guess that EOS creates/sets the local engineID (sysmac) when you create any snmpv3 user and dont explicitly set the engineID yourself.

For remote agents, EOS cannot do that (because it doesnt know the sysmac of the remote system) and hence you have to define it manually and before remote user creation

==

anout the compliance: it's true, if we allow AVD to render "snmp-server user ... v3 localized..." and push via CVP, this will lead to a compliance error, because EOS never have a line like "snmp-server user bla test v3 321313123131" in the running-config.

My personal feeling is to put as little logic in here as possible and let the user make the choices..

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: Addressed in f219642

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: Addressed in f219642

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
gmuloc: Addressed in f219642

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
```diff
@@ -57,6 +57,7 @@ interface Management1
 | ------- | -------- | ---------- | ----- |
 | DC1_OPS | DC1 | All | Enabled |
 
+
```
ClausHolbechArista: this extra space is caused by some template change. I will try to identify it.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -102,18 +117,27 @@ snmp-server location {{ snmp_server.location }}
...
+{%                     set user_cli = user_cli ~ " localized " ~ user.localized %}
+{%                 endif %}
 {%                 set user_cli = user_cli ~ " auth " ~ user.auth ~ " " ~ user.auth_passphrase %}
-{%             endif %}
-{%             if user.priv is arista.avd.defined
-                  and user.version is arista.avd.defined('v3')
-                  and user.priv_passphrase is arista.avd.defined %}
-{%                 set user_cli = user_cli ~ " priv " ~ user.priv ~ " " ~ user.priv_passphrase %}
+{%                 if user.priv is arista.avd.defined
+                      and user.version is arista.avd.defined('v3')
```
ClausHolbechArista: this check for v3 is redundant since it is nested under the `auth` which also checks for this.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,8 +24,28 @@
...
+{%                 set row_id = engine_id.id | arista.avd.default('-') %}
+{%                 set row_address = engine_id.address | arista.avd.default('-') %}
+{%                 set row_port = engine_id.port | arista.avd.default('-') %}
+| remote | {{ row_id }} | {{ row_address }} | {{ row_port }} |
+{%             endif %}
+{%         endfor %}
+{%     endif %}
 {%     if snmp_server.ipv4_acls is arista.avd.defined %}
 
+
```
ClausHolbechArista: remove this extra blank line

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
ccsnw: the extra spaces in the j2 of the doc, let me comment there

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,8 +24,28 @@
 | {{ row_contact }} | {{ row_location }} | {{ row_traps_disabled }} | Disabled |
 {%         endif %}
 {%     endif %}
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+
+
```
ccsnw: only one empty line here

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: Indeed you are right, by default the engineID is derived from the System Mac:
```
Leaf2#show run | i engine
Leaf2#show snmp engineID
Local SNMP EngineID: f5717f502200715bad00
Leaf2#show version | i MAC
Hardware MAC address: 5022.0071.5bad
```

so `f5117f + < system mac > + 00`.

Regarding the CVP test, CVP is actually removing the `localized` value from the config it pushes if there is no `auth` following so in the end the only location where the localized engineID would be present is in the AVD configuration

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: Addressed in f219642

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
ccsnw: "Regarding the CVP test, CVP is actually removing the localized value from the config it pushes if there is no auth following so in the end the only location where the localized engineID would be present is in the AVD configuration"
>> Yup, but cvp will not modify the configlet right?
>> so if our render has "... v3 localized xxx" then cvp only pushes the "... v3" part and it will complain that designed and running config are out of sync?

gmuloc: The configlet is not modified indeed but the designed configuration generated is with `... v3` only as in the screenshot above. Then CVP does not complain since the comparison is done between the designed config and the running config.

ccsnw: got it, thx

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/snmp.md
gmuloc: Thanks for catching this! addressed in 81d0d63

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
gmuloc: Addressed in 81d0d63 - thanks!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
gmuloc: Addressed in 81d0d63 - thanks!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: Missed this one - addressed in 81d0d63 as well

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
 | {{ row_contact }} | {{ row_location }} | {{ row_traps_disabled }} | Disabled |
 {%         endif %}
 {%     endif %}
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+
+### SNMP EngineID Configuration
+
+| Type | Name (Hex) | IP | Port |
```
ClausHolbechArista:
```suggestion
| Type | EngineID (Hex) | IP | Port |
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
...
+| ---- | ---------- | -- | ---- |
+{%         if snmp_server.engine_ids.local is arista.avd.defined %}
+| local |  {{ snmp_server.engine_ids.local }} | - | - |
```
ClausHolbechArista:
```suggestion
| local | {{ snmp_server.engine_ids.local }} | - | - |
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -102,18 +117,26 @@ snmp-server location {{ snmp_server.location }}
 {%             if user.group is arista.avd.defined %}
 {%                 set user_cli = user_cli ~ " " ~ user.group %}
 {%             endif %}
+{%             if user.remote_address is arista.avd.defined %}
```
ClausHolbechArista: Remotes only work with v3. Maybe I told you to remove this earlier :) but I thought it was nested inside the v3 check.
```suggestion
{%             if user.remote_address is arista.avd.defined and user.version is arista.avd.defined('v3') %}
```

ClausHolbechArista (changes requested):
Minor adjustments. I have added them as suggestions, so you can just add them to batch and commit everything on github.

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -1,6 +1,21 @@
...
 {% if snmp_server is arista.avd.defined %}
 !
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+{%         if snmp_server.engine_ids.local is arista.avd.defined %}
+snmp-server engineID local {{ snmp_server.engine_ids.local }}
+{%         endif %}
+{%         for engine_id in snmp_server.engine_ids.remotes %}
+{%             if engine_id.id is arista.avd.defined and engine_id.address is arista.avd.defined %}
+{%                 set remote_engine_ids_cli =  "snmp-server engineID remote " ~ engine_id.address %}
+{%                 if engine_id.port is arista.avd.defined %}
```
tgodaA: Should this knob be ```port``` or ```udp_port``` to match the cli? Any suggestions @ClausHolbechArista @carlbuchmann

tgodaA: IMO, udp_port sounds better as we used the same knob with in snmp_server.users

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
 | {{ row_contact }} | {{ row_location }} | {{ row_traps_disabled }} | Disabled |
 {%         endif %}
 {%     endif %}
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+
+### SNMP EngineID Configuration
+
+| Type | EngineID (Hex) | IP | Port |
+| ---- | ---------- | -- | ---- |
```
tgodaA: !! Not mandatory, just my opinion. Can you make the ```----``` lines in sync with the above row.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
ClausHolbechArista: I think we had udp_port under the users, but changed it to port to match the engine data model. If we change to udp_port, please change both engine_id and users models

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -1,6 +1,21 @@
 {# eos - SNMP Settings#}
 {% if snmp_server is arista.avd.defined %}
 !
+{%     if snmp_server.engine_ids is arista.avd.defined %}
+{%         if snmp_server.engine_ids.local is arista.avd.defined %}
+snmp-server engineID local {{ snmp_server.engine_ids.local }}
+{%         endif %}
```
tgodaA:
```suggestion
{%     if snmp_server.engine_ids.local is arista.avd.defined %}
snmp-server engineID local {{ snmp_server.engine_ids.local }}
{%     endif %}
```

tgodaA: To match the cli output we need to place the ```snmp_server.engine_ids.remotes``` after line 143. So I suggested the above change for lines 4-7.

FYI:
image

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -1,6 +1,21 @@
...
+{%         for engine_id in snmp_server.engine_ids.remotes %}
```
tgodaA, line 14: When you move this for loop after line143, please make sure to add defined test for ```snmp_server.engine_ids.remotes``` instead of ```snmp_server.engine_ids```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
...
+{%         if snmp_server.engine_ids.local is arista.avd.defined %}
+| local | {{ snmp_server.engine_ids.local }} | - | - |
+{%         endif %}
+{%         for engine_id in snmp_server.engine_ids.remotes %}
```
tgodaA:
```suggestion
{%         for engine_id in snmp_server.engine_ids.remotes | arista.avd.default([]) %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
tgodaA: The users model is having ```udp_port```.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
tgodaA: Currently you added defined test only for ```snmp_server.engine_ids``` but not ```snmp_server.engine_ids.remotes```. So when ever someone defines ```snmp_server.engine_ids``` without **remotes** the current code will throw error. So I suggested the above.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
...
+{%             if engine_id.id is arista.avd.defined and engine_id.address is arista.avd.defined %}
+{%                 set row_id = engine_id.id | arista.avd.default('-') %}
+{%                 set row_address = engine_id.address | arista.avd.default('-') %}
```
tgodaA: We don't need to set the lines 38,39 as you have already done the **defined** test for **engine_id.id** and **engine_id.address**, which will make sure they are defined as well as not None. You can directly use those vars in line 41.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/snmp-settings.j2
```diff
@@ -24,6 +24,24 @@
...
+{%                 set row_port = engine_id.port | arista.avd.default('-') %}
+| remote | {{ row_id }} | {{ row_address }} | {{ row_port }} |
+{%             endif %}
```
tgodaA:
```suggestion
{%             if engine_id.id is arista.avd.defined and engine_id.address is arista.avd.defined %}
{%                 set row_port = engine_id.port | arista.avd.default('-') %}
| remote | {{ engine_id.id }} | {{ engine_id.address }} | {{ row_port }} |
{%             endif %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,12 @@ sflow:
 
 ```yaml
 snmp_server:
+  engine_ids:
+    local: < engine_name in hex >
+    remotes:
+      - id: < engine_name in hex >
+        address: <  hostname_or_ip_of_remote_engine >
+        port: < udp_port >
```
tgodaA:
```suggestion
        udp_port: < integer >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2370,11 +2376,18 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
+      # remote_address and udp_port are used for remote users
+      remote_address: < hostname_or_ip_of_remote_engine >
+      # udp_port will not be used if no remote_address is configured
+      udp_port: < udp_port >
```
tgodaA, line 2382:
```suggestion
      udp_port: < integer >
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,12 @@ sflow:
 
 ```yaml
 snmp_server:
+  engine_ids:
+    local: < engine_name in hex >
```
tgodaA: local: < engine_id in hex >, this is engine ID right?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,12 @@ sflow:
...
+    remotes:
+      - id: < engine_name in hex >
```
tgodaA: id: < engine_id in hex >, this is engine ID right?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2318,6 +2318,12 @@ sflow:
...
+        address: <  hostname_or_ip_of_remote_engine >
```
tgodaA, line 2325: address: <  hostname | ip_of_remote_engine >

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2370,11 +2376,18 @@ snmp_server:
   users:
     - name: < username >
       group: < group_name >
+      # remote_address and udp_port are used for remote users
+      remote_address: < hostname_or_ip_of_remote_engine >
```
tgodaA, line 2380: remote_address:  < hostname | ip_of_remote_engine >

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2370,11 +2376,18 @@ snmp_server:
...
+      # udp_port will not be used if no remote_address is configured
+      udp_port: < udp_port >
       version: < v1 | v2c | v3 >
+      # For a local user (i.e. no remote_ip), use the local engine_id
+      # For a remote user, use the remote engine_id
+      localized: < engine_name in hex >
```
tgodaA: localized: < engine_id in hex >

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: I don't agree. Hostname of what. Normally we use the pipe to describe specific options. Like fixed strings. Here it is just a string.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: I assume you mean line 14 here right? but yeah good catch thanks!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: there is no need to test for existence - for loop on an empty list does not print anything

gmuloc: Should I test existence for both?

gmuloc: Nope - interesting ok :)! will do

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: yes - not sure what you mean by that comment maybe removing `hex`?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: so happy to change this but this is not aligned with the Vxlan udp_port description in the same document: `udp_port: < udp_port >` nor the cvaddrs suggested formatting: `< ip/fqdn >:<port>` or the vrf logging ports: `ports: - < custom_port_1 >`. Is there a rule regarding what the description should be?

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
gmuloc: hehe - one of the renaming did not make it through!

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: I vote for `udp_port: < udp_port >`

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ClausHolbechArista: `local: < engine_id_in_hex >`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
```diff
@@ -1,12 +1,27 @@
...
+{%             if engine_id.id is arista.avd.defined and engine_id.address is arista.avd.defined %}
+{%                 set remote_engine_ids_cli =  "snmp-server engineID remote " ~ engine_id.address %}
+{%                 if engine_id.udp_port is arista.avd.defined %}
+{%                     set remote_engine_ids_cli =  remote_engine_ids_cli ~ " udp-port " ~  engine_id.udp_port %}
+{%                 endif %}
+{%                 set remote_engine_ids_cli = remote_engine_ids_cli ~ " " ~ engine_id.id %}
+{{ remote_engine_ids_cli }}
+{%             endif %}
+{%         endfor %}
+{%     endif %}
```
tgodaA, line 24: Please move this block line13-24 (snmp_server.engineids.remotes) below line143(snmp_server.users) to match the cli output.

tgodaA: image

gmuloc: Actually it will be more complex than this as it needs to be between local and remote users:
```
Leaf1#show run | se snm
snmp-server engineID local 424242424242424242
snmp-server contact test
snmp-server user test3 test-group v3
snmp-server engineID remote 42.42.42.42 BABABABABABAB
snmp-server user giza hsajsd remote 1.1.1.1 v3
snmp-server user blah blahg remote 1.1.1.1 udp-port 13
```

tgodaA: Oh! Ok. That's no problem.

tgodaA (approved):
LGTM

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/snmp-settings.j2
gmuloc: to accommodate for this we would need to loop twice through the users, first print the local users, then the remote engineIDs and then the remote ones.

tgodaA: Thats ok. This order formatting is not mandatory. We will try to implement if it's possible.

PR #1719, opened 2022-04-13
Feat(eos_designs): Support for IPv6 overlay peerings with RFC5549 underlay

xaviramon:
## Change Summary

Adding features to build the overlay peering using IPv6 addresses. Supporting eBGP with RFC5549 underlay address scheme.

PR for underlay still required to make the entire functionality to work.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Create a new switch variable to control overlay routing in IPv6 space.
Shutdown the IPv6 address family.
Add the neighbor statement depending on overlay used (IPv4 vs IPv6).

## How to test
Molecule

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #1723, opened 2022-04-13
Feat(eos_designs): Support for IPv6 in underlay with RFC5549

xaviramon:
## Change Summary

PR required to make #1719 work. Underlay supports IPv6 with RFC5549 and transport of IPv4 VTEP addresses

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd. eos_designs `

## Proposed changes
Create new peer-groups for IPv6 AFs
Generate AFs as required
Modify PL and RM to support IPv6 addressing

## How to test
molecule

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista (changes requested):
I think you are doing too many changes to ipv4 things in here. Basically you should just be _adding_ IPv6 AF for the existing underlay protocol. It should _not_ change everything else.
If someone wants to use different peer-group names, they can just set the names on the existing keys.
Also since you are not removing IPv4, the new name with IPv6 is just as wrong as the IPv4 name.
I suggest to simplify this PR, to just add the underlay AF, loopback, PL etc, and then leave the naming as it is. We can add a new - more neutral - `underlay_peers` and `mlag_underlay_peer` later (separate PR)

ClausHolbechArista:
Also the requirement for rfc5549 is not clear to me. It should just work for regular ibgp/ebgp also right? I understand it will not work for OSPF/ISIS as part of this PR, but I think those are already covered by Emil in the MPLS design.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -47,6 +47,7 @@ p2p_uplinks_mtu: 9000
 
 underlay_routing_protocol: BGP
 overlay_routing_protocol: BGP
+underlay_ipv6: false
```
ClausHolbechArista, line 50: since we use `arista.avd.defined` checks, we don't need this.

ansible_collections/arista/avd/roles/eos_designs/templates/mlag/router-bgp.j2
```diff
@@ -14,6 +14,12 @@ router_bgp:
       send_community: all
 {%     if switch.mlag_ibgp_origin_incomplete == true %}
       route_map_in: RM-MLAG-PEER-IN
+{%     endif %}
+{%     if underlay_ipv6 is arista.avd.defined(true) %}
```
ClausHolbechArista, line 18: if rfc5549 is a strict requirement, we should set a switch.underlay_ipv6 fact, based on checks for both in the switch facts. This would make it possible to add other conditions later like when we get ipv6 p2p support some day.

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/ebgp/prefix-lists.j2
```diff
@@ -1,14 +1,19 @@
 {# prefix-lists #}
 prefix_lists:
+{% if underlay_ipv6 is arista.avd.defined(true) and switch.vtep is arista.avd.defined(false) %}
```
ClausHolbechArista: Please remove this. We are not removing ipv4. We are adding ipv6.

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/ebgp/prefix-lists.j2
```diff
@@ -1,14 +1,19 @@
...
+{% else %}
   PL-LOOPBACKS-EVPN-OVERLAY:
     sequence_numbers:
+{%     if underlay_ipv6 is arista.avd.defined(false) %}
```
ClausHolbechArista: Please remove this. We are not removing ipv4. We are adding ipv6.

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/ebgp/route-maps.j2
```diff
@@ -2,8 +2,17 @@
 route_maps:
   RM-CONN-2-BGP:
     sequence_numbers:
+{% if underlay_ipv6 is arista.avd.defined(true) and switch.vtep is arista.avd.defined(false) %}
```
ClausHolbechArista: Please remove this. We are not removing ipv4. We are adding ipv6.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
ClausHolbechArista: we will keep it for clarity

ClausHolbechArista (approved):

onurgashi (approved):

onurgashi:
LGTM!

PR #1728, opened 2022-04-20
Fix(eos_cli_config_gen): Adjust the address-family evpn host-flap detection

tgodaA:
## Change Summary

The host-flap detection command under address-family evpn is adding multiple commands instead it should be single command. This is fixed and added the expiry timer knob as well.

## Component(s) name

## Proposed changes

The command should be single line instead it was multiple commands previously. This will overwrite the first command.

Expected (tested on device):
```
router bgp 65000
   address-family evpn
      host-flap detection window 180.0 threshold 21
```

But the old model was generating:
```
router bgp 65000
   address-family evpn
      host-flap detection window 180.0
      host-flap detection threshold 21
```

Also expiry timeout command was supported. So the current model can yield an output:
```
router bgp 65000
   address-family evpn
      host-flap detection window 180.0 threshold 21 expiry timeout 34.0 seconds
```

## How to test

```molecule converge -s eos_cli_config_gen```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
@@ -145,9 +145,9 @@ interface Management1
 
 #### EVPN Host Flapping Settings
 
-| State | Window | Threshold |
-| ----- | ------ | --------- |
-| Enabled | 10 |  1 |
+| State | Window | Threshold | Expiry Timeout |
+| ----- | ------ | --------- | -------------- |
+| Enabled | 10 |  1 | 3 |
 
 #### EVPN DCI Gateway Summary
 
@@ -243,8 +243,7 @@ router bgp 65101
       vlan 112
    !
    address-family evpn
-      host-flap detection window 10
-      host-flap detection threshold 1
+      host-flap detection window 10 threshold 1 expiry timeout 3 seconds
       domain identifier 65101:0
       neighbor EVPN-OVERLAY-PEERS activate
       neighbor EVPN-OVERLAY-PEERS domain remote
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
@@ -147,7 +147,7 @@ interface Management1
 
 | State | Window | Threshold | Expiry Timeout |
 | ----- | ------ | --------- | -------------- |
-| Enabled | 10 |  1 | 3 |
+| Enabled | 10 Seconds | 1 | 3 Seconds |
 
 #### EVPN DCI Gateway Summary
 
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
@@ -357,7 +357,7 @@ router bgp {{ router_bgp.as }}
    address-family evpn
 {%         if router_bgp.address_family_evpn.evpn_hostflap_detection.enabled is arista.avd.defined(false) %}
       no host-flap detection
-{%         else %}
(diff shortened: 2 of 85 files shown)

### CODE REVIEW

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2 [hunk 1]
```diff
@@ -358,11 +358,18 @@ router bgp {{ router_bgp.as }}
 {%         if router_bgp.address_family_evpn.evpn_hostflap_detection.enabled is arista.avd.defined(false) %}
       no host-flap detection
 {%         else %}
+{%             set host_flap_detection_cli = '' %}
 {%             if router_bgp.address_family_evpn.evpn_hostflap_detection.window is arista.avd.defined %}
-      host-flap detection window {{ router_bgp.address_family_evpn.evpn_hostflap_detection.window }}
+{%                 set host_flap_detection_cli = host_flap_detection_cli ~ ' window ' ~ router_bgp.address_family_evpn.evpn_hostflap_detection.window %}
```
ClausHolbechArista: if either window or threshold is given, EOS automatically adds the other to the CLI line using the default value. So to be consistent I think we should do the same. Please test it out.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
```diff
@@ -358,11 +358,18 @@ router bgp {{ router_bgp.as }}
 {%         if router_bgp.address_family_evpn.evpn_hostflap_detection.enabled is arista.avd.defined(false) %}
       no host-flap detection
 {%         else %}
+{%             set host_flap_detection_cli = '' %}
```
ClausHolbechArista: I would prefer to have this be `host-flap detection` and then have an `elif` instead of else above, to test if any of the knobs have been set.

tgodaA: Sure, will change the logic for .enabled

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-bgp.j2
tgodaA: Tested but didn't include the default values as I am not sure if they will change anytime in future. Also, considering the logic if only expiry timer was provided then window and threshold doesn't pick up their default. So going with the current one.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
```diff
@@ -145,9 +145,9 @@ interface Management1
 
 #### EVPN Host Flapping Settings
 
-| State | Window | Threshold |
-| ----- | ------ | --------- |
-| Enabled | 10 |  1 |
+| State | Window | Threshold | Expiry Timeout |
```
emilarista, line 148: Would maybe be nice with units on these, so for window it's seconds, threshold is number of moves, expiry timeout is also seconds, so maybe | State | Window (seconds) | Threshold (moves) | Expiry Timeout (seconds) | ... Or just put the unit in the value rows... Either way.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
```diff
@@ -145,9 +145,9 @@ interface Management1
...
+| ----- | ------ | --------- | -------------- |
+| Enabled | 10 |  1 | 3 |
```
emilarista: Extra whitespace in threshold column here.

emilarista (dismissed):

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-evpn.md
tgodaA: Thats from previous. I will fix it.

carlbuchmann (approved):
LGTM!

lermilov:
Looks awesome!

PR #1737, opened 2022-04-21
Feat(eos_cli_config_gen): Add Tap Aggregation support

ccsnw:
## Change Summary

- Allow to configure a switch in "tap aggregation" exclusive mode, mixed mode can be added later
- Allow configuration of related global options

## Related Issue(s)

Fixes #N/A

## Component(s) name

## Proposed changes
```
tap_aggregation:
  mode:
    exclusive:
      enabled: < true | false >
      profile: < profile_name | default --> tap-aggregation-default >
      no-errdisable:
        - < EthernetX | Port-ChannelX >
  encapsulation_dot1br_strip: < true | false >
  encapsulation_vn_tag_strip: < true | false >
  protocol_lldp_trap: < true | false >
  # Allowed truncation_size values vary depending on the platform
  truncation_size: < size in bytes >
  mac:
    # mac.timestamp.replace_source_mac and mac.timestamp.header.format are mutually exclsuive. If both are defined, replace_source_mac takes precedence
    timestamp:
      replace_source_mac: < true | false >
      header:
        format: < "48-bit" | "64-bit" >
        eth_type: < integer >
    # mac_fcs_append and mac_fcs_error are mutually exclusive. If both are defined, mac_fcs_append takes precedence
    fcs_append: < true | false >
    fcs_error: < "correct" | "discard" | "pass-through" >
```

## How to test
Molecule test will be added once initial DM is considered Ok.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2470,6 +2470,23 @@ monitor_sessions:
       size: < bytes >
 ```
 
+#### Tap Aggregation
+
+```yaml
+tap_aggregation:
+  # The chassis switches DCS-7500R and DCS-7500R2 also support mixed-mode on lincard basis, support for this can be added later
```
ClausHolbechArista: We don't need this in the documentation.
```suggestion
```

ClausHolbechArista (commented):
Missing documentation template. You can decide to keep it simple, and just add the config.

ccsnw:
> Missing documentation template. You can decide to keep it simple, and just add the config.

will add doc and tests for this one and the range_prefix one, just wanted to verify that basic DM is Ok before adding it

ClausHolbechArista:
> > Missing documentation template. You can decide to keep it simple, and just add the config.
> 
> will add doc and tests for this one and the range_prefix one, just wanted to verify that basic DM is Ok before adding it

All good. We could add more nesting in the DM here, but the value is limited, so I agree with your suggestion.

ccsnw:
> > > Missing documentation template. You can decide to keep it simple, and just add the config.
> > 
> > 
> > will add doc and tests for this one and the range_prefix one, just wanted to verify that basic DM is Ok before adding it
> 
> All good. We could add more nesting in the DM here, but the value is limited, so I agree with your suggestion.

Argh, sorry. Think I've missed one point (for future use). In mixed mode, it should be possible to configure multiple linecards of a chassis in tap aggregation mode. So my suggested DM will fail.

Something like this might make more sense:

```
tap_aggregation:
  mode:
    exclusive: < true | false >
    mixed:
     - linecard1
     - linecard2
  encapsulation_dot1br_strip: < true | false >
  encapsulation_vn_tag_strip: < true | false >
  protocol_lldp_trap: < true | false >
  truncation_size: < size in bytes >
  mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
  mac_fcs_append: < true | false >
  mac_fcs_error: < correct | discar | pass-through >
```

The list entries under "mixed" could of course have multiple key/values, this is just an example.

What doy ou think?

ClausHolbechArista:
> > > > Missing documentation template. You can decide to keep it simple, and just add the config.
> > > 
> > > 
> > > will add doc and tests for this one and the range_prefix one, just wanted to verify that basic DM is Ok before adding it
> > 
> > 
> > All good. We could add more nesting in the DM here, but the value is limited, so I agree with your suggestion.
> 
> Argh, sorry. Think I've missed one point (for future use). In mixed mode, it should be possible to configure multiple linecards of a chassis in tap aggregation mode. So my suggested DM will fail.
> 
> Something like this might make more sense:
> 
> ```
> tap_aggregation:
>   mode:
>     exclusive: < true | false >
>     mixed:
>      - linecard1
>      - linecard2
>   encapsulation_dot1br_strip: < true | false >
>   encapsulation_vn_tag_strip: < true | false >
>   protocol_lldp_trap: < true | false >
>   truncation_size: < size in bytes >
>   mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
>   mac_fcs_append: < true | false >
>   mac_fcs_error: < correct | discar | pass-through >
> ```
> 
> The list entries under "mixed" could of course have multiple key/values, this is just an example.
> 
> What doy ou think?

https://www.arista.com/en/um-eos/eos-test-access-point-tap-aggregation#xx1132305

This shows other options for mode exclusive, so having a `mode` dict seems relevant. I know the no_errdisable is often used. It should probably be a list of interfaces.
I don't know the CLI for mixed mode, so can't comment on that.

ccsnw:
> > > > > Missing documentation template. You can decide to keep it simple, and just add the config.
> > > > 
> > > > 
> > > > will add doc and tests for this one and the range_prefix one, just wanted to verify that basic DM is Ok before adding it
> > > 
> > > 
> > > All good. We could add more nesting in the DM here, but the value is limited, so I agree with your suggestion.
> > 
> > 
> > Argh, sorry. Think I've missed one point (for future use). In mixed mode, it should be possible to configure multiple linecards of a chassis in tap aggregation mode. So my suggested DM will fail.
> > Something like this might make more sense:
> > ```
> > tap_aggregation:
> >   mode:
> >     exclusive: < true | false >
> >     mixed:
> >      - linecard1
> >      - linecard2
> >   encapsulation_dot1br_strip: < true | false >
> >   encapsulation_vn_tag_strip: < true | false >
> >   protocol_lldp_trap: < true | false >
> >   truncation_size: < size in bytes >
> >   mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
> >   mac_fcs_append: < true | false >
> >   mac_fcs_error: < correct | discar | pass-through >
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The list entries under "mixed" could of course have multiple key/values, this is just an example.
> > What doy ou think?
> 
> https://www.arista.com/en/um-eos/eos-test-access-point-tap-aggregation#xx1132305
> 
> This shows other options for mode exclusive, so having a `mode` dict seems relevant. I know the no_errdisable is often used. It should probably be a list of interfaces. I don't know the CLI for mixed mode, so can't comment on that.

@ClausHolbechArista Think this should cover it:

```
tap_aggregation:
  mode:
    exclusive:
      enabled: < true | false >
      profile: < profile_name | default --> tap-aggregation-default >
      no-errdisable:
        - < EthernetX | Port-ChannelX >
    mixed:
     - module: < linecard_number>
       profile: < profile_name | default --> tap-aggregation-default >
  encapsulation_dot1br_strip: < true | false >
  encapsulation_vn_tag_strip: < true | false >
  protocol_lldp_trap: < true | false >
  truncation_size: < size in bytes >
  mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
  mac_fcs_append: < true | false >
  mac_fcs_error: < correct | discar | pass-through >
```

Assumptions:

- In "exclusive mode" the profile applied is global
- no-errdisable will create one CLI line per interface (tested this) in addition to the global "exclusive mode"
- In mixed-mode my guess is that it's possible to apply different profiles on per module basis (can't test this at this point)
- Also can't test how the no-errdisable in mixed mode would look like

What I'd implement for now:

```
tap_aggregation:
  mode:
    exclusive:
      enabled: < true | false >
      profile: < profile_name | default --> tap-aggregation-default >
      no-errdisable:
        - < EthernetX | Port-ChannelX >
  encapsulation_dot1br_strip: < true | false >
  encapsulation_vn_tag_strip: < true | false >
  protocol_lldp_trap: < true | false >
  truncation_size: < size in bytes >
  mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
  mac_fcs_append: < true | false >
  mac_fcs_error: < correct | discar | pass-through >
```

This would allow to add mixed-mode at a later stage without breaking the model.

Do you agree?

ClausHolbechArista:
> 
> * In "exclusive mode" the profile applied is global

I'm not sure what this means.

> * no-errdisable will create one CLI line per interface (tested this) in addition to the global "exclusive mode"
> * In mixed-mode my guess is that it's possible to apply different profiles on per module basis (can't test this at this point)
> * Also can't test how the no-errdisable in mixed mode would look like

We will address mixed mode another day :) Let's ignore that for now.

> What I'd implement for now:
> 
> ```
> tap_aggregation:
>   mode:
>     exclusive:
>       enabled: < true | false >
>       profile: < profile_name | default --> tap-aggregation-default >
>       no-errdisable:
>         - < EthernetX | Port-ChannelX >
>   encapsulation_dot1br_strip: < true | false >
>   encapsulation_vn_tag_strip: < true | false >
>   protocol_lldp_trap: < true | false >
>   truncation_size: < size in bytes >
>   mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
>   mac_fcs_append: < true | false >
>   mac_fcs_error: < correct | discar | pass-through >
> ```
> 
> This would allow to add mixed-mode at a later stage without breaking the model.
> 
> Do you agree?

 Looks good! Thank you.

ccsnw:
> I'm not sure what this means.

I think in mixed mode you can have multiple line cards per chassis in TAP agg. mode and my guess is that each card can operate in a different profile.

 whereas in exclusive mode (non modular switch)  the whole switch needs to run the same profile.

But like you said, lets leave mixed-mode aside for now

ccsnw:
> Missing documentation template. You can decide to keep it simple, and just add the config.

@ClausHolbechArista Documentation template added and code adjusted to updated DM

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
```diff
@@ -0,0 +1,39 @@
...
+{%     if tap_aggregation.mode.exclusive is arista.avd.defined %}
+{%         if tap_aggregation.mode.exclusive.enabled is arista.avd.defined(true) %}
+{%             set mode_cli = "mode exclusive" %}
+{%         endif %}
+{%         if tap_aggregation.mode.exclusive.profile is arista.avd.defined %}
+{%             set mode_cli = mode_cli ~ " profile " ~ tap_aggregation.mode.exclusive.profile %}
+{%         endif %}
+   {{ mode_cli }}
+{%         if tap_aggregation.mode.exclusive.no_errdisable is arista.avd.defined %}
+{%             for interface in tap_aggregation.mode.exclusive.no_errdisable %}
```
tgodaA: Please use natural_sort here then we can remove the defined test also.

image

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2470,6 +2471,27 @@ monitor_sessions:
...
+        - < EthernetX | Port-ChannelX >
+  encapsulation_dot1br_strip: < true | false >
+  encapsulation_vn_tag_strip: < true | false >
+  protocol_lldp_trap: < true | false >
+  # Allowed truncation_size values vary depending on the platform
+  truncation_size: < size in bytes >
+  mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
+  # mac_fcs_append and mac_fcs_error are mutually exclusive. If both are defined, mac_fcs_append takes precedence
+  mac_fcs_append: < true | false >
+  mac_fcs_error: < correct | discar | pass-through >
```
tgodaA: typo **discard**

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2470,6 +2471,27 @@ monitor_sessions:
...
+      enabled: < true | false >
+      profile: < profile_name | default --> tap-aggregation-default >
+      no-errdisable:
+        - < EthernetX | Port-ChannelX >
+  encapsulation_dot1br_strip: < true | false >
+  encapsulation_vn_tag_strip: < true | false >
+  protocol_lldp_trap: < true | false >
+  # Allowed truncation_size values vary depending on the platform
+  truncation_size: < size in bytes >
+  mac_timestamp: < replace source-mac | header format 48-bit | header format 64-bit >
```
tgodaA: I think we can still add knobs under this:

```
mac:
  timestamp:
    replace_source_mac: < true | false >
    header:
      format: < "48-bit" | "64-bit" >
      eth_type: < integer > # I found this one to be in range 0-65535
  fcs_append: < true | false >
  fcs_error: < "correct" | "discard" | "pass-through" >
```

tgodaA: image

tgodaA: We can add a warning for mac timestamp. Although we can add three different timestamps only one will take precedence.

image

tgodaA: I understood that either ```mac timestamp replace <>``` or ```mac timestamp header <>``` should be configured. When we give both only **replace** is taking precedence although we can configure **header** timestamp.

ccsnw:
```
mac:
  timestamp:
    replace_source_mac: < true | false >
    header:
      format: < "48-bit" | "64-bit" >
      eth_type: < integer > # I found this one to be in range 0-65535
  fcs_append: < true | false >
  fcs_error: < "correct" | "discard" | "pass-through" >
```

>> Thanks! On 4.26.3M on a DCS-7280CR2A-30-F, I dont have eth-type available, so wasn't even aware of it. Do you have an info about it?

ccsnw: > I understood that either `mac timestamp replace <>` or `mac timestamp header <>` should be configured. When we give both only **replace** is taking precedence although we can configure **header** timestamp.

This is correct. You can configure replace_source_mac and one header fomat (48 | 64bit) together, in this case replace source mac will take precedence.  So I think it makes sense to adjust the DM like you suggested.

To shed some more light on this from the technical point:

a) When we configure: "mac timestamp replace source-mac" the 48bit src field of the frame will be replaced with a 48bit long timestamp, the rest of the frame remains as it is

b) When we use "mac timestamp header format 64-bit" or "mac timestamp header format 64-bit", the switch will insert additional fields (that contain a timestamp, either 48 or 64 bits long) to the frame, making it bigger.

==

It wouldn't make sense to combine a) and b) at the same time, because this would mean: Replace Source MAC field with a timestamp AND (in addition) add more fields to the frame that also include another timestamp.. we would end up with a frame that doesn't have source MAC info anymore, but has 2 timestamps (redundant).

@ClausHolbechArista fell free to correct me if the explanation above is incorrect

ccsnw: @tgodaA changes added. Do you have any documentation about eth_type? I cant find any. Would like to read it out of interest.
Thanks!

tgodaA: Hi @ccsnw, please use the below references for mac timestamps:

1) https://www.arista.com/assets/data/pdf/Whitepapers/Overview_Arista_Timestamps.pdf
2) https://aristanetworks.force.com/AristaCommunity/s/article/analyzing-packet-header-timestamps-in-wireshark
3) https://aristanetworks.force.com/AristaCommunity/s/article/timestamping-on-the-7150-series
4) https://damianzaremba.co.uk/2017/12/decoding-arista-ethertype-headers-with-gopacket/

In ref #**1** page 3 should have your required information.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
```diff
@@ -0,0 +1,42 @@
...
+   truncation size {{ tap_aggregation.truncation_size }}
+{%     endif %}
+{%     if tap_aggregation.mac.timestamp.replace_source_mac is arista.avd.defined(true) %}
+   mac timestamp replace source-mac
+{%     elif tap_aggregation.mac.timestamp.header.format is arista.avd.defined %}
+   mac timestamp header format {{ tap_aggregation.mac.timestamp.header.format }}
+{%     endif %}
+{%     if tap_aggregation.mac.timestamp.header.eth_type is arista.avd.defined %}
+   mac timestamp header eth-type {{ tap_aggregation.mac.timestamp.header.eth_type }}
+{%     endif %}
```
tgodaA: All of this should be in one block, under ```.header```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
ccsnw: thanks!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
ccsnw: thanks @tgodaA i've created the section

ccsnw:
@ClausHolbechArista can you rerun the check by chance? stuck on a dl from galaxy

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/tap-aggregation.j2
```diff
@@ -0,0 +1,45 @@
...
+| Truncation Size | {{ tap_aggregation.truncation_size }} |
+{%     endif %}
+{%     if tap_aggregation.mac.timestamp.replace_source_mac is arista.avd.defined(true) %}
+| Mac Timestamp | Replace Source-Mac |
+{%     elif tap_aggregation.mac.timestamp.header.format is arista.avd.defined %}
+| Mac Timestamp | Header Format {{ tap_aggregation.mac.timestamp.header.format }} |
+{%     endif %}
+{%     if tap_aggregation.mac.timestamp.header.eth_type is arista.avd.defined %}
+| Mac Timestamp | Header eth-type {{ tap_aggregation.mac.timestamp.header.eth_type }} |
+{%     endif %}
```
tgodaA: Please sync this as well according to the eos template.

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -2508,6 +2509,33 @@ monitor_sessions:
...
 ```
 
+#### Tap Aggregation
+
+```yaml
+tap_aggregation:
+  mode:
+    exclusive:
+      enabled: < true | false >
+      profile: < profile_name | default --> tap-aggregation-default >
```
tgodaA: We don't support default values in eos_cli_config_gen. As this default value can change anytime in future. So please remove this here as well as in documentation template.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/tap-aggregation.j2
```diff
@@ -0,0 +1,47 @@
+{% if tap_aggregation is arista.avd.defined and tap_aggregation.mode.exclusive.enabled is arista.avd.defined(true) %}
+
+## Tap Aggregation
+
+### Tap Aggregation Summary
+
+| Settings | Values |
+| -------- | ------ |
+| Mode Exclusive | {{ tap_aggregation.mode.exclusive.enabled }} |
+| Mode Exclusive Profile | {{ tap_aggregation.mode.exclusive.profile | arista.avd.default('tap-aggregation-default') }} |
```
tgodaA: Add this entry only if defined. I am not sure if this is going to be default on all platforms. Even this could change in future.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
```diff
@@ -0,0 +1,44 @@
...
+{%             set mode_cli = "mode exclusive" %}
+{%         endif %}
+{%         if tap_aggregation.mode.exclusive.profile is arista.avd.defined %}
+{%             set mode_cli = mode_cli ~ " profile " ~ tap_aggregation.mode.exclusive.profile %}
+{%         endif %}
+   {{ mode_cli }}
+{%         for interface in tap_aggregation.mode.exclusive.no_errdisable | arista.avd.natural_sort %}
+   mode exclusive no-errdisable {{ interface }}
+{%         endfor %}
+{%     endif %}
```
tgodaA: I think we need to agree on one thing here. Like print the config only if its **enabled**. So the line5 can be changed to ```tap_aggregation.mode.exclusive.enabled is arista.avd.defined(true)``` and then refactor the code accordingly.

Let me know your thought!

ccsnw: @tgodaA I did it like this on purpose, because EOS allows sth like this for example:

```
sh run sec tap agg
tap aggregation
   mode exclusive no-errdisable Ethernet1/1
   mac timestamp header format 64-bit
   mac fcs append
   mac fcs-error correct
```

so you can basically configure no errdisable on ifs, while the mode is not set..

but you cannot set a profile while the mode is false, these commands seem bundled.

====

keeping it like it is would come in handy for people that switch between tap and non tap mode and want to keep global tap config (so basically just switch it off and on). dunno if that happens in real life though

what do you think?

tgodaA: Then we need to refactor some code. 

For example:

```
tap_aggregation:
  mode:
    exclusive:
      no_errdisable:
        - Ethernet1
```

The above input will throw error! Because ```mode_cli``` will not be defined (as both line6 and line9) are failed.

I added a suggestion for this.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
```diff
@@ -0,0 +1,44 @@
...
+!
+tap aggregation
+{%     if tap_aggregation.mode.exclusive is arista.avd.defined %}
+{%         if tap_aggregation.mode.exclusive.enabled is arista.avd.defined(true) %}
+{%             set mode_cli = "mode exclusive" %}
+{%         endif %}
+{%         if tap_aggregation.mode.exclusive.profile is arista.avd.defined %}
+{%             set mode_cli = mode_cli ~ " profile " ~ tap_aggregation.mode.exclusive.profile %}
+{%         endif %}
+   {{ mode_cli }}
```
tgodaA:
```suggestion
{%         if tap_aggregation.mode.exclusive.enabled is arista.avd.defined(true) %}
{%             set mode_cli = "mode exclusive" %}
{%             if tap_aggregation.mode.exclusive.profile is arista.avd.defined %}
{%                 set mode_cli = mode_cli ~ " profile " ~ tap_aggregation.mode.exclusive.profile %}
{%             endif %}
   {{ mode_cli }}
{%         endif %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tap-aggregation.j2
ccsnw: I'm starting to feel that I've written this while half sleeping and half riding a bike, while speech typing with siri. 

let me fix it and thanks for reviewing!

tgodaA (approved):
LGTM

ClausHolbechArista (approved):
LGTM

PR #1740, opened 2022-04-22
Feat(plugins): Updated convert_dicts filter for list/string values in dictionary

ghost:
## Change Summary

Updated convert_dicts filter for list/string values in dictionary

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.convert_dicts`

## Proposed changes

- Checks if the nested dictionary value is list/string, a new key (secondary_key) will be inserted in the inner dictionary if secondary_key is provided.
- Checks if the input is not a dictionary but a list, primary key will be inserted for each element of the list.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/README.md
+++ b/ansible_collections/arista/avd/plugins/README.md
@@ -75,7 +75,7 @@ To use this filter:
 
 The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
 dictionaries. It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
-configurable) and if there is a list in the dictionary value,it inserts this value list to
+configurable) and if there is a non-dictionary value,it inserts this value to
 secondary key (key name is configurable), if secondary key is provided.
 
 This filter is intended for:
@@ -83,7 +83,7 @@ This filter is intended for:
 - Seemless data model migration from dictionaries to lists.
 - Improve Ansible's processing performance when dealing with large dictionaries by converting them to lists of dictionaries.
 
-Note: If the value is a list/string with no secondary key provided, it will pass through untouched
+Note: If there is a non-dictionary value with no secondary key provided, it will pass through untouched
 
 To use this filter:
 
--- a/ansible_collections/arista/avd/plugins/README.md
+++ b/ansible_collections/arista/avd/plugins/README.md
@@ -74,7 +74,7 @@ To use this filter:
 ### convert_dicts filter
 
 The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
-dictionaries.It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
+dictionaries. It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
 configurable) and if there is a list in the dictionary value,it inserts this value list to
 secondary key (key name is configurable), if secondary key is provided.
 
--- a/ansible_collections/arista/avd/plugins/README.md
+++ b/ansible_collections/arista/avd/plugins/README.md
@@ -73,24 +73,25 @@ To use this filter:
 
 ### convert_dicts filter
 
-The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of dictionaries
-It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is configurable) and if there is a list of dictionary in the dictionary value, it inserts this value list to secondary key "items" (key name is configurable).
+The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
+dictionaries.It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
+configurable) and if there is a list in the dictionary value,it inserts this value list to
+secondary key (key name is configurable), if secondary key is provided.
 
 This filter is intended for:
 
 - Seemless data model migration from dictionaries to lists.
 - Improve Ansible's processing performance when dealing with large dictionaries by converting them to lists of dictionaries.
 
-Note: If the variable is  a list of string/integer, it will pass through untouched
+Note: If the value is a list/string with no secondary key provided, it will pass through untouched
 
 To use this filter:
 
 ```jinja
-{# convert list of dictionary with default `name:` as the primary key and `items:` as secondary key #}
+{# convert list of dictionary with default `name:` as the primary key and None secondary key #}
 {% set example_list = example_dictionary | arista.avd.convert_dicts %}
 {% for example_item in example_list %}
 item primary key is {{ example_item.name }}
-item secondary key is {{ example_item.items }}
 {% endfor %}
 
 {# convert list of dictionary with `id:` set as the primary key and `types:` set as the secondary key #}
--- a/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
+++ b/ansible_collections/arista/avd/plugins/filter/convert_dicts.py
@@ -8,32 +8,31 @@ __metaclass__ = type
 import os
 
 
-def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
+def convert_dicts(dictionary, primary_key="name", secondary_key=None):
     """
     The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
-    dictionaries It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
-    configurable) and if there is a list of dictionary in the dictionary value,it inserts this value list to
-    secondary key "items" (key name is configurable).
+    dictionaries.It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
+    configurable) and if there is a list in the dictionary value,it inserts this value list to
+    secondary key (key name is configurable), if secondary key is provided.
 
     This filter is intended for:
 
     - Seemless data model migration from dictionaries to lists.
     - Improve Ansible's processing performance when dealing with large dictionaries by converting them to lists of dictionaries.
 
-    Note: If the variable is a list of string/integer, it will pass through untouched
+    Note: If the value is a list/string with no secondary key provided, it will pass through untouched
 
     To use this filter:
 
     ```jinja
-    {# convert list of dictionary with default `name:` as the primary key and `items:` as secondary key #}
+    {# convert list of dictionary with default `name:` as the primary key and None secondary key #}
     {% set example_list = example_dictionary | arista.avd.convert_dicts %}
     {% for example_item in example_list %}
     item primary key is {{ example_item.name }}
-    item secondary key is {{ example_item.items }}
     {% endfor %}
 
     {# convert list of dictionary with `id:` set as the primary key and `types:` set as the secondary key #}
-    {% set example_list = example_dictionary | arista.avd.convert_dicts('id') %}
+    {% set example_list = example_dictionary | arista.avd.convert_dicts('id','types') %}
     {% for example_item in example_list %}
     item primary key is {{ example_item.id }}
     item secondary key is {{ example_item.types }}
@@ -43,20 +42,27 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
     Parameters
     ----------
     dictionary : any
-        Nested Dictionary to convert - returned untouched if not a nested dictionary
+        Nested Dictionary to convert - returned untouched if not a nested dictionary and list
     primary_key : str, optional
         Name of primary key used when inserting outer dictionary keys into items.
     secondary_key : str, optional
-        Name of secondary key used when inserting dictionary values which are list of dictionaries into items.
+        Name of secondary key used when inserting dictionary values which are list into items.
 
     Returns
     -------
     any
-        Returns list of dictionaries or input variable untouched if not a nested dictionary/list of dictionaries
+        Returns list of dictionaries or input variable untouched if not a nested dictionary/list.
     """
-    if not isinstance(dictionary, dict) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
+    if not isinstance(dictionary, dict) and not isinstance(dictionary, list) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
         # Not a dictionary, return the original
         return dictionary
+    elif isinstance(dictionary, list):
+        output = []
+        for element in dictionary:
+            item = {}
+            item.update({primary_key: element})
+            output.append(item)
+        return output
     else:
         output = []
         for key in dictionary:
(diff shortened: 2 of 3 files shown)


ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,41 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
     -------
     any
-        Returns list of dictionaries or input variable untouched if not a nested dictionary/list of dictionaries
+        Returns list of dictionaries or input variable untouched if not a nested dictionary/list.
     """
-    if not isinstance(dictionary, dict) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
+    if not isinstance(dictionary, dict) and not isinstance(dictionary, list) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
         # Not a dictionary, return the original
         return dictionary
+    elif isinstance(dictionary, list):
```
ClausHolbechArista, line 59: we should not touch the input, if the list items are dicts. Only if they are something else.

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -73,24 +73,25 @@ To use this filter:
 
 ### convert_dicts filter
 
-The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of dictionaries
-It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is configurable) and if there is a list of dictionary in the dictionary value, it inserts this value list to secondary key "items" (key name is configurable).
+The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
+dictionaries.It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
```
gmuloc: missing a space between `dictionnaries.` and `It`

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -8,32 +8,31 @@
...
 
 
-def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
+def convert_dicts(dictionary, primary_key="name", secondary_key=None):
     """
     The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
-    dictionaries It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
-    configurable) and if there is a list of dictionary in the dictionary value,it inserts this value list to
-    secondary key "items" (key name is configurable).
+    dictionaries.It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
```
gmuloc: same comment as above - need to add back a space before `It`

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
Shivani-chourasiya: updated!

ansible_collections/arista/avd/plugins/README.md
Shivani-chourasiya: fixed.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
Shivani-chourasiya: fixed

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,42 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
+        Returns list of dictionaries or input variable untouched if not a nested dictionary/list.
     """
-    if not isinstance(dictionary, dict) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
+    if not isinstance(dictionary, dict) and not isinstance(dictionary, list) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
         # Not a dictionary, return the original
         return dictionary
+    elif isinstance(dictionary, list):
+        output = []
+        for element in dictionary:
+            if not isinstance(element, dict):
```
ClausHolbechArista, line 62: with this change you will remove items that are dicts (you are not adding them to output :)

Shivani-chourasiya: yeah, updated now.

Shivani-chourasiya: yeah, updated.

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/tests/unit/filters/test_convert_dicts.py
```diff
@@ -17,59 +19,95 @@ class TestConvertDicts():
...
-        assert resp == list
+        assert resp == [{'name': 'Test1'}, {'name': 'Test2'}, {'name': 'Test3'}]
 
     def test_convert_dicts_with_list_primary_and_secondary_key(self):
-        resp = convert_dicts(list, 'id', 'types')
-        assert resp == list
+        resp = convert_dicts(list, 'test', 'types')
+        assert resp == [{'test': 'Test1'}, {'test': 'Test2'}, {'test': 'Test3'}]
+
+    def test_convert_dicts_with_string_default(self):
```
gmuloc: This test names confused me at first as it looks like we are giving a string as input. Maybe we can make it clearer in the variable name / test names that we are testing a simple dict with only a string value

ansible_collections/arista/avd/tests/unit/filters/test_convert_dicts.py
```diff
@@ -17,59 +19,95 @@ class TestConvertDicts():
...
+    def test_convert_dicts_with_string_primary_key_and_secondary_key(self):
+        resp = convert_dicts(string, 'test', 'str')
+        assert resp == [{'test': 'dict', 'str': 'test_string'}]
+
+    def test_convert_dicts_with_list_of_dict_default(self):
+        resp = convert_dicts(list_of_dict)
+        assert resp == list_of_dict
+
+    def test_convert_dicts_with_list_of_dict_primary_key(self):
+        resp = convert_dicts(list_of_dict)
```
gmuloc: it should have a primary key as per test name

ansible_collections/arista/avd/tests/unit/filters/test_convert_dicts.py
```diff
@@ -17,59 +19,95 @@ class TestConvertDicts():
...
+        assert resp == list_of_dict
+
+    def test_convert_dicts_with_list_of_dict_secondary_key(self):
+        resp = convert_dicts(list_of_dict)
```
gmuloc: it should have a secondary key as per test name

ansible_collections/arista/avd/tests/unit/filters/test_convert_dicts.py
```diff
@@ -17,59 +19,95 @@ class TestConvertDicts():
...
+        assert resp == list_of_dict
+
+    def test_convert_dicts_with_list_of_dict_primary_key_and_secondary_key(self):
+        resp = convert_dicts(list_of_dict)
```
gmuloc: it should have both primary and secondary key as per test name

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,44 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
 
     Returns
     -------
     any
-        Returns list of dictionaries or input variable untouched if not a nested dictionary/list of dictionaries
+        Returns list of dictionaries or input variable untouched if not a nested dictionary/list.
     """
-    if not isinstance(dictionary, dict) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
+    if not isinstance(dictionary, dict) and not isinstance(dictionary, list) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
         # Not a dictionary, return the original
```
gmuloc: the comment is confusing compared to the if statement above - it seems we are asserting it is neither a dict NOR a list.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,44 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
+        Name of secondary key used when inserting dictionary values which are list into items.
 
     Returns
     -------
     any
-        Returns list of dictionaries or input variable untouched if not a nested dictionary/list of dictionaries
+        Returns list of dictionaries or input variable untouched if not a nested dictionary/list.
     """
-    if not isinstance(dictionary, dict) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
+    if not isinstance(dictionary, dict) and not isinstance(dictionary, list) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
```
gmuloc: To make it clear what we are doing I would refactor like this (it avoids to take a break thinking about the and/or operators precedence in python):
```
if not isinstance(dictionary, (dict, list)) or os.environ.get('AVD_DISABLE_CONVERT_DICTS'):
```

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,44 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
         # Not a dictionary, return the original
         return dictionary
+    elif isinstance(dictionary, list):
+        output = []
+        for element in dictionary:
```
gmuloc, line 61: in this bit - do we assume the list elements are all of same type?

it looks like this list:
```
[ "test", { "blah": 42} ]
```

would be converted to
```
[ {"name": "test"}, {"blah": 42}]
```

is this the expectation?

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -8,32 +8,31 @@
...
 
-def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
+def convert_dicts(dictionary, primary_key="name", secondary_key=None):
     """
     The `arista.avd.convert_dicts` filter will convert a dictionary containing nested dictionaries to a list of
-    dictionaries It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
-    configurable) and if there is a list of dictionary in the dictionary value,it inserts this value list to
-    secondary key "items" (key name is configurable).
+    dictionaries. It inserts the outer dictionary keys into each list item using the primary_key `name` (key name is
+    configurable) and if there is a list in the dictionary value,it inserts this value list to
```
gmuloc: technically the code does not check if the dictionary value is a list anymore - if you look at the tests, even if the value is a string and the user provides a secondary key, the string value will be inserted in the resulting dictionnary with the secondary key.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,44 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
     else:
         output = []
         for key in dictionary:
             if dictionary[key] is None:
                 # Catch cornercase where dictionary has no value because of old data models
                 output.append({primary_key: key})
             elif not isinstance(dictionary[key], dict):
-                if isinstance(dictionary[key], list):
-                    # Not a nested dictionary but a list of dictionaries, add secondary key for the list
+                # Not a nested dictionary, add secondary key for the list elements if secondary key is provided
```
gmuloc: same as above in the description - the current code does not care if the dictionary value is a list of elements or anything else

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
```diff
@@ -43,38 +42,44 @@ def convert_dicts(dictionary, primary_key="name", secondary_key="items"):
...
-                    secondary_element = []
                     item.update({primary_key: key})
-                    for element in dictionary[key]:
-                        secondary_element.append(element)
-                    item.update({secondary_key: secondary_element})
+                    item.update({secondary_key: dictionary[key]})
                     output.append(item)
                 else:
-                    # Not a nested dictionary but a string, return the original
                     return dictionary
```
gmuloc, line 83: I am not sure about this part - it seems to assume that all the elements are the same inside the dictionary (which may be true)

gmuloc (changes requested):

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
Shivani-chourasiya: @ClausHolbechArista , is this how it should be? (need a confirmation).
currently it works like this.

ansible_collections/arista/avd/tests/unit/filters/test_convert_dicts.py
Shivani-chourasiya: fixed

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
ClausHolbechArista: @gmuloc correct, that is what we have today. We do not enforce the primary key to be in the dict, but we convert it to _be_ a dict.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
Shivani-chourasiya: yes, this is how it works as of now.

ClausHolbechArista: Since we cannot fix the entry without having the secondary_key defined (case above) we just give up. It would not help us to do a partial convertion. Shivani had an error message here earlier, but I was not sure if we could give a meaningful error message, since we don't know the name of the outer var.
In general this case "should not happen" so either we can raise an exception, or we can ignore silently.

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
gmuloc: @ClausHolbechArista but so the expectation from the filter is that when we pass dicts or list they are all "containing" the same type of elements? If so should probably be added because in my example a list of `[<string>, <dict>]` becomes `[<dict>, <dict>]` but the second dict does not have `name` as a key (which probably the person running the filter could expect)

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
gmuloc: ok - understood I will resolve this

tgodaA (approved):
LGTM

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
gmuloc: looks like this is blocking the two approvals so will resolve here as it is the expected behavior for now.

gmuloc (approved):
LGTM - thanks for the changes

ansible_collections/arista/avd/plugins/filter/convert_dicts.py
ClausHolbechArista: We can try to improve the docs later.

PR #1745, opened 2022-04-25
Feat(eos_designs): Shutdown underlay links if the peer device is not deployed

### LINKED ISSUE #1744: Shutdown ports facing an undeployed device.
### Enhancement summary

The "is_deployed = false" setting excludes a device from being configured in CloudVision. This is intended for when a device is not present in the network. However, as expected, if the device is not present, the ports facing this device on a peer will end up being configured but will be down. Not only is this a security concern but it causes the validate role to raise an error as the port is expected to be up.

This enhancement causes the underlay peer port to be shutdown.

eos_designs, eos_config_deploy_cvp

### Use case example

Consider the simple topology below:

DC1_SPINE1 <Eth1---Eth1> DC1_BL2

Inventory file:

```
  DC1_BL2:
    hosts:
      DC1-BL2A:
        # Device configuration is generated by AVD
        # Device is not configured on Cloudvision (configlet is uploaded)
        is_deployed: false
```

Expected state on DC1_SPINE1:

```
interface Ethernet1
  shutdown
```

But config for DC1_BL2 is still generated.

As per the example. The peer underlay port should be shutdown.

Perhaps a global setting should be used to determine whether is_deployed = false causes the underlay port to be shutdown but this will add more toggles.

### Additional context

_No response_

### Contributing Guide

perimore:
## Change Summary

The host variable `is_deployed: false` setting excludes a device from being configured in CloudVision. This is intended for when a device is not present in the network. However, as expected, if the device is not present, the ports facing this device on a peer will end up being configured but will be down. Not only is this a security concern but it causes the validate role to raise an error as the port is expected to be up when using `eos_config_deploy_cvp`.

This enhancement causes the underlay peer port to be shutdown in `eos_designs` if a new fabric level variable `is_deployed_shutdown_peer_interfaces: true`. This seperate variable will avoid any breaking changes in `eos_designs`.

## Related Issue(s)

Fixes #1744 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Consider the simple topology below:

DC1_SPINE1 DC1_BL2

Inventory file:

```
  DC1_BL2:
    hosts:
      DC1-BL2A:
        # Device configuration is generated by AVD
        # Device is not configured on Cloudvision (configlet is uploaded)
        is_deployed: false
```

fabric variable:

```
`is_deployed_shutdown_peer_interfaces: true`
```

Expected state on DC1_SPINE1:

```
interface Ethernet1
  shutdown
```

## How to test
Set device variable to `is_deployed: false` and fabric variable `is_deployed_shutdown_peer_interfaces: true` and check if the peer port is configured to be shutdown.

Check the debug vars that `is_deployed = false` under the facts:

```
ansible_facts:
  switch:
...
    is_deployed: false
```

And that the peer config is shutting down the port:

```
ethernet_interfaces:
  Ethernet1:
    peer: DC1-LEAF1AA
    peer_interface: Ethernet1
    peer_type: l3leaf
    description: P2P_LINK_TO_DC1-LEAF1AA_Ethernet1
    mtu: 1500
    type: routed
    shutdown: true
    ip_address: 172.31.255.0/31
```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_config_deploy_cvp/README.md
```diff
@@ -127,7 +127,7 @@ tasks:
 
 ### Ignore devices not provisioned in Cloudvision
 
-When you want to provision a complete topoplogy and devices are not already in Cloudvision, you can configure inventory to ignore these devices by using a host variable: `is_deployed`
+When you want to provision a complete topology and devices are not already in Cloudvision, you can configure inventory to ignore these devices by using a host variable: `is_deployed`. This will also cause the ports facing the undeployed device to be shutdown in order to satisfy the validate role when checking if a port should be enabled but is down due to the other device not being present.
```
ClausHolbechArista: This documentation is only for the `eos_config_deploy_cvp` role, so we should not document the `eos_designs` behavior here. Please add documentation for this knob under `eos_designs` `Fabric Variables` as well as a new knob `is_deployed_shutdown: < true | false | default -> false >` and implement check for this new knob before you enable this new behavior. Make sure it is clear that the `is_deployed` should be a host_var and `is_deployed_shutdown` can be a group_var.

It is a nice idea, but I would consider this a breaking change for existing deployments, hence the feature knob. If you have better ideas for naming you can implement that.

perimore: Agreed, I'll make the changes and re push. Thanks.

perimore: Created `is_deployed_shutdown_peer_interfaces` a fabric level variable as I don't think it makes sense for this to be group level as it would be easy to forget to enable each time a new group is added.

Documented this toggle under `eos_designs` and updated `eos_config_deploy_cvp` docs to make it clear the host level `is_deployed: false` toggle will only affect actual device configs when fabric level `is_deployed_shutdown_peer_interfaces: true`. Host level `is_deployed: false` will still cause the device to be ignored in CVP.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -4,6 +4,9 @@
 
 fabric_name: DC1_FABRIC
 
+# If a device "is_deployed: false" at host_vars level, then shutdown facing peer ports.
```
gmuloc: the comment is a bit confusing.

```
# When a peer device is configured with `is_deployed: false` at host_vars level, setting this to true
# will shut down the interfaces facing this peer 
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -4,6 +4,9 @@
...
+is_deployed_shutdown_peer_interfaces: true
```
gmuloc: I would probably also rename the variable to make it clearer what it does e.g.
```
shutdown_interfaces_towards_undeployed_peers
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -1,6 +1,8 @@
 # Fabric Variables
 
 - The fabric underlay and overlay topology variables, define the elements related to build the L3 Leaf and Spine fabric.
+- It is possible to provision configurations for a complete topology but flag devices as not deployed using the host variable `is_deployed`.
+  - Interfaces on devices that face an undeployed device can also be shutdown using the variable `is_deployed_shutdown_peer_interfaces`. This is also useful when trying to satisfy the `eos_validate_state` interface tests.
```
gmuloc: 1. extra space at the beginning
2. regarding the `eos_validate_state` - I would be more explicit stating that if this setting is not used AND some Device Under Test (DUT) has some peer devices are using `is_deployed: false`, `eos_validate_state` role will fail on the tests trying to verify that DUT interfaces are up.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -21,6 +23,9 @@
 **Variables and Options:**
 
 ```yaml
+# Shutdown peer interfaces of a device that is not deployed.
```
gmuloc: same comments as above for the variable name and the description

ansible_collections/arista/avd/roles/eos_designs/templates/base/main.j2
```diff
@@ -56,3 +56,5 @@
 {% include 'base/eos-cli.j2' %}
 
 {% include 'base/struct-cfg.j2' %}
+
+{% include 'base/is_deployed_shutdown_peer_interfaces.j2' %}
```
gmuloc: `struc-cfg.j2` should be the latest included template (as this is custom_structured_configuration that should be able to override anything)

However - does this setting needs to be included in the structured_configuration? what is more important is to have the shutdown interfaces right? (like it is done in the facts)

ansible_collections/arista/avd/roles/eos_designs/templates/facts/topology/p2p-uplinks.j2
```diff
@@ -12,6 +12,7 @@ topology:
 {%             do topology_peers.append(uplink_switch) %}
     {{ uplink_interface }}:
       peer: {{ uplink_switch }}
+      peer_is_deployed: {{ uplink_switch_facts.is_deployed }}
```
gmuloc: would it make sense to have this information included in the topology csv fail?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
perimore: Agreed, originally I wanted the variable name to include "is_deployed" but there is no pretty way of doing this.

ansible_collections/arista/avd/roles/eos_designs/templates/base/main.j2
perimore: I didn't think about the ordering here but yes, I do believe it is more important to have the interfaces shutdown so any structured configs that enable the port should be overridden by this variable.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
perimore: Tried to make this clearer with the below:
```
# If a device is set to "is_deployed: false" at the host_vars level, then shutdown interfaces towards undeployed peers.
shutdown_interfaces_towards_undeployed_peers: true
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: Agreed

ansible_collections/arista/avd/roles/eos_designs/templates/base/main.j2
gmuloc: so I don't necessarily agree - the principle of custom_struct is to be able to override values generated by `eos_designs` and hence it should be possible to override this one as well (and yes `eos_validate_state` would fail on some tests but at some point you need to know what you are doing when using custom_struct)

but my second point was actually maybe more relevant and making the first one moot - looking at the template `is_deployed_shutdown_peer_interfaces.j2` it is actually putting in the `intended/structured_configuration` a variable that is not going to be used afterwards - is it actually needed here? Basically you have already used in the facts the value to decide to shutdown or not the interfaces right? So even if it is overriden at this stage - it is "already too late" because the facts have already generated the configuration with the interfaces toward the undeployed peers shutdown.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: I've tried to explain this in it's own section

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: should be clearer now

ansible_collections/arista/avd/roles/eos_designs/templates/base/main.j2
perimore: Understood, it's a unnecessary variable at this stage :). I'll remove the j2 file and update main.j2.

ansible_collections/arista/avd/roles/eos_designs/templates/facts/topology/p2p-uplinks.j2
perimore: Yes, added this in as we present it in the MD docs, under the column "Provisioned in CloudVision"

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,24 @@
...
+- To overcome this and shutdown interfaces towards undeployed peers, the variable `shutdown_interfaces_towards_undeployed_peers` can be used. This will override any setting that enables a port across the fabric and will satisfy the `eos_validate_state` role interface tests.
+
+Note: The variables should be applied to all devices in the fabric.
+
+
+
 **Variables and Options:**
 
 ```yaml
+# Shutdown interfaces on devices that face undeployed peers.
```
ClausHolbechArista: please add the `is_deployed: < true | false | default -> true >` as well. It is only documented inside another role.

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
```diff
@@ -1,4 +1,4 @@
-Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface
+Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface,Provisioned in CloudVision
```
ClausHolbechArista: Instead of documenting the `is_deployed` state per link, when it is really a device parameter for one or the other device, I think we should just show the per-link `shutdown` with True/False as value

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
```diff
@@ -12,6 +12,14 @@ switch:
...
 {% endif %}
 
+{# switch.is_deployed #}
+{% if hostvars[inventory_hostname]['is_deployed'] is arista.avd.defined(false) %}
+{%     set is_deployed = false %}
+{% else %}
+{%     set is_deployed = true %}
+{% endif %}
+  is_deployed: {{ is_deployed }}
+
```
ClausHolbechArista:
```suggestion
{# switch.is_deployed #}
  is_deployed: {{ is_deployed | arista.avd.default(true)

```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,24 @@
...
 *Only supported with core_interfaces data model.
 
+## Flagging a Device as Not Deployed
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`. I.e. via the Ansible inventory file.
+- By default, this will have no impact, except cause the device to be ignored by CloudVision when used in conjunction with the `eos_config_deploy_cvp` role. If the device is not present in the network, `eos_validate_state` role will fail tests on peers of the undeployed device trying to verify that interfaces are up.
+- To overcome this and shutdown interfaces towards undeployed peers, the variable `shutdown_interfaces_towards_undeployed_peers` can be used. This will override any setting that enables a port across the fabric and will satisfy the `eos_validate_state` role interface tests.
+
+Note: The variables should be applied to all devices in the fabric.
+
+
```
ClausHolbechArista, line 33: Too many blank lines. Remove two

ClausHolbechArista (changes requested):
This will conflict with some other eos_designs changes we are working on, but @carlbuchmann and I can take care of the rebasing of this one if we merge that before this.
A few comments/requests.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: ```is_deployed``` is a host level variable and is not a fabric variable so I don't think it should be explained here. 

Although I have added an explanation in it's own section above.

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
perimore: agreed. Rather than having to work out whether true/false means shutdown = true or enabled = true, I'm going to make this explicit. The cell will contain "Enabled" or "Shutdown".

ansible_collections/arista/avd/roles/eos_designs/templates/facts/switch/switch.j2
perimore: Will update, can't commit directly as ```}}``` is missing

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,22 @@
   - IBGP (only with OSPF or ISIS variants in underlay)
 - Only summary network addresses need to be defined. IP addresses are then assigned to each node, based on its unique device id.
   - To view IP address allocation and consumption, a summary is provided in the auto-generated fabric documentation in Markdown and CSV format.
-- The variables should be applied to all devices in the fabric.
 
 *Only supported with core_interfaces data model.
 
+## Flagging a Device as Not Deployed
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`. I.e. via the Ansible inventory file.
```
gmuloc: is it `i.e.` or `e.g.` here? there are multiple options for the `is_deployed` variables. I would say either `E.g. via the ansible inventory file` or remove the last sentence.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,22 @@
...
+- By default, this will have no impact, except cause the device to be ignored by CloudVision when used in conjunction with the `eos_config_deploy_cvp` role. If the device is not present in the network, `eos_validate_state` role will fail tests on peers of the undeployed device trying to verify that interfaces are up.
```
gmuloc: so here it means that `is_deployed` would not have any effect if `eos_config_deploy_cli` is used and the config would still be deployed - is it correct? If so probably make it clear the behavior when using the `eos_config_deploy_cli` role.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: Agreed and I've attempted to make the documentation clearer

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: Correct and agreed. I've attempted to make the documentation clearer

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,24 @@
   - IBGP (only with OSPF or ISIS variants in underlay)
 - Only summary network addresses need to be defined. IP addresses are then assigned to each node, based on its unique device id.
   - To view IP address allocation and consumption, a summary is provided in the auto-generated fabric documentation in Markdown and CSV format.
-- The variables should be applied to all devices in the fabric.
 
 *Only supported with core_interfaces data model.
 
+## Flagging a Device as Not Deployed
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`. 
+- By default, this will have no impact within the `eos_designs` role. Configs will still be generated by the `eos_config_deploy_cli` role.
```
gmuloc: so I doublechecked technically this is `eos_cli_config_gen` which generates the config - whether it is then deployed through CVP or EAPI directly to the devices.
 
the `eos_config_deploy_eapi` (I think I got the name wrong in my previous review) is what I had in mind in the sense that `is_deployed` is ignored by this role it seems (when reading the doc) but then the validate tests would not fail provided either I think with this PR for the deployed hosts.

perimore: Makes sense, I'll clarify

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
ClausHolbechArista: Using to specific strings makes it easier for humans but harder for computers. This is a CSV, so more likely to be consumed by a script of some sort. Personally I have not used CSVs anywhere, but I know @carlbuchmann have, so  hopefully he can chip in.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
gmuloc: @perimore - let's still add the variable somewhere in `eos_designs` role (and indicate that it is used as well in the `eos_config_deploy_cvp` to make it clear that it can be used by `eos_designs` in conjunction with the new knob you are adding otherwise it can be a bit confusing. I think it can be enough to just add the variable clearly to the paragraph you wrote above:

```
# Use at the host level
is_deployed: < true | false | default -> true >
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,24 @@
...
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`. 
+- By default, this will have no impact within the `eos_designs` role. Configs will still be generated by the `eos_cli_config_gen` role and will still be pushed by the `eos_config_deploy_eapi` directly to devices if used.
+- However, if the `eos_config_deploy_cvp` role is used to push configurations, CloudVision will ignore the devices flagged  as `is_deployed: false` and not attempt to configure them. 
+- If the device is not present in the network due to CloudVision not configuring the device, `eos_validate_state` role will fail tests on peers of the undeployed device trying to verify that interfaces are up.
+- To overcome this and shutdown interfaces towards undeployed peers, the variable `shutdown_interfaces_towards_undeployed_peers` can be used, satisfying the `eos_validate_state` role interface tests. Again, this is only an issue if `eos_config_deploy_cvp` is used and the devices are not present in the network.
+
+Note: The variables should be applied to all devices in the fabric.
+
 **Variables and Options:**
```
gmuloc: Given that you have added the `Flagging a Device as Not Deployed` section, I wonder if this line  should not be replaced with:

```suggestion
## Variables and Options:
```

Otherwise it looks like the Variables and Options are part of the `Flagging a Device as Not Deployed` section when rendering the markdown.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: I've added the variable as below to "Variables and Options:". Is this enough? `is_deployed` is explained in the paragraph.

```Use at the host level to flag a device as undeployed. See the explanation above as to how this is used in combination with `shutdown_interfaces_towards_undeployed_peers` below.

is_deployed: < true | false | default -> true >```

Comment (gmuloc) [reply ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md]: I think for clarity's sake I would leave this "special" variable definition in the section where you explain it - to be clear that this is not one that should be applied to the whole Fabric

File: ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md [hunk 16]
```diff
@@ -14,13 +14,28 @@
...
+## Flagging a Device as Not Deployed
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`. 
+- By default, this will have no impact within the `eos_designs` role. Configs will still be generated by the `eos_cli_config_gen` role and will still be pushed by the `eos_config_deploy_eapi` directly to devices if used.
+- However, if the `eos_config_deploy_cvp` role is used to push configurations, CloudVision will ignore the devices flagged  as `is_deployed: false` and not attempt to configure them. 
+- If the device is not present in the network due to CloudVision not configuring the device, `eos_validate_state` role will fail tests on peers of the undeployed device trying to verify that interfaces are up.
+- To overcome this and shutdown interfaces towards undeployed peers, the variable `shutdown_interfaces_towards_undeployed_peers` can be used, satisfying the `eos_validate_state` role interface tests. Again, this is only an issue if `eos_config_deploy_cvp` is used and the devices are not present in the network.
+
+Note: The variables should be applied to all devices in the fabric.
```
gmuloc: I think the note here is not relevant in that sections - either it should go back to the previous position (line 17) or it should be inserted just after the `Variables and Options` section

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
carlbuchmann: Yes, these CSV are typically consumed by scripts, i.e. batfish in my case, so I would go with enabled true/false.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: Agreed and moved after `Variables and Options`

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
perimore: We're going back and forth between add the variable and don't add the variable in this section :). 

I'm going to remove it again and leave it in the explanation. Hopefully this is final.

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
```diff
@@ -1,4 +1,4 @@
-Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface
+Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface,Enabled/Shutdown
```
ClausHolbechArista:
```suggestion
Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface,Enabled
```

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
```diff
@@ -11,6 +11,7 @@ Node Type,Node,Node Interface,Peer Type,Peer Node,Peer Interface
 {%             do csv_line.append(node_hostvars.ethernet_interfaces[ethernet_interface].peer_type | arista.avd.default("")) %}
 {%             do csv_line.append(node_hostvars.ethernet_interfaces[ethernet_interface].peer | arista.avd.default("")) %}
 {%             do csv_line.append(node_hostvars.ethernet_interfaces[ethernet_interface].peer_interface | arista.avd.default("")) %}
+{%             do csv_line.append('Shutdown' if node_hostvars.ethernet_interfaces[ethernet_interface].shutdown is arista.avd.defined(true) else 'Enabled') %}
```
ClausHolbechArista:
```suggestion
{%             do csv_line.append(not node_hostvars.ethernet_interfaces[ethernet_interface].shutdown | arista.avd.default(false)) %}
```

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
ClausHolbechArista: or if you prefer Shutdown, fine with me as well, just remove the `not` from line 14 suggestion

perimore: Fixed in my latest push. Making it clear with "Node Interface Enabled" as to not confuse with the "Peer Interface" state

ansible_collections/arista/avd/roles/eos_designs/templates/documentation/fabric-topology.j2
perimore: This is now `Node Interface Enabled: True|False`

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/logic.j2
```diff
@@ -14,6 +14,7 @@
                        "peer": peer,
                        "peer_interface": peer_uplink.interface,
                        "peer_type": peer_facts.switch.type,
+                       "peer_is_deployed": peer_facts.switch.is_deployed | arista.avd.default,
```
tgodaA:
```suggestion
                       "peer_is_deployed": peer_facts.switch.is_deployed,
```

the **is_deployed** is already set a default value (True) in the **eos_designs_facts.py**

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,25 @@
   - IBGP (only with OSPF or ISIS variants in underlay)
 - Only summary network addresses need to be defined. IP addresses are then assigned to each node, based on its unique device id.
   - To view IP address allocation and consumption, a summary is provided in the auto-generated fabric documentation in Markdown and CSV format.
-- The variables should be applied to all devices in the fabric.
 
 *Only supported with core_interfaces data model.
 
-**Variables and Options:**
+## Flagging a Device as Not Deployed
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`.
```
gmuloc:
```suggestion
## Flagging a Device as Not Deployed

- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`.

  ```yaml
  # Use at the host level
  is_deployed: < true | false | default -> true >
  ```
```

gmuloc (approved):
LGTM

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -14,13 +14,30 @@
...
-- The variables should be applied to all devices in the fabric.
 
 *Only supported with core_interfaces data model.
 
-**Variables and Options:**
+## Flagging a Device as Not Deployed
+
+- It is possible to provision configurations for a complete topology but flag devices as undeployed using the host level variable `is_deployed: false`.
+
+  ```yaml
```
tgodaA, line 24:
```suggestion
```yaml
```

tgodaA: I don't see end quotes **```** for this

ClausHolbechArista: sorry I missed this before merging. It renders correctly, so let's fix this in another PR touching this file (I have a few pending, so I will add it there)

gmuloc: ah yeah copy paste issue I had them in my test .. I check the rendered file and it looked ok

however we should not remove the two spaces in front of ```yaml as we want the snippet to be indented under the bulletpoint.

ClausHolbechArista: agreed. I have added the closing ``` to the trunk_groups PR.

Issue #1593, opened 2022-03-16
Remove BGP rendering on irrelevant nodes


emilarista:
There was a hack implemented in the mpls design PR to keep rendering BGP for irrelevant nodes such as spines in the isis-ibgp molecule example in order to make the merge non-breaking. This hack should be removed for 4.0. 

eos_designs/templates/facts/switch/switch.j2

```
{# Hack to make mpls PR non-breaking, adds empty bgp to igp topology spines #}
{%     elif switch_underlay_routing_protocol in ['isis', 'ospf'] and switch_evpn_role == 'none' and bgp_as is arista.avd.defined %}
  bgp_as: "{{ bgp_as }}"
{%     endif %}
```

ClausHolbechArista:
@emilarista

PR #1754, opened 2022-04-29
Refactor(eos_designs): Move eos_designs facts to Python

ClausHolbechArista:
## Change Summary

Move eos_designs facts to Python

## Why ##

Enable future support for --limit
Enable future support for Studios

## Component(s) name

`arista.avd.eos_designs`
`arista.avd.eos_designs_facts`

## Proposed changes

- Refactor all facts templates from jinja2 to python functions
- Create new action plugin `eos_designs_facts` to set facts:
  - avd_switch_facts
  - avd_topology_peers
  - avd_overlay_peers
- Update role tasks to leverage new plugin
- Cleanup old templates
- Refactor network_services logic, to simplify facts stage
- `eos_designs_facts` will `run_once`
  - so the same facts are set on all devices
  - these facts can be used by various templates accessing information about peers

## How to test

- No changes in molecule.
- Tested on various demo and customer projects.

### Repository Checklist


ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -0,0 +1,178 @@
...
+        set_avd_topology_facts = self._task.args.get('avd_topology_facts', False)
+
+        groups = task_vars.get('groups', {})
+        fabric_name = task_vars.get('fabric_name', '')
+        fabric_hosts = groups.get(fabric_name, [])
+        ansible_play_hosts_all = task_vars.get('ansible_play_hosts_all', [])
+
+        # Check if fabric_name is set and that all play hosts are part Ansible group set in "fabric_name"
+        if fabric_name is None or not set(ansible_play_hosts_all).issubset(fabric_hosts):
+            raise AnsibleActionFail("Invalid/missing 'fabric_name' variable. \
```
gmuloc: https://peps.python.org/pep-0008/#maximum-line-length - here it is better to use implied continuation as per pep8 standard - not sure what linter we are using here?

```
raise AnsibleActionFail("Invalid/missing 'fabric_name' variable. All hosts "
                                        "in the play must have the same 'fabric_name' value "
                                        "which points to an Ansible Group containing the hosts.")
```

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -0,0 +1,178 @@
...
+                        if peer not in avd_overlay_peers:
+                            avd_overlay_peers[peer] = []
+                        avd_overlay_peers[peer].append(host)
+
+                backup_hostvars[host] = host_hostvars
+
+            for host in fabric_hosts:
+                host_hostvars = backup_hostvars[host]
+
+                if set_avd_topology_facts:
```
gmuloc: here would maybe be better to reverse the for loop and if check as the facts are not set per host and of `set_avd_topology_facts` is not set the for loop will check for each host that it is empty. 

Probbly same for the previous loop.

ansible_collections/arista/avd/plugins/filter/list_compress.py
```diff
@@ -8,15 +8,15 @@
 from ansible.errors import AnsibleFilterError
 
 
-class FilterModule(object):
+def list_compress(list_to_compress):
```
gmuloc, line 11: would be nice to add a docstring to explain what this function is doing.

out of curiosity, is there a reason to move it out of the class (except removing `self`)?

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -9,49 +9,50 @@
...
+            # Group 12                                              (\d+)                              matches first interface
+            # Group 13                                                      (\d+)                      matches last interface ex. 1, 4, 48
+            # Group 14                                                            (\.((\d+)-)?(\d+))?  matches dot and sub-interface(s) ex. .141, .12-15
+            # Group 15                                                               ((\d+)-)?         matches first sub-interface and dash ex. 12-
+            # Group 16                                                                (\d+)            matches first sub-interface ex. 12
+            # Group 17                                                                        (\d+)    matches last sub-interface ex. 141, 15
+            # Remember that the groups() object is 0-based and the group numbers above are 1-based
+            search_result = re.search(regex, one_range)
+            if search_result:
+                if len(search_result.groups()) == 17:
```
gmuloc: 17 feel a bit like a magic number here - maybe some kind of constant like MAX_GROUP or something like this

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -127,9 +128,12 @@ def expand_module(interface_string):
 
                 else:
```
gmuloc, line 131: so here we are moving the exception down one level - so if search_result is empty we will not raise the exception anymore - it looks ok from the logic but just checking

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/node_type.py
```diff
@@ -0,0 +1,115 @@
...
+        Hostvars for one host
+
+    Returns
+    -------
+    dict
+        switch.*: Facts based on node_type_keys data model
+    """
+
+    switch = vars.get('switch', {})
+    switch['type'] = vars['type']
```
gmuloc: `KeyError` if there is no`'type'` in vars. It is caught `eos_design_facts.py` so probably ok but maybe lets add in the doc string that it is expected

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/node_type.py
```diff
@@ -0,0 +1,115 @@
...
+    node_type_key_data = vars['node_type_keys'][node_type_key]
+
+    switch.update(get_facts(NODE_TYPE_FACTS, node_type_key_data))
+
+    return {"switch": switch}
+
+
+def get_node_type_key(vars):
+    """
+    Get node_type_key by searching for the corrosponding "type" under "node_type_keys"
```
gmuloc:
```suggestion
    Get node_type_key by searching for the corresponding "type" under "node_type_keys"
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/node_type.py
```diff
@@ -0,0 +1,115 @@
...
+
+    for key in node_type_keys:
+        if node_type_keys[key]['type'] == type:
+            node_type_key = key
+            break
+
+    if not node_type_key:
+        raise KeyError(f"node_type_keys.<>.type=={type}")
+
+    return node_type_key
```
gmuloc:
```suggestion
    for key, value in node_type_keys.items():
      if value['type'] == type:
        return key
    
    # Not found 
    raise KeyError(f"node_type_keys.<>.type=={type}")
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/templates.py
```diff
@@ -0,0 +1,61 @@
...
+        "get": "templates.ip_addressing",
+        "node_type_data_key": "ip_addressing",
+        "default": {}
+    }, {
+        "set": "interface_descriptions",
+        "get": "templates.interface_descriptions",
+        "node_type_data_key": "interface_descriptions",
+        "default": {}
+    }
+]
```
gmuloc:
```suggestion
TEMPLATE_FACTS = [
    {
        "set": "ip_addressing",
        "get": "templates.ip_addressing",
        "node_type_data_key": "ip_addressing",
        "default": {},
    },
    {
        "set": "interface_descriptions",
        "get": "templates.interface_descriptions",
        "node_type_data_key": "interface_descriptions",
        "default": {},
    },
]
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/templates.py
```diff
@@ -0,0 +1,61 @@
...
+    node_type_key = switch['node_type_key']
+
+    for template_fact in TEMPLATE_FACTS:
+        get_key = template_fact['get']
+        set_key = template_fact['set']
+        node_type_data_key = template_fact['node_type_data_key']
+        default_value = template_fact['default']
+
+        var_templates = get(vars, get_key, {})
+        node_type_templates = get(vars, f"node_type_keys.{node_type_key}.{node_type_data_key}", {})
```
gmuloc: Lots of opportunities for `KeyError` here as well - wondering how to give back meaningful error messages back to the user as the current error message would look like:
```
raise AnsibleActionFail(f"{e} is required but was not found for host '{host}'") from e
```

but here it could be a problem in some internal dictionary (TEMPLATE_FACTS missing a key) or some missing node_type_key in the switch dictionary (though it would have raised before when trying to set the type in our current flow)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+import ipaddress
+from hashlib import sha256
+from ansible_collections.arista.avd.plugins.module_utils.utils import get, default, get_facts, template_var
+
+
+# List of switch facts to set based on keys read from "Fabric Topology" data (ex. "l3leaf"). Use dot-notation for nested dicts in "key".
+# "set_key" is optional unless "key" contains dot-notation. "set_key" will default to same value as "key"
+GENERAL_FACTS = [
+    {
+        "key": "id",
```
gmuloc: is there some kind of documentation regarding all these dictionaries with the potential vales inside (set_key, required, default, ...) for developers to take part in?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+        Jinja filters loaded in main module
+    template_lookup_module : func
+        Instance of Ansible 'template' lookup module used to render ip addresses with user-defined templates
+
+    Returns
+    -------
+    dict
+        switch.*: Facts based on fabric_topology data model
+
+    Raises
```
gmuloc: This section is missing in other `.py` file and I think it would be good to add it

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+    if get(switch_data_combined, 'link_tracking.enabled') is True:
+        switch['link_tracking_groups'] = get_link_tracking_groups(switch, switch_data_combined)
+
+    if get(switch_data_combined, 'lacp_port_id_range.enabled') is True:
+        node_group_length = max(len(switch_data_node_group_nodes), 1)
+        switch['lacp_port_id'] = get_lacp_port_id_range(switch, switch_data_combined, node_group_length)
+
+    # ##############################
+    # Network services related facts
+    if switch.get('network_services_l2') is True \
```
gmuloc: PEP8 - better to remove the `\`

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+            or switch.get('network_services_l3') is True \
+            or switch.get('network_services_l1') is True:
```
gmuloc:
```suggestion
    if (
        switch.get("network_services_l2") is True
        or switch.get("network_services_l3") is True
        or switch.get("network_services_l1") is True
    ):
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+            switch['ipv6_router_id'] = get_ipv6_router_id(vars, template_lookup_module)
+
+        switch['evpn_role'] = default(switch_data_combined.get('evpn_role'), switch['default_evpn_role'])
+
+        switch['mpls_overlay_role'] = default(switch_data_combined.get('mpls_overlay_role'),
+                                              switch['default_mpls_overlay_role'])
+
+        if switch['underlay_routing_protocol'] == 'ebgp' or switch['evpn_role'] != 'none' or switch['mpls_overlay_role'] != 'none':
+            switch['bgp_as'] = get_bgp_as(vars, switch_data_combined)
+        # Hack to make mpls PR non-breaking, adds empty bgp to igp topology spines
```
gmuloc: will this hack be fixable at some point or is it here to stay?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+
+        if switch['mpls_overlay_role'] in ["client", "server"]:
+            switch['mpls_route_reflectors'] = switch_data_combined.get('mpls_route_reflectors')
+
+        if switch['underlay_routing_protocol'] in ['isis', 'isis-ldp', 'isis-sr', 'isis-sr-ldp']:
+            switch.update(get_isis_facts(vars, switch_data_combined))
+
+    # #######################
+    # MLAG related facts
+    if switch['mlag_support'] is True and switch_data_combined.get('mlag', True) is True and len(switch_data_node_group_nodes) == 2:
```
gmuloc: Here woudl we want the fact role to output warnings if there are more than 2 switches and Mlag is supposed to be enabled?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+
+        elif list(switch_data_node_group_nodes.keys()).index(vars['inventory_hostname']) == 1:
+            switch['mlag_role'] = "secondary"
+            switch['mlag_peer'] = list(switch_data_node_group_nodes.keys())[0]
+
+        if switch.get('underlay_router') is True:
+            switch['mlag_l3'] = True
+            mlag_peer_vlan = switch['mlag_peer_vlan']
+            if switch_data_combined.get('mlag_peer_l3_vlan', 4093) not in [None, False, mlag_peer_vlan]:
+                switch['mlag_peer_l3_vlan'] = switch_data_combined.get('mlag_peer_l3_vlan', 4093)
```
gmuloc:
```suggestion
            combined_peer_l3_vlan = switch_data_combined.get('mlag_peer_l3_vlan', 4093)
            if combined_peer_l3_vlan not in [None, False, switch['mlag_peer_vlan']:
                switch['mlag_peer_l3_vlan'] = combined_peer_l3_vlan
```

remove one `get` call

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+        switch['inband_management_role'] = 'child'
+        switch['inband_management_parents'] = switch['uplink_switches']
+        switch.update(get_facts(INBAND_MANAGEMENT_FACTS, switch_data_combined))
+        switch.update(get_inband_management_facts(switch))
+
+    # Remove None values.
+    sanitized_switch = {}
+    for key in switch:
+        if switch[key] is not None:
+            sanitized_switch[key] = switch[key]
```
gmuloc:
```suggestion
    sanitized_switch = {k: v for k,v in switch.items() if v is not None}
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+    else:
+        for node_group in node_type_config.get('node_groups', {}):
+            if host in node_type_config['node_groups'][node_group].get('nodes', {}):
+                switch_data['node'] = node_type_config['node_groups'][node_group]['nodes'][host]
+                switch_data['node_group'] = node_type_config['node_groups'][node_group]
+                switch_data['group'] = node_group
+                break
+
+    if not switch_data.get('node'):
+        raise KeyError(f"{node_type_key}.(node_groups.)nodes.{host}")
```
gmuloc: I wonder if this error message makes it to the user given that the KeyError will be caught in upper methods and a different message is used

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+    Returns
+    -------
+    dict : platform_settings
+    """
+
+    for platform_setting in vars.get('platform_settings', []):
+        if platform in platform_setting.get('platforms', []):
+            return platform_setting
+
+    for platform_setting in vars.get('platform_settings', []):
```
gmuloc: Here if I have defined platform_settings in multiple places I will end up picking up the first occurence of `default`?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+        Default value
+
+    Returns
+    -------
+    str : underlay_routing_protocol
+    """
+
+    underlay_routing_protocol = str(default(vars.get('underlay_routing_protocol'),
+                                            default_underlay_routing_protocol)).lower()
+    if underlay_routing_protocol not in ['ebgp', 'isis', 'isis-ldp', 'isis-sr', 'isis-sr-ldp', 'ospf', 'ospf-ldp']:
```
gmuloc: probably should document this list somewhere - typically there is no `ibgp` possible here

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+    match_tags = switch['filter_tags']
+    if switch['group'] is not None:
+        match_tags.append(switch['group'])
+
+    network_services_keys = vars.get('network_services_keys', [])
+    for network_services_key in avd_natural_sort(network_services_keys, 'name'):
+        network_services_key_name = network_services_key.get('name')
+        if network_services_key_name is not None and vars.get(network_services_key_name) is not None:
+            tenants = vars.get(network_services_key_name)
+            # Support legacy data model by converting nested dict to list of dict
```
gmuloc: so here more a question than a comment - is this support going to be removed at some point?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
```diff
@@ -0,0 +1,856 @@
...
+        isis_system_id_prefix = switch_data_combined['isis_system_id_prefix']
+        facts['isis_net'] = f"{isis_area_id}.{isis_system_id_prefix}.{switch_id:04d}.00"
+
+    default_is_type = vars.get('isis_default_is_type', 'level-2')
+    is_type = str(switch_data_combined.get('is_type', default_is_type)).lower()
+    if is_type not in ["level-1", "level-2", "level-1-2"]:
+        is_type = "level-2"
+    facts['is_type'] = is_type
+
+    if switch.get('mpls_lsr') is True:
```
gmuloc: here for the "CORE" and "EVPN_UNDERLAY" - I wonder if it would not be better to move these default strings to some centralized place (like CONSTANTS or whatever) in order to make it simple to look it up rather than inside the python files (where changing it in one place could mean missing it in another).
Another option would be to have this defined in YAML though I am not sure we want the defaults value we allocate when there is no specified value to be easily overriden as it will makes things more complex.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+    Returns
+    -------
+    dict
+        topology.links: List of uplinks with all parameters
+        topology.peers: List of all uplink peers
+    """
+
+    switch = vars.get('switch', {})
+
+    if switch.get('uplink_type') == 'p2p':
```
gmuloc: probably some comment that will be addressed only once we put in place the schema but what happens if the the user put another value than p2p or port-channel in the `uplink_type`?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+
+    # Special cases of setting ips during topology facts run,
+    # since these functions relies on data set in avd_switch_facts
+    if switch.get('vtep') is True:
+        switch['vtep_ip'] = get_vtep_ip(vars, template_lookup_module)
+
+    if switch.get('mlag') is True:
+        switch.update(get_mlag_ips(vars, template_lookup_module))
+
+    sanitized_topology = {}
```
gmuloc: can use the same oneliner to remove `None` values or maybe make some method in utils if we start using it in multiple places

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+    for key in topology:
+        if topology[key] is not None:
+            sanitized_topology[key] = topology[key]
```
gmuloc:
```suggestion
sanitized_switch = {k: v for k,v in switch.items() if v is not None}
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+    topology = vars.get('topology', {})
+
+    topology['links'] = []
+    topology['peers'] = []
+
+    uplink_interfaces = switch.get('uplink_interfaces', [])
+
+    for uplink_index, uplink_interface in enumerate(uplink_interfaces):
+        uplink_switches = switch.get('uplink_switches', [])
+        if len(uplink_switches) > uplink_index:
```
gmuloc: so what happens if there are more `uplink_switches` than uplink_interfaces`?  Looks like we would just ignore the end of the list but maybe worth outputting some warning if we can

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+    switch = vars['switch']
+    topology = vars.get('topology', {})
+
+    topology['links'] = []
+    topology['peers'] = []
+
+    uplink_interfaces = switch.get('uplink_interfaces', [])
+
+    for uplink_index, uplink_interface in enumerate(uplink_interfaces):
+        uplink_switches = switch.get('uplink_switches', [])
```
gmuloc: you can move this line out of the for loop I think so that you do the get call only once rather than for each interface

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+    avd_range_expand = filters['avd_range_expand']
+    avd_list_compress = filters['avd_list_compress']
+
+    topology['links'] = []
+    topology['peers'] = []
+
+    uplink_interfaces = switch.get('uplink_interfaces', [])
+
+    for uplink_index, uplink_interface in enumerate(uplink_interfaces):
+        uplink_switches = switch.get('uplink_switches', [])
```
gmuloc: probably can get this variable assignment out of the for loop

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
```diff
@@ -0,0 +1,332 @@
...
+
+def get_vtep_ip(vars, template_lookup_module):
+    """
+    Run template lookup to render ipv4 address for vtep_ip
+
+    Since some templates might contain certain legacy variables (switch_*),
+    those are mapped from the switch.* model
+
+    Even though this will set switch.vtep_ip, it is run as part of the topology facts,
+    since we rely on "avd_switch_facts", which is not set until the switch facts phase has completed.
```
gmuloc: if I read this correctly it means the code is doing some computation twice, once for topology facts and once for switch facts for the special cases where the topology facts require some info that is not yet available?

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+def get(dictionary, key, default=None, required=False, org_key=None):
+    """
+    Get vlans list based on network_services_keys and filters defined in fabric_topology data model
```
gmuloc: is it only vlans list? the methode look more generic

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    Returns
+    -------
+    any
+        Value or default value
+    """
+
+    if value is not None:
+        return value
+    else:
+        for default_value in default_values:
```
gmuloc: are there cases where we receive multiple values and some of them are None? I am not sure I get the need for handling multiple default values here.

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+                return default_value
+
+    return None
+
+
+def unique(in_list):
+    """
+    Return list of unique items from the in_list
+
+    If Value is not None, the default values will be checked one by one.
```
gmuloc: Not sure this is relevant here

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        return value
+    else:
+        for default_value in default_values:
+            if default_value is not None:
+                return default_value
+
+    return None
+
+
+def unique(in_list):
```
gmuloc, line 84: the function will raise an exception if `in_list` is `None`

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+        Unique list items
+    """
+
+    list_set = set(in_list)
+    return (list(list_set))
+
+
+def get_facts(facts, input_data):
+    """
+    Get vlans list based on network_services_keys and filters defined in fabric_topology data model
```
gmuloc: Need to change the docstring

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+
+    facts: [
+        {
+            "key": <>
+            "set_key": <optional - >
+            "default": <optional - default value
```
gmuloc: add `required` here

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,166 @@
...
+    The result is forced into a string and leading/trailing newlines and whitespaces are removed.
+
+    Parameters
+    ----------
+    template_file : str
+        Filename to pass to template_lookup_module
+    template_vars : any
+        Vars to pass to template_lookup_module
+    template_lookup_module : func
+        Instance of Ansible 'template' lookup module used to render ip addresses with user-defined templates
```
gmuloc: is it only used to render ip addresses?

ansible_collections/arista/avd/plugins/modules/eos_designs_facts.py
```diff
@@ -17,32 +17,35 @@
...
     description: |
-      Set avd_overlay_peers fact with a list per device of EVPN Route Server Clients,
-      meaning devices pointing to this device as EVPN Route Server
+      Set avd_switch_facts fact with a dict per device of topology.* facts
+      Set avd_topology_peers fact with a list per device of downstream devices,
+      meaning devices pointing to this device as uplink switch
+      Most topology_facts are based on switch.* facts, so those must be set in the same
+      or a previous task.
     required: False
     type: bool
```
gmuloc, line 41: not sure here but should we also indicate "why" it is better to run this task with `run_once` in the documentation?

ansible_collections/arista/avd/roles/eos_designs/doc/node-types.md
```diff
@@ -107,12 +107,12 @@ vtep_ip:
 - `{{ switch_id }}`
 - `{{ loopback_ipv4_offset }}`
 
-`p2p_uplinks_ip` & `p2p_uplinks_peer_ip` can leverage the switch facts (switch.*) to customize the IP adressing.
-All other templates can only leverage the input variables and variables set in switch facts template.
+While all templates can leverage the internal switch facts (switch.*) to customize the interface descriptions,
```
gmuloc, line 110: Maybe let's emphasise this (bold or something). Are the internal switch facts mentioned here the same as the `contextual variables` from line 76?

ansible_collections/arista/avd/roles/eos_designs/templates/inband_management/ip-virtual-router-mac-address.j2
```diff
@@ -1,4 +1,6 @@
 {# Set virtual router mac address unless it will be set in network services #}
 {% if not (switch.network_services_l2 is arista.avd.defined(true) and switch.network_services_l3 is arista.avd.defined(true)) %}
+{%     if switch.virtual_router_mac_address is arista.avd.defined(fail_action="error", var_name="virtual_router_mac_address") %}
```
gmuloc, line 3: it looks like this make `virtual_router_mac_address` mandatory when either one (or both) of the l2 or l3 network services is not defined - is it the intention here?

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
```diff
@@ -1 +1,10 @@
-{{ switch.overlay_loopback_description }}
+{% if overlay_loopback_description is not arista.avd.defined %}
```
gmuloc, line 1: for this file I am not really able to review but I am not sure if this is part of the move to python?

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
```diff
@@ -1,32 +1,63 @@
...
+{%             set whitelist.vlans = whitelist.vlans | select('in', uplink_switch_vlans) %}
+{%         endif %}
+{%     endfor %}
+{% endif %}
 {% set tmp_hostvars = hostvars[inventory_hostname] %}
 {% for network_services_key in network_services_keys | arista.avd.natural_sort('name') %}
 {%     if network_services_key.name is arista.avd.defined and tmp_hostvars[network_services_key.name] is arista.avd.defined %}
 {# tenants #}
 {%         for tenant in tmp_hostvars[network_services_key.name] | arista.avd.convert_dicts | arista.avd.natural_sort('name') %}
-{%             if tenant.name in switch.tenants | arista.avd.default([]) %}
```
gmuloc, line 7: looks like this change is also not related to the facts moving to python but rather adding some functionality to the tenants  - it looks good

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
```diff
@@ -1,32 +1,63 @@
 {% set tmp_tenants = [] %}
+{% set match_tags = switch.filter_tags %}
+{% if switch.group is arista.avd.defined %}
+{%     do match_tags.append(switch.group) %}
+{% endif %}
+{% set whitelist = namespace(vlans=switch.vlans | arista.avd.default([]) | arista.avd.range_expand | map('int')) %}
```
gmuloc, line 6: is this whitelist some kind of protection against using wrong VLAN id in tenant files?
If so it could be good to output some warning that the vlans are being skipped

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vlans.j2
```diff
@@ -4,11 +4,9 @@ vlans:
 {%     for vrf in tenant.vrfs %}
 {# Tenant VLANs w/SVIs #}
 {%         for svi in vrf.svis %}
-{%             if svi.id | int in topology.vlans %}
```
gmuloc, line 7: this check is removed because there is no more `topology.vlans` variables - how is this behavior conserved (or is it removed on purpose?)

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/ethernet-interfaces.j2
```diff
@@ -17,8 +16,7 @@ ethernet_interfaces:
     service_profile: {{ p2p_uplinks_qos_profile }}
 {%         endif %}
 {%         if link.ptp is arista.avd.defined %}
-    ptp:
-      enable: {{ link.ptp }}
+    ptp: {{ link.ptp }}
```
gmuloc, line 19: there are some templates in `eos_cli_config_gen` which look like:
```
{%         if ethernet_interface.ptp.enable is arista.avd.defined(true) %}
```

https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2#L393

and this is not changed in this PR - would not this break?

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/port-channel-interfaces.j2
```diff
@@ -14,7 +13,7 @@ port_channel_interfaces:
 {%         endif %}
     type: switched
     shutdown: false
-    vlans: {{ link.vlans | arista.avd.list_compress }}
```
gmuloc, line 17: Does removing compress here change the behavior?

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -539,3 +541,83 @@ templates:
           list_merge: "{{ custom_structured_configuration_list_merge }}"
           strip_empty_keys: false
 ```
+
+### EOS Designs Facts
```
gmuloc, line 553: I think to allow people to contribute to this part it will require an explanation on how to modify the files, the structure of the _FACTS dictionary (though it is more or less explains in utils.py)

gmuloc (commented):
I have been through all the files though it may not be a completely thorough review - a couple of the comments are really questions.

A couple of generic points:
* The README.md part should be increased to help understanding how to contribute in particular the structure of the  *_FACTS dictionaries and how to augment them / with what options
* the python pre-commit linter is pylint though if we start using more python in the repo we should probably take a moment to think about python formatting (black/flake8/..)
* Many different functions can raise `KeyError` and I am not sure how the error would be presented back to the user as there is a kind of catch all KeyError in `eos_designs_facts.py` that could make troubleshooting a bit difficult: https://github.com/aristanetworks/ansible-avd/blob/ec2e88887f0c201005eabed5e1730851d905e05d/ansible_collections/arista/avd/plugins/action/eos_designs_facts.py#L144 
* Some changes looks not related to the PR (modification of ptp behavior / addition of some behavior for the VRFs ...) and it does not seem that they are documented
* in several places there is a support in place for legacy models (all the convert dict to list essentially) - maybe would be good to start announcing when it is not going to be supported (like a deprecation warning being printed by the plugin maybe?)

ansible_collections/arista/avd/plugins/filter/list_compress.py
ClausHolbechArista: I wanted to be able to import it easily in module_utils, without having to handle the different Ansible classes.

ansible_collections/arista/avd/plugins/filter/range_expand.py
ClausHolbechArista: It is the number of regex match groups in the elaborate description above. I can add a comment.

ansible_collections/arista/avd/plugins/filter/range_expand.py
ClausHolbechArista: Yes exactly. So in the case of a blank string in input, we want to return an empty list

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/node_type.py
ClausHolbechArista: `type` is one of our required vars. We have many others - set by required=True here and there. I don't think we should document the data model directly in docstring, but I can add the `Raises` section and describe that we raise KeyError if a required key is missing (Should match the required keys from avd.sh)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/templates.py
ClausHolbechArista: The contents of `e` should still lead us on the way. We could make our error message instead of relying on `KeyError`, and then we catch on that in the wrapper function. We could use our own `get` with `required=True` everywhere to make both ensure the correct error is raised, but also to make it clear to the reader that this is a datamodel variable we have as `required`.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
ClausHolbechArista: They will not get a topology built :)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
ClausHolbechArista: I have a strip_empties already, which we should probably use here instead. So we remove None values recusively.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
ClausHolbechArista: We try to forgive missing entries in either list here. So just silently ignoring the missing entries. This is similar to what we have in Jinja today.
We have some incoming change which will set some of these values based on the platform. With that change, we will need the `uplink_switches` to be the "main" iterator, and then just ignore if there are more interfaces available. Only if we have fewer interfaces, we should eject a warning.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/topology.py
ClausHolbechArista: Not really. We are waiting with a few switch facts, until we have the topology facts set.
So:
1. Switch facts
2. Topology facts
3. A few more switch facts (not the same as 1)

I will try to improve the description here.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: This is a wrong docstring. :) I will fix it.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: We have that in jinja today, and it is widely used. Where we check a primary input var, fallback to a secondary input var, maybe a tertiary and so on. Often we then have a static value at the end.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: typo. Thanks

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: Any suggestions? Check specifically for `None` and return `[]` maybe?

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: Mostly yes. But this could be used for anything. I will fix.

ansible_collections/arista/avd/roles/eos_designs/doc/node-types.md
ClausHolbechArista: It varies a bit. We need to change our templates, to use the `switch.*` values instead of some temporary variables. But that would break any customer templates, so we need to maintain these until next major version.
So I don't want to spend too much time highlighting all the legacy variables. I think I will change the current templates, and update the docs to suggest only using the `switch.*` facts (which are still subject to change). Existing deployments using these vars will still work because of the extra care we take in the template functions.

ansible_collections/arista/avd/roles/eos_designs/templates/inband_management/ip-virtual-router-mac-address.j2
ClausHolbechArista: Only if this template runs, which means only if `inband_management_subnet` is set.
There is no behavioral change here, we just output a "nicer" error message instead of a generic "Undefined variable" from Jinja.

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
ClausHolbechArista: I am moving this logic away from the facts. Previously it was setting `switch.overlay_loopback_description` in the `switch.j2` which was then read here only. So with this I am moving the logic to the only place where it matters. This will also make it simpler to override.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
ClausHolbechArista: Maybe I should add a description. This logic is used for _not_ deploying a vlan on a `l2leaf` if the "parent" does not have this vlan. This is the existing behavior, but it was implemented elsewhere.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
ClausHolbechArista: I have removed the `switch.tenants` "tree` to avoid carrying a large variable tree around in facts. So this logic was previously in `switch.j2` (look at the bottom), where it implemented the filtering as part of facts.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vlans.j2
ClausHolbechArista: This is now implemented in the `network_services/logic.j2` directly (the whitelist stuff)

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/ethernet-interfaces.j2
ClausHolbechArista: I am changing the content of `link.ptp` to be the full dictionary from `uplink_ptp`, where previously it was only taking the `enabled` key. This way we can add more keys to the input data model (read by eos_cli_config_gen) without having to change eos_designs.
Snip from `topology.py`:
```python
                if switch.get('uplink_ptp') is not None:
                    topology_link['ptp'] = switch['uplink_ptp']
```

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/port-channel-interfaces.j2
ClausHolbechArista: I have moved the compress to the fact function, so we don't have to carry a list of 2000 vlans just to compress them into a string here. Instead we compresss it "at the source".

ansible_collections/arista/avd/plugins/filter/range_expand.py
ClausHolbechArista: move to constants

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/node_type.py
ClausHolbechArista: get(vars, 'type', required=True)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
ClausHolbechArista: tracked in [#1593](https://github.com/aristanetworks/ansible-avd/issues/1593)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
ClausHolbechArista: will change to strip_empties

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py
ClausHolbechArista: We plan to move such defaults out to yaml in the future

ansible_collections/arista/avd/plugins/module_utils/utils.py
gmuloc: returning `[]` from `None` could indeed be a solution except if there are cases in which we want to make the difference between passing `[]` and passing `None` as input and in that case probably should `raise`. Now as a first approach I think returning `[]` is probably good enough.

ansible_collections/arista/avd/roles/eos_designs/templates/inband_management/ip-virtual-router-mac-address.j2
gmuloc: thanks - resolving this!

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
gmuloc: Ok so maybe the only comment based on our call would be to make the default part of the "default_values.yml" file rather than hardcoding them in the template here.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
gmuloc: ok - I did not read through all of switch.j2 during my review. Resolving this

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/logic.j2
gmuloc: probably worth adding some comment in the template indeed :)

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vlans.j2
gmuloc: ack - resolving this

ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/port-channel-interfaces.j2
gmuloc: for record keeping: https://github.com/aristanetworks/ansible-avd/blob/ec2e88887f0c201005eabed5e1730851d905e05d/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/switch.py#L718

Resolving

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts/eos_designs_facts.py
```diff
@@ -0,0 +1,1201 @@
...
+        self.message = message
+        super().__init__(self.message)
+
+
+class EosDesignsFactsMissingVariableError(EosDesignsFactsError):
+    pass
+
+
+class EosDesignsFacts:
+    _hostvars = {}
```
gmuloc: just a quick question - using mutable objects like list and dict as class variables can have surprising effects in python (cf https://docs.python.org/3/tutorial/classes.html#class-and-instance-variables) is this the intention here?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+        if self._any_network_services:
+            vlans = []
+            match_tags = self.filter_tags
+            if self.group is not None:
+                match_tags.append(self.group)
+
+            network_services_keys = get(self._hostvars, "network_services_keys", default=[])
+            for network_services_key in self._natural_sort(network_services_keys, "name"):
+                network_services_key_name = network_services_key.get("name")
+                if network_services_key_name is not None and get(self._hostvars, network_services_key_name) is not None:
```
gmuloc: for this kind of methods with a lots of nested block, a static code analysis tool would generally end up complaining about too many nested if/for/..

a solution here to save on indentation in the for loop could be something like
```
for ...
  if network_service_key_name is None or get(self._hostvars, network_services_key_name) is None:
    continue # move to next element of the list
  tenants = self._convert_dicts(tenants, 'name')
```

sometimes it make things more readable rather than too many nested for/if loop

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -0,0 +1,154 @@
...
+                    if peer not in avd_overlay_peers:
+                        avd_overlay_peers[peer] = []
+                    avd_overlay_peers[peer].append(host)
+
+                host_topology_peers = avd_switch_facts[host]['switch'].get('uplink_peers', [])
+
+                for peer in host_topology_peers:
+                    if peer not in avd_topology_peers:
+                        avd_topology_peers[peer] = []
+                    avd_topology_peers[peer].append(host)
```
gmuloc:
```suggestion
for peer in host_topology_peers:
  avd_topology_peers.setdefault(peer, []).append(host)
```

gmuloc: can be applied above as well for the host_evpn_route_servers

ansible_collections/arista/avd/plugins/module_utils/utils.py
gmuloc: hmmm not sure if this was changed yet - looks like it was not in my Github review view

ansible_collections/arista/avd/plugins/module_utils/utils.py
gmuloc: same here not sure if this is only a display problem on my end

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+        """
+        switch_data = {}
+        hostname = self.hostname
+        node_type_config = get(self._hostvars, f"{self.node_type_key}", required=True)
+
+        if hostname in node_type_config.get('nodes', {}):
+            node_config = node_type_config['nodes'][hostname]
+            switch_data['node_group'] = {}
+        else:
+            for node_group in node_type_config.get('node_groups', {}):
```
gmuloc, line 326: it seems we have quite often this pattern:
```
for element in config.get(elements):
  if lookup_value in config[elements][element].get('sthing')
    do stuff
```

I am wondering if this could not be replaced with some lookup function un utils
like
```
def find(config, parent_key, lookup_key, lookup_value):
  return parent_value or None
```

to be used in multiple places

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+            template_vars = {"ansible_search_path": self._ansible_search_path}
+            # Copying __dict__ will expose all switch facts cached up until this function is run.
+            # TODO: We should probably find and document a list of supported context variables instead.
+            template_vars['switch'] = {k: self.__dict__.get(k) for k in self.keys()}
+            template_vars['switch_id'] = self.id
+            template_vars['loopback_ipv4_pool'] = self.loopback_ipv4_pool
+            template_vars['loopback_ipv4_offset'] = self.loopback_ipv4_offset
+            template_path = get(self.ip_addressing, "router_id", required=True)
+            return template_var(template_path, template_vars, self._template_lookup_module)
+        return None
```
gmuloc: typically here returning None if not True first  allow to have less nested code (though we don't have too many nesting in this one granted)
```
if self.underlay_router is not True:
  return None
template_vars = {"ansible_search_path": self._ansible_search_path}
# TODO: We should probably find and document a list of supported context variables instead.
template_vars['switch'] = {k: self.__dict__.get(k) for k in self.keys()}
template_vars['switch_id'] = self.id
template_vars['loopback_ipv4_pool'] = self.loopback_ipv4_pool
template_vars['loopback_ipv4_offset'] = self.loopback_ipv4_offset
template_path = get(self.ip_addressing, "router_id", required=True)
return template_var(template_path, template_vars, self._template_lookup_module)
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+
+    @cached_property
+    def bgp_cluster_id(self):
+        if self.underlay_router is True:
+            return get(self._switch_data_combined, "bgp_cluster_id")
+        return None
+
+    @cached_property
+    def bgp_peer_groups(self):
```
gmuloc, line 719: random thought not for this PR but we could leverage the code somewhere to verify the passwords look correct here (would save some time when trying to push wrong config to a box)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+    def mpls_route_reflectors(self):
+        if self.underlay_router is True:
+            if self.mpls_overlay_role in ["client", "server"]:
+                return get(self._switch_data_combined, "mpls_route_reflectors")
+        return None
+
+    @cached_property
+    def isis_net(self):
+        if self.underlay_router is True:
+            if self.underlay_routing_protocol in ['isis', 'isis-ldp', 'isis-sr', 'isis-sr-ldp']:
```
gmuloc, line 833: wondering if for these four methods it would be interesting to add 

```python
def _is_underlay_routing_protocol_in(self, input_list):
  return underlay_router is True and self.underlay_routing_protocol in input_list
```

to then have the four methods look like:

```python
def <name>(self):
  if _is_underlay_routing_protocol_in(['isis', 'isis-ldp', 'isis-sr', 'isis-sr-ldp']):
    do_stuff
    return value
  return None
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+
+    @cached_property
+    def mlag(self):
+        if (
+            self.mlag_support is True
+            and get(self._switch_data_combined, "mlag", default=True) is True
+            and len(self._switch_data_node_group_nodes) == 2
+        ):
+            return True
+        return False
```
gmuloc:
```suggestion
    def mlag(self):
        return (
            self.mlag_support is True
            and get(self._switch_data_combined, "mlag", default=True) is True
            and len(self._switch_data_node_group_nodes) == 2
        )
```

no need to return True False as we literally return the boolean value if the comparison

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+        Run template lookup to render ipv4 address for vtep_ip
+
+        Since some templates might contain certain legacy variables (switch_*),
+        those are mapped from the switch.* model
+        '''
+        if self.vtep is True:
+            template_vars = {"ansible_search_path": self._ansible_search_path}
+            # Copying __dict__ will expose all switch facts cached up until this function is run.
+            # TODO: We should probably find and document a list of supported context variables instead.
+            template_vars['switch'] = {k: self.__dict__.get(k) for k in self.keys()}
```
gmuloc: would it not be possible to use `self.render()` here?

the difference is that render remove None but would it not work in this case?:
```
return {k: getattr(self, k) for k in self.keys() if getattr(self, k) is not None}
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -0,0 +1,1306 @@
...
+            template_vars['switch_vtep_loopback_ipv4_pool'] = self.vtep_loopback_ipv4_pool
+            template_vars['loopback_ipv4_offset'] = self.loopback_ipv4_offset
+
+            if self.mlag is True:
+                if self.mlag_role == 'primary':
+                    template_vars['mlag_primary_id'] = self.id
+                    template_vars['mlag_secondary_id'] = self._mlag_peer_id
+                elif self.mlag_role == 'secondary':
+                    template_vars['mlag_secondary_id'] = self.id
+                    template_vars['mlag_primary_id'] = self._mlag_peer_id
```
gmuloc, line 1218:
```suggestion
            if self.mlag is True:
                template_vars['mlag_primary_id'] = self.id if self.mlag_role == "primary" else self._mlag_peer_id
                template_vars['mlag_secondary_id'] = self._mlag_peer_id if self.mlag_role == "primary" else self.id
```

Only a suggestion - not everyone loves ternary operators
though in any case the case where role is neither `primary` not `secondary` is not handled (but would be handled by schema)

gmuloc (approved):
Added a couple of questions / suggestions to reformat but nothing major - probably should fix the two descriptions in utils.py if this is not a display glitch but otherwise LGTM.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: I will not spend too much time on this for now, since we will refactor all this to lists of dicts instead of nested dicts. As this happens, we will have to write some helper functions indeed.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: If we move that around, we should do it everywhere. I am not sure I agree that it is more readable in general.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: Please investigate how, and submit an issue with your findings/thoughts

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: I don't think we save a lot with such a function. Instead we could do something like `_need_isis` and `_need_bgp` and implement the checks there, and just check the single fact everywhere else.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: The point is we do _not_ want it to self-render, since that would also trigger this function itself, or other similar functions, which would then in turn trigger this one.
For now I think this is a "cheap" solution, and we need to solve this generally (later)

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: We normally prefer the more elaborate style.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: I forgot... I will fix it

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: this could have been named `get_first_value_which_is_not_none(*values)` :)
I will change this to just take a list of values and check them one by one.

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
ClausHolbechArista: This would be a larger initiative, hence not part of this refactoring.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
ClausHolbechArista: Nice. Didn't know that trick.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: I have tried to fix the worst ones

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
ClausHolbechArista: I will push those improvements to later.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -0,0 +1,150 @@
...
+from ansible_collections.arista.avd.plugins.filter.list_compress import list_compress
+from ansible_collections.arista.avd.plugins.filter.natural_sort import natural_sort
+from ansible_collections.arista.avd.plugins.module_utils.eos_designs_facts import EosDesignsFacts
+from ansible_collections.arista.avd.plugins.module_utils.utils import AristaAvdMissingVariableError
+
+
+class ActionModule(ActionBase):
+    def run(self, tmp=None, task_vars=None):
+        if task_vars is None:
+            task_vars = {}
```
tgodaA, line 21:
```
    def run(self, tmp=None, task_vars={}):
```

I believe we can write it this way.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -0,0 +1,150 @@
...
+                for peer in host_evpn_route_servers:
+                    avd_overlay_peers.setdefault(peer, []).append(host)
+
+                host_topology_peers = avd_switch_facts[host]['switch'].get('uplink_peers', [])
+
+                for peer in host_topology_peers:
+                    avd_topology_peers.setdefault(peer, []).append(host)
+
+            result['ansible_facts'] = {}
+            if set_avd_switch_facts:
```
tgodaA: Any reason to check this variable ```set_avd_switch_facts``` as this is already inside ```if set_avd_switch_facts``` (line52)

ansible_collections/arista/avd/plugins/module_utils/utils.py
tgodaA: I found one problem that could arise. What if this filter was used after a default filter and default filter is having default value as a string?

Does anyone think this is possible?

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
ClausHolbechArista: Nope. Pep8 violation.
image

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
ClausHolbechArista: No reason anymore. It looked differently earlier. I will clean up.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: I am not sure what you want me to do here. It is a function that works with lists only. It will raise an error if `set` cannot convert the input. It should not be used with other types of input.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
tgodaA: Yes, thank you for clarifying this. I found this article: https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

ansible_collections/arista/avd/plugins/module_utils/utils.py
tgodaA: Agreed.

jonxstill (approved):
Tested a variety of test repos against this change, issues were noted with one of these (a customer repo) that @ClausHolbechArista has now fixed.  LGTM.

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class AristaAvdError(Exception):
+    def __init__(self, message="An Error has occured in an arista.avd plugin"):
+        self.message = message
+        super().__init__(self.message)
+
+
+class AristaAvdMissingVariableError(AristaAvdError):
+    pass
```
tgodaA, line 12: Aren't we throwing any error?

ClausHolbechArista: This is wrapping AristaAvdError, and caught in the calling function of the action plugin, where it raises another exception. As far as I know, this is how you wrap more generic exceptions.

carlbuchmann (changes requested):
- When running `make unit-tests` locally I see errors. 
- I will investigate why CI didn't run these tests.
- I ran against the latest devel branch, and all test are successful. 



```shell
==================================== short test summary info ====================================
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT4]
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[]
FAILED tests/unit/filters/test_list_compress.py::TestListCompressFilter::test_list_compress[LIST_TO_COMPRESS_VALID0]
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[None]
FAILED tests/unit/filters/test_list_compress.py::TestListCompressFilter::test_list_compress[LIST_TO_COMPRESS_VALID1]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Et1-2/3-4/5-6]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID6]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Ethernet1]
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT2]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID10]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID8]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID7]
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT5]
FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT1]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID4]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Ethernet1-2]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[vlan1-3]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[1-3]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID5]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Eth 3-5,7-8]
FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[et2-6,po1-2]
```

ClausHolbechArista:
> * When running `make unit-tests` locally I see errors.
> * I will investigate why CI didn't run these tests.
> * I ran against the latest devel branch, and all test are successful.
> 
> ```shell
> ==================================== short test summary info ====================================
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT4]
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[]
> FAILED tests/unit/filters/test_list_compress.py::TestListCompressFilter::test_list_compress[LIST_TO_COMPRESS_VALID0]
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[None]
> FAILED tests/unit/filters/test_list_compress.py::TestListCompressFilter::test_list_compress[LIST_TO_COMPRESS_VALID1]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Et1-2/3-4/5-6]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID6]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Ethernet1]
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT2]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID10]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID8]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID7]
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT5]
> FAILED tests/unit/filters/test_natural_sort.py::TestNaturalSortFilter::test_natural_sort_invalid[ITEM_TO_NATURAL_SORT1]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID4]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Ethernet1-2]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[vlan1-3]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[1-3]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[RANGE_TO_EXPAND_VALID5]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[Eth 3-5,7-8]
> FAILED tests/unit/filters/test_range_expand.py::TestListCompressFilter::test_range_expand[et2-6,po1-2]
> ```

Issues fixed.

carlbuchmann (approved):
LGTM!

PR #1765, opened 2022-05-05
Feat(eos_designs): Add ability to set mlag port-channel id

### LINKED ISSUE #1764: Ability to set mlag peer link port-channel ID
### Enhancement summary

Give users the ability to deterministically set the mlag peer link port-channel id.  If the key is not set, the default method of port-channel id generation is used.

eos_designs

### Use case example

A chassis switch's mlag interfaces may occur on Ethernet3/47/1. This means the generated port-channel id will be Port-Channel3471, which is not a valid port-channel number.  For this reason, we need to give users the ability to set the port-channel id themselves. 

 I'm suggesting we add a field called `mlag_port_channel_id` which users can use in the node type defaults section or node group defaults section to set the mlag peer link port-channel id.  If the key is not set, the default method of port-channel id generation is used.  So if the `mlag_port_channel_id` key is set to `2000`, the mlag peer link the device uses is Port-Channel2000.  If the `mlag_port_channel_id` key is not present, the mlag peer link is set using the current method.

_No response_

### Additional context

_No response_

### Contributing Guide

jrecchia1029:
## Change Summary

Gives users the ability to set the id of the port-channel used for the mlag peer link.

## Related Issue(s)

Fixes #1764

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Adding a field called `mlag_port_channel_id` which users can use in the node type defaults section or node group defaults section to set the mlag peer link port-channel id. If the key is not set, the default method of port-channel id generation is used.

Add a key called `mlag_port_channel_id` which can be set under node type defaults or node group defaults section.
```
< node_type_key >:
  mlag_port_channel_id: < 1-2000 > | Optional
  node_groups:
    <node_goup>:
      mlag_port_channel_id: < 1-2000 > | Optional
```

## How to test

With an inventory consisting of a pair of 7280s serving as l3leafs and a pair of 7308s serving as a custom node type called splines,  I added the key the splines defaults section only.  MLAG came up on the spline switches, which had the mlag port-channel id manually set, and MLAG also came up on the leaf switches which did NOT have the mlag port-channel id manually set. Both configurations matched the expected output.

### User Checklist

- N/A

### Repository Checklist

- [ x ] My code has been rebased from devel before I start
- [ x ] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [ x ] My change requires a change to the documentation and documentation have been updated accordingly.
- [ x ] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)


ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -363,6 +363,10 @@ defaults <- node_group <- node_group.node <- node
     # MLAG interfaces speed | Optional and depends on mlag_interfaces to be defined
     mlag_interfaces_speed: < interface_speed | forced interface_speed | auto interface_speed >
 
+    # MLAG peer link port-channel id: | Optional; if not set, the mlag port-channel id 
+    # is generated from the digits of the first interface present in 'mlag_interfaces'
+    mlag_port_channel_id: < 1-2000 >
```
gmuloc: The limit `1-2000` is only valid for Chassis - maybe need to indicate this in the ranges option.

7500 switch:
```
s7500(config)#int po ?
  $         list end
  <1-2000>  Lag interface number
```

 vs for fixed switch:

```
sw(config)#int po ?
  $           list end
  <1-999999>  Lag interface number
```

gmuloc (commented):
Thanks a lot for the contribution! it looks indeed useful to go around the limitation you mentioned for modular switches.

Regarding the tests - when adding a knob like this it is better to modify `molecule/eos_designs_unit_test` rather than modifying the specific designs. Could you please move your tests to this directory?

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
jrecchia1029: I did a little digging and it actually looks like this is only a limitation for versions of EOS prior to 4.25.0F where support for port channel ID configuration up to 999,999 was introduced (RFE 408088). I will update the documentation accordingly.

carlbuchmann: in terms of the documentation for such ranges, due to the fact that various EOS versions platforms have different limits and can change over time, it's more practical to simply state that it needs to be an integer? 

```yaml
mlag_port_channel_id: < integer >
```

jrecchia1029: I think that's good. If the user winds up using an invalid port-channel id, the error should be pretty obvious when they go to validate/push the config. I can make the change if we're 100% set on using `< integer>`.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -363,6 +363,10 @@ defaults <- node_group <- node_group.node <- node
     # MLAG interfaces speed | Optional and depends on mlag_interfaces to be defined
     mlag_interfaces_speed: < interface_speed | forced interface_speed | auto interface_speed >
 
+    # MLAG peer link port-channel id: | Optional; if not set, the mlag port-channel id
+    # is generated from the digits of the first interface present in 'mlag_interfaces'
+    mlag_port_channel_id: < 1-2000 > for EOS < 4.25.0F | < 1 - 999999 > for EOS >= 4.25.0F
```
gmuloc: agree with @carlbuchmann  use `< integer >` here and document the limitation in some comment above the parameter

gmuloc (changes requested):

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
ClausHolbechArista: Just a note for general awareness. Even though we can use a higher PO number, we cannot increase the MLAG id above 2000. Not relevant for this feature, but worth mentioning.
Another point on this PR. The whole reason for it, was the problem of PO > 2000 on modulars. Since as found out here >=4.25.0F can support >2000 do we still want it? It could lead to conflicts if the selected PO is the same as something generated dynamically for servers.

ClausHolbechArista: any way of course we can add this knob to offer more control.

jrecchia1029: Yea. I think it's worth keeping just to make things easier if a switch gets shipped to a customer with an EOS version prior to 4.25.0F.  Customers can temporarily use a custom port-channel id number just to verify MLAG comes up properly before they have the opportunity to upgrade.

gmuloc (approved):
LGTM

ClausHolbechArista:
@jrecchia1029 I will rebase and refactor this to work with the new eos_designs_facts module.

PR #1799, opened 2022-05-18
Feat(eos_config_deploy_cvp): support for !vault value in inventory file

### LINKED ISSUE #1781: Support for encrypted individual variables via Ansible Vault in inventory
### Enhancement summary

When using ansible vault encrypted individual variables -  https://docs.ansible.com/ansible/latest/user_guide/vault.html#encrypting-individual-variables-with-ansible-vault in the inventory file, the eos_config_deploy_cvp role is crashing on the serialisation of the inventory:

```
ansible_collections/arista/avd/plugins/modules/inventory_to_container.py\", line 291, in serialize\nAttributeError: 'str' object has no attribute 'items'\n",
```

a little troubleshooting leads to this line: https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py#L431 where the `yaml.safe_load` exception which happens because of the vault syntax is going to be caught by the `except` statement and silently ignored. This will lead to an empty string for `inventory_content` and none of the further functions will catch it (the empty string is Iterable).

This issue is opened as a feature request rather than a bug in order to be able to support encrypted value.

The solution is to use instead: https://github.com/ansible/ansible/blob/2cbfd1e350cbe1ca195d33306b5a9628667ddda8/lib/ansible/parsing/utils/yaml.py#L59 
as described in this SO answer: https://stackoverflow.com/questions/69817464/pyyaml-error-could-not-determine-a-constructor-for-the-tag-vault 

eos_config_deploy_cvp

### Use case example

please detail your use case

Use ansible_password inline encrypted variable inside the inventory

It is possible today to leverage vault using encrypted files rather than inline encrypted passwords cf https://github.com/arista-netdevops-community/avd-evpn-webinar-june-11/tree/master for an example.

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

This PR aims to fix the current silent failure behavior that occurs
when using inline !vault variable in the inventory file.

## Related Issue(s)

Fixes #1781

## Component(s) name

### Current behavior:

When the inventory_to_container.py plugin tries to load such inventory,
the yaml safe_load function fails BUT the plugin catch the exception and
does not raise afterwards (only module.debug). The inventory_content
however is set to "" (empty string) which leads to an error in serialize
later on.

### Commit behavior:

* Introduce a special constructor when parsing via yaml to understand
  the `!vault` construct. This replaces the !vault values by
  AnsibleVaultEncryptedUnicode type in the parsed file which allows for
  the plugins to continue
* Change the behavior if another parsing error to raise an AnsibleError
  with the instruction rather than swallowing the Exception and failing
  later.
* Update the unit tests to use the same constructor

### Notes:

* the author tried to use decrypted vault values when parsing the
  inventory but it turns out to be difficult as the plugin does not
  receive the vault password when call from AVD `eos_config_deploy_cvp`
  and hence it would be require to add some logic to retrieve the vault
  password(s) to be able to decrypt the vault value.
  Given that the vault values are actually not required (as far as the
  author can see) - the decision to simply load the inventory with
  encoded vault values as AnsibleVaultEncryptedUnicode seems a good
  compromise.

## How to test

Run the unit tests

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py
+++ b/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py
@@ -441,7 +441,7 @@ def main():
                 inventory_content = yaml.safe_load(stream)
             except yaml.YAMLError as exc:
                 raise AnsibleValidationError(
-                    f"Failed to parse inventory file, original exception:{exc}"
+                    "Failed to parse inventory file"
                 ) from exc
         result["CVP_TOPOLOGY"] = get_containers(inventory_content=inventory_content,
                                                 parent_container=parent_container,
--- a/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py
+++ b/ansible_collections/arista/avd/plugins/modules/inventory_to_container.py
@@ -97,8 +97,7 @@ import glob
 import os
 import traceback
 from ansible.module_utils.basic import AnsibleModule
-from ansible.errors import AnsibleError
-from ansible.parsing.yaml.objects import AnsibleVaultEncryptedUnicode
+from ansible.module_utils.errors import AnsibleValidationError
 
 TREELIB_IMP_ERR = None
 try:
@@ -439,7 +438,7 @@ def main():
                 inventory_content = yaml.safe_load(stream)
             except yaml.YAMLError as exc:
                 module.debug(exc)
-                raise AnsibleError(
+                raise AnsibleValidationError(
                     f"Failed to parse inventory file, original exception:{exc}"
                 ) from exc
         result["CVP_TOPOLOGY"] = get_containers(inventory_content=inventory_content,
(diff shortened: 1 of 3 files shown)


ansible_collections/arista/avd/plugins/modules/inventory_to_container.py
```diff
@@ -433,7 +438,10 @@ def main():
                 inventory_content = yaml.safe_load(stream)
             except yaml.YAMLError as exc:
                 module.debug(exc)
-        result['CVP_TOPOLOGY'] = get_containers(inventory_content=inventory_content,
+                raise AnsibleValidationError(
+                    f"Failed to parse inventory file, original exception:{exc}"
```
ClausHolbechArista: Do you need to print out the exception? Normally, when raising with `from` it will already print out both exceptions.

gmuloc: I did not want to remove the `module.debug` log that was already there in case it was needed by someone but happy to remove it indeed

gmuloc: Hmm ok - read too fast - I have now removed the `{exc}` in the error message..

ClausHolbechArista (approved):
LGTM

carlbuchmann (approved):
LGTM

PR #1812, opened 2022-05-23
Fix(eos_designs): ipv6_underlay should not apply for l2 switches

ClausHolbechArista:
## Change Summary

`ipv6_underlay` should not apply for l2 switches

## Related Issue(s)

`ipv6_underlay` should not require `loopback_ipv6_pool` on l2-only switches (ex. l2leaf).

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Implemented `switch.ipv6_underlay` set based on checks for `underlay_router` and `ipv6_underlay`.

## How to test

Added molecule scenario for `l2leaf` with `ipv6_underlay: true`.
Verified failure
Applied fix
Verified no failure.
No other molecule changes.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
@@ -869,14 +869,21 @@ class EosDesignsFacts:
         return None
 
     @cached_property
+    def underlay_ipv6(self):
+        if self.underlay_router is True:
+            return get(self._hostvars, "underlay_ipv6")
+        return None
+
+
+    @cached_property
     def loopback_ipv6_pool(self):
-        if get(self._hostvars, "underlay_ipv6") is True:
+        if self.underlay_ipv6 is True:
             return get(self._switch_data_combined, "loopback_ipv6_pool", required=True)
         return None
 
     @cached_property
     def loopback_ipv6_offset(self):
-        if get(self._hostvars, "underlay_ipv6") is True:
+        if self.underlay_ipv6 is True:
             return get(self._switch_data_combined, "loopback_ipv6_offset", default=0)
         return None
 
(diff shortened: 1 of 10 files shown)


ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -868,15 +868,21 @@ def node_sid(self):
                 return self.id + node_sid_base
         return None
 
+    @cached_property
+    def underlay_ipv6(self):
+        if self.underlay_router is True:
+            return get(self._hostvars, "underlay_ipv6")
+        return None
```
gmuloc, line 875: According to documentation, there is a default value for `underlay_ipv6` which is `False` so should we really return `None` here?

ClausHolbechArista: Strictly speaking, this is an internal var which is not necessarily following the input, but I understand your point. The None value will be removed from the switch facts before output, so that is why we have `is arista.avd.defined(true)` in the templates. If we change this to False, we could also simplify the templates to just use `if switch.underlay_ipv6`

onurgashi (approved):

onurgashi (approved):
Tested on a customer repository and it works as expected.

LGTM after the comments from @gmuloc

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
gmuloc: ok I see - so the template will be interpreting the absence of value as `false` in the end as is documented. 
To some extent it means we are never explicitly setting the documented default value just interpreting it implicitly and it may come back to bite us someday but in this specific scenario I think it is quite safe to stay like this.

Another downside may come the day we decide to add type hint in the code though we can get out with `Optional` convenience definition from PEP-0484.

gmuloc (approved):
LGTM

PR #1827, opened 2022-05-27
Feat(eos_designs): Relax requirement of node-specific configuration

ClausHolbechArista:
## Change Summary

Relax requirement of node-specific configuration

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Currently all nodes must be defined under `nodes`. Ex for a l3leaf:
```
l3leaf:
  nodes:
    MY-L3LEAF-123:
      id: 123
      mgmt_ip: 10.20.30.40/24
```
or
```
l3leaf:
  node_groups:
    MY_NODE_GROUP:
      nodes:
        MY-L3LEAF-123:
          id: 123
          mgmt_ip: 10.20.30.40/24
```

With this PR we relax this requirement, so all the node vars can be set on any level. Ex.
```
l3leaf:
  defaults:
    id: 123
    mgmt_ip: 10.20.30.40/24
```

This is useful when the inventory is using in-line jinja or if the inventory is created from external scripts/roles setting host_vars per device.
Ex:
```
l3leaf:
  defaults:
    id: {{ inventory_hostname | split('-') | last }}
    mgmt_ip: {{ lookup('my_ipam', inventory_hostname, 'mgmt_ip') }}
```

## How to test

Moved topology vars defined in hostvars to `defaults` for one host. Failed before this change, but is accepted with no changes to output.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/device.with.dots.in.hostname.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/device.with.dots.in.hostname.yml
@@ -1,9 +1,8 @@
 ---
 type: spine
 spine:
-  nodes:
-    device.with.dots.in.hostname:
-      id: 1
-      loopback_ipv4_pool: 1.2.3.4/24
-      bgp_as: 1234
+  defaults:
+    id: 1
+    loopback_ipv4_pool: 1.2.3.4/24
+    bgp_as: 1234
 mgmt_gateway: 192.168.0.1
--- a/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
@@ -316,13 +316,13 @@ class EosDesignsFacts:
             combined : dict
                 Combined configuration after inheritance from all levels
         """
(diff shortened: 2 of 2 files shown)


gmuloc (commented):
Probably need to change the documentation as well to reflect the new possibility:

https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md?plain=1#L126

Also we could end up with multiple devices with the same `id` which could lead to some similar IPs being generated down the line -not sure if we could raise an error in that case though.

ClausHolbechArista:
> Probably need to change the documentation as well to reflect the new possibility:
> 
> https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md?plain=1#L126
> 
> Also we could end up with multiple devices with the same `id` which could lead to some similar IPs being generated down the line -not sure if we could raise an error in that case though.

Most users would still use it per device, and above the section we state:
```
All node types have the same structure based on `defaults`, `node_group`, `node` and all variables can be defined in any section and support inheritance like this:
```

gmuloc (approved):
LGTM

emilarista (approved):
LGTM

PR #1833, opened 2022-05-27
Feat(plugins): Add the avd ipaddress filter

tgodaA:
## Change Summary

Add arista.avd.ipaddr filter to remove the dependency on ansible.netcommon collection. Currently adding functionality only for the used methods of ansible.netcommon.ipaddr, like ('address', 'prefix', 'net', 'netmask', 'size') etc. similarly for ipmath. This plugin can be optimized later.

## Component(s) name

`arista.avd.plugins`

## How to test

```shell
ansible-test units -v tests/unit/filters/test_ipaddr.py
```

Sanity test
```shell
ansible-test sanity --color yes -v --requirements --docker --python 3.9
```

## Checklist

- [x] Replace ansible.netcommon.ipaddr with arista.avd.ipaddr
- [x] Replace ansible.netcommon.ipmath with arista.avd.ipmath
- [x] **make refresh-facts** should not generate/alter any facts
- [x] **make refresh-facts** should not generate any run-time errors
- [x] Add test cases for all methods
- [x] All test cases must pass
- [x] Remove ansible.netcommon, netaddr dependencies from the requirements

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_validate_function(self, VALIDATE_INPUT, VALIDATE_EXPECTED):
+        resp = f.validate(VALIDATE_INPUT)
+        if resp:
+            resp = str(resp)
+        assert resp == VALIDATE_EXPECTED
+
+    @pytest.mark.parametrize("IP_ADDRESS", IP_ADDRESSES)
+    def test_ipaddr_validate_item(self, IP_ADDRESS):
+        resp = f.ipaddr(IP_ADDRESS)
+        expected_ip = f.validate(IP_ADDRESS)
```
gmuloc, line 85: is it a good test given that `ipaddr` is actually calling `validate` so we are calling twice validate and expecting the output to be the same?

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
```diff
@@ -39,8 +39,8 @@ ztp:
   general:
     subnets:
 {% if ztp_network_summary is arista.avd.defined %}
-      - network: {{ ztp_network_summary | ansible.netcommon.ipaddr('network') }}
```
gmuloc, line 42: if we remove netcommon from everywhere we also need to update the collection.yml and galaxy.yml files

tgodaA:
Will add the tests for ipv4 and ipv6.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,151 @@
...
+INPUT_IPMATH_OFFSET = [1, '10', -1]
+IPMATH_ADDRESSES_EXPECTED = [['192.24.2.2', '::2', '192.168.32.1', 'fe80::101', '192.168.32.2', '192.168.2.0', '192.168.1.35', '192.168.1.35'],
+                             ['192.24.2.11', '::b', '192.168.32.10', 'fe80::10a', '192.168.32.11', '192.168.2.9', '192.168.1.44', '192.168.1.44'],
+                             ['192.24.2.0', '::', '192.168.31.255', 'fe80::ff', '192.168.32.0', '192.168.1.254', '192.168.1.33', '192.168.1.33']
+                             ]
+IPV4_ADDRESSES_VALIDATE_EXPECTED = ['192.24.2.1', '192.168.32.0/24', '192.168.32.1/24', '192.168.1.255/24', '192.168.1.34/28',
+                                    '192.168.1.34/25']
+
+
+def get_by_method(IP_ADDRESS_INPUT_LIST=None, method='validate'):
```
ClausHolbechArista, line 45: I fear that adding all this logic to the tests, will just make it pass more or less all the time.
IMO we should set up specific asserts for the `isinstance` according to our expectations.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,151 @@
...
+        ipv4_ipv6_assert(resp_list, IP_ADDRESS_LIST_EXPECTED)
+
+    @pytest.mark.parametrize("IP_ADDRESS_LIST_INPUT,IP_ADDRESS_LIST_EXPECTED", [(IP_ADDRESSES_INPUT, IP_ADDRESSES_EXPECTED)])
+    def test_ipaddr_network_address_list(self, IP_ADDRESS_LIST_INPUT, IP_ADDRESS_LIST_EXPECTED):
+        resp_list = f.ipaddr(IP_ADDRESS_LIST_INPUT, method='address')
+        assert resp_list == IP_ADDRESS_LIST_EXPECTED
+        assert get_by_method(IP_ADDRESS_LIST_INPUT, 'address') == IP_ADDRESS_LIST_EXPECTED
+        ipv4_ipv6_assert(resp_list, IP_ADDRESS_LIST_EXPECTED)
+
+    @pytest.mark.parametrize("INPUT_LISTS,IP_ADDRESS_LISTS_EXPECTED", [((IP_ADDRESSES_INPUT, DEFAULTATTR_INPUT), IP_ADDRESSES_DEFAULTATTR_EXPECTED)])
```
ClausHolbechArista: you should really add a third input parameter instead of overloading on of them like this.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,151 @@
...
+    def test_ipaddr_defaultattr_offset_list(self, INPUT_LISTS, IP_ADDRESS_LISTS_EXPECTED):
+        IP_ADDRESS_LIST_INPUT = INPUT_LISTS[0]
+        for indx, INPUT_VAL in enumerate(INPUT_LISTS[1]):
+            EXPECTED_OUTPUT = IP_ADDRESS_LISTS_EXPECTED[indx]
+            resp_list = f.ipaddr(IP_ADDRESS_LIST_INPUT, method=INPUT_VAL)
+            print(resp_list, EXPECTED_OUTPUT)
+            assert resp_list == EXPECTED_OUTPUT
+
+    @pytest.mark.parametrize("INPUT_LISTS,IP_ADDRESS_LISTS_EXPECTED", [((IP_ADDRESSES_INPUT, INPUT_IPMATH_OFFSET), IPMATH_ADDRESSES_EXPECTED)])
+    def test_ipaddr_network_address_list(self, INPUT_LISTS, IP_ADDRESS_LISTS_EXPECTED):
```
ClausHolbechArista: you have 3 functions with the same name.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,151 @@
...
+
+    @pytest.mark.parametrize("IP_ADDRESS_LIST_INPUT,IP_ADDRESS_LIST_EXPECTED", [(IP_ADDRESSES_INPUT, IP_ADDRESSES_NET_EXPECTED)])
+    def test_ipaddr_network_list(self, IP_ADDRESS_LIST_INPUT, IP_ADDRESS_LIST_EXPECTED):
+        resp_list = f.ipaddr(IP_ADDRESS_LIST_INPUT, method='net')
+        assert resp_list == IP_ADDRESS_LIST_EXPECTED
+        assert get_by_method(IP_ADDRESS_LIST_INPUT, 'net') == IP_ADDRESS_LIST_EXPECTED
+        ipv4_ipv6_assert(resp_list, IP_ADDRESS_LIST_EXPECTED)
+
+    @pytest.mark.parametrize("IP_ADDRESS_LIST_INPUT,IP_ADDRESS_LIST_EXPECTED", [(IP_ADDRESSES_INPUT, IP_ADDRESSES_SUBNETWORK_EXPECTED)])
+    def test_ipaddr_subnetwork_list(self, IP_ADDRESS_LIST_INPUT, IP_ADDRESS_LIST_EXPECTED):
```
ClausHolbechArista: It is not apparent what we are testing in each scenario. It would be great with a docstring for each test function.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
```diff
@@ -0,0 +1,151 @@
...
+    @pytest.mark.parametrize("INPUT_LISTS,IP_ADDRESS_LISTS_EXPECTED", [((IP_ADDRESSES_INPUT, DEFAULTATTR_INPUT), IP_ADDRESSES_DEFAULTATTR_EXPECTED)])
+    def test_ipaddr_defaultattr_offset_list(self, INPUT_LISTS, IP_ADDRESS_LISTS_EXPECTED):
+        IP_ADDRESS_LIST_INPUT = INPUT_LISTS[0]
+        for indx, INPUT_VAL in enumerate(INPUT_LISTS[1]):
+            EXPECTED_OUTPUT = IP_ADDRESS_LISTS_EXPECTED[indx]
+            resp_list = f.ipaddr(IP_ADDRESS_LIST_INPUT, method=INPUT_VAL)
+            print(resp_list, EXPECTED_OUTPUT)
+            assert resp_list == EXPECTED_OUTPUT
+
+    @pytest.mark.parametrize("INPUT_LISTS,IP_ADDRESS_LISTS_EXPECTED", [((IP_ADDRESSES_INPUT, INPUT_IPMATH_OFFSET), IPMATH_ADDRESSES_EXPECTED)])
```
ClausHolbechArista: as discussed, these inputs don't really need `parametrize`

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
ClausHolbechArista: You could also consider grouping the inputs more around how you expect the outputs. So different inputs expected to return None etc.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: I have been continuously adding and making changes, I will change these.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: Actually, I don't need parametrize here. Just removing it.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: Basically we need to test, f.ipaddr(ip, method) and f.method(ip). Both of them yield different in some cases based on input data-type. For example, a=ipaddress.ip_interface('192.168.0.1/24'). f.ipaddr(a, method) yields correct output but simply passing f.method(a) yields None. So I need a data-type check for the second test.

That's why I was doing tests for both f.ipaddr(ip, method) and f.method(ip), asserting them against the provided expected output. Let me know your thoughts.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: Sure. Not an expert but I will add doc string to explain as much as I can.

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
 
 !!! Note this is not using the same range syntax as EOS for modular or break-out ports. On EOS `et1/1-2/4` gives you `et1/1, et1/2, et1/3, et1/4, et2/1, et2/2, et2/3, et2/4` on a fixed switch, but a different result on a modular switch depending on the module types. In AVD the same range would be `et1-2/1-4`
 
+### IP Address filters
+
+Four filters are currently supported to handle all of the IP Address related calculations.
+
+#### ipaddr filter
+
+The ```arista.avd.ipaddr``` filter provides the capabilities to calculate all IP address related functionality like, `subnet`, `address`, `subnet size`, `network ip` etc. The input can be an IP address or a list of IP addresses and the output will be of similar data-type.
```
gmuloc:
```suggestion
The `arista.avd.ipaddr` filter provides capabilities to manipulate IP addresses. It supports operations like `subnet`, `address`, `subnet size`, `network ip`, etc... The input can be an IP address or a list of IP addresses and the output will be of similar data-type.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+This filter supports both `IPv4` and `IPv6` version types. If no IP version is specified, then the filter calculates for both of the versions. This filter takes two arguments apart from the IP address(es) as input, which are `method` and `version`. The supported versions are `4`, and `6` (integers), which represent the **ipv4** and **ipv6** respectively. The supported `methods` are:
```
gmuloc:
```suggestion
This filter supports both `IPv4` and `IPv6` version types. If no IP version is specified, then the filter calculates for both of the versions. This filter takes two arguments apart from the IP address(es) as input, which are `method` and `version`. The supported versions are `4`, and `6` (integers), which represent the **ipv4** and **ipv6** address families respectively. The supported `methods` are:
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+1) `validate`: Returns the valid IP addresses or Network IPs based on the version type. If no version number is provided then both IPv4 and IPv6 IPs are returned based on the input data-type(string or list). This is the default method in the `ipaddr` filter.
```
gmuloc:
```suggestion
1) `validate`: Returns the valid IP addresses or Network IPs based on the version type. If no version number is provided then both IPv4 and IPv6 IPs are returned based on the input data-type (string or list). This is the default method in the `ipaddr` filter.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', '192.168.0', 'fe08::1', '192.168.1.0/24', '', True, 'test12.com']
+    {{ ip_pool | arista.avd.ipaddr }}
+    Output: ['192.168.0.1', 'fe08::1', '192.168.1.0/24']
+    ```
+
+    The below method is same as above.
```
gmuloc:
```suggestion
    which is equivalent to:
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', '192.168.0', 'fe08::1', '192.168.1.0/24', '', True, 'test12.com']
+    {{ ip_pool | arista.avd.ipaddr('validate') }}
+    Output: ['192.168.0.1', 'fe08::1', '192.168.1.0/24']
+    ```
+
+1) `address`: Returns the valid IP address(es) only.
```
gmuloc:
```suggestion
2) `address`: Returns the valid IP address(es) only.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', '192.168.0', 'fe08::1', '192.168.1.0/24', '', True]
+    {{ ip_pool | arista.avd.ipaddr('address') }}
+    Output: ['192.168.0.1', 'fe08::1']
+    ```
+
+1) `net`: Returns only the Network IPs.
```
gmuloc:
```suggestion
3) `net`: Returns only the Network IPs.
```

gmuloc: Looks like it returns also the `Network/Mask` according to the example

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', '192.168.0', 'fe08::1', '192.168.1.0/24', '', True]
+    {{ ip_pool | arista.avd.ipaddr('net') }}
+    Output: ['192.168.1.0/24']
+    ```
+
+1) `subnet`: Calculates and returns the Network IP of the provided input.
```
gmuloc:
```suggestion
4) `subnet`: Calculates and returns the Network IP of the provided input. If no subnet mask if present, this assumes `/32` for IPv4 and `/128` for IPv6.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', 'fe08::1', '192.168.1.0/24', '192.168.1.20/28']
+    {{ ip_pool | arista.avd.ipaddr('subnet') }}
+    Output: ['192.168.0.1/32', 'fe08::1/128', '192.168.1.0/24', '192.168.1.16/28']
+    ```
+
+1) `netmask`: Calculates the subnet mask of the provided IP address(es).
```
gmuloc:
```suggestion
5) `netmask`: Calculates the subnet mask of the provided IP address(es). If no subnet mask if present, this assumes `/32` for IPv4 and `/128` for IPv6.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', 'fe08::1', '192.168.1.0/24', '192.168.0.70/26']
+    {{ ip_pool | arista.avd.ipaddr('netmask') }}
+    Output: ['255.255.255.255', 'ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff', '255.255.255.0', '255.255.255.192']
+    ```
+
+1) `prefix`: Calculates and returns the Network Prefix(es) for the provided IP address(es). Return value is an integer or list of integers.
```
gmuloc:
```suggestion
6) `prefix`: Calculates and returns the Network Prefix(es) for the provided IP address(es). Return value is an integer or list of integers.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', 'fe08::1', '192.168.1.0/24', '192.168.0.4/26']
+    {{ ip_pool | arista.avd.ipaddr('prefix') }}
+    Output: [32, 128, 24, 26]
+    ```
+
+1) `size`: Calculates and returns the number of IP address(es) possible in the provided pool(s). Return value is an integer or list of integers.
```
gmuloc:
```suggestion
7) `size`: Calculates and returns the number of IP address(es) possible in the provided pool(s). Return value is an integer or list of integers.
```

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', 'fe08::1', '192.168.1.0/24', '192.168.0.4/26']
+    {{ ip_pool | arista.avd.ipaddr('size') }}
+    Output: [1, 1, 256, 64]
+    ```
+
+1) `network`: Calculates and returns the Network IP address(es).
```
gmuloc:
```suggestion
8) `network`: Calculates and returns the Network IP address(es).
```

ansible_collections/arista/avd/plugins/README.md
gmuloc: it looks like it returns the network prefix length cf definition of the network prefix here: https://en.wikipedia.org/wiki/Subnetwork

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -204,6 +210,142 @@ To use this filter:
...
+
+    ```example
+    ip_pool = ['192.168.0.1', 'fe08::1', '192.168.1.0/24', '192.168.0.70/26']
+    {{ ip_pool | arista.avd.ipaddr('network') }}
+    Output: ['192.168.0.1', 'fe08::1', '192.168.1.0', '192.168.1.64']
+    ```
+
+1) `defaultattr`: When the provided input method type doesn't with any of the above, then this method is invoked. Useful when the input method is not actually a method but an IP pool or a number. When the input argument is an integer then IP address corresponding to that number is returned. When an IP pool is provided as an argument then all of the input addresses which are subset of the IP pool are returned (last example).
```
gmuloc:
```suggestion
9) `defaultattr`: When the provided input method type doesn't match with any of the above, then this method is invoked. Useful when the input method is not actually a method but an IP pool or a number. When the input argument is an integer then IP address corresponding to that number is returned. When an IP pool is provided as an argument then all of the input addresses which are subset of the IP pool are returned (last example).
```

gmuloc: Looks like this is not complete - from examples the argument can be an integer or an integer given as a string ?

gmuloc: This method looks to be doing too many things at once with a very different outcome depending on the type or parseable type of the input - do we have use cases for this?

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def net(self, val, version=None):
+        if isinstance(val, int) or '/' not in val:
+            return
+        try:
+            network_address = ipaddress.ip_network(val, strict=True)
+            if isinstance(version, int) and network_address.version != version:
+                return
+            return network_address
+        except ValueError:
```
gmuloc, line 19: silent failure?

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+import ipaddress
+from jinja2.runtime import Undefined
+
+
+class FilterModule(object):
```
gmuloc, line 9: Maybe let's add docstrings to the methods in there

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+    def netmask(self, val, version=None):
+        valid_subnet = self.subnet(val, version)
+        if valid_subnet:
+            return valid_subnet.netmask
+        return
+
+    def prefix(self, val, version=None):
+        valid_subnet = self.subnet(val, version)
+        if valid_subnet:
+            return valid_subnet.prefixlen
```
gmuloc, line 36: so it is prefixlen and not prefix

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+        except ValueError:
+            return
+
+    def defaultattr(self, input_val, ip_address):
+        try:
+            input_val_int = int(input_val)
+            if self.size(ip_address) == 1:
+                return ip_address
+            else:
+                max_ips = self.size(ip_address)
```
gmuloc:
```suggestion
            max_ips = self.size(ip_address)
            if max_ips == 1:
                return ip_address
            else:
```

one less computation

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+                if input_val_int < max_ips:
+                    network_address = ipaddress.ip_network(ip_address, strict=False)
+                    calc_ip = str(network_address.network_address + input_val_int)
+                    prefix = network_address.prefixlen
+                    return f'{calc_ip}/{prefix}'
+                else:
+                    return
+        except ValueError:
```
gmuloc, line 93: so this assume that the ValueError will only comes from `int()` ?

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+    def ipaddr(self, ip_addresses, method='validate', version=None):
+        result = None
+        if isinstance(ip_addresses, (Undefined, bool)) or ip_addresses in (None, '', []):
+            return
+
+        if isinstance(ip_addresses, list):
+            result = []
+            for ip_address in ip_addresses:
+                if single_entry := self.ipaddr(ip_address, method=method, version=version):
+                    result.append(single_entry)
```
gmuloc, line 114: what if the function is given nested lists?

(nice usage of the walrus := )

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            if not isinstance(ip_addresses, (str, int)):
+                ip_addresses = str(ip_addresses)
```
gmuloc, line 117: Hmmm so here what would we expect then something thats not Undefined, not a bool, not a list, not a string and not an int. could be a set / tuple / dict ... Would it make sense to cast these?

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+            try:
+                result = getattr(self, method)(ip_addresses, version=version)
+            except (TypeError, AttributeError):
+                if self.validate(ip_addresses):
+                    result = self.defaultattr(method, str(self.validate(ip_addresses, version)))
```
gmuloc: Could it become to compute only once? (adding version in if so not sure here):
```suggestion
                if validated_addresses := self.validate(ip_addresses, version):
                    result = self.defaultattr(method, str(validated_addresses))
```

ansible_collections/arista/avd/plugins/filter/ipaddr.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def ipmath(self, ip_addresses, add_val):
+        result = None
+        if isinstance(ip_addresses, (Undefined, bool)) or ip_addresses in (None, '', []):
+            return
+        if isinstance(ip_addresses, list):
+            result = []
+            for ip_address in ip_addresses:
+                if self.ipmath(ip_address, add_val):
+                    result.append(self.ipmath(ip_address, add_val))
```
gmuloc:
```suggestion
                if res := self.ipmath(ip_address, add_val):
                    result.append(res)
```

gmuloc (changes requested):

ansible_collections/arista/avd/plugins/README.md
tgodaA: Yeah, but this filters only the Network IPs from the given list.

tgodaA: This is an md hack for numbering. Use 1) always and it will auto calculate the numbers.

gmuloc: Does it work in mkdocs?

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: The reason for this test, is to verify that both f.ipaddr(IP_ADDRESS) and f.validate(IP_ADDRESS) are returning the same. Because we have some extra logic in ipaddr before it calls validate. This is just to make sure all the desired data-types are passing through that extra logic to the validate. Only in such cases both of them return same.

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
tgodaA: Yes, that's in my to-do list.

tgodaA: I mean the checklist of this PR.

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: I did add that before but didn't make sense to me just because we don't wanna overload our filter will all data-type tests. Mostly we use this to pass int, str, list, ipaddress data-types only.

This string conversion is for the input data-type of ipaddress. May be we used ansible.netcommon before using one of our filters.

tgodaA: I can check for all other data-types on line 102.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
gmuloc: That's not what the docstring or the name of the test are saying - it seems to indicate that we are actually testing validate

```
Test single input to ipaddr('validate')
```

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
gmuloc: So as we will be relaxing the requirements I guess it will not require to wait for a major version bump in this case. @ClausHolbechArista  thoughts?

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: Yes, we don't want to raise an error. Here we wanted to return None.

ansible_collections/arista/avd/plugins/README.md
gmuloc: Just tested it and seems to work as well in mkdocs (though the expected syntax is `1.` rather than `1)` and the Github syntax also recommends using `1.` syntax: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#lists

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
tgodaA: Also, we need to ask other customers to use our filters in their custom jinja templates as well.

ClausHolbechArista: Yeah that should be added to release notes similar to what we did with 3.3.0
image

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: It is prefixlen. But that's how it was defined in the ansible.netcommon.ipaddr module. I will change it to prefixlen, because that seems fair to me like you suggested.

ansible_collections/arista/avd/plugins/README.md
tgodaA: Yes, it is. Let me keep this discussion open. The reason to have this prefix is to preserve the old name (from ansible.netcommon.ipaddr)

tgodaA: But I agree, this should be prefixlen. I will change it.

ansible_collections/arista/avd/plugins/README.md
tgodaA: Yes, we need this filter. Because we will be able to calculate the **i**th IP just by ipaddr(i).

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: This shouldn't raise any error because this is a filter to filter only the Network IP among the list of provided ip/pools

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: Good catch. Fixing it now.

ansible_collections/arista/avd/plugins/filter/ipaddr.py
tgodaA: But we won't be giving such input.

ansible_collections/arista/avd/tests/unit/filters/test_ipaddr.py
tgodaA: Yes, the first test is for 'validate' and this one is for ipaddr('validate'). As the first test looks good, I am using f.validate to test ipaddr('validate').

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
tgodaA: I removed the dependency from collections and galaxy. Let me know if they shouldn't be removed.

jonxstill (approved):
LGTM

tgodaA:
+ Closing this PR as this is not the way we wanted to implement this feature.
+ Will work on a new PR instead of working on resolving the existing conflicts.

Issue #1771, opened 2022-05-10
Support for EVPN single-active in eos_designs role


dgonzalez-arista:
### Enhancement summary

We do support EVPN single-active in eos_cli_config_gen #1330, this is to extend the support to eos_designs role.

Today we can configure active-active multihoming connected endpoints as follows:
https://avd.sh/en/latest/roles/eos_designs/doc/connected-endpoints.html#evpn-aa-esi-dual-attached-endpoint-scenario

As discussed with @gmuloc, ideally we would need to extend the model to support EVPN A/S, both at the port-channel level with a new redundancy var and a preference:

```
servers:
  server01:
    rack: RackB
    adapters:
      - endpoint_ports: [ E0, E1 ]
        switch_ports: [ Ethernet10, Ethernet10 ]
        switches: [ DC1-SVC3A, DC1-SVC4A ]
        profile: VM_Servers
        port_channel:
          description: PortChanne1
          redundancy: single-active
          designated_forwarder_preference: [ 100, 0]
          short_esi: 0303:0202:0101
```

This also can be configured in ethernet interfaces, for example to connect to separate CE devices:

```
servers:
  server01:
    rack: RackB
    adapters:
      - endpoint_ports: [ E0, E1 ]
        switch_ports: [ Ethernet10, Ethernet10 ]
        switches: [ DC1-SVC3A, DC1-SVC4A ]
        redundancy: single-active
        designated_forwarder_preference: [ 100, 0]
        short_esi: 0303:0202:0101
```

An example of eos cli configuation would be would be the following:
```
interface Ethernet1/1
   no shutdown
   speed 100g
   switchport
   switchport trunk allowed vlan 700
   switchport mode trunk
   switchport vlan translation 11 700
   spanning-tree portfast
   link tracking group LT_GROUP1 downstream
   evpn ethernet-segment
      identifier 0000:0000:0000:0102:0001
      redundancy single-active
      designated-forwarder election algorithm preference 100
      route-target import 00:00:01:02:00:01
```

@gmuloc proposed something around the following as a data model. Please add anything that i might be missing:
```
< endpoint_2 >:
    rack: RackC
    adapters:
      - speed: < interface_speed | forced interface_speed | auto interface_speed >
        endpoint_ports: [ < interface_name > ]
        switch_ports: [ < switchport_interface > ]
        switches: [ < device > ]
        profile: < port_profile_name >
        ethernet_segment:
          short_esi: < 0000:0000:0000 | auto >
[...]

PR #1780, opened 2022-05-11
Refactor(eos_cli_config_gen): New data model for port-channel ESI variables

emilarista:
## Change Summary

Adds new data model for EVPN ESI variables on port-channel ethernet segments, with fallback to old model.

## Related Issue(s)

Fixes #1702 

## Component(s) name

`arista.avd.eos_cli_config_gen`

## Proposed changes

```yaml
port_channel_interfaces:
  < Port-Channel_interface >:
    evpn_ethernet_segment:
      identifier: < EVPN Ethernet Segment Identifier (Type 1 format) >
      route_target: < EVPN Route Target for ESI with format xx:xx:xx:xx:xx:xx >
```

## How to test
Tested with molecule

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #1834, opened 2022-05-31
Feat(eos_designs): Support for short_esi: auto in port profiles & use of new eos_cli_config_gen data model

jonxstill:
## Change Summary

Allow use of short_esi: auto in port_profiles for connected_endpoints.
Make use of new eos_cli_config_gen evpn_ethernet_segment data model.

## Related Issue(s)

This is part of the fix for #1771 - a 2nd PR will be forthcoming for the remainder of this
This is a replacement for #1772 which I've closed.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
This change builds on the existing short_esi functionality but references adapter_settings instead of adapter - allowing for short_esi to be inherited from a port_profile.

This change also switches both underlay and connected_endpoints port-channels to using the new `evpn_ethernet_segment` key in structured config.

Data model change:
```yaml
port_profiles:
  < port_profile_1 >:
    port_channel:
      short_esi: auto
```

## How to test
New test added in `eos_designs_unit_tests` to inherit from a port_profile that defines short_esi: auto.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/documentation/devices/SITE1-LER1.md
+++ b/ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/documentation/devices/SITE1-LER1.md
@@ -298,6 +298,17 @@ interface Ethernet8
 | Port-Channel8.222 | - | l2dot1q | 222 | False | 222 | - | - | True | - | - | - |
 | Port-Channel8.333 | - | l2dot1q | 434 | False | 333 | - | - | True | - | - | - |
 
+#### EVPN Multihoming
+
+##### EVPN Multihoming Summary
+
+| Interface | Ethernet Segment Identifier | Multihoming Redundancy Mode | Route Target |
+| --------- | --------------------------- | --------------------------- | ------------ |
+| Port-Channel8 | 0000:0000:0303:0202:0101 | all-active | 03:03:02:02:01:01 |
+| Port-Channel8.111 | 0000:0000:0303:0202:0111 | all-active | 03:03:02:02:01:11 |
+| Port-Channel8.222 | 0000:0000:0303:0202:0222 | all-active | 03:03:02:02:02:22 |
+| Port-Channel8.333 | 0000:0000:0303:0202:0333 | all-active | 03:03:02:02:03:33 |
+
 ### Port-Channel Interfaces Device Configuration
 
 ```eos
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/MH_SERVERS.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/MH_SERVERS.yml
@@ -6,6 +6,11 @@ port_profiles:
     link_tracking:
       enabled: true
 
+  Tenant_ESI_Auto:
+    parent_profile: Tenant_X
+    port_channel:
+      short_esi: auto
+
   Tenant_X_LT:
     mode: access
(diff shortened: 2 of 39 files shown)


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/MH_SERVERS.yml
```diff
@@ -49,6 +54,17 @@ servers:
           mode: active
           short_esi: auto
 
+  server04_AUTO_ESI_Profile:
```
gmuloc, line 57: Maybe lets add another test here to make sure that if the profile value is overwritten at the adapter level, the adapter value takes precedence (tested it and it works as expected of course but would be good to be able to catch this if we ever break it)

ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/documentation/devices/SITE1-LER1.md
```diff
@@ -298,6 +298,17 @@ interface Ethernet8
...
+
+##### EVPN Multihoming Summary
+
+| Interface | Ethernet Segment Identifier | Multihoming Redundancy Mode | Route Target |
+| --------- | --------------------------- | --------------------------- | ------------ |
+| Port-Channel8 | 0000:0000:0303:0202:0101 | all-active | 03:03:02:02:01:01 |
+| Port-Channel8.111 | 0000:0000:0303:0202:0111 | all-active | 03:03:02:02:01:11 |
+| Port-Channel8.222 | 0000:0000:0303:0202:0222 | all-active | 03:03:02:02:02:22 |
+| Port-Channel8.333 | 0000:0000:0303:0202:0333 | all-active | 03:03:02:02:03:33 |
+
```
gmuloc, line 311: I was curious to understand why this was not there before given that we are "only" moving around values from one key to another. The template should be feeding the default values but actually it never worked for legacy values before PR #1780 because the list of the way we build the list of interfaces in the template: https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/port-channel-interfaces.j2#L139 

With "legacy values" the `evpn_ethernet_segment` key was never set.

Not sure it is worth spending time "fixing" this (in another PR) given that we will deprecate it later

gmuloc (approved):
LGTM - the additional test I mentioned is optional

ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/documentation/devices/SITE1-LER1.md
jonxstill: That was my feeling exactly - if we're going to deprecate the "legacy" method I don't feel it's worth investing time in now.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/MH_SERVERS.yml
jonxstill: Yep, that makes sense.  Will add that shortly.

jonxstill: Have added a test for this now.

gmuloc (approved):
LGTM

emilarista (approved):
LGTM!

PR #1864, opened 2022-06-09
Feat(eos_designs): Single-Active EVPN Multihoming

### LINKED ISSUE #1771: Support for EVPN single-active in eos_designs role

jonxstill:
## Change Summary

Support single-active EVPN Multihoming.

## Related Issue(s)

Fixes #1771 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
This PR introduces a new data model for ESIs in AVD which affects port-channel and ethernet interfaces.  Subinterfaces of port-channels are unaffected.

```yaml
< connected_endpoints_keys.key >:
  < endpoint_1 >:
    adapters:
        # Settings for all- or single-active EVPN multihoming
      - ethernet_segment:

          # Define a manual short-esi (be careful using this on profiles) or auto-generate an ESI | required
          short_esi: < 0000:0000:0000 | auto >

          # Configure this Ethernet Segment for all-active or single-active forwarding | optional
          # If omitted, Port-Channels use the EOS default of all-active
          # If omitted, Ethernet interfaces are explicitly configured as single-active
          redundancy: < all-active | single-active >

          # Configure DF algorithm and preferences | optional
          #  - auto: Use preference-based algorithm and assign preference based on position of 
          #          the device in the 'switches' list
          #          e.g. assuming a list of three switches, this would assign a preference of
          #          200 to the first switch, 100 to the 2nd and 0 to the third
          #  - preference: Set preference for each switch manually using designated_forwarder_preferences key
          #  - modulus: Use the default modulus-based algorithm
          #  - Alternatively a list of preferences can be provided - [500, 250] for example which will be
          #    mapped according the 'switches' list
          # If omitted, Port-Channels use the EOS default of modulus
          # If omitted, Ethernet interfaces default to the 'auto' mechanism detailed above
          designated_forwarder_algorithm: < "auto" | "modulus" | "preference" >
          # manual preference as described above | required only for preference algorithm
          designated_forwarder_preferences: [ < df_preference_for_each_switch > ]
          # Disable preemption for single-active forwarding when auto/manual DF preference is configured | optional
          dont_preempt: < true | false >
```

If a `short_esi` is defined under the `port_channel` key, it will be overridden by the `short_esi` defined under `ethernet_segment`.  Configuration using the new scheme is identical regardless of whether a port-channel or ethernet interface is being configured.  Port-channels still require setting a `port_channel` key and `mode` subkey as before.

This model is designed to assume sensible defaults, allowing minimal configuration to be set in most cases, but always allowing for this behaviour to be overridden.

## How to test
* A number of new molecule test cases have been added to the MH* leaf switches.
* A separate AVD inventory was used to test generated configuration in vEOS-lab.

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -47,6 +47,38 @@ ethernet_interfaces:
...
+{%                             do adapter_settings.ethernet_segment.update({ 'short_esi': short_esi_hash }) %}
+{%                         endif %}
+{%                         if adapter_settings.ethernet_segment.short_esi.split(':') | length == 3 %}
+    evpn_ethernet_segment:
+      identifier: {{ adapter_settings.ethernet_segment.short_esi | arista.avd.generate_esi(evpn_short_esi_prefix) }}
+{%                             if adapter_settings.ethernet_segment.redundancy is arista.avd.defined %}
+      redundancy: {{ adapter_settings.ethernet_segment.redundancy }}
+{%                             else %}
+      redundancy: single-active
+{%                             endif %}
```
gmuloc: I feel this would be equivalent:
```suggestion
redundancy: {{ adapter_settings.ethernet_segment.redundancy | arista.avd.default("single-active") }}
```

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -47,6 +47,38 @@ ethernet_interfaces:
...
+        algorithm: modulus
+{%                             elif adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined('auto') or
+                                    adapter_settings.ethernet_segment.designated_forwarder_preferences is not arista.avd.defined %}
+{%                                 set auto_preferences = range(((adapter_settings.switches | length) * 100) - 100, 0, -100) %}
+        algorithm: preference
+        preference_value: {{ auto_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
+{%                             else %}
+        algorithm: preference
+        preference_value: {{ adapter_settings.ethernet_segment.designated_forwarder_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
+{%                             endif %}
```
gmuloc: To change a bit the logic around + adding a check that preferences are a list in the last case (maybe we can even raise a warning if it is not a list?)

PS: this is typed inside github desktop so not 100% sure about indent..

```suggestion
      designated_forwarder_election:
{%                             if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined('modulus') %}
        algorithm: modulus
{%                             else %} 
        algorithm: preference
{%                                 if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined('auto') or
                                      adapter_settings.ethernet_segment.designated_forwarder_preferences is not arista.avd.defined %}
{%                                     set auto_preferences = range(((adapter_settings.switches | length) * 100) - 100, 0, -100) %}
        preference_value: {{ auto_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
{%                                 elif adapter_settings.ethernet_segment.designated_forwarder_preferences | arista.avd.defined(var_type='list) %}
        preference_value: {{ adapter_settings.ethernet_segment.designated_forwarder_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
{%                                 endif %}
{%                             endif %}
```

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
```diff
@@ -76,21 +76,37 @@ port_channel_interfaces:
...
     evpn_ethernet_segment:
-      identifier: {{ adapter_settings.port_channel.short_esi | arista.avd.generate_esi(evpn_short_esi_prefix) }}
-      route_target: {{ adapter_settings.port_channel.short_esi | arista.avd.generate_route_target }}
+      identifier: {{ adapter_settings.ethernet_segment.short_esi | arista.avd.generate_esi(evpn_short_esi_prefix) }}
+{%                                     if adapter_settings.ethernet_segment.redundancy is arista.avd.defined %}
+      redundancy: {{ adapter_settings.ethernet_segment.redundancy }}
+{%                                     endif %}
+{%                                     if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined %}
+      designated_forwarder_election:
+        algorithm: preference
```
gmuloc: So from what I read it may be on purpose but I is it not possible to configure `modulus` here? (or do we rely on the fact that by default it is modulus maybe?)

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
jonxstill: Agreed; fixed in latest commit.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
jonxstill: This also makes sense - I just didn't see a nice way to do it; thanks for the var_type tip!  Incorporated in latest commit.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
jonxstill: Yes, I had a mental note to add that into port-channel interfaces; wasn't a big deal as the EOS default is modulus, but it's nice to have things consistent between the two.

jonxstill:
Added support for`dont_preempt`.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/MH-LEAF1A.cfg
```diff
@@ -155,6 +207,86 @@ interface Ethernet14
...
+   switchport mode trunk
+   switchport
+   evpn ethernet-segment
+      identifier 0000:0000:885b:86cc:8bac
+      redundancy single-active
+      designated-forwarder election algorithm preference 500
+      route-target import 88:5b:86:cc:8b:ac
+!
+interface Ethernet20
+   description server11_Single_Active_Port_Channel_Manual_DF_Preempt_Eth1
```
gmuloc: so here it is fun that the description mentions "Preempt" when the knob is actually "dont-preempt" - I understand why (as preempt is the default) but it could feel confusing in the long run so probably worth adding extra 4th characters so that it makes sense in the future :)

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
```diff
@@ -47,6 +47,40 @@ ethernet_interfaces:
...
+        preference_value: {{ auto_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default('0') }}
+{%                                 elif adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined(var_type='list') %}
+        preference_value: {{ adapter_settings.ethernet_segment.designated_forwarder_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default('0') }}
+{%                                 endif %}
+{%                             endif %}
+{%                             if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined and
+                                  adapter_settings.ethernet_segment.designated_forwarder_preferences is not arista.avd.defined('modulus') and
+                                  adapter_settings.ethernet_segment.dont_preempt is arista.avd.defined(true) %}
+        dont_preempt: true
+{%                             endif %}
```
gmuloc: if you move the dont_preempt block in the previous `if / else` block you can already be sure it is not `modulus` and so simplify the if condition like this:

```suggestion
{%                                 if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined and
                                      adapter_settings.ethernet_segment.dont_preempt is arista.avd.defined(true) %}
        dont_preempt: true
{%.                                endif %}
{%                             endif %}
```

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
```diff
@@ -76,21 +76,45 @@ port_channel_interfaces:
...
+{%                                         elif adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined(var_type='list') %}
+        algorithm: preference
+        preference_value: {{ adapter_settings.ethernet_segment.designated_forwarder_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
+{%                                         endif %}
+{%                                         if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined and
+                                              adapter_settings.ethernet_segment.designated_forwarder_preferences is not arista.avd.defined('modulus') and
+                                              adapter_settings.ethernet_segment.dont_preempt is arista.avd.defined(true) %}
+        dont_preempt: true
+{%                                         endif %}
+{%                                     endif %}
```
gmuloc: Same as for `ethernet-interface`.

line 97 you have already checked if the `designated_forwarder_precedence` is defined. so we could rewrite it like this (and I feel it could maybe even be simplified further but I am not finding how right now):

```suggestion
{%                                     if adapter_settings.ethernet_segment.redundancy is arista.avd.defined %}
      redundancy: {{ adapter_settings.ethernet_segment.redundancy }}
{%                                     endif %}
{%                                     if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined %}
      designated_forwarder_election:
{%                                         if adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined('modulus') %}
        algorithm: modulus
{%                                         elif adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined('auto') %}
{%                                             set auto_preferences = range(((adapter_settings.switches | length) * 100) - 100, 0, -100) %}
        algorithm: preference
        preference_value: {{ auto_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
{%                                             if adapter_settings.ethernet_segment.dont_preempt is arista.avd.defined(true) %}
        dont_preempt: true
{%                                             endif %}
{%                                         elif adapter_settings.ethernet_segment.designated_forwarder_preferences is arista.avd.defined(var_type='list') %}
        algorithm: preference
        preference_value: {{ adapter_settings.ethernet_segment.designated_forwarder_preferences[adapter_settings.switches.index(inventory_hostname)] | arista.avd.default("0") }}
{%                                             if adapter_settings.ethernet_segment.dont_preempt is arista.avd.defined(true) %}
        dont_preempt: true
{%                                             endif %}
{%                                         endif %}
{%                                     endif %}
```

ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
```diff
@@ -197,6 +203,29 @@ port_profiles:
...
+          redundancy: < all-active | single-active >
+
+          # Configure DF algorithm and preferences | optional
+          #  - auto: Use preference-based algorithm and assign preference based on position of device in the 'switches' list
+          #          e.g. assuming a list of three switches, this would assign a preference of 200 to the first switch, 100 to the 2nd and 0 to the third
+          #  - modulus: Use the default modulus-based algorithm
+          #  - Alternatively a list of preferences can be provided - [500, 250] for example which will be mapped according the 'switches' list
+          # If omitted, Port-Channels use the EOS default of modulus
+          # If omitted, Ethernet interfaces default to the 'auto' mechanism detailed above
+          designated_forwarder_preferences: < auto | modulus | [ <df preference for each switch> ] >
```
ClausHolbechArista: Let's try to avoid using different value types for the same key.
I suggest:
```yaml
# If omitted, Port-Channels use the EOS default of modulus
# If omitted, Ethernet interfaces default to the 'auto' mechanism detailed above
designed_forwarder_algorithm: < "auto" | "modulus" >
designated_forwarder_preferences: [ < df_preference_for_each_switch > ]
```

jonxstill: Agreed; that will probably be clearer.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
jonxstill: Thanks for the suggestions.  I've refactored the logic because of @ClausHolbechArista 's schema recommendations and tried to get down to fewer levels of nesting and compound conditionals.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/MH-LEAF1A.cfg
jonxstill: Fixed :)

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
```diff
@@ -76,21 +76,47 @@ port_channel_interfaces:
     mtu: {{ adapter_settings.mtu }}
 {%                         endif %}
 {%                         if adapter_settings.switches | unique | list | length > 1 %}
-{%                             if adapter_settings.port_channel.short_esi is arista.avd.defined %}
-{%                                 if adapter_settings.port_channel.short_esi | lower == 'auto' %}
+{%                             if adapter_settings.port_channel.short_esi is arista.avd.defined and adapter_settings.ethernet_segment.short_esi is not arista.avd.defined %}
+{%                                 do adapter_settings.update({ 'ethernet_segment': { 'short_esi': adapter_settings.port_channel.short_esi }  }) %}
```
ClausHolbechArista: This update will overwrite everything else defined under `ethernet_segment`. Is that on purpose?
I would prefer to set a `short_esi` variable using the `arista.avd.default` filter and then reference/overwrite that throughout the template.

ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
jonxstill: These changes have now been made.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/ethernet-interfaces.j2
jonxstill: As below, I've now refactored this.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
jonxstill: This was very much deliberate.  The old mechanism on port-channel interfaces only ever let you configure all-active, with no other knobs.  If you want to do anything other than that, use the new mechanism.  I'd rather keep it as a binary old/new rather than mix the two.

ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
```diff
@@ -76,13 +76,21 @@ port_profiles:
     port_channel:
       description: < port_channel_description >
       mode: < "active" | "passive" | "on" >
-      # Allocates an automatic short_esi to all ports using this profile
-      # Please see the notes under "EVPN A/A ESI dual-attached endpoint scenario" before setting short_esi: auto.
-      short_esi: auto
+      # Please see the notes under "EVPN A/A ESI dual- and single-attached endpoint scenarios" before setting short_esi: auto
+      # Also note that defining short_esi under ethernet_segment overrides short_esi defined at the port_channel level
+      short_esi: < 0000:0000:0000 | auto >
```
gmuloc: just some interrogation would it be better to use `xxxx:xxxx:xxxx` instead of `0000:0000:0000` to make it clear about the format? in `eos_cli_config_gen` the help states: `identifier: < EVPN Ethernet Segment Identifier (Type 1 format) >` but not sure it would be rendering nicely here

gmuloc (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/doc/connected-endpoints.md
jonxstill: Thanks, fixed.

ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2
jonxstill: Have made this change - thanks for the feedback, makes it a lot more readable.

ClausHolbechArista (approved):
LGTM

PR #1868, opened 2022-06-10
Feat(eos_designs): Support for computing SNMP v3 engineid and hashes

### LINKED ISSUE #1846: (eos_design): auto-generate SNMP local engined_id to be able to compute encrypted keys for SNMPv3
### Enhancement summary

The current situation is that it is not possible to hide completely the SNMPv3 password inside an AVD repo relying on eos_designs without removing the `intended` folder for Github (Which is possible of course but can lead to other operational difficulties as the changes to intended configuration are not visible anymore in the commits).

SNMPv3 passphrases (auth and priv) can be encrypted but they are dependent on a local value which is the local engine_id as defined in RFCs.

By default the local engine_id for an Arista switch is some function of the system Mac - however in AVD we are not aware of the system Macs (contrary to CVP for instance).

The idea around this enhancement is the following:
* enables an AVD user to set a knob to allow autogeneration of the local engine_id based on a hash of say (happy to debate this): hostname + mgmt_ip
* leverage the auto generated engine_id to encrypt locally using an ansible fiter the SNMPv3 passphrases as they would be on the EOS switch (the code to do this is available inside the EOS switch bash prompt)
* This would imply replicating to `snmp_settings` of the `eos_designs` some of the `eos_cli_config_gen` keys to make them available to be added to the `structured_configuration`

eos_designs, plugins

### Use case example

With these the usage would then be:
* use ansible vault to encrypt a file containing centralized passphrases for SNMPv3 user passphrases (both auth and priv)
* in `eos_designs`, enable the knob to autogenerate the local engine_id + another knob to required to encrypt the passphrases for v3 user
* the intended structured_configuration and configuration would then contain the SNMPv3 command with the localized encrypted passphrases, ready to be pushed toward the devices.

a **caveat** is that everytime the local engine_id is changed, the localized encrypted passphrases need to be udpated, so using this on brownfield devices would require to update the configuration for pre-existing SNMPv3 users.

cf description above

* removing intended folder from git
* accepting that the keys are in clear in the config
* for each device go to the switch to get back the system Mac and compute the localized keys and insert them in AVD (works for a small number of devices)

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

* add two knobs in eos_designs
    * one to allow to auto-generate local engineID for SNMP
    * one to auto compute the localized auth and priv keys for v3 users based on the previously generated engineID 
* add the users definition to `eos_designs` `snmp_settings` knob. 
* Add an Ansible filter to compute the localized keys
* Add pytest tests for the filter
* add molecule tests

With these changes, it is now possible using a vault file to completely hide the SNMP passwords.

**Note**: because the `snmp_server` knob in `eos_cli_config_gen` is one block, this PR may bring conflicting behavior to enable additional keys. Probably need to add some tests there.

## Related Issue(s)

Fixes #1846

## Component(s) name

`arista.avd.eos_designs`
`arista.avd.plugins`

## Proposed changes

cf Change Summary

Adding the following to `snmp_settings`

```yaml
 snmp_settings:
   contact: < contact_info >
   location: < boolean | default -> false > # Formatted as: {{ fabric_name }} {{ dc_name }} {{ pod_name }} {{ switch_rack }} {{ inventory_hostname }}
   # Generate a local engineId for SNMP by hashing via SHA1 the string
   # generated via the concatenation of the hostname plus the management IP.
   # {{ inventory_hostname }} + {{ switch.mgmt_ip }}
   auto_compute_local_engineid: < boolean | default -> false >
   # Requires auto_compute_local_engineid to be `true`
   # if enabled, the SNMPv3 passphrases for auth and priv are transfromed
   # using RFC 2574, matching the value they would take in EOS cli
   # the algorithm requires a local engineId which is unknown to AVD
   # hence the necessity to generate one beforehand.
   compute_v3_localized_key: < boolean | default -> false >
   users:
     - name: < username >
       group: < group >
       version: < v1 | v2c | v3 >
       auth: < md5 | sha | sha256 | sha384 | sha512 > # optional
       auth_passphrase: < clear_passphrase >          # requires auth, recommended to use vault
       priv: < des | aes | aes192 | aes256 >          # optional
       priv_passphrase: < clear_pasphrase >           # requires priv, recommended to use vault
 ```

## How to test

* Filter:  
    ```
    # pwd: /ansible_collections/arista/avd
    ansible-test units --venv   --python <YOUR_VERSION> tests/unit/filters/test_snmp_hash.py
    ```
* eos_designs changes: 
    ```bash
    # pwd: ansible_collections/arista/avd/molecule
    make converge MOLECULE=eos_designs_unit_tests  ANSIBLE_OPTIONS="--limit=mgmt_interface_default"`
    ```

### Repository Checklist

### CODE REVIEW

File: ansible_collections/arista/avd/roles/eos_designs/templates/base/snmp-server.j2 [hunk 1]
```diff
@@ -1,6 +1,11 @@
 {# snmp settings #}
 snmp_server:
 {% if snmp_settings is arista.avd.defined %}
+{%     if snmp_settings.auto_compute_local_engineid is arista.avd.defined(true) %}
+{%         set local_engine_id = (inventory_hostname ~ switch.mgmt_ip) | hash('sha1') %}
+    engineid:
+    local: {{ local_engine_id }}
```
ClausHolbechArista: something looks off here. Missing indentation?

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
```diff
@@ -72,6 +72,24 @@ name_servers:
 snmp_settings:
   contact: < contact_info >
   location: < boolean | default -> false > # Formatted as: {{ fabric_name }} {{ dc_name }} {{ pod_name }} {{ switch_rack }} {{ inventory_hostname }}
+  # Generate a local engineId for SNMP by hashing via SHA1 the string
+  # generated via the concatenation of the hostname plus the management IP.
+  # {{ inventory_hostname }} + {{ switch.mgmt_ip }}
+  auto_compute_local_engineid: < boolean | default -> false >
```
ClausHolbechArista: Please align `auto_compute_.....` vs. `compute_.....`

ansible_collections/arista/avd/roles/eos_designs/templates/base/snmp-server.j2
```diff
@@ -18,4 +23,40 @@ snmp_server:
...
+      localized:  {{ local_engine_id }}
+{%             endif %}
+{%             if user.version is arista.avd.defined("v3")
+                and user.auth is arista.avd.defined
+                and user.auth_passphrase is arista.avd.defined %}
+      auth: {{ user.auth }}
+{%                 if snmp_settings.auto_compute_local_engineid is arista.avd.defined(true)
+                      and snmp_settings.compute_v3_user_localized_key is arista.avd.defined(true) %}
+{%                     set hash_filter = {"passphrase": user.auth_passphrase, "auth": user.auth, "engine_id": local_engine_id} %}
+      auth_passphrase: {{ hash_filter | hash_passphrase }}
```
ClausHolbechArista: please use FQCN ex. `arista.avd.hash_passphrase`. Same elsewhere. Also make sure if you are adding new requirements.

ansible_collections/arista/avd/plugins/filter/snmp_hash.py
```diff
@@ -0,0 +1,128 @@
...
+            localized_key = localized_key[: PRIV_KEY_LENGTH[priv_type] // 4]
+        except KeyError as error:
+            raise AnsibleFilterError(
+                f"{priv_type} is not a valid Priv algorithm for SNMPv3"
+            ) from error
+    return localized_key
+
+
+class FilterModule(object):
+    def hash_passphrase(self, input_dict):
```
ClausHolbechArista: Please move this outside the class, so we can import it from other code bits.

ClausHolbechArista (commented):
Nice feature. A few requests.

ansible_collections/arista/avd/roles/eos_designs/templates/base/snmp-server.j2
gmuloc: I am not sure I understand this part 
> Also make sure if you are adding new requirements. 

What do you mean?

ClausHolbechArista: If your new filter requires some extra libraries or if you require other ansible colections. It was not clear how much of this was from your new filter, and if something is ansible builtin.

ansible_collections/arista/avd/roles/eos_designs/templates/base/snmp-server.j2
```diff
@@ -1,6 +1,11 @@
 {# snmp settings #}
 snmp_server:
 {% if snmp_settings is arista.avd.defined %}
+{%     if snmp_settings.compute_local_engineid is arista.avd.defined(true) %}
+{%         set local_engine_id = (inventory_hostname ~ switch.mgmt_ip) | hash('sha1') %}
```
ClausHolbechArista: Please use `FQCN` on `hash`

gmuloc: This is this built-in one: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#hashing-and-encrypting-strings-and-passwords 
we are using it in other places without FQCN so probably need to open an issue:
https://github.com/aristanetworks/ansible-avd/blob/2252eb1c6c598a81f9f4b4ca1b21a9c7abab2f04/ansible_collections/arista/avd/roles/eos_designs/templates/connected_endpoints/port-channel-interfaces.j2#L84

ansible_collections/arista/avd/roles/eos_designs/templates/base/snmp-server.j2
gmuloc: so `hash_passphrase` is the new filter and `hash` is the ansible builtin one nothing else present. Once I figure out the FQCN for the builtin filter I guess we are good for this one?

ClausHolbechArista (approved):
LGTM

emilarista (approved):
LGTM!

PR #1879, opened 2022-06-13
Fix(eos_designs): Correct pruning of vlans when no vlans are allowed

### LINKED ISSUE #1822: Vlan pruning on l2leaf uplinks does not work when no vlans are allowed
### Issue Summary

In the case of no vlans defined on an `l2leaf`, the uplink will be configured without the `switchport trunk allowed vlan` command on the `l2leaf` as well as the parent. This means that _all_ vlans will be allowed - contrary to the intent of pruning vlans.

`arista.avd.eos_designs`

### How do you run AVD ?

_No response_

### Steps to reproduce

```shell
define an `l2leaf` with no vlans (ex. filter.tags: [])
check uplink config.

There is an example in molecule scenario `evpn_underlay_isis_overlay_ibgp`, where no vlans are defined on the l2leafs, but uplinks are wide open.
```

### Relevant log output

_No response_

### Contributing Guide

ClausHolbechArista:
## Related Issue(s)

Fixes #1822

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

The current pruning logic works down to 1 vlan, but if no vlans are allowed, it does not print `switchport trunk allowed vlan none`. Instead it prints skips the allow command completely, which in EOS means to allow all vlans.

This fix catches the issue with no vlans allowed, between AVD controlled switches, where we know the list of defined vlans of both ends. This will only change the allowed vlans, if the child switch (`l2leaf`) has no vlans defined.

## How to test

Since many of our molecule scenarios already contain L2 switches with no vlans, we can see this problem and fix working in existing molecule scenarios.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/evpn_underlay_isis_overlay_ibgp/documentation/devices/DC1-L2LEAF1A.md
+++ b/ansible_collections/arista/avd/molecule/evpn_underlay_isis_overlay_ibgp/documentation/devices/DC1-L2LEAF1A.md
@@ -211,8 +211,8 @@ vlan internal order ascending range 1006 1199
 
 | Interface | Description | Mode | VLANs | Native VLAN | Trunk Group | Channel-Group |
 | --------- | ----------- | ---- | ----- | ----------- | ----------- | ------------- |
-| Ethernet1 | DC1-LEAF2A_Ethernet7 | *trunk | *- | *- | *- | 1 |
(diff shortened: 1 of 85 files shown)


carlbuchmann (approved):
LGTM!

tgodaA (approved):
LGTM

PR #1883, opened 2022-06-14
Feat(eos_config_deploy_cvp): Updated cv_collection version to v3 by default

### LINKED ISSUE #1796: Update version to v3 in eos_config_deploy_cvp role
### Enhancement summary

`cv_collection` version needs to be updated to `v3` in main.yml in `eos_config_deploy_cvp/defaults/main.yml` 

eos_config_deploy_cvp

### Use case example

please detail your use case

change line 
`cv_collection: v1` to 
`cv_collection: v3` 
in `https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/defaults/main.yml`

_No response_

### Additional context

_No response_

### Contributing Guide

sugetha24:
## Change Summary
Updated `cv_collection: v3` variable to use v3 cv collection by default

## Related Issue(s)

Fixes #1796

## Component(s) name

## Proposed changes
cv_collection: v1 to
cv_collection: v3
in https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/defaults/main.yml

## How to test
`molecule converge --scenario-name eos_config_deploy_cvp`
Also, manually tested the role.

### User Checklist
- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_config_deploy_cvp/intended/structured_configs/cvp/cv_server.yml
```diff
@@ -1,95 +1,71 @@
 ---
 CVP_DEVICES:
-  DC1-BL1A:
-    name: DC1-BL1A
+  - fqdn: DC1-BL1A
```
gmuloc: is this change from `name` to `fqdn` expected? it does not look like a fqdn here so would this mean that users now need to use fqdn somewhere?

tgodaA: This is expected. And have been using this already but the default values are now set to v3. I believe v1 is deprecated/will be deprecated soon.

ClausHolbechArista: but `name` != `fqdn`.

tgodaA: Yes, but when I used v3 in my playbook. The config is generated in the same format. In v1 I could see the **name** as key. Not sure if this was the intended design, but it is working for me.

noredistribution: v3 uses https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2

v1 uses:
 https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices.j2
 
 so we should have `fqdn` with v3

gmuloc (commented):
BTW this PR for now does not address the README - it only changes the molecule 
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/README.md?plain=1#L92

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/README.md
```diff
@@ -89,7 +89,7 @@ For complete list of authentication options available with Cloudvision Ansible c
 - __`state`__: `present` / `absent`. Support creation or cleanup topology on CV server. Default: `present`.
 - __`execute_tasks`__:  `true` / `false`. Support automatically excuting pending tasks. Default: `false`.
 - __`cvp_configlets`__: Structure to add additional configlets to those automatically generated by AVD roles.
-- __`cv_collection`__: Version of Cloudvision collection to use. Can be `v1` or `v3`. Default is `v1`.
+- __`cv_collection`__: Version of Cloudvision collection to use.
```
tgodaA:
```suggestion
- __`cv_collection`__: Version of Cloudvision collection to use. Default is `v3`.
```

gmuloc:
Adding what I posted on google chat here:

So I did not have time to conduct extended testing - I was questioning in my first comment if using “non fqdn” in the fqdn field would cause an issue. After some troubleshooting there is cuurently no way to tell ansible-avd to call the ansible-cvdptask(s) with a different search_key so by default we end up using host name as search_key (as per ansible-cvp code) so I don’t know the implications of the pr 

From my initial it “looks to be working” except that we end up with “non valid fqdn” in the fqdn fields. For now it seems it does not have any impact (given that we will always use hostname as search key) but I think it could get some when we modify ansible-avd role to use different search_key in the case where the fqdn is different from the host name (I.e. there is a dns domain configured on the device)

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/main.yml
```diff
@@ -1,4 +1,16 @@
 ---
+- name: "Collection {{ ansible_collection_name }} version {{ version }}{{ ('(git ' ~ git_tag ~ ')') if git_tag }} loaded from {{ collection_path }}"
+  tags: [always, avd_version]
+  set_fact:
+    avd_collection_version: version
+  vars:
+    versions: "{{ lookup('pipe', 'ansible-galaxy collection list --format yaml ' ~ ansible_collection_name) | from_yaml }}"
```
tgodaA:
```suggestion
    versions: "{{ lookup('pipe', 'ansible-galaxy collection list -p ' ~ collection_path ~ ' --format yaml ' ~ ansible_collection_name) | from_yaml }}"
```

tgodaA:
This branch has conflicts. Needs a rebase, I couldn't do it with github **Update branch**

tgodaA:
So in this PR we are removing both V1 and V3. But is there a reason to still keep the v1 and v3 directories. I don't see any type of call from the actual main.yml.

If so we can delete the v1 and v3 directories as they still have some main.yml files.

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -11,7 +11,11 @@ CVP_DEVICES:
 {% for device in groups[container_root] | arista.avd.natural_sort %}
 {%     if device | arista.avd.is_in_filter(local_var.device_filter) %}
 {%         if hostvars[device]['is_deployed'] is arista.avd.defined(true) or hostvars[device]['is_deployed'] is not arista.avd.defined %}
-  - fqdn: {{ device }}
+{%             if cv_collection is arista.avd.defined("v3") %}
+{#                 This if block is a workaround to keep old v3 behavior of splitting hostname on "." #}
+{%                 set device = device | split('.') | first %}
+{%             endif %}
+  - hostname: {{ device }}
```
tgodaA, line 18: Did we mention anywhere that we must provide only hostnames but not the full qualified domain names?

Also, we need to specify both old versions are supported.

I don't exactly know how this fqdn, hostname, <hostname> things are working for old versions as well as for the latest.

noredistribution: we never used fqdns in reality, it was always hostname in both v1 and v3 (on ansible-cvp side), the key in v3 was just named incorrectly, or more correctly said whatever we input in inventory/host_vars as the device identifier we consider that as the hostname of the device regardless what it is

below is my understanding:

## behavior in v1
in v1 this is how the var looks like:

```yaml
CVP_DEVICES:
 tp-avd-leaf1:
   name: tp-avd-leaf1
   parentContainerName: TP_LEAF1
   configlets:
     - tp-avd_tp-avd-leaf1
   imageBundle: []
```

the `name` key there is populated from the inventory/host_vars in the jinja template:

https://github.com/aristanetworks/ansible-avd/blob/releases/v3.5.x/ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices.j2#L13-L14

which is then sent by the v1 tasks and when it lands on ansible-cvp side where we use that as the `hostname` of the devices
few refs:
https://github.com/aristanetworks/ansible-cvp/blob/devel/ansible_collections/arista/cvp/plugins/modules/cv_device.py#L408
https://github.com/aristanetworks/ansible-cvp/blob/169074fa8993aedfc0b45b1abdd74d0db90737d4/ansible_collections/arista/cvp/plugins/modules/cv_device.py#L473
https://github.com/aristanetworks/ansible-cvp/blob/devel/ansible_collections/arista/cvp/plugins/modules/cv_device.py#L483


## behavior in v3

in v3 we have had this model:

```yaml
CVP_DEVICES:
 - fqdn: tp-avd-leaf1
   parentContainerName: TP_LEAF1
   configlets:
     - tp-avd_tp-avd-leaf1
```

however `fqdn` there is not really fqdn and in fact we set the hostname from that key (in case the user really specified an fqdn, we split it by the dots and return the first element)

https://github.com/aristanetworks/ansible-cvp/blob/169074fa8993aedfc0b45b1abdd74d0db90737d4/ansible_collections/arista/cvp/plugins/module_utils/device_tools.py#L75

with that said the v3 schema has `fqdn` or `serialNumber` keys as a must, so we cannot use `hostname` key in the j2 template today, we should set that back to `fqdn` and then update ansible-cvp later to allow `hostname` key as well
ref: https://github.com/aristanetworks/ansible-cvp/blob/devel/ansible_collections/arista/cvp/plugins/module_utils/resources/schemas/v3.py#L95-L108

I tried it as well and if we leave `hostname` the playbook fails with:

```
fatal: [cv_server]: FAILED! => changed=false
  msg: |-
    Error, your input is not valid against current schema:
     {'hostname': 'tp-avd-leaf1', 'parentContainerName': 'TP_LEAF1', 'configlets': ['tp-avd_tp-avd-leaf1']}
```

noredistribution (commented):
for ansible_collections/arista/avd/roles/eos_config_deploy_cvp/tasks/absent.yml

we could do something like this:

```
- name: "Load CVP device information for {{ inventory_hostname }}"
  tags: [build, provision, online]
  include_vars: "{{ structured_cvp_dir }}/{{ inventory_hostname }}.yml"

- name: "Factory resetting devices on {{ inventory_hostname }}"
  tags: [reset]
  arista.cvp.cv_device_v3:
    devices: "{{ CVP_DEVICES }}"
    state: factory_reset
  register: CVP_DEVICES_RESULTS

- name: "Execute pending tasks on {{ inventory_hostname }}"
  tags: [reset]
  arista.cvp.cv_task_v3:
    tasks: "{{ CVP_DEVICES_RESULTS.taskIds }}"

- name: "Deleting Containers topology on {{ inventory_hostname }}"
  tags: [reset]
  arista.cvp.cv_container_v3:
    topology: "{{ CVP_CONTAINERS }}"
    state: absent

- name: "Delete configlets from CVP {{ inventory_hostname }}."
  tags: [reset]
  arista.cvp.cv_configlet_v3:
    configlets: "{{ CVP_VARS.CVP_CONFIGLETS }}"
    configlet_filter: ["{{ configlets_prefix }}"]
    state: absent
```

the above worked for me using veos-lab

the only problem is that we'd need to introduce a 12 min sleep at least between resetting the devices and deleting the containers, in the future would be better than to move to change control module and move only to the next tasks when we're sure the cc completed successfully and all devices are in the undefined container

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
tgodaA: This explains everything. So we need to make changes on CVP and then only we can make the current adjustments. Let's keep this PR in DRAFT until everything is sorted.

ClausHolbechArista:
Replaced with #2882 for now, since we are not ready to remove v1 support until we have feature parity.

PR #1887, opened 2022-06-17
Feat(eos_designs): EVPN to MPLS-VPN Gateway Overlay Support

emilarista:
## Change Summary

Adds support in eos_designs for configuring EVPN to MPLS-VPN interworking. Overlay only, underlay will come in a separate PR.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

```yaml
<node_type>:
  node_groups:
    <grpname>:
      mpls_vpn_gateway:
        enabled: true
        evpn_domain_id: "65000:1" # Default -> "65000:1"
        mpls_domain_id: "65000:2" # Default -> "65000:2"
        enable_d_path: < true | false | Default -> true >
        maximum_routes: 10000 # Default = 0
        local_as: 65001 # Default is False
        address_families: [ vpn-ipv4, vpn-ipv6 ] # default is [ vpn-ipv4 ]
        remote_peers:
          - hostname: <hostname of MPLS RR1 >
            ip_address: < ip >
            bgp_as: < bgp_as >
          - hostname: <hostname of MPLS RR1 >
            ip_address: < ip >
            bgp_as: < bgp_as >
        mpls_tunnel_source: Loopback0 # Default is Loopback0
```

## How to test
Tested with molecule and virtual lab.

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -345,6 +345,48 @@ defaults <- node_group <- node_group.node <- node
...
+
+```yaml
+< node_type_key >:
+
+  defaults:
+    # Node is acting as MPLS-VPN Gateway for EVPN to MPLS-VPN Interworking | Optional.
+    # A new BGP peer-group is used for this neighborship -> "bgp_peer_groups.mpls_vpn_gateway_overlay_peers".
+    # L3 Reachability is required for this to work, the preferred method to establish underlay connectivity is to use core_interfaces.
+    mpls_vpn_gateway:
+        enabled: < true | false | Default -> False >
```
gmuloc: In general we mostly use `default` instead of `Default` in the doc (except the snippet just above ... :) )

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -345,6 +345,48 @@ defaults <- node_group <- node_group.node <- node
...
+        mpls_domain_id: < "nn:nn" | Default -> "65000:2" >
+
+        # D-path can be turned off for the inter-vpn export if desired.
+        enable_d_path: < true | false | Default -> true >
+
+        # Maximum number of routes to allow from the MPLS domain.
+        maximum_routes: < integer | Default -> 0 >
+
+        # Local-as to use when peering to the MPLS domain.
+        local_as: < bgp asn | Default -> False >
```
gmuloc: We are trying to avoid mixing types in variables so it would be better to maybe not have a default value and if absent just not render it (I guess this is even maybe what the code is doing but I will check).

Additionally as a general comment on doc, the common format for variable name is to used somehting like `bgp_asn` rather than using spaces (same remark for the following variables)

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -127,6 +129,9 @@ bgp_peer_groups:
   evpn_overlay_core:
     name: < name of peer group | default -> EVPN-OVERLAY-CORE >
     password: "< encrypted password >"
+  mpls_vpn_gateway_overlay_peers:
+    name: < name of peer group | Default -> MPLS-VPN-GATEWAY-OVERLAY-PEERS >
+    password: "< encrypted password >"
```
gmuloc:
```suggestion
    name: < peer_group_name | default -> MPLS-VPN-GATEWAY-OVERLAY-PEERS >
    password: "< encrypted_password >"
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -705,6 +705,51 @@ def evpn_gateway_remote_peers(self):
...
+
+    @cached_property
+    def mpls_vpn_gateway_maximum_routes(self):
+        if self.underlay_router is True and self.mpls_vpn_gateway is True:
+            return get(self._switch_data_combined, "mpls_vpn_gateway.maximum_routes", default=0)
+
+    @cached_property
+    def mpls_vpn_gateway_local_as(self):
+        if self.underlay_router is True and self.mpls_vpn_gateway is True:
+            return get(self._switch_data_combined, "mpls_vpn_gateway.local_as", default=False)
```
gmuloc: We should not mix types for variables (default False) when expecting an AS
Better to do something like this which would return None if not present and be ignored in the template later:

```suggestion
    def mpls_vpn_gateway_local_as(self):
        if self.underlay_router is True and self.mpls_vpn_gateway is True:
            return get(self._switch_data_combined, "mpls_vpn_gateway.local_as")
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -705,6 +705,51 @@ def evpn_gateway_remote_peers(self):
             return get(self._switch_data_combined, "evpn_gateway.remote_peers")
         return None
 
+    @cached_property
+    def mpls_vpn_gateway(self):
```
ClausHolbechArista: Instead of adding 9 new facts for one rarely used feature, I think it would be much simpler to just add a single `mpls_vpn_gateway`. You can implement the defaults in that function, and return a dict with settings(or defaults) set as expected.

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ebgp/neighbors.j2
```diff
@@ -20,3 +20,9 @@
       description: {{ evpn_gw_remote_peer }}
       remote_as: "{{ overlay_data.evpn_gateway_remote_peers[evpn_gw_remote_peer].bgp_as }}"
 {% endfor %}
+{% for mpls_vpn_gw_remote_peer in switch.mpls_vpn_gateway_remote_peers | arista.avd.natural_sort %}
+    {{ mpls_vpn_gw_remote_peer.ip_address }}:
+      peer_group: {{ switch.bgp_peer_groups.mpls_vpn_gateway_overlay_peers.name }}
+      description: {{ mpls_vpn_gw_remote_peer.hostname }}
+      remote_as: "{{ mpls_vpn_gw_remote_peer.bgp_as }}"
```
gmuloc, line 27: This makes both `description` and `remote_as` mandatory - maybe lets add it in the doc or add if statements in here

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ibgp/address-family-vpn-ipv6.j2
```diff
@@ -1,14 +1,23 @@
-{% if switch.mpls_overlay_role in ["server", "client"] %}
+{% if switch.mpls_overlay_role in ["server", "client"] or (switch.mpls_vpn_gateway is arista.avd.defined(true) and "vpn-ipv6" in switch.mpls_vpn_gateway_address_families) %}
```
ClausHolbechArista: You are adding special checks everywhere for the gateway feature. It seems like you should just set the mpls_overlay_role to something meaningful (ex. client). I guess a gateway could also work as client/server ?

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ebgp/peer-groups.j2
```diff
@@ -24,3 +24,18 @@
...
+      type: mpls-vpn
+      update_source: {{ switch.mpls_vpn_gateway_mpls_tunnel_source }}
+      bfd: true
+      ebgp_multihop: "{{ mpls_vpn_gateway_ebgp_multihop }}"
+{%     if switch.bgp_peer_groups.mpls_vpn_gateway_overlay_peers.password is arista.avd.defined %}
+      password: {{ switch.bgp_peer_groups.mpls_vpn_gateway_overlay_peers.password }}
+{%     endif %}
+      send_community: all
+      maximum_routes: {{ switch.mpls_vpn_gateway_maximum_routes }}
+{%     if switch.mpls_vpn_gateway_local_as is arista.avd.defined %}
```
gmuloc: What happens here if you don't define it - does it end up being `False`  (which should be changed according to previous comments) ? However probably worth adding a test where it is not defined to make sure this is rendered correctly

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ibgp/router-bgp.j2
```diff
@@ -3,7 +3,11 @@ router_bgp:
 {% if switch.evpn_role == "server" or switch.mpls_overlay_role == "server" %}
   bgp_cluster_id: {{ switch.bgp_cluster_id | arista.avd.default(switch.router_id) }}
 {% endif %}
-
+{% if switch.mpls_vpn_gateway is arista.avd.defined(true) and switch.mpls_vpn_domain_dpath_enabled is arista.avd.defined(true) %}
```
ClausHolbechArista: all these double checks could be simplified if you create a single fact that has all the needed logic done already.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -345,6 +345,48 @@ defaults <- node_group <- node_group.node <- node
...
 
+### BGP & MPLS-VPN Gateway
+
+```yaml
+< node_type_key >:
+
+  defaults:
+    # Node is acting as MPLS-VPN Gateway for EVPN to MPLS-VPN Interworking | Optional.
+    # A new BGP peer-group is used for this neighborship -> "bgp_peer_groups.mpls_vpn_gateway_overlay_peers".
+    # L3 Reachability is required for this to work, the preferred method to establish underlay connectivity is to use core_interfaces.
```
ClausHolbechArista, line 428: You could add something like 
"requires `evpn_role` _and_ `mpls_overlay_role` to be set"
to simplify a lot of the logic. Checks for these could be included in the fact function.

gmuloc (changes requested):

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
emilarista: Fixed this. Added a dict instead.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
emilarista: Changed to default.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
emilarista: Changed doc to none instead of False. Also changed bgp asn to bgp_asn

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
emilarista: Implemented this in new "singe fact" dict for mpls_vpn_gateway.

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ebgp/peer-groups.j2
emilarista: maximum_routes for this defaults to 0, included it in the molecule test in latest commit.

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ebgp/neighbors.j2
emilarista: Added a note in the doc saying hostname, ip_address and remote_as are all mandatory for each remote peer.

ansible_collections/arista/avd/roles/eos_designs/templates/overlay/ibgp/address-family-vpn-ipv6.j2
emilarista: If i reuse the mpls_overlay_role, it would break ibgp design due to these lines in the ibgp evpn template:

```jinja2
  address_family_evpn:
{% if switch.mpls_overlay_role in ["server", "client"] %}
    neighbor_default:
      encapsulation: mpls
{%     if switch.mpls_overlay_role == "client" %}
      next_hop_self_source_interface: Loopback0
{%     endif %}
{% endif %}
```

I suppose i could test for the inverse here: if server/client and not mpls_vpn_gateway.enabled. There are many more instances where this would be required in the ibgp overlay templates. Not sure which is better, adding checks for gateway knobs where i need mpls functionality turned on or adding checks for gateway knobs where i need to prevent mpls functionality from getting turned on as a result of mpls_overlay_role reuse...

emilarista:
Converted back to draft pending some mpls logic refactoring.

PR #1922, opened 2022-07-04
Feat(eos_designs): Fabric EVPN multicast

### LINKED ISSUE #1913: Feat(eos_designs): Fabric variable to enable EVPN multicast
### Enhancement summary

Make it easy to configure fabric to support EVPN Multicast.
Applies to L3LS EVPN Scenario.

eos_designs

Note: Both  `underlay_multicast` and new `evpn_multicast` must be enabled in for L3LS EVPN scenario, this may be different for MPLS for scenario.

**Existing Multicast knob:**

```yaml
# Enable Multicast in the underlay on all p2p uplink interfaces and mlag l3 peer interface.
# Specifically PIM Sparse-Mode will be configured on all routed underlay interfaces.
# No other configuration is added, so the underlay will only support Source-Specific Multicast (SSM)
# The configuration is intended to be used as multicast underlay for EVPN OISM overlay
underlay_multicast: <boolean | default -> false >
```

**New additional knob in fabric services:**

```yaml
# Enable Fabric to support EVPN Multicast
# Requires "underlay_multicast: true" and IGMP snooping enabled globally.
# For MLAG devices Route Distinguisher must be unique since this feature will create multi-vtep configuration.

evpn_multicast: <boolean | default -> false >
```

Pseudocode logic:

avd_facts
```
If node_type_keys.<node_type_key>.vtep_key is true and underlay_multicast is true
   check that igmp snooping is not disable globally  -> provide error
         set_fact -> switch.evpn_multicast: true
```

structured configuration templates

```
if switch.evpn_multicast is  true
```

Configure software-forwarding sfe:
```yaml
router_multicast:
  ipv4:
    routing: true
    software_forwarding: sfe
```

if switch.evpn_multicast is  true and mlag is true

Enable multi-vtep:
```yaml
vxlan_interface:
  Vxlan1:
    vxlan:
      source_interface: Loopback0
      mlag_source_interface: Loopback1
```

### Contributing Guide

carlbuchmann:
## Change Summary

Provide knob to globally enable EVPN multicast functionality

## Related Issue(s)

Fixes #1913 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

```yaml
# Enable Fabric to support EVPN Multicast
# Requires "underlay_multicast: true" and IGMP snooping enabled globally.
# For MLAG devices Route Distinguisher must be unique since this feature will create multi-vtep configuration.
evpn_multicast: <boolean | default -> false >
```
Additional details documented in issue #1913 

## How to test

See Molecule test case

### User Checklist

- [x] Update documentation
- [x] Create feature
  - [x]  Provide error when IGMP snooping is not enabled globally
- [x] Create Molecule Scenario

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -213,6 +210,15 @@ uplink_ptp:
 # The configuration is intended to be used as multicast underlay for EVPN OISM overlay
 underlay_multicast: < boolean | default -> false >
 
+# Enable Fabric to support EVPN Multicast
+# Requires "underlay_multicast: true" and IGMP snooping enabled globally.
+# For MLAG devices Route Distinguisher must be unique since this feature will create multi-vtep configuration.
+evpn_multicast: <boolean | default -> false >
+
+# Disable IGMP snooping at fabric level.
+# If set, it overrides per vlan settings
```
lermilov: >>If set, it overrides per vlan settings

I would suggest to specify what does it mean "set" here. Because if set to "true", then it will not override per vlan settings. right?

lermilov (changes requested):
According to the https://www.arista.com/en/support/toi/eos-4-26-0f/14669-layer-2-multicast-evpn


For 7050X3/7300X3/720XP platforms, we also need to add additional platform command to make EVPN L2 multicast to work:
`platform trident forwarding-table partition flexible exact-match 16384 l2-shared 98304 l3-shared 131072`

Exact values inside command should be configurable because they depend on the scaling. For default, we can use numbers above for X3 platforms.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -663,6 +663,24 @@ def spanning_tree_root_super(self):
...
+        if get(self._hostvars, "evpn_multicast") is True and self.vtep is True:
+            if not (
+                self.underlay_multicast is True
+                and self.igmp_snooping_enabled is not False
+            ):
+                raise AristaAvdError(
+                    "'evpn_multicast: True' is only supported in "
+                    "combination with 'underlay_multicast: True' and 'igmp_snooping_enabled : True'"
+                )
+            return get(self._hostvars, "evpn_multicast", default=False)
```
ClausHolbechArista, line 782: could we add a check that if mlag we also check for hostvars `overlay_rd_type.admin_subfield` is either "overlay_lookback", unset (default="overlay_loopback") or that the value is different that the value of the mlag peer (same key). You can find the mlag peer's value by accessing `avd_switch_facts` with the `get` function. See peer_mlag_id for an example.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
carlbuchmann: I believe that is correct, This knob was only moved location but not created in this PR.
@ClausHolbechArista

carlbuchmann:
> According to the https://www.arista.com/en/support/toi/eos-4-26-0f/14669-layer-2-multicast-evpn
> 
> For 7050X3/7300X3/720XP platforms, we also need to add additional platform command to make EVPN L2 multicast to work: `platform trident forwarding-table partition flexible exact-match 16384 l2-shared 98304 l3-shared 131072`
> 
> Exact values inside command should be configurable because they depend on the scaling. For default, we can use numbers above for X3 platforms.

Address with commit: a9122aa

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
carlbuchmann: Yes, will add this check!

ansible_collections/arista/avd/roles/eos_designs/templates/base/platform.j2
```diff
@@ -1,7 +1,12 @@
 {# platform #}
-{% if switch.platform_settings.lag_hardware_only is arista.avd.defined %}
 platform:
+{% if switch.platform_settings.lag_hardware_only is arista.avd.defined %}
   sand:
     lag:
       hardware_only: {{ switch.platform_settings.lag_hardware_only }}
 {% endif %}
+{% set trident3_platforms = ["7050X3", "7300X3", "720XP", "722XP"] %}
```
ClausHolbechArista: please don't put platforms into the code. We should refactor this change to use platform settings.

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vxlan-interface.j2
```diff
@@ -3,7 +3,12 @@ vxlan_interface:
   Vxlan1:
     description: {{ inventory_hostname }}_VTEP
     vxlan:
+{% if switch.evpn_multicast is arista.avd.defined(true) and switch.mlag is arista.avd.defined(true)  %}
```
ClausHolbechArista: may I suggest a separate `switch.multi_vtep` fact which for now is calculated based on those two other ones, but in the future could have it's own knob to enable multi_vtep.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
ClausHolbechArista: correct, it will only override if set to false. Remember also that this default_ is only a fallback if not set on the node_type. So ex. `l3leaf.defaults.igmp_snooping_enabled` would take precedence.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
```diff
@@ -83,6 +83,8 @@ mlag_ibgp_peering_vrfs:
 # Platform Settings
 platform_settings: "{{ default_platform_settings }}"
 
+# Default platform settings for trident3 platforms - from EVPN multicast
+platform_settings_trident_forwarding_table_partition: " flexible exact-match 16384 l2-shared 98304 l3-shared 131072"
```
ClausHolbechArista: Let's not bloat the defaults with on-off special things like this. I suggest to move to platform_settings where we have other similar knobs that are platform specific, but only leveraged with some features.

ansible_collections/arista/avd/roles/eos_designs/templates/base/platform.j2
carlbuchmann: moved to platform settings

ansible_collections/arista/avd/roles/eos_designs/defaults/main/main.yml
carlbuchmann: addressed in latest commit

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vxlan-interface.j2
carlbuchmann: yes, will move that calculation to python!

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
carlbuchmann: updated with `If set to false`, please feel free to modify as needed!

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
carlbuchmann: Completed

ansible_collections/arista/avd/roles/eos_designs/templates/network_services/vxlan-interface.j2
carlbuchmann: Completed

ansible_collections/arista/avd/roles/eos_designs/templates/base/router-multicast.j2
```diff
@@ -0,0 +1,6 @@
+{% if switch.evpn_multicast is arista.avd.defined(true) %}
+router_multicast:
+  ipv4:
+    routing: true
```
ClausHolbechArista, line 4: I think we forgot to do the first three lines of this based on `underlay_multicast`.
Lets refactor this one to:
```jinja
{% if switch.underlay_multicast is arista.avd.defined(true) %}
router_multicast:
  ipv4:
    routing: true
{%     if switch.evpn_multicast is arista.avd.defined(true) %}
    software_forwarding: sfe
{%     endif %}
{% endif %}
```

ansible_collections/arista/avd/roles/eos_designs/doc/platform-settings.md
```diff
@@ -44,49 +46,65 @@ The reload delay values should be reviewed and tuned to the specific environment
...
       queue_monitor_length_notify: false
-  - platforms: [ '7280R', '7280R2' ]
+  - platforms: ['7050X3', '720XP', '722XP']
+    # Only applied when enabling EVPN multicast
+    trident_forwarding_table_partition: "flexible exact-match 16384 l2-shared 98304 l3-shared 131072"
+    reload_delay:
+      mlag: 300
+      non_mlag: 330
+    feature_support:
+      # "queue-monitor length notify" is only valid for R-Series so should be disabled on default platform.
```
ClausHolbechArista:
```suggestion
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -213,6 +209,20 @@ uplink_ptp:
 # The configuration is intended to be used as multicast underlay for EVPN OISM overlay
 underlay_multicast: < boolean | default -> false >
 
+# Enable Fabric to support EVPN Multicast
+# Requires "underlay_multicast: true" and IGMP snooping enabled globally.
```
ClausHolbechArista:
```suggestion
# Requires "underlay_multicast: true" and IGMP snooping enabled globally (default).
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-variables.md
```diff
@@ -213,6 +209,20 @@ uplink_ptp:
 # The configuration is intended to be used as multicast underlay for EVPN OISM overlay
 underlay_multicast: < boolean | default -> false >
 
+# Enable Fabric to support EVPN Multicast
```
ClausHolbechArista, line 212:
```suggestion
# Enable Fabric to support EVPN Multicast
# General Configuration required for EVPN Multicast. "evpn_l2_multicast" must also be configured under the Network Services (tenants).
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/router-multicast.j2
carlbuchmann: good catch, fixed!

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -732,6 +732,62 @@ def spanning_tree_root_super(self):
...
+                        "For MLAG devices Route Distinguisher must be unique when 'evpn_multicast: True' "
+                        "since it will create a multi-vtep configuration."
+                    )
+            return get(self._hostvars, "evpn_multicast", default=False)
+        return None
+
+    @cached_property
+    def multi_vtep(self):
+        if self.mlag is True and self.evpn_multicast is True:
+            return True
```
lermilov, line 788: Not sure if I understan this function correctly. If intention is to define general "multi_vtep" fact, then instead of checking "self.evpn_multicast", it is better to check that vxlan_source is lo0 and vxlan_mlag_source is lo1.

ClausHolbechArista: the _result_ of this knob is that those two will be as you say. It is not the trigger.

ClausHolbechArista (approved):
LGTM

lermilov (approved):

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -732,6 +732,62 @@ def spanning_tree_root_super(self):
...
 
+    @cached_property
+    def underlay_multicast(self):
+        return get(self._hostvars, "underlay_multicast")
+
+    @cached_property
+    def overlay_rd_type_admin_subfield(self):
+        tmp_overlay_rd_type_admin_subfield = default(
+            get(self._hostvars, "evpn_rd_type.admin_subfield"),
+            get(self._hostvars, "overlay_rd_type.admin_subfield")
```
tgodaA, line 743: Because the new key is ```overlay_rd_type```, shouldn't we use it first?

tgodaA: NVM, this is correct.

PR #1931, opened 2022-07-07
Fix(eos_designs): remove the need for mgmt_gateway

### LINKED ISSUE #1886: Bug(eos_designs): Invalid config if no mgmt_ip is given for a node.
### Issue Summary

`eos_designs` should not require mgmt_gateway to be set, and should not `no shut` the default management interface, if no `mgmt_ip` is set for a given device.

eos_designs

### How do you run AVD ?

_No response_

### Steps to reproduce

```shell
don't set a `mgmt_ip` for device
don't set `mgmt_gateway`
This will return an `undefined variable` error because `mgmt_gateway` is not found.

After setting `mgmt_gateway` the playbook runs, but the config produced includes a default route, enabled management interface, but no IP address.
```

### Relevant log output

_No response_

### Contributing Guide

gmuloc:
## Change Summary

This PR relax the need for a `mgmt_gateway`. Currently the behavior is that `eos_designs` will still try to install a static default route to the `mgmt_gateway` even if it is not defined.

Note: however the default behavior is still to configure a default `MGMT` vrf so this PR may need some more adjustments.

## Related Issue(s)

Fixes #1886 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

in the `eos_designs/template/base/static-routes.j2`, attempt to install a default mgmt route only if `mgmt_gateway` is defined.

## How to test

molecule with a new device `no_mgmt_interface`

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
@@ -1,12 +1,14 @@
 static_routes:
-{% if mgmt_destination_networks is arista.avd.defined and mgmt_destination_networks | length > 0 %}
-{%     for mgmt_destination_network in mgmt_destination_networks %}
+{% if mgmt_gateway is arista.avd.defined %}
+{%     if mgmt_destination_networks is arista.avd.defined and mgmt_destination_networks | length > 0 %}
+{%         for mgmt_destination_network in mgmt_destination_networks %}
   - vrf: {{ mgmt_interface_vrf }}
     destination_address_prefix: {{ mgmt_destination_network }}
     gateway: {{ mgmt_gateway }}
-{%     endfor %}
-{% else %}
+{%         endfor %}
+{%     else %}
   - vrf: {{ mgmt_interface_vrf }}
     destination_address_prefix: 0.0.0.0/0
     gateway: {{ mgmt_gateway }}
+{%     endif %}
 {% endif %}
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/no_mgmt_interface.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/no_mgmt_interface.cfg
@@ -1,0 +1,26 @@
+!RANCID-CONTENT-TYPE: arista
+!
+vlan internal order ascending range 1006 1199
+!
+transceiver qsfp default-mode 4x10G
+!
+service routing protocols model multi-agent
+!
+hostname no_mgmt_interface
+!
+no enable password
+no aaa root
+!
+vrf instance MGMT
+!
+ip routing
+no ip routing vrf MGMT
+!
+management api http-commands
+   protocol https
+   no shutdown
+   !
+   vrf MGMT
+      no shutdown
+!
+end
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/no_mgmt_interface.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/no_mgmt_interface.yml
@@ -1,0 +1,16 @@
+service_routing_protocols_model: multi-agent
+ip_routing: true
+vlan_internal_order:
+  allocation: ascending
+  range:
+    beginning: 1006
+    ending: 1199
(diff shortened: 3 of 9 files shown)


ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
```diff
@@ -1,12 +1,14 @@
 static_routes:
-{% if mgmt_destination_networks is arista.avd.defined and mgmt_destination_networks | length > 0 %}
-{%     for mgmt_destination_network in mgmt_destination_networks %}
+{% if mgmt_gateway is arista.avd.defined %}
```
ClausHolbechArista, line 2: maybe we should also check for `switch.mgmt_ip`. And then also check for `switch.mgmt_ip` before configuring the MGMT vrf. We can add those in this PR (and rename it) or create a seperate one. Up to you.

ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
```diff
@@ -1,12 +1,14 @@
 static_routes:
```
ClausHolbechArista, line 1: The docs should also be updated to have this as "Optional"
```yaml
# Management interface configuration | Required
mgmt_vrf_routing: < boolean | default -> false >
mgmt_interface: < mgmt_interface | default -> Management1 >
mgmt_interface_vrf: < vrf_name | default -> MGMT >
mgmt_gateway: < IPv4 address >
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
gmuloc: Just to verify, would this be the correct behavior in these cases:
* ip + gateway == route
* ip only = no route (because we don't know the next hop?)
* gateway only = ? (right now we have `gateway only = route` but then if there is no IP it may not be very useful indeed)
* neither == no route

Regarding the MGMT vrf, it becomes sneaky because we have in some places like name-servers (and maybe other I would need to analyze further) the mgmt_vrf expected to be defined and because we are giving a default value this is always generated:
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/templates/base/vrfs.j2#L3

Example where the VRF name will be used even if we don't render it:
* https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/templates/base/name-servers.j2#L5

ClausHolbechArista: Agree on your behavior descriptions.
You are right on the other templates. Would it stop the config of MGMT vrf if I set `mgmt_interface_vrf: null`?

gmuloc: `mgmt_interface_vrf: null` raises a not very explicit error:

```
TASK [arista.avd.eos_designs : Generate device configuration in structured format] ***
An exception occurred during task execution. To see the full traceback, use -vvv. The error was:       ^
fatal: [no_mgmt_interface -> localhost]: FAILED! => {"msg": "Unexpected failure during module execution.", "stdout": ""}
```

ClausHolbechArista: Let's keep this PR only looking at mgmt_gateway and punt the issue of mgmt_interface_vrf to another PR.

ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
gmuloc: addressed in 5b1eef5

ansible_collections/arista/avd/roles/eos_designs/templates/base/static-routes.j2
gmuloc: So i had started on something in here: https://github.com/gmuloc/ansible-avd/tree/mgmt-interface-vrf building upon 5b1eef5 but this was not completed yet we can postpone this to a new issue / PR indeed

ClausHolbechArista (approved):
LGTM.

tgodaA (approved):
LGTM

PR #1946, opened 2022-07-14
Feat(cvp_configlet_upload): document option of configlet_build_config

### LINKED ISSUE #1945: Feat(cvp_configlet_upload): document the option of configlet_build_config
### Enhancement summary

Explain the syntax to apply configs without adding a common prefix

cvp_configlet_upload

### Use case example

please detail your use case

Update doc

Modify the logic to support other reasonable assumptions.  

### Additional context

_No response_

### Contributing Guide

pvinci-arista:
## Change Summary
Document the 'none' option

## Related Issue(s)

Fixes #1945 

## Component(s) name

`arista.avd.cvp_configlet_upload`

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/modules/configlet_build_config.py
```diff
@@ -37,6 +37,7 @@
   configlet_prefix:
     description: Prefix to append on configlet.
     required: true
+    choices: ['none', <prefix to prepend>]
```
tgodaA, line 40: If you are using choices, then I believe we are limiting the the input options, like the input should be only one of it.

tgodaA: Also, when we define choices in the document, we also need to define it here (line113), in argument_spec: https://github.com/aristanetworks/ansible-avd/pull/1946/files#diff-f0f5953060115687a96986b8e4921454e70944b4695b8c3c0172b32bec5bef9aR113

pvinci-arista: Is there a better way to resolve this?  Alternatively, I can relax the test to:
```
if prefix and prefix != 'none': 
```

ansible_collections/arista/avd/plugins/modules/configlet_build_config.py
```diff
@@ -83,6 +84,7 @@ def get_configlet(src_folder="", prefix='AVD', extension='cfg'):
         Path where to find configlet, by default ""
     prefix : str, optional
```
gmuloc, line 85: Not sure why we have `optional` here when we state that it is `required` line 39. Given that we have a default of `AVD` should we relax the `required` above? @tgodaA

ansible_collections/arista/avd/plugins/modules/configlet_build_config.py
gmuloc: Reading through this: https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html I agree with @tgodaA and we should not use choices here. Probably just make it clearer in the description that `"none"` is a specific keyword to indicate "no keyword"

ansible_collections/arista/avd/plugins/modules/configlet_build_config.py
pvinci-arista: configlet_prefix: 'AVD_' at line 59 in the example is also misleading.  it would render AVD__<configlet_name>

pvinci-arista: I think it is more intuitive to have a default of 'none', but that would be a breaking change of the existing behavior, yes?

gmuloc: > configlet_prefix: 'AVD_' at line 59 in the example is also misleading. it would render AVD__<configlet_name>

Indeed for this one we can fix it inside this PR

gmuloc: > I think it is more intuitive to have a default of 'none', but that would be a breaking change of the existing behavior, yes?

It would indeed..

tgodaA: I believe here it is stated as optional because of the default parameter. But indeed this is still confusing. There is no sync between what is described and what's used in examples.

tgodaA: @gmuloc , line39 is defined as required because of line 113. But I think this is mis-interpreted, because this should have been optional (reason being, we already have a default value set in line 76). There is something wrong with this module. I believe this is a bug. We need to open a bug and fix it. But not in this PR because it's a Feat.

gmuloc: https://github.com/aristanetworks/ansible-avd/issues/1956

PR #1948, opened 2022-07-18
Feat(eos_designs): Auto BGP ASN

jonxstill:
## Change Summary

Automatically assign BGP ASNs to a node type from a range of ASNs

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Amends the existing eos_designs bgp_as variable to support ranges of ASNs.  This allows a `bgp_as` range to be defined under a node type's `defaults` clause so that ASNs can be allocated to (for example) l3leafs using the following mechanism:

The bgp_as range is expanded into a list of ASNs and ASNs allocated according to the following:

* Standalone nodes: device's `id` is used to index into that list.
* A/A EVPN MH nodes: device's `id` is used to index into that list.
* MLAG nodes: The device ID of the first node in the node group of the MLAG pair is used to index into the list.

Bare (non-range) ASN behaviour is unaffected.  Inheritance is also unaffected.
```
# Node BGP AS | Required.
# For eBGP topologies, bgp_as can be defined as a range of ASNs under defaults.  This range will be expanded and ASNs allocated
# according to the following rules:
#  - Standalone nodes and Non-MLAG (i.e. EVPN A/A multi-homing) node groups will be allocated unique ASNs from the range based on
#    the node's ID.
#  - Switches in MLAG pairs will be allocated a single ASN, based on the node ID of the first node in the node group.
# Examples:
# bgp_as: 65101-65110                 Contiguous range of ASNs.
# bgp_as: 65001,65010,65020,65030     Non-contiguous range of ASNs.
# bgp_as: 65000.101-125               Contiguous range of 4-byte ASNs, increment in the last 2 bytes.
# bgp_as: 64512                       Single AS, no range.
# Commas and hyphens can be combined in more-or-less any combination.
```
## How to test
New molecule test switches defined to test just this functionality.  Also tested against my standard demo repos.

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
@@ -41,3 +41,18 @@ l3leaf:
         AUTO_BGP_ASN_LEAF3B:
           id: 6
           mgmt_ip: 192.168.202.110/24
+    AUTO_BGP_ASN_LEAF4_MLAG_OVERRIDE:
+      bgp_as: 65222
+      nodes:
+        AUTO_BGP_ASN_LEAF4A:
+          id: 7
+          mgmt_ip: 192.168.202.111/24
+        AUTO_BGP_ASN_LEAF4B:
+          id: 8
+          mgmt_ip: 192.168.202.112/24
+    AUTO_BGP_ASN_LEAF5_NODE_OVERRIDE:
+      nodes:
+        AUTO_BGP_ASN_LEAF5A:
+          id: 9
+          bgp_as: 65333
+          mgmt_ip: 192.168.202.113/24
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
@@ -1,0 +1,43 @@
+type: l3leaf
+mgmt_gateway: 192.168.202.1
+
+l3leaf:
+  defaults:
+    platform: vEOS-LAB
+    loopback_ipv4_pool: 192.168.255.0/24
+    loopback_ipv4_offset: 2
+    vtep_loopback_ipv4_pool: 192.168.254.0/24
+    uplink_interfaces: []
+    uplink_switches: []
+    uplink_ipv4_pool: 172.31.255.0/24
+    bgp_defaults: ['no bgp default ipv4-unicast', 'distance bgp 20 200 200']
+    mlag_interfaces: [ Ethernet3, Ethernet4 ]
+    virtual_router_mac_address: 00:dc:00:00:00:0a
+    mlag_peer_l3_ipv4_pool: 10.255.251.0/24
+    mlag_peer_ipv4_pool: 10.255.252.0/24
+    spanning_tree_mode: mstp
+    spanning_tree_priority: 4096
+    bgp_as: 65101-65110
+  node_groups:
+    AUTO_BGP_ASN_LEAF1_AA:
+      mlag: false
+      nodes:
+        AUTO_BGP_ASN_LEAF1A:
+          id: 1
+          mgmt_ip: 192.168.202.105/24
+        AUTO_BGP_ASN_LEAF1B:
+          id: 2
+          mgmt_ip: 192.168.202.106/24
+    AUTO_BGP_ASN_LEAF2_STANDALONE:
+      nodes:
+        AUTO_BGP_ASN_LEAF2:
+          id: 3
+          mgmt_ip: 192.168.202.107/24
+    AUTO_BGP_ASN_LEAF3_MLAG:
+      nodes:
+        AUTO_BGP_ASN_LEAF3A:
+          id: 5
+          mgmt_ip: 192.168.202.109/24
+        AUTO_BGP_ASN_LEAF3B:
+          id: 6
+          mgmt_ip: 192.168.202.110/24
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
@@ -56,3 +56,7 @@ l3leaf:
           id: 9
           bgp_as: 65333
           mgmt_ip: 192.168.202.113/24
+  nodes:
+    AUTO_BGP_UNGROUPED_LEAF6:
+      id: 10
+      mgmt_ip: 192.168.202.114/24
--- a/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
@@ -1122,14 +1122,27 @@ class EosDesignsFacts:
         '''
         Get global bgp_as or fabric_topology bgp_as.
 
-        This will fail if none of these are found.
+        This will fail if none of these are found. AS ranges will be expanded and allocated to devices based on:
+         - ID of node (for standalone or A/A MH devices)
+         - ID of node 0 in node group (for MLAG)
+         - Bare ASNs work as per previous behaviour
         '''
         if self.underlay_router is True:
             if self.underlay_routing_protocol == 'ebgp' or self.evpn_role != 'none' or self.mpls_overlay_role != 'none':
                 if get(self._hostvars, "bgp_as") is not None:
                     return str(get(self._hostvars, "bgp_as"))
                 else:
-                    return str(get(self._switch_data_combined, "bgp_as", required=True))
+                    bgp_as_range_expanded = self._range_expand(str(get(self._switch_data_combined, "bgp_as", required=True)))
+                    try:
+                        if len(bgp_as_range_expanded) == 1:
+                            return str(bgp_as_range_expanded[0])
+                        elif self.mlag:
+                            return str(bgp_as_range_expanded[self._switch_data_node_group_nodes[0]["id"] - 1])
+                        else:
+                            return str(bgp_as_range_expanded[self.id - 1])
+                    except IndexError as exc:
+                        raise AristaAvdError("unable to allocate bgp as: please expand the bgp_as range") from exc
+
             # Hack to make mpls PR non-breaking, adds empty bgp to igp topology spines
             # TODO: Remove this as part of AVD4.0
             elif (
--- a/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
@@ -1141,11 +1141,13 @@ class EosDesignsFacts:
                         if len(bgp_as_range_expanded) == 1:
                             return bgp_as_range_expanded[0]
                         elif self.mlag:
-                            return bgp_as_range_expanded[self._switch_data_node_group_nodes[0]["id"] - 1]
+                            return bgp_as_range_expanded[self.mlag_switch_ids["primary"] - 1]
                         else:
                             return bgp_as_range_expanded[self.id - 1]
                     except IndexError as exc:
-                        raise AristaAvdError("Unable to allocate bgp as: bgp_as range is too small for number of devices") from exc
+                        raise AristaAvdError(
+                            "Unable to allocate BGP AS: bgp_as range is too small "
+                            f"({len(bgp_as_range_expanded)}) for the id of the device") from exc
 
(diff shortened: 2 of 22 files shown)


ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -962,14 +962,27 @@ def bgp_as(self):
...
         if self.underlay_router is True:
             if self.underlay_routing_protocol == 'ebgp' or self.evpn_role != 'none' or self.mpls_overlay_role != 'none':
                 if get(self._hostvars, "bgp_as") is not None:
                     return str(get(self._hostvars, "bgp_as"))
                 else:
-                    return str(get(self._switch_data_combined, "bgp_as", required=True))
+                    bgp_as_range_expanded = self._range_expand(str(get(self._switch_data_combined, "bgp_as", required=True)))
+                    try:
+                        if len(bgp_as_range_expanded) == 1:
+                            return str(bgp_as_range_expanded[0])
```
gmuloc: Do we need to cast to string "again" when it was done line 975? Looking at the code of range_expand it returns strings

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -962,14 +962,27 @@ def bgp_as(self):
         '''
         Get global bgp_as or fabric_topology bgp_as.
 
-        This will fail if none of these are found.
+        This will fail if none of these are found. AS ranges will be expanded and allocated to devices based on:
+         - ID of node (for standalone or A/A MH devices)
+         - ID of node 0 in node group (for MLAG)
+         - Bare ASNs work as per previous behaviour
```
gmuloc: Referring to previous behavior in the docstring does not indicate what is the previous behavior :)

probably better to reword this by stating the different cases.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -962,14 +962,27 @@ def bgp_as(self):
...
+                    bgp_as_range_expanded = self._range_expand(str(get(self._switch_data_combined, "bgp_as", required=True)))
+                    try:
+                        if len(bgp_as_range_expanded) == 1:
+                            return str(bgp_as_range_expanded[0])
+                        elif self.mlag:
+                            return str(bgp_as_range_expanded[self._switch_data_node_group_nodes[0]["id"] - 1])
+                        else:
+                            return str(bgp_as_range_expanded[self.id - 1])
+                    except IndexError as exc:
+                        raise AristaAvdError("unable to allocate bgp as: please expand the bgp_as range") from exc
```
gmuloc: maybe even saying `bgp_as range too small` instead of `please expand bgp_as range`

```suggestion
                        raise AristaAvdError("Unable to allocate bgp as: please expand the bgp_as range") from exc
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -333,7 +333,18 @@ default_interfaces:
 < node_type_key >:
 
   defaults:
-    # node BGP AS | Required.
+    # Node BGP AS | Required.
+    # For eBGP topologies, bgp_as can be defined as a range of ASNs under defaults.  This range will be expanded and ASNs allocated
+    # according to the following rules:
+    #  - Standalone nodes and Non-MLAG (i.e. EVPN A/A multi-homing) node groups will be allocated unique ASNs from the range based on
```
gmuloc, line 339: So let's assume I have 3 nodes EVPN A/A that I want to put in a node groups for other purposes (shared variables .. whatever). It seems this won't be supported by this PR right?

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -333,7 +333,18 @@ default_interfaces:
...
+    #    the node's ID.
+    #  - Switches in MLAG pairs will be allocated a single ASN, based on the node ID of the first node in the node group.
```
gmuloc:
```suggestion
    #  - Switches in MLAG pairs will be allocated a single ASN for both, based on the node ID of the first node in the node group.
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -333,7 +333,18 @@ default_interfaces:
...
+    # Examples:
+    # bgp_as: 65101-65110                 Contiguous range of ASNs.
+    # bgp_as: 65001,65010,65020,65030     Non-contiguous range of ASNs.
+    # bgp_as: 65000.101-125               Contiguous range of 4-byte ASNs, increment in the last 2 bytes.
+    # bgp_as: 64512                       Single AS, no range.
+    # Commas and hyphens can be combined in more-or-less any combination.
```
gmuloc: not a native English speaker here so just a question regarding hyphenated `more-or-less` - is it supposed to be spelled like this?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
```diff
@@ -0,0 +1,58 @@
...
+    uplink_ipv4_pool: 172.31.255.0/24
+    bgp_defaults: ['no bgp default ipv4-unicast', 'distance bgp 20 200 200']
+    mlag_interfaces: [ Ethernet3, Ethernet4 ]
+    virtual_router_mac_address: 00:dc:00:00:00:0a
+    mlag_peer_l3_ipv4_pool: 10.255.251.0/24
+    mlag_peer_ipv4_pool: 10.255.252.0/24
+    spanning_tree_mode: mstp
+    spanning_tree_priority: 4096
+    bgp_as: 65101-65110
+  node_groups:
```
gmuloc, line 21: I am 99.9% sure it works already given that it has a `switch.id` but maybe lets add a test with a `nodes` not inside a `node_groups`

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
jonxstill: The cast in 975 is to cast the input to range_expand, which is explicitly looking for lists or strings and a bare AS in the YAML (with no quotes) will be interpreted as an int.  You're right about the extraneous str on 978 though - range_expand will always return a list of strings (due to the use of f-strings).  Will fix.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
jonxstill: That sounds better for sure.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
jonxstill: 3 A/A nodes in a group will be allocated 3 ASNs based on their ID.  It's only MLAG that gets treated differently (uses ID of first node in group to determine ASN for all nodes).  If you want those nodes in a group for some other reason and don't want this BGP behaviour then just assign ASNs manually as before and they will override any defaults.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
jonxstill: Agreed.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
jonxstill: So I just looked this up and I think this is just a habitual thing of mine.  Will remove.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
gmuloc: Hehe - I thought I deleted this as I came to the same conclusion by reading your tests.. Resolving it.. apologies it stayed at the end of my review process!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
jonxstill: Good catch, will do.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
jonxstill: Yes, will do.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTO_BGP_ASN.yml
jonxstill: Done, works great!

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -962,14 +962,31 @@ def bgp_as(self):
...
                 if get(self._hostvars, "bgp_as") is not None:
                     return str(get(self._hostvars, "bgp_as"))
                 else:
-                    return str(get(self._switch_data_combined, "bgp_as", required=True))
+                    bgp_as_range_expanded = self._range_expand(str(get(self._switch_data_combined, "bgp_as", required=True)))
+                    try:
+                        if len(bgp_as_range_expanded) == 1:
+                            return bgp_as_range_expanded[0]
+                        elif self.mlag:
+                            return bgp_as_range_expanded[self._switch_data_node_group_nodes[0]["id"] - 1]
```
tgodaA: use the cached_property: mlag_switch_ids

return bgp_as_range_expanded[self.mlag_switch_ids["primary"] - 1]

gmuloc (commented):
After applying the suggested change from @tgodaA I will approve this.

Can you also indicate what happens if you put a range in the top level `bgp_as`?

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -962,14 +962,31 @@ def bgp_as(self):
...
+                        else:
+                            return bgp_as_range_expanded[self.id - 1]
+                    except IndexError as exc:
+                        raise AristaAvdError("Unable to allocate bgp as: bgp_as range is too small for number of devices") from exc
```
ClausHolbechArista:
```suggestion
                        raise AristaAvdError(f"Unable to allocate BGP AS: bgp_as range is too small ({len(bgp_as_range_expanded)}) for the id of the device") from exc
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -333,7 +333,18 @@ default_interfaces:
 < node_type_key >:
 
   defaults:
-    # node BGP AS | Required.
+    # Node BGP AS | Required.
+    # For eBGP topologies, bgp_as can be defined as a range of ASNs under defaults.  This range will be expanded and ASNs allocated
```
ClausHolbechArista: It would accept the range anywhere in the hierarchy.
```suggestion
    # For eBGP topologies, bgp_as can be defined as a range of ASNs.  This range will be expanded and ASNs allocated
```

ClausHolbechArista (commented):
Would be nice if you could add this feature to on of the sites in the 5stage molecule scenario, in a way that is not producing any changes. This way we get some feature interaction tests as well.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
jonxstill: Done, thanks.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
jonxstill: Done, thanks.

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
jonxstill: Done, thanks.

tgodaA:
LGTM, but would it possible to test on 5stage-2DC-clos scenario, instead of creating a new one.

ClausHolbechArista:
> LGTM, but would it possible to test on 5stage-2DC-clos scenario, instead of creating a new one.

we need the unit tests as well, but for feature interaction it would be good to see it on one of the other examples. One change it will give though is that mlag leaf switches will only use every other ASN from a list.

tgodaA:
Yeah, but I would comment that it is based on switch, primary ID.

ClausHolbechArista (approved):
LGTM. The testing done is thorough and checking many different variations.

gmuloc (approved):
LGTM

PR #1949, opened 2022-07-18
Feat(cvp_configlet_upload): Expand prefix options to not modify configlet name

pvinci-arista:
## Related Issue(s)

Fixes #1946 

## Component(s) name

`arista.avd.cvp_configlet_upload`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
+++ b/ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
@@ -51,8 +51,9 @@ class TestConfigletBuildConfig():
         output = get_configlet()
         assert output == dict()
 
-    def test_get_configlet_none_prefix(self):
+    @pytest.mark.parametrize("prefix", ['none', "", None])
+    def test_get_configlet_none_prefix(self, prefix):
         extension = 'cfg'
-        output = get_configlet(src_folder=CONFIGLETS_DIR, prefix='none')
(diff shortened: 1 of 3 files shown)


gmuloc (commented):
Looks good - could you add some tests for `None` and `""` in this file as well: https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py#L54

ClausHolbechArista (commented):
Please update the PR description and title to reflect the updates you are suggesting. (I agree on your changes).
The PR title is directly used in the release-notes, so it should be precise.

ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
```diff
@@ -51,8 +51,9 @@ def test_get_configlet_invalid_source(self):
         output = get_configlet()
         assert output == dict()
 
-    def test_get_configlet_none_prefix(self):
+    @pytest.mark.parametrize("prefix", ['none', "", None])
+    def test_get_configlet_none_prefix(self, prefix):
```
ClausHolbechArista, line 55: There is another test "test_get_configlet_blank_prefix" which fails now. Since you are now including that scenario here, the other test should be removed. But this also means that the changes are actually breaking in case someone has a blank prefix today. 
 @carlbuchmann can we accept this for minor?

PR #1955, opened 2022-07-21
Fix(eos_designs): fix the Loopback0 interface description

### LINKED ISSUE #1954: Bug(eos_designs): fix loopback0 description
### Issue Summary

Loopback0 description is erroring when using design type: L2LS.  This bug needs fixed before PR1938 can be merged.
```
TASK [arista.avd.eos_designs : Generate device configuration in structured format] *****************************************************************************************
Wednesday 20 July 2022  14:57:24 +0000 (0:00:00.302)       0:00:03.441 ******** 
Wednesday 20 July 2022  14:57:24 +0000 (0:00:00.302)       0:00:03.440 ******** 
fatal: [SPINE1 -> localhost]: FAILED! => 
  msg: '''dict object'' has no attribute ''overlay_loopback_interface'''
fatal: [SPINE2 -> localhost]: FAILED! => 
  msg: '''dict object'' has no attribute ''overlay_loopback_interface'''
 ```
 Current template (eos_designs/templates/underlay/interfaces/loopback-interfaces.j2) requires that overlay_loopback_interface is set.  In L2LS design, there is no overlay, and overlay_loopback_interface description is not set.
 
```
loopback_interfaces:
{% if switch.router_id is arista.avd.defined %}
{# BGP Loopback #}
  Loopback0:
    description: "{% include switch.interface_descriptions.overlay_loopback_interface %}"
```

Suggested fix is:
```
loopback_interfaces:
{% if switch.router_id is arista.avd.defined %}
{# BGP Loopback #}
  Loopback0:
{% if switch.interface_descriptions.overlay_loopback_interface is arista.avd.defined %}
    description: "{% include switch.interface_descriptions.overlay_loopback_interface %}"
{% endif %}
```

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
set
design:
  type: "l2ls"

comment out:
      # overlay_loopback_interface: 'interface_descriptions/loopback_interfaces/overlay-loopback.j2'

make build
```

### Relevant log output

```shell
TASK [arista.avd.eos_designs : Generate device configuration in structured format] *****************************************************************************************
Wednesday 20 July 2022  14:57:24 +0000 (0:00:00.302)       0:00:03.441 ******** 
Wednesday 20 July 2022  14:57:24 +0000 (0:00:00.302)       0:00:03.440 ******** 
fatal: [SPINE1 -> localhost]: FAILED! => 
  msg: '''dict object'' has no attribute ''overlay_loopback_interface'''
fatal: [SPINE2 -> localhost]: FAILED! => 
  msg: '''dict object'' has no attribute ''overlay_loopback_interface'''
```

### Contributing Guide

tgodaA:
## Change Summary

In l2ls design there is no overlay. But when the template for overlay_loopback_interface description is not provided in the designs, it errors out as the Loopback0 interface description is dependent on `interface_descriptions.overlay_loopback_interface`, which is null for l2ls design. So this is fixed with a defined test.

## Related Issue(s)

Fixes #1954 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Added a defined test before including the description template for loopback interface.

## How to test

`make refresh-facts` -> should not generate any new facts or delete any. Also, there should be no error while running this.
The change was tested for a l2ls network locally.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/loopback-interfaces.j2
+++ b/ansible_collections/arista/avd/roles/eos_designs/templates/underlay/interfaces/loopback-interfaces.j2
@@ -2,7 +2,9 @@ loopback_interfaces:
 {% if switch.router_id is arista.avd.defined %}
 {# BGP Loopback #}
   Loopback0:
+{%     if switch.interface_descriptions.overlay_loopback_interface is arista.avd.defined %}
     description: "{% include switch.interface_descriptions.overlay_loopback_interface %}"
+{%     endif %}
     shutdown: false
     ip_address: {{ switch.router_id }}/32
 {%     if switch.ipv6_router_id is arista.avd.defined %}


gmuloc (commented):
What will be the Loopback0 description for L2LS then? should we maybe find some `elif` block to add some information?

mthiel117:
> What will be the Loopback0 description for L2LS then? should we maybe find some `elif` block to add some information?

@gmuloc Loopback addresses used in L2LS are typically only defined on the Spines when they run a routing protocol to provide another interface to reach the device.  Typically no description is given.  I think providing a description would be ok, but not needed in this case.  The fix provided in this PR is to validate that overlay_loopback_interface is set before the description is rendered.

IMO... Providing a native loopback interface description for L2LS should be part of another feature PR, and should be called something other than overlay_loopback_interface, since it is not an overlay interface.

gmuloc:
> > What will be the Loopback0 description for L2LS then? should we maybe find some `elif` block to add some information?
> 
> @gmuloc Loopback addresses used in L2LS are typically only defined on the Spines when they run a routing protocol to provide another interface to reach the device. Typically no description is given. I think providing a description would be ok, but not needed in this case. The fix provided in this PR is to validate that overlay_loopback_interface is set before the description is rendered.
> 
> IMO... Providing a native loopback interface description for L2LS should be part of another feature PR, and should be called something other than overlay_loopback_interface, since it is not an overlay interface.

Fair enough - so let's add a test to this PR to capture the new behavior and we are good to go

tgodaA:
@gmuloc its hard to add a test case for this, reason being all of the current evpn designs require the overlay_loopback_description. So, we can add a test for this after l2ls is added.

gmuloc (approved):
LGTM

mthiel117 (approved):
LGTM as well.

carlbuchmann:
> @gmuloc its hard to add a test case for this, reason being all of the current evpn designs require the overlay_loopback_description. So, we can add a test for this after l2ls is added.

Agreed @tgodaA - LGTM!

PR #1970, opened 2022-08-02
Fix(eos_designs): Correct underlay routing for overlay_routing_protocol: "none"

### LINKED ISSUE #1884: EVPN overlay variables generated although EVPN_role is none
### Issue Summary

tried creating l3 router node type and keys where the router will not be running EVPN. 

Defined the node type as :

```yaml
l3router:
    type: l3router
    connected_endpoints: true
    default_evpn_role: none
    mlag_support: true
    ip_addressing:
      mlag_ip_primary: 'templates/ip_addressing/mlag_ip_primary.j2'
      mlag_ip_secondary: 'templates/ip_addressing/mlag_ip_secondary.j2'
      mlag_l3_ip_primary: 'templates/ip_addressing/mlag_l3_ip_primary.j2'
      mlag_l3_ip_secondary: 'templates/ip_addressing/mlag_l3_ip_secondary.j2'    
    network_services: 
      l2: true
      l3: true
    underlay_router: true
    uplink_type: p2p  
```

So, the **evpn_role** is **None**.

End result which appeared wrong of the device type that is not running EVPN
 Name of the prefix list containing EVPN OVERLAY that could be deceiving to the user and does not seem like there is a way to override this name

```eos
   seq 10 permit 2.2.2.0/24 eq 32
route-map RM-CONN-2-BGP permit 10
   match ip address prefix-list PL-LOOPBACKS-EVPN-OVERLAY
```

Also noticed the below under debug vars for this node and appeared in accurate as EVPN role is none:

```yaml
        overlay_address_families:
        - evpn
        overlay_routing_protocol: ebgp
        overlay_routing_protocol_address_family: ipv4
```

eos_designs

### How do you run AVD ?

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Contributing Guide

ClausHolbechArista:
## Change Summary

Correct underlay routing for overlay_routing_protocol: "none"

## Related Issue(s)

Recently we added support for `overlay_routing_protocol: "none"`. The PR did not include changes to underlay routing so the eos_designs behavior was to only redistribute the loopback IP into the underlay. This is wrong when not using an overlay, so instead we should redistribute all connected routes in default VRF to the underlay.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

When `overlay_routing_protocol: "none"` redistribute all routes in default VRF to underlay routing protocol.

## How to test

Will be covered in future L2LS / L3LS molecule scenarios.

No changes to current molecule scenarios.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs-l2ls/intended/configs/BGP-SPINE1.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs-l2ls/intended/configs/BGP-SPINE1.cfg
@@ -99,9 +99,6 @@ ip virtual-router mac-address 00:1c:73:00:00:99
 ip routing
 no ip routing vrf MGMT
 !
-ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
-   seq 10 permit 192.168.255.0/24 eq 32
-!
 mlag configuration
    domain-id BGP_SPINES
    local-interface Vlan4094
@@ -111,9 +108,6 @@ mlag configuration
    reload-delay non-mlag 330
 !
 ip route vrf MGMT 0.0.0.0/0
-!
-route-map RM-CONN-2-BGP permit 10
-   match ip address prefix-list PL-LOOPBACKS-EVPN-OVERLAY
 !
 route-map RM-MLAG-PEER-IN permit 10
    description Make routes learned over MLAG Peer-link less preferred on spines to ensure optimal routing
@@ -137,7 +131,7 @@ router bgp 65001
    neighbor 192.168.253.5 description DUMMY-CORE
    neighbor 192.168.254.1 peer group MLAG-IPv4-UNDERLAY-PEER
    neighbor 192.168.254.1 description BGP-SPINE2
-   redistribute connected route-map RM-CONN-2-BGP
+   redistribute connected
    !
    address-family ipv4
       neighbor IPv4-UNDERLAY-PEERS activate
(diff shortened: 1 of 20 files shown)


ansible_collections/arista/avd/roles/eos_designs/templates/underlay/ebgp/router-bgp.j2
```diff
@@ -25,8 +25,12 @@ router_bgp:
         activate: true
 {% endif %}
   redistribute_routes:
+{% if switch.overlay_routing_protocol == "none" %}
+    connected: {}
+{% else %}
     connected:
       route_map: RM-CONN-2-BGP
+{% endif %}
```
tgodaA, line 33: Any reason, why we are not doing this way?

```suggestion
{% if switch.overlay_routing_protocol == "none" %}
  redistribute_routes:
    connected:
      route_map: RM-CONN-2-BGP
{% endif %}
```

ClausHolbechArista: Those two are  not the same logic. Before this PR we had the redistribution of connected _with_ route-map. In the case on overlay_protocol: none I now want to remove the route_map.

onurgashi (approved):

ansible_collections/arista/avd/molecule/eos_designs-l2ls/intended/configs/BGP-SPINE1.cfg
```diff
@@ -137,7 +131,7 @@ router bgp 65001
    neighbor 192.168.253.5 description DUMMY-CORE
    neighbor 192.168.254.1 peer group MLAG-IPv4-UNDERLAY-PEER
    neighbor 192.168.254.1 description BGP-SPINE2
-   redistribute connected route-map RM-CONN-2-BGP
+   redistribute connected
```
tgodaA, line 134: @ClausHolbechArista , I got what you said with this example.

tgodaA (approved):
LGTM

PR #1971, opened 2022-08-02
Refactor(eos_designs): Move eos_designs structured_config templates to Python

ghost:
## Change Summary

Move eos_designs structured_config to Python

## Related Issue(s)

Fixes #https://github.com/aristanetworks/avd-internal/issues/36
#https://github.com/aristanetworks/avd-internal/issues/37

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,45 @@
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+from functools import cached_property
+
+
+class EosDesigns:
```
ClausHolbechArista: Please call this `AvdFacts`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,148 @@
+from ansible_collections.arista.avd.plugins.module_utils.eos_designs import EosDesigns
+from functools import cached_property
+from ansible_collections.arista.avd.plugins.module_utils.utils import get
+
+
+class AVDStructConfig(EosDesigns):
+    def __init__(
```
ClausHolbechArista: please use __init__ from base class. Notice it uses _hostvars, so you will have to change that reference in the other functions here.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -6,9 +6,10 @@
 from functools import cached_property
 from hashlib import sha256
 from ansible_collections.arista.avd.plugins.module_utils.utils import AristaAvdError, get, default, template_var, AristaAvdMissingVariableError
+from ansible_collections.arista.avd.plugins.module_utils.eos_designs import EosDesigns
 
 
-class EosDesignsFacts:
+class EosDesignsFacts(EosDesigns):
     def __init__(
```
ClausHolbechArista: Move this __init__ to base class

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -33,17 +34,7 @@ def __init__(
...
-        Return the list of "keys"
-
-        Actually the returned list are the names of attributes not starting with "_" and using cached_property class.
-        The "_" check is added to allow support for "internal" cached_properties storing temporary values.
-        '''
-
-        return [key for key in dir(cls) if not key.startswith('_') and isinstance(getattr(cls, key), cached_property)]
+        # super().__init__()
 
     @classmethod
```
ClausHolbechArista: Move this to base class

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,148 @@
...
+    def router_bgp(self):
+        if "bgp_as" in self._switch_data:
+            router_bgp = {'as': self._switch_bgp_as,
+                          'router_id': self._switch_router_id,
+                          'bgp_defaults': []}
+            for bgp_default in self._switch_bgp_defaults:
+                router_bgp['bgp_defaults'].append(bgp_default)
+
+            if "bgp_maximum_paths" in self._host_vars and "bgp_ecmp" in self._host_vars:
+                router_bgp['bgp_defaults'].append("maximum-paths " + str(self._bgp_maximum_paths)
```
ClausHolbechArista: please use f-strings everywhere for variable insertion in strings. It makes it more readable.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,148 @@
...
+        return get(self._range, 'beginning')
+
+    @cached_property
+    def _ending(self):
+        return get(self._range, 'ending')
+
+    @cached_property
+    def static_routes(self):
+        static_routes = []
+        if "mgmt_destination_networks" in self._host_vars and len(self._mgmt_destination_networks) > 0:
```
ClausHolbechArista: It would be simpler to read by just setting `mgmt_destinations_networks = get(self._hostvars, "mgmt_destination_networks")` inside the `static_routes` function and then run `if mgmt_destination_networks` since it will be None if not set.
Same elsewhere. All the small internal properties will only slow things down, since they are written to the internal __dict__.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -5,15 +5,11 @@
...
-from ansible.plugins.filter.core import combine
 from ansible.plugins.loader import lookup_loader
 from ansible.template import Templar
-from ansible_collections.arista.avd.plugins.filter.convert_dicts import convert_dicts
-from ansible_collections.arista.avd.plugins.filter.list_compress import list_compress
-from ansible_collections.arista.avd.plugins.filter.range_expand import range_expand
-from ansible_collections.arista.avd.plugins.filter.natural_sort import natural_sort
 from ansible_collections.arista.avd.plugins.module_utils.eos_designs_facts import EosDesignsFacts
 from ansible_collections.arista.avd.plugins.module_utils.utils import AristaAvdMissingVariableError
+from ansible.plugins.filter.core import combine
```
gmuloc: lets move this back to its original line (line 8)

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -118,12 +114,8 @@ def create_avd_switch_facts_instances(self, fabric_hosts: list, hostvars: object
             avd_switch_facts[host] = {
                 "switch": EosDesignsFacts(
                     hostvars=host_hostvars,
-                    template_lookup_module=template_lookup_module,
                     combine=combine,
-                    list_compress=list_compress,
-                    range_expand=range_expand,
-                    natural_sort=natural_sort,
-                    convert_dicts=convert_dicts,
+                    template_lookup_module=template_lookup_module,
```
gmuloc: same for this one, let's put it back where it was originally

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,82 @@
...
+        hostvars=None,
+        combine=None,
+        template_lookup_module=None
+    ):
+        if hostvars:
+            self._hostvars = hostvars
+        if combine:
+            self._combine = combine
+        if template_lookup_module:
+            self._template_lookup_module = template_lookup_module
```
gmuloc: What is the rational behind not setting the instance attributes if they are `None`? Not saying this is wrong, just trying to understand why we would rather **not** set an attribute than havig it being `None`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,403 @@
...
+                          'bgp_defaults': []}
+            for bgp_default in get(self._switch_data, 'bgp_defaults'):
+                router_bgp['bgp_defaults'].append(bgp_default)
+
+            if get(self._hostvars, 'bgp_maximum_paths') is not None and get(self._hostvars, 'bgp_ecmp') is not None:
+                router_bgp['bgp_defaults'].append(f"maximum-paths {str(get(self._hostvars, 'bgp_maximum_paths'))} "
+                                                  f"ecmp {str(get(self._hostvars, 'bgp_ecmp'))}")
+            elif get(self._hostvars, 'bgp_maximum_paths') is not None:
+                router_bgp['bgp_defaults'].append(f"maximum-paths {str(get(self._hostvars, 'bgp_maximum_paths'))}")
+            return router_bgp
```
gmuloc: Let's try to make the code not a one to one copy of Jinja2 template as we have been using some cumbersome logic in Jinja that can be alleviated by python syntax. E.g.:

```suggestion
    def router_bgp(self):
        bgp_as = get(self._switch_data, "bgp_as")
        if bgp_as is None:
            return

        bgp_defaults = [bgp_default for bgp_default in get(self._switch_data, 'bgp_defaults')]
        if get(self._hostvars, 'bgp_maximum_paths') is not None:
            max_paths_str = f"maximum-paths {str(get(self._hostvars, 'bgp_maximum_paths'))}"
            if get(self._hostvars, 'bgp_ecmp') is not None:
                max_paths_str += f" ecmp {str(get(self._hostvars, 'bgp_ecmp'))}")
            
            bgp_defaults.append(max_path_str)

        return {'as': get(self._switch_data, 'bgp_as'),
                     'router_id': get(self._switch_data, 'router_id'),
                     'bgp_defaults': bgp_defaults}
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,403 @@
...
+    @cached_property
+    def ip_routing(self):
+        return True
+
+    @cached_property
+    def ipv6_unicast_routing(self):
+        if get(self._hostvars, 'underlay_rfc5549') is True:
+            return True
+        elif get(self._switch_data, 'underlay_ipv6') is True:
+            return True
```
gmuloc:
```suggestion
        return get(self._hostvars, 'underlay_rfc5549') is True or get(self._switch_data, 'underlay_ipv6') is True
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,403 @@
...
+
+    @cached_property
+    def hardware_counters(self):
+        hardware_counter = get(self._hostvars, 'hardware_counters')
+        if hardware_counter is not None:
+            hardware_counters = {'features': []}
+            if get(hardware_counter, 'features') is not None:
+                for feature, direction in get(hardware_counter, 'features').items():
+                    hardware_counters['features'].append({feature: direction})
+            return hardware_counters
```
gmuloc:
```suggestion
        if hardware_counter is None:
            return None
        features = []
        if get(hardware_counter, 'features') is not None:
            features = [{feature: direction} for feature, direction in get(hardware_counter, 'features').items()]
 
        return {"features": features}
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,403 @@
...
+        vlan_internal_order = {'allocation': get(self._internal_vlan_order, 'allocation'),
+                               'range': {'beginning': get(range, 'beginning'),
+                                         'ending': get(range, 'ending')}}
+        return vlan_internal_order
+
+    @cached_property
+    def event_monitor(self):
+        event_monitor = get(self._hostvars, 'event_monitor')
+        if event_monitor is not None:
+            event_monitors = {'enabled': 'true'}
```
gmuloc: This is not what is in the template:
```
{% if event_monitor is defined and event_monitor == true %}
event_monitor:
  enabled: true
{% endif %}
```

here we return true if it is defined even if it is false

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
Shivani-chourasiya: Thanks for the review @gmuloc , I will look into your comments today.

ClausHolbechArista: Please rebase. We no longer use this combine. Instead we use our own merge function.

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
ClausHolbechArista: That is to avoid overwriting variables set by a previous template in the template stack. (In this case "template" == "python class"). Since we run a merge on the data, we don't want to get a `nxt` with `None`, since that will blank out any data we already have.
Only exception is the custom_structured_config stuff, where we allow the user to null any variable.

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
Shivani-chourasiya: yes, I have rebased it. I can see we are still having combine in eos_designs_facts , shall I refactor it in this PR?

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
ClausHolbechArista: I missed the area where we are doing this. You are correct, we can just set to the passed value. Actually we should probably remove the default value since the class will not work without these set correctly.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -24,6 +26,15 @@ def run(self, tmp=None, task_vars=None):
 
         root_key = ""
 
+        # READ ANSIBLE VARIABLES AND PERFORM TEMPLATING TO SUPPORT INLINE JINJA
```
gmuloc: No need to have this comment in upper case letters

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -63,7 +74,7 @@ def run(self, tmp=None, task_vars=None):
 
         # If the argument 'debug' is set, a 'avd_yaml_templates_to_facts_debug' list will be added to the output.
         # This list contains timestamps from every step for every template. This is useful for identifying slow templates.
-        # Here we pull in the list from any previous tasks, so we can just add the the list.
+        # Here we pull in the list from any previous tasks, so we can just add the list.
         if debug:
             avd_yaml_templates_to_facts_debug = template_vars.get('avd_yaml_templates_to_facts_debug', [])
```
gmuloc, line 98: Let's add a comment that a this stage, the template_list can contains either templates or python_modules

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @cached_property
+    def _hostname(self):
+        """
+        switch.hostname fact set based on inventory_hostname variable
+        """
+        return get(self._hostvars, "inventory_hostname", required=True)
+
+    @cached_property
+    def router_bgp(self):
```
gmuloc: @ClausHolbechArista , Should we add docstrings for non oneliners?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    max_paths_str += f" ecmp {bgp_ecmp}"
+
+                bgp_defaults.append(max_paths_str)
+
+            return {
+                'as': get(self._switch_data, 'bgp_as'),
+                'router_id': get(self._switch_data, 'router_id'),
+                'bgp_defaults': bgp_defaults,
+            }
+        return None
```
gmuloc:
```suggestion
        if bgp_as is None:
            return None
            
        bgp_defaults = get(self._switch_data, 'bgp_defaults')
        bgp_maximum_paths = get(self._hostvars, 'bgp_maximum_paths')
        bgp_ecmp = get(self._hostvars, 'bgp_ecmp')
        if bgp_maximum_paths is not None:
            max_paths_str = f"maximum-paths {bgp_maximum_paths}"
            if get(self._hostvars, 'bgp_ecmp') is not None:
                max_paths_str += f" ecmp {bgp_ecmp}"

            bgp_defaults.append(max_paths_str)

        return {
            'as': get(self._switch_data, 'bgp_as'),
            'router_id': get(self._switch_data, 'router_id'),
            'bgp_defaults': bgp_defaults,
        }
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                for mgmt_destination_network in self._mgmt_destination_networks:
+                    static_routes.append({'vrf': self._mgmt_interface_vrf,
+                                          'destination_address_prefix': mgmt_destination_network,
+                                          'gateway': self._mgmt_gateway})
+            else:
+                static_routes.append({'vrf': self._mgmt_interface_vrf,
+                                      'destination_address_prefix': '0.0.0.0/0',
+                                      'gateway': self._mgmt_gateway})
+            return static_routes
+        return None
```
gmuloc: is there a reason to return `None` vs `[]` depending upon which variable is not present?

```suggestion
        if self._mgmt_gateway is None:
            return None
        static_routes = []
        mgmt_destination_networks = get(self._hostvars, 'mgmt_destination_network')
        if mgmt_destination_networks is not None:
            for mgmt_destination_network in self._mgmt_destination_networks:
                static_routes.append({'vrf': self._mgmt_interface_vrf,
                                      'destination_address_prefix': mgmt_destination_network,
                                      'gateway': self._mgmt_gateway})
        else:
            static_routes.append({'vrf': self._mgmt_interface_vrf,
                                  'destination_address_prefix': '0.0.0.0/0',
                                  'gateway': self._mgmt_gateway})
        return static_routes
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @cached_property
+    def hardware_counters(self):
+        hardware_counter = get(self._hostvars, 'hardware_counters')
+        if hardware_counter is not None:
+            features = []
+            if get(hardware_counter, 'features') is not None:
+                features = [{feature: direction} for feature, direction in get(hardware_counter, 'features').items()]
+            return {"features": features}
+        return None
```
gmuloc:
```suggestion
        if hardware_counter is not None:
            features = []
            if get(hardware_counter, 'features') is None:
                return None
            features = [{feature: direction} for feature, direction in get(hardware_counter, 'features').items()]
        return {"features": features}
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                for platform_item in platform_speed_groups:
+                    if "platform" in platform_item and "speeds" in platform_item:
+                        platform = platform_item
+                    else:
+                        platform = {"platform": platform_item, "speeds": platform_speed_groups[platform_item]}
+                    if platform['platform'] == switch_platform:
+                        speeds = self._convert_dicts(get(platform, 'speeds'), 'speed', 'speed_groups')
+                        for speed in self._natural_sort(speeds, 'speed'):
+                            for speed_group in speed['speed_groups']:
+                                tmp_speed_groups.update({speed_group: speed['speed']})
```
gmuloc: where is the return statement for this block? am I missing something?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            if tmp_speed_groups:
+                hardware = {'speed_groups': {}}
+                for speed_group in self._natural_sort(tmp_speed_groups):
+                    hardware['speed_groups'].update({speed_group: {'serdes': tmp_speed_groups[speed_group]}})
+                return hardware
```
gmuloc: is this maybe too indented?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        hardware_counter = get(self._hostvars, 'hardware_counters')
+        if hardware_counter is not None:
+            features = []
+            if get(hardware_counter, 'features') is not None:
+                features = [{feature: direction} for feature, direction in get(hardware_counter, 'features').items()]
+            return {"features": features}
+        return None
+
+    @cached_property
+    def hardware(self):
```
gmuloc: refactor to have less indentation if possible

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -120,10 +116,6 @@ def create_avd_switch_facts_instances(self, fabric_hosts: list, hostvars: object
                     hostvars=host_hostvars,
                     template_lookup_module=template_lookup_module,
                     combine=combine,
```
ClausHolbechArista: Let's also move combine to the class (if needed).

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -8,9 +8,11 @@
 from ansible.errors import AnsibleActionFail
 from ansible.utils.vars import isidentifier
 from ansible.plugins.loader import lookup_loader
+from ansible.parsing.yaml.dumper import AnsibleDumper
 from ansible_collections.arista.avd.plugins.module_utils.strip_empties import strip_null_from_data
 from ansible_collections.arista.avd.plugins.plugin_utils.merge import merge
 from datetime import datetime
+import importlib
```
ClausHolbechArista: According to pep-8 (https://peps.python.org/pep-0008/#imports) we should move this above line 7

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -63,7 +74,7 @@ def run(self, tmp=None, task_vars=None):
 
         # If the argument 'debug' is set, a 'avd_yaml_templates_to_facts_debug' list will be added to the output.
         # This list contains timestamps from every step for every template. This is useful for identifying slow templates.
-        # Here we pull in the list from any previous tasks, so we can just add the the list.
+        # Here we pull in the list from any previous tasks, so we can just add the list.
```
ClausHolbechArista: _to_ the list

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -88,24 +95,39 @@ def run(self, tmp=None, task_vars=None):
...
+                    if debug:
+                        debug_item['timestamps']['strip_empty_keys'] = datetime.now()
 
-            # Load data from the template result.
-            template_result_data = yaml.safe_load(template_result[0])
+                    template_result_data = strip_null_from_data(template_result_data)
+
+            elif 'python_module' in template_item:
+                module_path = template_item.get('python_module')
+                class_name = self._task.args.get("class_name")
```
ClausHolbechArista: 1. We should set class_name per python_module instead of per task.
2. We need a default class name in case it is not set.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -88,24 +95,39 @@ def run(self, tmp=None, task_vars=None):
...
+                cls = getattr(importlib.import_module(module_path), class_name)
+                obj = cls(hostvars=template_vars)
```
ClausHolbechArista: Could we rename `obj` to `cls_instance`

ClausHolbechArista: Also add a check  that `cls` is an instance of our base class.

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -88,24 +95,39 @@ def run(self, tmp=None, task_vars=None):
...
 
-            # If the argument 'strip_empty_keys' is set, remove keys with value of null / None from the resulting dict (recursively).
-            if strip_empty_keys:
                 if debug:
-                    debug_item['timestamps']['strip_empty_keys'] = datetime.now()
+                    debug_item['timestamps']['render_python_class'] = datetime.now()
 
-                template_result_data = strip_null_from_data(template_result_data)
+                template_result_data = obj.render()
```
ClausHolbechArista: 1. Let's add an `if` before that checks with `getattr` if `render` is available, and if not give an error.
2. Enclose the render in a try/except and raise `AnsibleActionFail`

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,82 @@
...
+    from ansible_collections.arista.avd.plugins.filter.list_compress import list_compress
+    from ansible_collections.arista.avd.plugins.filter.range_expand import range_expand
+    from ansible_collections.arista.avd.plugins.filter.natural_sort import natural_sort
+except ImportError as imp_exc:
+    FILTER_IMPORT_ERROR = imp_exc
+else:
+    FILTER_IMPORT_ERROR = None
+
+
+class AVDFacts:
```
ClausHolbechArista: Rename to `AvdFacts`

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __init__(
+        self,
+        hostvars=None,
+        combine=None,
```
ClausHolbechArista: as mentioned in the other file, move combine to use the same method as the rest.

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -6,80 +6,15 @@
...
-    def render(self):
-        '''
-        Return a dictionary of all @cached_property values.
-
-        If the value is cached, it will automatically get returned from cache
-        If the value is not cached, it will be resolved by the attribute function first.
-        Empty values are removed from the returned data.
-        '''
-        return {key: getattr(self, key) for key in self.keys() if getattr(self, key) is not None}
+    def __init__(self, hostvars=None,
```
ClausHolbechArista: Why do we need this?

ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
```diff
@@ -5,7 +5,8 @@ default_templates:
     structured_config:
       # Render Structured Configuration
       # Base features
-      - template: "base/main.j2"
+#      - template: "base/main.j2"
```
ClausHolbechArista: Remove the old line. Same below

ansible_collections/arista/avd/roles/eos_designs/templates/base/hardware-counters.j2
```diff
@@ -1,7 +1,7 @@
-{% if hardware_counters is defined and hardware_counters is not none %}
+{% if hardware_counters is arista.avd.defined %}
```
ClausHolbechArista: Since the new base.py should cover all the base templates, I think we should just delete them in this PR.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+from functools import cached_property
+from ansible_collections.arista.avd.plugins.module_utils.utils import get
+from ansible_collections.arista.avd.plugins.filter.snmp_hash import hash_passphrase
+from hashlib import sha1
+
+
+class AVDStructConfig(AVDFacts):
+
+    def __init__(self, hostvars=None):
+        self._hash_passphrase = hash_passphrase
```
ClausHolbechArista: since it is imported in this file, I don't think we need this. Just reference the imported name directly in the functions.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+    @cached_property
+    def _mgmt_interface_vrf(self):
+        return get(self._hostvars, 'mgmt_interface_vrf')
+
+    @cached_property
+    def _mgmt_gateway(self):
+        return get(self._hostvars, 'mgmt_gateway')
+
+    @cached_property
+    def _switch_data(self):
```
ClausHolbechArista: 1. please call it `_switch` to make it as close to `switch.xxx` in the old templates.
2. I would asume this is required, so we should error out if it is missing. (Just set `required=True`)

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+    @cached_property
+    def _platform_settings(self):
+        return get(self._switch_data, 'platform_settings')
+
+    @cached_property
+    def _mgmt_ip(self):
+        return get(self._switch_data, 'mgmt_ip')
+
+    @cached_property
+    def _hostname(self):
```
ClausHolbechArista: Please take the hostname from `switch.hostname` instead of `inventory_hostname`.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            bgp_ecmp = get(self._hostvars, 'bgp_ecmp')
+            if bgp_maximum_paths is not None:
+                max_paths_str = f"maximum-paths {bgp_maximum_paths}"
+                if get(self._hostvars, 'bgp_ecmp') is not None:
+                    max_paths_str += f" ecmp {bgp_ecmp}"
+
+                bgp_defaults.append(max_paths_str)
+
+            return {
+                'as': get(self._switch_data, 'bgp_as'),
```
ClausHolbechArista: Should use the `bgp_as` variable set in line 47

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+    def _hostname(self):
+        """
+        switch.hostname fact set based on inventory_hostname variable
+        """
+        return get(self._hostvars, "inventory_hostname", required=True)
+
+    @cached_property
+    def router_bgp(self):
+        bgp_as = get(self._switch_data, "bgp_as")
+        if bgp_as is not None:
```
ClausHolbechArista: Change to
```python
        if bgp_as is None:
            return None

        bgp_defaults = ....
```

ClausHolbechArista: Same style elsewhere. It will give less indentation and hopefully easier to follow.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                'router_id': get(self._switch_data, 'router_id'),
+                'bgp_defaults': bgp_defaults,
+            }
+        return None
+
+    @cached_property
+    def static_routes(self):
+        if self._mgmt_gateway is not None:
+            static_routes = []
+            mgmt_destination_networks = get(self._hostvars, 'mgmt_destination_network')
```
ClausHolbechArista: I think you are missing the final `s` in `mgmt_destination_networks`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            if mgmt_destination_networks is not None:
+                for mgmt_destination_network in self._mgmt_destination_networks:
```
ClausHolbechArista: This one should be
```python
for mgmt_destination_network in mgmt_destination_networks:
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        ):
+            router_multicast = {'ipv4': {'routing': True}}
+            if get(self._switch_data, 'evpn_multicast') is True:
+                router_multicast['ipv4'].update({'software_forwarding': 'sfe'})
+            return router_multicast
+        return None
+
+    @cached_property
+    def hardware_counters(self):
+        hardware_counter = get(self._hostvars, 'hardware_counters')
```
ClausHolbechArista: Change this to read `hardware_counters.features` directly. Call the variable `hardware_counters_features`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @cached_property
+    def hardware(self):
+        platform_speed_groups = get(self._hostvars, 'platform_speed_groups')
+        if platform_speed_groups is not None:
+            tmp_speed_groups = {}
+            switch_platform = get(self._switch_data, 'platform')
+            if switch_platform is not None:
+                for platform_item in platform_speed_groups:
+                    if "platform" in platform_item and "speeds" in platform_item:
```
ClausHolbechArista: please add a few comments here, so we know why we accept two different data models here. Some of it can be removed when we deprecate the old dicts.

ClausHolbechArista: Actually it would be better to use self._convert_dicts directly with primary key "platform" and secondary key "speeds", so we can remove them all by just searching later.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    if "arista.io" in cvp_ip:
+                        cvaas['ips'].append(cvp_ip)
+                    else:
+                        cvp_on_prem['ips'].append(cvp_ip)
+
+            daemon_terminattr = {'cvaddrs': []}
+            if cvp_on_prem['ips']:
+                for cvp_instance_ip in cvp_on_prem['ips']:
+                    daemon_terminattr['cvaddrs'].append(f"{cvp_instance_ip}:"
+                                                        f"{get(self._hostvars, 'terminattr_ingestgrpcurl_port')}")
```
ClausHolbechArista: would be more readable by setting a variable above.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                daemon_terminattr.update({'cvauth': {'method': 'key', 'key': get(self._hostvars,
+                                                                                 'cvp_ingestauth_key')}})
+            elif cvaas['ips']:
+                for cvp_instance_ip in cvaas['ips']:
+                    daemon_terminattr['cvaddrs'].append(f"{cvp_instance_ip}:443")
+                daemon_terminattr.update({'cvauth': {'method': 'token-secure',
+                                                     'token_file': get(self._hostvars, 'cvp_token_file',
+                                                                       '/tmp/cv-onboarding-token')}})
+
+            daemon_terminattr.update({'cvvrf': self._mgmt_interface_vrf})
```
ClausHolbechArista: Please change to a single update with a larger dict, that you can unfold to still make it more readable.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            if cvp_on_prem['ips']:
+                for cvp_instance_ip in cvp_on_prem['ips']:
+                    daemon_terminattr['cvaddrs'].append(f"{cvp_instance_ip}:"
+                                                        f"{get(self._hostvars, 'terminattr_ingestgrpcurl_port')}")
+                daemon_terminattr.update({'cvauth': {'method': 'key', 'key': get(self._hostvars,
+                                                                                 'cvp_ingestauth_key')}})
+            elif cvaas['ips']:
+                for cvp_instance_ip in cvaas['ips']:
+                    daemon_terminattr['cvaddrs'].append(f"{cvp_instance_ip}:443")
+                daemon_terminattr.update({'cvauth': {'method': 'token-secure',
```
ClausHolbechArista: please unfold the json to make it more readable.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                        cvaas['ips'].append(cvp_ip)
+                    else:
+                        cvp_on_prem['ips'].append(cvp_ip)
+
+            daemon_terminattr = {'cvaddrs': []}
+            if cvp_on_prem['ips']:
+                for cvp_instance_ip in cvp_on_prem['ips']:
+                    daemon_terminattr['cvaddrs'].append(f"{cvp_instance_ip}:"
+                                                        f"{get(self._hostvars, 'terminattr_ingestgrpcurl_port')}")
+                daemon_terminattr.update({'cvauth': {'method': 'key', 'key': get(self._hostvars,
```
ClausHolbechArista: please unfold the json to make it more readable

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            daemon_terminattr.update({'cvvrf': self._mgmt_interface_vrf})
+            daemon_terminattr.update({'smashexcludes': get(self._hostvars, 'terminattr_smashexcludes')})
+            daemon_terminattr.update({'ingestexclude': get(self._hostvars, 'terminattr_ingestexclude')})
+            daemon_terminattr.update({'disable_aaa': get(self._hostvars, 'terminattr_disable_aaa', False)})
+            return daemon_terminattr
+        return None
+
+    @cached_property
+    def vlan_internal_order(self):
+        range = get(self._internal_vlan_order, 'range')
```
ClausHolbechArista: We should favor readability here, so please just get `range.beginning` and `range.ending` directly.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+    @cached_property
+    def event_monitor(self):
+        if get(self._hostvars, 'event_monitor') is True:
+            return {'enabled': 'true'}
+        return None
+
+    @cached_property
+    def event_handlers(self):
+        event_handlers = get(self._hostvars, 'event_handlers')
+        if event_handlers is not None and hasattr(event_handlers, '__iter__'):
```
ClausHolbechArista: We don't have a check like this anywhere else. Why do we need it here?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            event_handlers = self._convert_dicts(event_handlers, 'name')
+            for handler in event_handlers:
+                handler_name = get(handler, 'name')
+                event_handler = {handler_name: {}}
```
ClausHolbechArista: you are overwriting this variable on each loop, so it will only contain the last event handler.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                action = get(handler, 'action')
+                action_type = get(handler, 'action_type')
```
ClausHolbechArista: for these simple gets, we can just use the regular `<dict>.get()`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        return None
+
+    @cached_property
+    def queue_monitor_length(self):
+        queue_monitor_length = get(self._hostvars, 'queue_monitor_length')
+        if queue_monitor_length is not None:
+            queue_monitor_length_dict = {'enabled': True}
+            queue_monitor_length_notifying = get(queue_monitor_length, 'notifying')
+            feature_support = get(self._platform_settings, 'feature_support')
+            queue_monitor_length_notify = ""
```
ClausHolbechArista: let's call this variable something that more clearly tell that this is about feature support. Ex. `notify_supported = get(self._platform_settings, "feature_support.queue_monitor_length_notify")`.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    {
+                        'vrf': self._mgmt_interface_vrf
+                    },
+                'nodes': name_servers
+            }
+        return None
+
+    @cached_property
+    def redundancy(self):
+        redundancy = get(self._hostvars, 'redundancy')
```
ClausHolbechArista: refactor to get `redundancy.protocol` directly. You can just `return get(self._hostvars, 'redundancy.protocol')` since it will return `None` if not found.

ClausHolbechArista: Same elsewhere

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        if snmp_settings is not None:
+            snmp_server = {}
+            compute_local_engineid = get(snmp_settings, 'compute_local_engineid')
+            if compute_local_engineid is True:
+                local_engine_id = sha1(f"{self._hostname}{self._mgmt_ip}".encode('utf-8')).hexdigest()
+                snmp_server.update({'engineid': {'local': local_engine_id}})
+            contact = get(snmp_settings, 'contact')
+            if contact is not None:
+                snmp_server.update({'contact': contact})
+            if get(snmp_settings, 'location') is not None:
```
ClausHolbechArista: please refactor this to add the elements to a list and then use a list comprehension to check for `is not None` and finally use `" ".join(location_elements)`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        snmp_settings = get(self._hostvars, 'snmp_settings')
+        if snmp_settings is not None:
+            snmp_server = {}
+            compute_local_engineid = get(snmp_settings, 'compute_local_engineid')
+            if compute_local_engineid is True:
+                local_engine_id = sha1(f"{self._hostname}{self._mgmt_ip}".encode('utf-8')).hexdigest()
+                snmp_server.update({'engineid': {'local': local_engine_id}})
+            contact = get(snmp_settings, 'contact')
+            if contact is not None:
+                snmp_server.update({'contact': contact})
```
ClausHolbechArista: is this faster/better than just `snmp_server['contact'] = contact`?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    queue_monitor.update({'vrf': vrf})
+                return queue_monitor
+        return None
+
+    @cached_property
+    def management_api_http(self):
+        management_eapi = get(self._hostvars, 'management_eapi')
+        if management_eapi is not None:
+            management_api_http = {'enable_vrfs': {self._mgmt_interface_vrf: {}}}
+            if get(management_eapi, 'enable_http') is not None:
```
ClausHolbechArista: We are not adding anything here, so we could consider using `management_api_http = management_eapi.fromkeys(['enable_http','enable_https','default_services'])` and then update with the enable_vrfs.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,82 @@
...
+        '''
+
+        return [key for key in dir(cls) if not key.startswith('_') and isinstance(getattr(cls, key), cached_property)]
+
+    @classmethod
+    def internal_keys(cls):
+        '''
+        Return the list of "keys" starting with _
+
+        Actually the returned list are the names of attributes starting with "_" and using cached_property class.
```
gmuloc:
```suggestion
        Return a list containing the names of attributes starting with "_" and using cached_property class.
```

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
```diff
@@ -0,0 +1,82 @@
...
+        if key in self.keys():
+            return getattr(self, key)
+        return default_value
+
+    def reset(self):
+        '''
+        Reset the cached attribute values
+        '''
+
+        self.__dict__ = {}
```
gmuloc: Have we tested this? This would remove every single attribute not only the cached ones right?

I think we should rather do something like:

```
for key in self.keys():
  del self.__dict__[key]
for key in self.internal_keys():
  del self.__dict__[key]

```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                for speed_group in self._natural_sort(tmp_speed_groups):
+                    hardware['speed_groups'].update({speed_group: {'serdes': tmp_speed_groups[speed_group]}})
+                return hardware
+        return None
+
+    @cached_property
+    def daemon_terminattr(self):
+        cvp_instance_ip = get(self._hostvars, 'cvp_instance_ip')
+        cvp_instance_ips = get(self._hostvars, 'cvp_instance_ips')
+        if cvp_instance_ip is not None or cvp_instance_ips is not None:
```
gmuloc: Lets return None to decrease indentation of the method as in other methods

```suggestion
        if cvp_instance_ip is None and cvp_instance_ips is None:
          return None
        cvp_on_prem = {"ips": []}
        ...
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                if "arista.io" in cvp_instance_ip:
+                    cvaas['ips'].append(cvp_instance_ip)
+                else:
+                    cvp_on_prem['ips'].append(cvp_instance_ip)
+            elif cvp_instance_ips is not None:
+                for cvp_ip in cvp_instance_ips:
+                    if "arista.io" in cvp_ip:
+                        cvaas['ips'].append(cvp_ip)
+                    else:
+                        cvp_on_prem['ips'].append(cvp_ip)
```
gmuloc: So if both `cvp_instance_ip` and `cvp_instance_ips` are set, we only consider the first variable?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: If we are only adding one key in the dictionary why don't we simply do this:
```suggestion
                daemon_terminattr['cvauth'] = {
                    'method': 'key', 
                    'key': get(self._hostvars, 'cvp_ingestauth_key')
                  }
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: What happens if we have both cvaas and cvp_on_prems IPs ?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                                tmp_speed_groups.update({speed_group: speed['speed']})
+            if tmp_speed_groups:
+                hardware = {'speed_groups': {}}
+                for speed_group in self._natural_sort(tmp_speed_groups):
+                    hardware['speed_groups'].update({speed_group: {'serdes': tmp_speed_groups[speed_group]}})
+                return hardware
+        return None
+
+    @cached_property
+    def daemon_terminattr(self):
```
gmuloc: I feel this function is a bit cumbersome because of the cvaas / non cvaas distinction. I am wondering if it would not be worth to set some flag `is_cvaas` early on when we parse the two variables (cvp_instance_ip / cvp_instance_ips) that we can use afterwards.

Also I think we should be able to warn if there is a mix of cvaas and non cvaas IPs

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: so here I am not sure we are in it for the performance but looking at this: https://discuss.codecademy.com/t/should-we-use-update-or-value-assignment/461981/3 we could also just use regular assignment rather than `update`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    event_handler[handler_name].update({'regex': regex})
+            return event_handler
+        return None
+
+    @cached_property
+    def load_interval(self):
+        load_interval_default = get(self._hostvars, 'load_interval_default')
+        if load_interval_default is not None:
+            return {'default': load_interval_default}
+        return None
```
gmuloc: Starting python 3.8 we could write this
```suggestion
        if (load_interval_default := get(self._hostvars, 'load_interval_default')) is not None:
            return {'default': load_interval_default}
        return None
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    snmp_location = f"{snmp_location} {pod_name}"
+                rack = get(self._switch_data, 'rack')
+                if rack is not None:
+                    snmp_location = f"{snmp_location} {rack}"
+                snmp_location = f"{snmp_location} {self._hostname}"
+                snmp_server.update({'location': snmp_location})
+            users = get(snmp_settings, 'users')
+            if users is not None:
+                snmp_server.update({'users': []})
+                for user in users:
```
gmuloc: lets do something like this:
```
users_dict = {}
for user in users:
  ... current for loop
if users_dict:
  snmp_server["users"] = users_dict
```


Additionally the for loop can be simplified - there were some things written only because of Jinja

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                                user_dict.update({'priv_passphrase': priv_passphrase})
+                    snmp_server['users'].append(user_dict)
+            return snmp_server
+        return None
+
+    @cached_property
+    def spanning_tree(self):
+        spanning_tree_root_super = get(self._switch_data, 'spanning_tree_root_super')
+        spanning_tree_mode = get(self._switch_data, 'spanning_tree_mode')
+        if spanning_tree_root_super is True or spanning_tree_mode is not None:
```
gmuloc: as usual let's prefer a syntax like

```suggestion
        if spanning_tree_root_super is False and spanning_tree_mode is None:
            return None
        ...
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            spanning_tree = {}
+            if spanning_tree_root_super is True:
+                spanning_tree.update({'root_super': True})
+            if spanning_tree_mode is not None:
+                spanning_tree.update({'mode': spanning_tree_mode})
```
gmuloc: use a variable to assign `get(self._switch_data, 'spanning_tree_priority', ;32768')` and use it in the code for readbility

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+        unsupported_transceiver = get(self._hostvars, 'unsupported_transceiver')
+        if unsupported_transceiver is not None:
+            return {'license_name': get(unsupported_transceiver, 'license_name'),
+                    'license_key': get(unsupported_transceiver, 'license_key')}
+        return None
+
+    @cached_property
+    def local_users(self):
+        local_users = get(self._hostvars, 'local_users')
+        if local_users is not None:
```
gmuloc:
```
if local_users is None:
  return None
...
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+                    local_users_dict[name].update({'ssh_key': ssh_key})
+            return local_users_dict
+        return None
+
+    @cached_property
+    def clock(self):
+        timezone = get(self._hostvars, 'timezone')
+        if timezone is not None:
+            return {'timezone': timezone}
+        return None
```
gmuloc: @ClausHolbechArista at the end do we remove vairables like `{'timezone': None}`? Just wondering if we could just replace this with

```
return {'timezone': get(self._hostvars, 'timezone')}
```
and let the rest of the code deal with it.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            platform.update({'sand': {'lag': {'hardware_only': lag_hardware_only}}})
+        if (
+                get(self._platform_settings, 'trident_forwarding_table_partition') is not None and
+                get(self._switch_data, 'evpn_multicast') is True
+        ):
+            platform.update({'trident': {'forwarding_table_partition':
+                            get(self._platform_settings, 'trident_forwarding_table_partition')}})
+        if platform:
+            return platform
+        return None
```
gmuloc: Hmmm so we want to return `None` rather than `{}` - @ClausHolbechArista  maybe when we strip for empty variable we could also include empty lists and dicts? or are there use cases I am missing?

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @cached_property
+    def mac_address_table(self):
+        mac_address_table = get(self._hostvars, 'mac_address_table')
+        if mac_address_table is not None:
+            aging_time = get(mac_address_table, 'aging_time')
+            if aging_time is not None:
+                return {'aging_time': aging_time}
+        return None
```
gmuloc:
```suggestion
        mac_address_table = get(self._hostvars, 'mac_address_table')
        if mac_address_table is None or (aging_time := get(mac_address_table, 'aging_time')) is None:
            return None
        return {'aging_time': aging_time}
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+
+    @cached_property
+    def queue_monitor_streaming(self):
+        queue_monitor_streaming = get(self._hostvars, 'queue_monitor_streaming')
+        if queue_monitor_streaming is not None:
```
gmuloc: reverse the logic
```
if queue_monitor_streaming is None:
    return None
...
```

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,559 @@
...
+            ):
+                return {
+                    'port_id':
+                        {'range':
+                            {'begin': begin,
+                             'end': end
+                             }
+                         }
+                }
+        return None
```
gmuloc:
```suggestion
    def lacp(self):
        lacp_port_id = get(self._switch_data, 'lacp_port_id')
        if lacp_port_id is None:
            return None
 
        begin = get(lacp_port_id, 'begin')
        end = get(lacp_port_id, 'end')
        if begin is not None and end is not None:
            return {
                'port_id':
                    {'range':
                        {'begin': begin,
                         'end': end,
                         }
                     }
                }
            return None
```

gmuloc (dismissed):

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
ClausHolbechArista: It will not remove attributes. It will only remove the underlying __dict__ construct, which is how @cached_property works. I have tested it once.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: This is not really making it simpler to read.

ClausHolbechArista: Instead we should use `get` feature of dot-notation, so you can directly read `mac_address_table.aging_time`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: We have some data models that require an empty dict, but I don't remember any root-level keys. To avoid any changes to existing behavior, I think we should wait with such changes.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: No we will not "dive in" to the returned data models. Only check the outermost key. This is a slight deviation from the previous YAML method, but is much simpler and quicker. Also it was required to skip the "remove_empties" stuff from the python area because of some old issues/bugs.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: this goes in general. We need more helper variables.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: If we choose to do this here, we should do it all over.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: IMO We should change this to support mixing the types. We could just run through the list of servers and append a full dict per server instead of iterating twice. More changes will be required here, since we also need to support token auth for on-prem.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: agreed

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: The block is basically just preparing the `tmp_speed_groups` dict which is then parsed below.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: lets do the full check for `hardware_counters.features` instead of multiple levels.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: Would be nice, but I am not sure how much value they would bring.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: Only situation where we would return [] is when mgmt_destination_networks is []. Only downside is that it will trigger some empty documentation tables in eos_cli_config_gen.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: `[bgp_default for bgp_default in get(self._switch_data, 'bgp_defaults')]` should have a default value in case `bgp_defaults` is not set. `[bgp_default for bgp_default in get(self._switch_data, 'bgp_defaults', [])]`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
ClausHolbechArista: This change would return `False` and not `None` in the non-positive case. That might not be hitless.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: yeah it was like this before the suggestion as well - maybe we can test on the empty list at the end and return None then

ansible_collections/arista/avd/plugins/module_utils/eos_designs.py
gmuloc: As agreed with Claus in our chat @Shivani-chourasiya let's remove the reset method for now

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: hmm so maybe not worse doing it then

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
gmuloc: ok resolving this then

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
Shivani-chourasiya: it was - `{% if event_handlers is defined and event_handlers is iterable %}`

ClausHolbechArista: Ok, I think that is an early attempt at better input validation. Since we will get schemas now, I think we can skip this check. Just check if not None like everything else. Thanks

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import (absolute_import, division, print_function)
```
ClausHolbechArista: Let's call this file avdfacts.py

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,658 @@
...
+
+    @cached_property
+    def ip_routing(self):
+        """
+        ip_routing set to True
+        """
+        return True
+
+    @cached_property
+    def ipv6_unicast_routing(self):
```
ClausHolbechArista: some of these functions were cleaner before, where we did not invert to logic. I understand the change, since we want to return instead of indent larger sections of code, but in this one and some of the others, it ends up being harder to read because we detect for case None and return True by default. We should only do the inverted var check and return None, when it is larger code sections that would be indented.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,658 @@
...
+    def hardware(self):
+        """
+        hardware set based on platform_speed_groups variable and switch.platform fact.
+        Converting nested dict to list of dict to support avd_v4.0
+        """
+        if (platform_speed_groups := get(self._hostvars, 'platform_speed_groups')) is None:
+            return None
+
+        tmp_speed_groups = {}
+        if (switch_platform := get(self._switch, 'platform')) is not None:
```
ClausHolbechArista: this check should be moved together with the check above with an `or`

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,658 @@
...
+        cvp_instance_ips = get(self._hostvars, 'cvp_instance_ips')
+        if cvp_instance_ip is None and cvp_instance_ips is None:
+            return None
+        cvp_instance_ip_list = []
+        if cvp_instance_ip is not None:
+            cvp_instance_ip_list.append(cvp_instance_ip)
+        elif cvp_instance_ips is not None:
+            for cvp_ip in cvp_instance_ips:
+                cvp_instance_ip_list.append(cvp_ip)
+
```
ClausHolbechArista: This can be refactored and simplified a bit:
```suggestion
        cvp_instance_ip = get(self._hostvars, 'cvp_instance_ip')
        cvp_instance_ip_list = get(self._hostvars, 'cvp_instance_ips', [])
        if cvp_instance_ip is not None:
            cvp_instance_ip_list.append(cvp_instance_ip)

        if not cvp_instance_ip_list:
            return None   
```
And then remove the if check below, since it is included here.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,658 @@
...
+            event_handler[handler_name] = {}
+            action = handler.get('action')
+            action_type = handler.get('action_type')
+            if action is not None and action_type is not None:
+                event_handler[handler_name]['action_type'] = action_type
+                event_handler[handler_name]['action'] = action
+            if (delay := handler.get('delay')) is not None:
+                event_handler[handler_name]['delay'] = delay
+            if get(handler, 'asynchronous') is True:
+                event_handler[handler_name]['asynchronous'] = 'handler.asynchronous'
```
ClausHolbechArista: This was a bug in the old code. This should just be the value of the input value like all the others.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,658 @@
...
+    def event_monitor(self):
+        """
+        event_monitor set based on event_monitor data-model
+        """
+        if get(self._hostvars, 'event_monitor') is not True:
+            return None
+        return {'enabled': 'true'}
+
+    @cached_property
+    def event_handlers(self):
```
ClausHolbechArista: there is no difference between the input and output data models, so we can just relay it through without modifications.
```
        return get(self._hostvars, 'event_handlers')
```

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -76,38 +96,59 @@ def run(self, tmp=None, task_vars=None):
...
+                if strip_empty_keys:
+                    if debug:
+                        debug_item['timestamps']['strip_empty_keys'] = datetime.now()
+
+                    template_result_data = strip_null_from_data(template_result_data)
+
+            elif 'python_module' in template_item:
+                module_path = template_item.get('python_module')
+                class_name = template_item.get('python_class_name', DEFAULT_PYTHON_CLASS_NAME)
+                cls = getattr(importlib.import_module(module_path), class_name)
```
ClausHolbechArista: please add a try/except around this import - in case the file is not there.

ansible_collections/arista/avd/roles/eos_designs/templates/base/base.py
```diff
@@ -0,0 +1,671 @@
...
+        Returns the value for switch.mgmt_ip fact used in snmp_server and management_interfaces data-models
+        """
+        return get(self._switch, 'mgmt_ip')
+
+    @cached_property
+    def _hostname(self):
+        """
+        hostname variable set based on switch.hostname fact
+        """
+        return get(self._switch, "hostname", required=True)
```
ClausHolbechArista: everywhere where we pull a *required* variable from a local _switch or similar, we need to have the full key name in `org_key`. So in this example add `org_key="switch.hostname"`. This is required to produce useful error messages

ClausHolbechArista (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/base.py
```diff
@@ -0,0 +1,664 @@
...
+        we will use hash_passphrase filter to create an instance of hashlib._hashlib.HASH corresponding to the auth_type
+        value based on various snmp_settings.users information.
+        """
+        if (snmp_settings := get(self._hostvars, 'snmp_settings')) is None:
+            return None
+
+        snmp_server = {}
+        if (compute_local_engineid := snmp_settings.get('compute_local_engineid')) is True:
+            local_engine_id = sha1(f"{self._hostname}{self._mgmt_ip}".encode('utf-8')).hexdigest()
+            snmp_server['engineid'] = {'local': local_engine_id}
```
curtisarista:
```suggestion
            snmp_server['engine_ids'] = {'local': local_engine_id}
```
@ClausHolbechArista we see that eos_cli_config_gen snmp-settings.j2 expects snmp_server['engine_ids'] as input but eos_designs is outputting snmp_server['engineid']

reference from eos_cli_config_gen/templates/eos/snmp-settings.j2:

```jinja2
{# eos - SNMP Settings#}
{% if snmp_server is arista.avd.defined %}
!
{%     if snmp_server.engine_ids.local is arista.avd.defined %}
snmp-server engineID local {{ snmp_server.engine_ids.local }}
{%     endif %}
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/base.py
```diff
@@ -0,0 +1,664 @@
...
+
+    @cached_property
+    def snmp_server(self):
+        """
+        snmp_server set based on snmp_settings data-model, using various snmp_settings information.
+
+        if snmp_settings.compute_local_engineid is True we will use sha1 to create a
+        unique local_engine_id value based on switch.hostname and switch.mgmt_ip facts.
+
+        If user.version is set to 'v3', compute_local_engineid and compute_v3_user_localized_key are set to 'True'
```
curtisarista, line 333: either the documentation or the template is incorrect.  the template uses `compute_v3_user_localized_key` but the document references `compute_v3_localized_key`

[management settings doc](https://github.com/aristanetworks/ansible-avd/blob/58d3a1e61a0715951f0d9a490b38c6601c4cb040/ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md?plain=1#L84)

curtisarista: @ClausHolbechArista

carlbuchmann: @gmuloc, which one should we honor, the template or the document?

ClausHolbechArista: let's use the template key, so we need to update the docs.

carlbuchmann (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/base.py
ClausHolbechArista: @carlbuchmann will submit a separate PR on the doc fix

PR #1997, opened 2022-08-12
Refactor(eos_designs): Optimize templating and variable handling

ClausHolbechArista:
## Change Summary

Optimize templating and variable handling

By leveraging Python ChainMap, we can avoid copying/merging variables but instead chain them together. This can speed up the overall operation of eos_designs, but especially with templating, where the `ansible.builtin.template` lookup module makes a deepcopy of all variables, we can now run our own variant which will not copy the variables.

This in turn allows us to expose the full hostvars to the custom ip_addressing and interface_descriptions templates with no performance penalty, rendering the `custom_template_extra_vars` option redundant.

## Component(s) name

`arista.avd.eos_designs`
`arista.avd.yaml_templates_to_facts`
`arista.avd.eos_designs_facts`

## Proposed changes

- Replace merging of variables with chainmaps
- Replace the use of `ansible.builtin.template` lookup module with our own variant. We will still leverage the Ansible `Templar` which is the wrapper around Jinja2 that loads extensions and plugins.
- Remove the now redundant `custom_template_extra_vars` key
- Update release notes

## How to test

- Removed the `custom_template_extra_vars` key.
- No changes to molecule output.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,70 @@ def unique(in_list):
...
+    template_data, unused = loader._get_file_contents(template_file_path)
+    template_data = to_text(template_data)
+
+    templar.available_variables = template_vars
+    result = templar.template(template_data, convert_data=False, escape_backslashes=False)
+    return result
+
+
+def compile_searchpath(searchpath: list):
+    newsearchpath = []
```
ClausHolbechArista, line 193: Add docstring

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,70 @@ def unique(in_list):
...
     -------
     str
         The rendered template
     """
 
-    result = template_lookup_module.run([template_file], template_vars)[0]
-    return str(result).strip()
+    loader = templar._loader
+    template_file_path = loader.path_dwim_relative_stack(searchpath, 'templates', template_file)
+    template_data, unused = loader._get_file_contents(template_file_path)
```
ClausHolbechArista: test sanity-test with _
rename template_data to jinja2_template or "template"

ansible_collections/arista/avd/plugins/module_utils/eos_designs_facts.py
```diff
@@ -1229,14 +1245,11 @@ def ipv6_router_id(self):
         '''
         if self.underlay_ipv6 is True:
             template_vars = self._template_vars
```
ClausHolbechArista: add a chainmap around this to avoid writing the vars below to the original

ansible_collections/arista/avd/docs/release-notes/3.x.x.md
```diff
@@ -8,6 +8,19 @@
...
+This section provides an overview of the data models that have ***changed*** from the previous release, but are backwards compatible.
+
+No changes are required for existing deployments.
+
+#### Changes to eos_designs role
+
+- __Node Types variables:__
+
+  - `node_type_keys.<key>.custom_templates_extra_vars` introduced in 3.7.0 has been removed. This is non-breaking since the templating environment
+  used for custom templates now supports all host- and groupvars by default.
```
gmuloc: This look like a very germanic way of writing things :) I would suggest rewriting it along these lines.

```suggestion
  used for custom templates now supports all host and group vars by default.
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/molecule.yml
```diff
@@ -33,6 +33,7 @@ provisioner:
     defaults:
       jinja2_extensions: 'jinja2.ext.loopcontrols,jinja2.ext.do,jinja2.ext.i18n'
       gathering: explicit
+      callbacks_enabled: profile_tasks
```
gmuloc: Should this be left in the molecule pipeline?

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -55,17 +56,25 @@ def run(self, tmp=None, task_vars=None):
...
-        output = {}
-
-        template_lookup_module = lookup_loader.get('ansible.builtin.template', loader=self._loader, templar=self._templar)
+        # Create a new Ansible "templar" instance to be passed along to our simplified "templater"
+        searchpath = compile_searchpath(task_vars.get('ansible_search_path'))
+        templar = self._templar.copy_with_new_env(searchpath=searchpath, available_variables={})
 
-        template_vars = task_vars
+        # If the argument 'root_key' is set, output will be assigned to this variable. If not set, the output will be set at as "root" variables.
+        # We use ChainMap to avoid large amounts of data around, mapping in avd_switch_facts and protecting task_vars from changes.
```
gmuloc:
```suggestion
        # We use ChainMap to avoid copying large amounts of data around, mapping in avd_switch_facts and protecting task_vars from changes.
```

ansible_collections/arista/avd/plugins/action/yaml_templates_to_facts.py
```diff
@@ -55,17 +56,25 @@ def run(self, tmp=None, task_vars=None):
...
+        hostname = task_vars['inventory_hostname']
+        switch_facts = get(task_vars, f"avd_switch_facts..{hostname}", default={}, separator="..")
+        if root_key:
+            template_vars = ChainMap({root_key: output}, switch_facts, task_vars)
+        else:
+            template_vars = ChainMap(output, switch_facts, task_vars)
 
         # If the argument 'debug' is set, a 'avd_yaml_templates_to_facts_debug' list will be added to the output.
         # This list contains timestamps from every step for every template. This is useful for identifying slow templates.
         # Here we pull in the list from any previous tasks, so we can just add the the list.
```
gmuloc:
```suggestion
        # Here we pull in the list from any previous tasks, so we can just add to the list.
```

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,92 @@ def unique(in_list):
...
     ----------
     template_file : str
-        Filename to pass to template_lookup_module
-    template_vars : any
-        Vars to pass to template_lookup_module
-    template_lookup_module : func
-        Instance of Ansible 'template' lookup module
+        Path containing template
+    extra_vars : any
+        Vars to pass to templar
```
gmuloc: no `extra_vars` + missing `searchpath`

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,92 @@ def unique(in_list):
...
+    - ...
+
+    Parameters
+    ----------
+    template_file : str
+        Path containing template
+    extra_vars : any
+        Vars to pass to templar
+    templar : func
+        Instance of Ansible Templar class
```
gmuloc: same comment as above

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,92 @@ def unique(in_list):
...
+        The rendered template
+    """
+    return str(template(template_file, template_vars, templar, searchpath)).strip()
+
+
+def template(template_file, template_vars, templar, searchpath):
+    """
+    Run Ansible Templar with template read from file
+
+    This function has does not support all Ansible features.
```
gmuloc:
```suggestion
    This function does not support all Ansible features.
```

gmuloc: not sure I understand all the implications here but it may be worth documenting this somewhere  in our doc in case users would rely on these features for their custom templates

gmuloc (commented):
Need to remove the three prof binary files

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: with the latest changes I think I will just remove this line. This docstring was written 3-4 months ago - before ChainMaps.

ClausHolbechArista: The list of unsupported features was below. I will reword things a bit to make it more clear.

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: I don't follow this comment?

ClausHolbechArista: nvm

gmuloc (approved):
LGTM - the failure in the pipeline is again the mkdoc fluke

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -1,6 +1,9 @@
 from __future__ import (absolute_import, division, print_function)
 __metaclass__ = type
 
+from ansible.module_utils._text import to_text
+import os
```
tgodaA, line 5:
```suggestion
from os.path import join as os_path_join
```

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,98 @@ def unique(in_list):
...
+    -------
+    list of str
+        List of both original and extra paths with "/templates" added.
+    """
 
-    result = template_lookup_module.run([template_file], template_vars)[0]
-    return str(result).strip()
+    newsearchpath = []
+    for p in searchpath:
+        newsearchpath.append(os.path.join(p, 'templates'))
```
tgodaA, line 195:
```suggestion
        newsearchpath.append(os_path_join(p, 'templates'))
```

ansible_collections/arista/avd/plugins/module_utils/utils.py
```diff
@@ -102,29 +105,98 @@ def unique(in_list):
...
     """
-    Run Ansible Template Lookup Plugin
+    Wrap "template" for single values like IP addresses
 
     The result is forced into a string and leading/trailing newlines and whitespaces are removed.
 
+    Parameters
+    ----------
     Parameters
     ----------
```
tgodaA:
```suggestion
    Parameters
    ----------
```

ansible_collections/arista/avd/plugins/module_utils/utils.py
ClausHolbechArista: From testing I don't see any difference in memory use and I don't think this makes it more readable. @gmuloc what is your take on this?

gmuloc: When I read `os_path_join` in the middle of the file (which I did first as I saw the comments in reverse order), I ended up looking for a custom function implemented by us. Then I ended up on this line. I feel like in this case having `os.path.join` is not a drawback and from what I read it does not have any impact. 

If we were using another function than `join` (clashing with the built-in join), I would probably be going with @tgodaA's suggestion but in our case it does not bring much so I would keep what is currently in there.

gmuloc (commented):
Tested the change in this commit with an inline Jinja template built for 3.6.0 that looks like:
```
   custom_structured_configuration_loopback_interfaces:
      loopback40:
        description: inband_access
        shutdown: false
        vrf: tranrz
        ip_address: "{{ loopback40_ip_pool | ansible.netcommon.ipmath(switch.id) }}/32"
```

Before this commit it was failing, after this commit it passes

tgodaA (approved):
LGTM

PR #2000, opened 2022-08-16
Feat(eos_cli_config_gen): add support for SSL profile cipher-list

### LINKED ISSUE #1973: Add support to specify cipher-list in management_security model
### Enhancement summary

Add support to specify the cipher-list in management_security data model in eos_cli_config_gen

```shell
management security
   ssl profile eAPI
      cipher-list ECDHE-RSA-AES256-GCM-SHA384
      certificate eAPI.crt key eAPI.key
```

https://avd.sh/en/stable/roles/eos_cli_config_gen/index.html#management-security

eos_cli_config_gen

### Use case example

This is currently required when using python 3.10 for enabling strong ciphers on EOS side to avoid SSLV3_ALERT_HANDSHAKE_FAILURE error when connecting via eAPI.

```yaml
management_security:
  entropy_source: < entropy_source >
  password:
    minimum_length: < 1-32 >
    encryption_key_common: < true | false >
    encryption_reversible: < aes-256-gcm >
  ssl_profiles:
    - name: <ssl_profile_1>
      tls_versions: < list of allowed tls versions as string >
      cipher-list: < valid-cipher-string >        <-------- New addition
      certificate:
        file: < certificate filename >
        key: < key filename >
    - name: <ssl_profile_2>
      tls_versions: < list of allowed tls versions as string >
```

since there is no `eos_cli` available under `management_security` data model, I have to `eos_cli` to render full management security config under root level.

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

Add support for cipher-list

## Related Issue(s)

Fixes #1973 

## Component(s) name

## Proposed changes

Added `cipher_list` parameter for `ssl_profiles`

```yaml
  management_security:
    entropy_source: < entropy_source >
    password:
      minimum_length: < 1-32 >
      encryption_key_common: < true | false >
      encryption_reversible: < aes-256-gcm >
    ssl_profiles:
      - name: <ssl_profile_1>
        tls_versions: < list of allowed tls versions as string >
        # cipher_list syntax follows the openssl cipher strings format
        cipher_list: < column separated list of allowed ciphers as a string >
        certificate:
          file: < certificate filename >
          key: < key filename >
      - name: <ssl_profile_2>
        tls_versions: < list of allowed tls versions as string >
```

## How to test

molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-security.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-security.yml
@@ -5,3 +5,12 @@ management_security:
     minimum_length: 17
     encryption_key_common: true
     encryption_reversible: aes-256-gcm
+  ssl_profiles:
+    - name: tls-versions-profile
+      tls_versions: "1.0 1.1"
+    - name: cipher-list-profile
+      cipher_list: ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384
+    - name: certificate-profile
+      certificate:
+        file: eAPI.crt
+        key: eAPI.key
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/mac-security-eth-po-entropy.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/mac-security-eth-po-entropy.md
@@ -66,9 +66,9 @@ interface Management1
 
 ## Management Security SSL Profiles
 
-| SSL Profile Name | TLS protocol accepted | Certificate filename | Key filename |
-| ---------------- | --------------------- | -------------------- | ------------ |
-| SSL_PROFILE | 1.1 1.2 | SSL_CERT | SSL_KEY |
+| SSL Profile Name | TLS protocol accepted | Certificate filename | Key filename | Cipher List |
+| ---------------- | --------------------- | -------------------- | ------------ | ----------- |
+| SSL_PROFILE | 1.1 1.2 | SSL_CERT | SSL_KEY | - |
 
 ## Management Security Configuration
 
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
@@ -6,6 +6,7 @@
 - [Authentication](#authentication)
 - [Management Security](#management-security)
   - [Management Security Summary](#management-security-summary)
+  - [Management Security SSL Profiles](#management-security-ssl-profiles)
   - [Management Security Configuration](#management-security-configuration)
 - [Monitoring](#monitoring)
 - [Internal VLAN Allocation Policy](#internal-vlan-allocation-policy)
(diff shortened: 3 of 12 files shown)


tgodaA:
In the proposed changes of the template. Please mention the old and new or add only the new knobs that will be added. Easy for us to review.

gmuloc:
it is clearly stated just above the snippet:
> Added `cipher_list` parameter for `ssl_profiles`

tgodaA:
Yeah, I saw that. I am just saying that it's not required (I agree that it's completely up to you) to add all knobs.

tgodaA (approved):
LGTM

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-security.yml
```diff
@@ -5,3 +5,12 @@ management_security:
     minimum_length: 17
     encryption_key_common: true
     encryption_reversible: aes-256-gcm
+  ssl_profiles:
+    - name: tls-versions-profile
+      tls_versions: "1.0 1.1"
+    - name: cipher-list-profile
+      cipher_list: ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384
```
ClausHolbechArista, line 12: Wouldn't it make more sense to use a list in the data model and then create the colon separated list in the template?
so
```yaml
ssl_profiles:
  - name: xxx
    ciphers:
      - ECDHE-RSA-AES256-GCM-SHA384
      - ECDHE-RSA-AES256-SHA384
```
In the docs, we should then print as comma separated to make it more readable.
Or is following OpenSSL's format better?

gmuloc: thought about this as well - basically the only thing that is important is the order - cf: https://www.mkssoftware.com/docs/man1/openssl_ciphers.1.asp so we could have a list instead.

List may be easier to read though in some occasions it could be confusing:

e.g. for `ALL:!COMPLEMENTOFDEFAULT:!eNULL.`

we would end up with
```
ciphers:
  - ALL
  - !COMPLEMENTOFDEFAULT
  - !eNULL
```

Which is kind of weird when reading as it looks like we include all ciphers.

ClausHolbechArista: IMO those two are equally confusing :)

gmuloc: haha well I don't have a string feeling for any - might be simpler for people to use the string if they already have it but happy to go with the list version as well. There may be things I don't foresee doing this as I have not played that much with the syntax but we can always change at a later stage if need be. 

The only thing is that each new element of the list impact the "selected" list of ciphers so the order is paramount.
As long as we keep the order of the list in the template it stays usable.

ClausHolbechArista: Just skimmed the examples from openssl, and since it can have characters like `= + -` etc I think we should just stick with a single string like you have. Sorry for the noise :)

ClausHolbechArista (approved):
LGTM

PR #2007, opened 2022-08-19
Optimize(eos_cli_config_gen): Remove lower method on boolean variables

tgodaA:
## Change Summary

The templates used for documentation are using lower method on boolean variables, which is not required. So optimizing the code. This is not a bug fix, just removing that extra layer of functionality, which is not required.

## Component(s) name

## Proposed changes

Removed the `| lower` on all boolean variables that are used in documentation templates only.

## How to test

`cd molecule`
`make refresh-facts`

## Checklist

- [x] Verify that artifacts are updated only on md files and no changes in cfg files.
- [x] Fixing `false` to `False` and 'true` to `True`, which is regular boolean, when printed in the md files. No other changes should be observed.

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista (approved):
LGTM

gmuloc (approved):
LGTM

PR #2010, opened 2022-08-23
Fix: Remove ansible-test enforced GPL3 header from modules

ClausHolbechArista:
## Change Summary

Remove ansible-test enforced GPL3 header from modules

Ansible-test sanity enforces a GPL3 license in all modules. This is not intended for collections, but only for Ansible Core, so we will add ignores for these rules and remove the GPL3 line.
The collection is licensed under Apache 2.0 as it is mentioned everywhere else.

### DIFF
--- a/ansible_collections/arista/avd/plugins/module_utils/strip_empties.py
+++ b/ansible_collections/arista/avd/plugins/module_utils/strip_empties.py
@@ -1,9 +1,7 @@
 #!/usr/bin/env python
 # coding: utf-8 -*-
 #
-# GNU General Public License v3.0+
-#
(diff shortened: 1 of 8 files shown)


gmuloc (approved):
LGTM

only wondering if it should be `Copyright 2019-2022`

ClausHolbechArista:
> LGTM
> 
> only wondering if it should be `Copyright 2019-2022`

To avoid that discussion, I will remove my change to the dates, and just remove the team, since that has no separate entity.

tgodaA (approved):
LGTM

PR #2032, opened 2022-09-07
Refactor: Move python libraries to plugin_utils

ClausHolbechArista:
Moving internal libraries from `module_utils` to `plugin_utils` to avoid some Ansible sanity-test restrictions.
Since we only consume the libraries from Action Plugins, we don't need them to be in the `module_utils` folder.


gmuloc (approved):
LGTM

ghost (approved):
LGTM!

PR #2040, opened 2022-09-09
Feat(plugins): New arista.avd.include_vars plugin

ClausHolbechArista:
## Change Summary

Feat(plugin): New arista.avd.include_vars plugin

## Related Issue(s)

When using `ansible.builtin.include_vars` the vars are loaded with a higher var precedence than anything we can set during the play. This blocks us from doing any variable type conversion during execution, since the original imported vars always take precedence.

## Component(s) name

`arista.avd.include_vars`

## Proposed changes

Since Ansible detects the module FQCN `ansible.builtin.include_vars` and handles the returned `ansible_facts` differently than any other module, we just need another name, to avoid this behavior.
The plugin `arista.avd.include_vars` simply wraps the builtin class to fool Ansible.

## How to test

When importing vars from file, debug var before and after validation to see if validation can overwrite the vars or not.

Also tested with PR #2024 where the issue was seen first.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/docs/release-notes/3.x.x.md
+++ b/ansible_collections/arista/avd/docs/release-notes/3.x.x.md
@@ -8,6 +8,19 @@
 
 ## Release 3.8.0
 
+!!! info "Variable precedence changed in eos_cli_config_gen"
+    In cases where `eos_cli_config_gen` is invoked in a separate playbook, variables are imported from "structured_config" files. These variables are now imported as regular facts with `host_facts` precedence instead of the special `include_vars` precedence.
+
+    This change means that any other vars included with the `ansible.builtin.include_vars` module may take precedence over the imported "structured_config".
+
+    It is recommended to replace `include_vars` or `ansible.builtin.include_vars` with `arista.avd.include_vars` in any tasks running
+    before AVD.
+
+    The purpose of this change is to allow AVD to override and perform automatic type conversion of input variables.
+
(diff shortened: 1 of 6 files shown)


gmuloc (approved):
LGTM - this is a good evolution especially for those using include_vars today who need to come up with custom_structured_configuration_prefix ordering trick to get some group_vars/host_vars to take precedence

My only comment would be that we should document this change as "potentially breaking" for the rare people out there who would rely on running in different playbook `eos_designs` and `eos_cli_config_gen` as this _could_ have an impact on their variables precedence.

tgodaA (approved):
lgtm

ansible_collections/arista/avd/docs/release-notes/3.x.x.md
```diff
@@ -8,6 +8,20 @@
 
 ## Release 3.8.0
 
+!!! info "Variable precedence changed in eos_cli_config_gen"
+
+    In cases where `eos_cli_config_gen` is invoked in a separate playbook, variables are imported from "structured_config" files. These variables are now imported as regular facts with `host_facts` precedence instead of the special `include_vars` precedence.
```
carlbuchmann:
```suggestion
    In cases where `eos_cli_config_gen` is invoked in a separate playbook, variables are imported from "structured_config" files. These variables are now imported as regular facts with `host_facts` precedence instead of the higher `include_vars` precedence.
```

carlbuchmann (approved):
LGTM, just a small suggestion on the wording in release notes!

PR #1984, opened 2022-08-08
CI: Markdownlint Addition

JulioPDX:
## Change Summary

This PR adds markdownlint to our pre-commit and CI workflow process. This will assist contributors and maintainers when working on our documentation and prevent minor syntax issues. Current workflow will try and fix all basic errors present, if this is not possible users will get an error and rule that is broken . This PR also includes any fixes triggered by this linter.  Also adds Flake8 summary on pre-commit workflow.

## Proposed changes

Add markdownlint to pre-commit workflow. Any markdown rule overrides have been set to adhere to current documentation practices. For example, enabling inline HTML in documents. A full list of rule edits can be viewed at `.github/.markdownlint.yaml`. 

This PR also adds `.github/.markdownlintignore`. This will disable the linter from running on the directories listed below. If any others should be included, please notify contributor. 

```shell
ansible_collections/arista/avd/molecule/
ansible_collections/arista/avd/examples/single-dc-l3ls/documentation/
ansible_collections/arista/avd/tests/unit/filters/toc_files/
```

## How to test

Checkout this PR and run pre-commit to view example output with no issues. 

```shell
> pre-commit run --all-files markdownlint
markdownlint.............................................................Passed

ansible-avd on markdown-lint                                                           py mk-venv
> 
```

Example when issue is present.

```shell
> pre-commit run --all-files markdownlint
markdownlint.............................................................Failed
- hook id: markdownlint
- exit code: 1

ansible_collections/arista/avd/README.md:5 MD025/single-title/single-h1 Multiple top-level headings in the same document [Context: "# About"]


ansible-avd on markdown-lint !1                                                        py mk-venv
> 
```

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #2051, opened 2022-09-13
Fix: pr labeler action

gmuloc:
* AVD list of labels is too long and only the first 30 in alphabetical orders are returned. This led to the action failing for PR #1984

## Change Summary

The github action is supposed to label the PRs when merged with a release note labels.
It turns out that the command line that checks before the label already exists before creating it uses some github cli command that will return ONLY the first 30 labels. This explains the failure for labeling PR #1984 when it merged.

This PR intends to fix this by adding a search string for only the "rn:" labels and extend the number of retrieved labels to 100 which should cover the use case.

## Component(s) name

`ci`

## Proposed changes

Add extra parameters to verify if a label already exists.

### Repository Checklist

### DIFF
--- a/.github/actions/rn-pr-labeler-action/action.yml
+++ b/.github/actions/rn-pr-labeler-action/action.yml
@@ -25,7 +25,7 @@ runs:
       # the color of the label though so it may not be desirable.
     - name: Check if label exist
       run: |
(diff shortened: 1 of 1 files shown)


ClausHolbechArista (approved):
LGTM

PR #2091, opened 2022-09-22
Refactor(eos_designs): Move IP and description logic to Python (step1)

ClausHolbechArista:
## Change Summary

Refactor(eos_designs): Move IP and description logic to Python (step1)

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Add support for loading configurable python modules with IP addressing and Interface Description logic:
```
templates:
  ip_addressing:
    python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing'
    python_class_name: 'AvdIpAddressing'
  interface_descriptions:
    python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.interface_descriptions'
    python_class_name: 'AvdInterfaceDescriptions'
```

- Refactor `eos_designs_facts` to leverage these new python modules.
  - The new logic uses python builtin `ipaddress` and enforces correct sizing of pools.
  - Errors will be raised if we try to fetch IPs outside the pool. Historically we just ignored the pool mask and assigned outside it.
- The default python modules still respect any custom templates, and since these take precedence, the python modules will not have any effect, until we remove the default templates.
- In this PR we remove the default IP templates that are only leveraged inside `eos_designs_facts`

## How to test

No changes to molecule. Tested locally with some refactored code.

- had to change a bad mask on an IP pool, since we now enforce taking from the pool instead of overflowing.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
+++ b/ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
@@ -30,12 +30,8 @@ default_templates:
           strip_empty_keys: false
 
     ip_addressing:
-      router_id: 'ip_addressing/router-id.j2'
-      ipv6_router_id: 'ip_addressing/router-id-ipv6.j2'
-      mlag_ip_primary: 'ip_addressing/mlag-ip-primary.j2'
-      mlag_ip_secondary: 'ip_addressing/mlag-ip-secondary.j2'
-      mlag_l3_ip_primary: 'ip_addressing/mlag-l3-ip-primary.j2'
-      mlag_l3_ip_secondary: 'ip_addressing/mlag-l3-ip-secondary.j2'
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing'
+      python_class_name: 'AvdIpAddressing'
       mlag_ibgp_peering_ip_primary: 'ip_addressing/mlag-ibgp-peering-ip-primary.j2'
       mlag_ibgp_peering_ip_secondary: 'ip_addressing/mlag-ibgp-peering-ip-secondary.j2'
       p2p_uplinks_ip: 'ip_addressing/p2p-uplinks-ip.j2'
@@ -44,6 +40,8 @@ default_templates:
       vtep_ip: 'ip_addressing/vtep-ip.j2'
 
     interface_descriptions:
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.interface_descriptions'
+      python_class_name: 'AvdInterfaceDescriptions'
       underlay_ethernet_interfaces: 'interface_descriptions/underlay/ethernet-interfaces.j2'
       underlay_port_channel_interfaces: 'interface_descriptions/underlay/port-channel-interfaces.j2'
       mlag_ethernet_interfaces: 'interface_descriptions/mlag/ethernet-interfaces.j2'
@@ -77,11 +75,11 @@ default_templates:
           list_merge: "{{ custom_structured_configuration_list_merge }}"
           strip_empty_keys: false
     ip_addressing:
-      router_id: 'ip_addressing/router-id.j2'
-      p2p_uplinks_ip: 'ip_addressing/p2p-uplinks-ip.j2'
-      p2p_uplinks_peer_ip: 'ip_addressing/p2p-uplinks-peer-ip.j2'
-      ipv6_router_id: 'ip_addressing/router-id-ipv6.j2'
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing'
+      python_class_name: 'AvdIpAddressing'
     interface_descriptions:
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.interface_descriptions'
+      python_class_name: 'AvdInterfaceDescriptions'
       underlay_ethernet_interfaces: 'interface_descriptions/underlay/ethernet-interfaces.j2'
       underlay_port_channel_interfaces: 'interface_descriptions/underlay/port-channel-interfaces.j2'
       connected_endpoints_ethernet_interfaces: 'interface_descriptions/connected_endpoints/ethernet-interfaces.j2'
@@ -116,16 +114,12 @@ default_templates:
           list_merge: "{{ custom_structured_configuration_list_merge }}"
           strip_empty_keys: false
     ip_addressing:
-      router_id: 'ip_addressing/router-id.j2'
-      ipv6_router_id: 'ip_addressing/router-id-ipv6.j2'
-      mlag_ip_primary: 'ip_addressing/mlag-ip-primary.j2'
-      mlag_ip_secondary: 'ip_addressing/mlag-ip-secondary.j2'
-      mlag_l3_ip_primary: 'ip_addressing/mlag-l3-ip-primary.j2'
-      mlag_l3_ip_secondary: 'ip_addressing/mlag-l3-ip-secondary.j2'
-      p2p_uplinks_ip: 'ip_addressing/p2p-uplinks-ip.j2'
-      p2p_uplinks_peer_ip: 'ip_addressing/p2p-uplinks-peer-ip.j2'
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing'
+      python_class_name: 'AvdIpAddressing'
 
     interface_descriptions:
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.interface_descriptions'
+      python_class_name: 'AvdInterfaceDescriptions'
       underlay_ethernet_interfaces: 'interface_descriptions/underlay/ethernet-interfaces.j2'
       underlay_port_channel_interfaces: 'interface_descriptions/underlay/port-channel-interfaces.j2'
       mlag_ethernet_interfaces: 'interface_descriptions/mlag/ethernet-interfaces.j2'
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
@@ -104,9 +104,7 @@ class AvdInterfaceDescriptions(AvdFacts):
             self._hostvars,
             "switch.interface_descriptions.connected_endpoints_ethernet_interfaces",
         ):
-            return self._template(
-                template_path, peer=peer, peer_interface=peer_interface
-            )
+            return self._template(template_path, peer=peer, peer_interface=peer_interface)
 
         elements = [peer, peer_interface]
         return "_".join([str(element) for element in elements if element is not None])
@@ -169,8 +167,6 @@ def load_interfacedescriptions(hostvars, templar) -> AvdInterfaceDescriptions:
 
     # if not isinstance(cls, AvdInterfaceDescriptions):
     if not issubclass(cls, AvdInterfaceDescriptions):
-        raise AristaAvdError(
-            f"{cls} is not a subclass of AvdInterfaceDescriptions class"
-        )
+        raise AristaAvdError(f"{cls} is not a subclass of AvdInterfaceDescriptions class")
 
     return cls(hostvars=hostvars, templar=templar)
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
@@ -104,14 +104,12 @@ class AvdInterfaceDescriptions(AvdFacts):
             self._hostvars,
             "switch.interface_descriptions.connected_endpoints_ethernet_interfaces",
         ):
-            template_vars = ChainMap({}, self._hostvars)
-            template_vars["peer"] = peer
-            template_vars["peer_interface"] = peer_interface
-            return self.template_var(template_path, template_vars)
+            return self._template(
+                template_path, peer=peer, peer_interface=peer_interface
+            )
 
         elements = [peer, peer_interface]
-        elements = [str(element) for element in elements if element is not None]
-        return "_".join(elements)
+        return "_".join([str(element) for element in elements if element is not None])
 
     def connected_endpoints_port_channel_interfaces(self, peer: str = None, adapter_port_channel_description: str = None) -> str:
         if template_path := get(
@@ -125,8 +123,7 @@ class AvdInterfaceDescriptions(AvdFacts):
             )
 
         elements = [peer, adapter_port_channel_description]
-        elements = [str(element) for element in elements if element is not None]
-        return "_".join(elements)
+        return "_".join([str(element) for element in elements if element is not None])
 
     def overlay_loopback_interface(self, overlay_loopback_description) -> str:
         if template_path := get(self._hostvars, "switch.interface_descriptions.overlay_loopback_interface"):
@@ -172,6 +169,8 @@ def load_interfacedescriptions(hostvars, templar) -> AvdInterfaceDescriptions:
 
     # if not isinstance(cls, AvdInterfaceDescriptions):
     if not issubclass(cls, AvdInterfaceDescriptions):
-        raise AristaAvdError(f"{cls} is not an instance of AvdFacts class")
+        raise AristaAvdError(
+            f"{cls} is not a subclass of AvdInterfaceDescriptions class"
+        )
 
     return cls(hostvars=hostvars, templar=templar)
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
@@ -60,7 +60,7 @@ class AvdInterfaceDescriptions(AvdFacts):
         self,
         link_peer: str,
         link_peer_channel_group_id: int,
-        link_channel_description: str | None,
+        link_channel_description: str,
     ) -> str:
         if template_path := get(self._hostvars, "switch.interface_descriptions.underlay_port_channel_interfaces"):
             template_vars = ChainMap({}, self._hostvars)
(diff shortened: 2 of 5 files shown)


ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
```diff
@@ -0,0 +1,165 @@
+
+import importlib
+from collections import ChainMap
+from functools import cached_property
```
gmuloc, line 3: mayb we can start enforcing some ordering on the imports between similar classes to enforce consistency? (that's cosmetic)

ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
```diff
@@ -30,12 +30,8 @@ default_templates:
           strip_empty_keys: false
 
     ip_addressing:
-      router_id: 'ip_addressing/router-id.j2'
-      ipv6_router_id: 'ip_addressing/router-id-ipv6.j2'
-      mlag_ip_primary: 'ip_addressing/mlag-ip-primary.j2'
-      mlag_ip_secondary: 'ip_addressing/mlag-ip-secondary.j2'
-      mlag_l3_ip_primary: 'ip_addressing/mlag-l3-ip-primary.j2'
-      mlag_l3_ip_secondary: 'ip_addressing/mlag-l3-ip-secondary.j2'
+      python_module: 'ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing'
```
gmuloc, line 33: Given that we have default variables in the classes for python modules and classes - do we actually need this?

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
```diff
@@ -0,0 +1,346 @@
...
+        '''
+        if template_path := get(self._hostvars, "switch.ip_addressing.mlag_l3_ip_primary"):
+            template_vars = ChainMap({}, self._hostvars)
+            template_vars['mlag_primary_id'] = self._mlag_primary_id
+            template_vars['mlag_secondary_id'] = self._mlag_secondary_id
+            template_vars['switch_data'] = {
+                'combined': {
+                    'mlag_peer_l3_ipv4_pool': self._mlag_peer_l3_ipv4_pool
+                }
+            }
```
gmuloc: Wondering if somehow we could manage to abstract this "repetitive" block in some function ?

looks like in most cases we need to
1. check what is the template
2. add some template_vars which we could list
3. the combined/ vrf part is a bit mort tricky as I guess this is to not deprecate the existing templates?

ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
```diff
@@ -44,6 +40,8 @@ default_templates:
       vtep_ip: 'ip_addressing/vtep-ip.j2'
```
gmuloc, line 40: so for now in this PR only some templates are passed over but it looks like all have been implemented. In the proposed changes section it states that the PR only removes the default templates only in use in eos_designs_facts. Does it mean there is an issue when the template are also used elsewehere?

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    @cached_property
+    def _mpls_lsr(self) -> str:
+        return get(self._hostvars, "switch.mpls_lsr")
+
+    @cached_property
+    def _mlag_peer(self) -> str:
+        return get(self._hostvars, "switch.mlag_peer", required=True)
+
+    @cached_property
+    def _mlag_port_channel_id(self) -> str:
```
gmuloc, line 41: Given that we compute these cached_properties in these classes - would this mean that if the same was needed in another classes we would need to add the cached_property definition there as well?

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+        default=DEFAULT_AVD_INTERFACE_DESCRIPTIONS_PYTHON_CLASS_NAME
+    )
+    try:
+        cls = getattr(importlib.import_module(module_path), class_name)
+    except ImportError as imp_exc:
+        raise AristaAvdError(imp_exc) from imp_exc
+
+    # if not isinstance(cls, AvdInterfaceDescriptions):
+    if not issubclass(cls, AvdInterfaceDescriptions):
+        raise AristaAvdError(f"{cls} is not an instance of AvdFacts class")
```
gmuloc: This error message should probably be more specific - stating clearly that we expect the class to inherit from AvdInterfaceDescriptions not just AvdFacts (otherwise I would be confused if I tried to import a class inheriting from AvdFacts implementing all the different templates).


Also out of curiosity, what happens if a class does not implement properly one of the necessary methods for description that would be needed for a specific AVD repo? Should we add checks to verify some method are present at the minimum?

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+    def connected_endpoints_port_channel_interfaces(self, peer: str = None, adapter_port_channel_description: str = None) -> str:
+        if template_path := get(self._hostvars, "switch.interface_descriptions.connected_endpoints_port_channel_interfaces"):
+            template_vars = ChainMap({}, self._hostvars)
+            template_vars['peer'] = peer
+            template_vars['adapter_port_channel_description'] = adapter_port_channel_description
+            return self.template_var(template_path, template_vars)
+
+        elements = [peer, adapter_port_channel_description]
+        elements = [str(element) for element in elements if element is not None]
+        return '_'.join(elements)
```
gmuloc:
```suggestion
        elements = [peer, adapter_port_channel_description]
        return '_'.join(str(element) for element in elements if element is not None)
```

(almost made it a oneliner but that may be less readable)

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
```diff
@@ -0,0 +1,165 @@
...
+            template_vars['link'] = {
+                "type": link_type,
+                "peer": link_peer,
+                "peer_interface": link_peer_interface,
+            }
+            return self.template_var(template_path, template_vars)
+
+        link_peer = str(link_peer).upper()
+        if link_type == "underlay_p2p":
+            return f"P2P_LINK_TO_{link_peer}_{link_peer_interface}"
```
gmuloc, line 57: wondering if we would not be better using "_".join by default in our scenarios to make sure that even if a variable is None we don't end up with a weird `DESCRIPTION__PO_`. Though in theory I agree we "know" some of these variables can never be empty

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
```diff
@@ -0,0 +1,346 @@
...
+    those are mapped from the switch.* model
+
+    This class is imported adhoc based on the variable `templates.ip_addressing.python_module` so it can
+    be overridden by a custom python class.
+    '''
+
+    def _ip(self, pool: str, prefixlen: int, subnet_offset: int, ip_offset: int) -> str:
+        network = ipaddress.ip_network(pool, strict=False)
+        subnets = network.subnets(new_prefix=prefixlen)
+        if isinstance(subnets, list):
```
gmuloc: This sounds weird! :) the doc from 3.6 document iterators - do you have some traceback on this?

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
```diff
@@ -0,0 +1,346 @@
...
+        if isinstance(ips, list):
+            # It seems that some versions of python return a list instead of the documented iterator
+            ips = iter(ips)
+        ip_index = 0
+        try:
+            while ip_index <= ip_offset:
+                ip = next(ips)
+                ip_index += 1
+        except StopIteration as e:
+            raise AristaAvdError(f"Unable to get {ip_offset+1} hosts in subnet {subnet} taken from pool {pool}") from e
```
gmuloc: it looks like math on IP works (python 3.10) so could this maybe be made simple?
```
>>> ipaddress.IPv4Address("42.42.42.42")
IPv4Address('42.42.42.42')
>>> ii = _
>>> ii+1
IPv4Address('42.42.42.43')
>>> ii+2
IPv4Address('42.42.42.44')
>>> ii+3
IPv4Address('42.42.42.45')
>>> ii+4
IPv4Address('42.42.42.46')
```

Like checking if ip_offset is longer than subnet length maybe? - this while loop seem unecessary except if I am missing something?

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
```diff
@@ -0,0 +1,346 @@
...
+        if isinstance(subnets, list):
+            # It seems that some versions of python return a list instead of the documented iterator
+            subnets = iter(subnets)
+        subnet_index = 0
+        try:
+            while subnet_index <= subnet_offset:
+                subnet = next(subnets)
+                subnet_index += 1
+        except StopIteration as e:
+            raise AristaAvdError(f"Unable to get {subnet_offset+1} /{prefixlen} subnets from pool {pool}") from e
```
gmuloc: This could probably be replaced with something like (maybe need some verificatin at the boundaries):

```
subnets = list(network.subnets(new_prefix=prefixlen)
try:
  subnet = subnets[subnet_offset]
except OutOfRange:
   AristaAvdError ..
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
ClausHolbechArista: I don't. But I *think* it was because of this: https://github.com/python/cpython/issues/71870
I was running some 3.8 version at the time.

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
ClausHolbechArista: I did as you suggest at first, but with a /64 in ipv6 it will try to produce a list with 64-bit entries, which takes quite a while :)
So instead I decided to use this iterator since we often need one of the first few hundred addresses. list() would be doing the same thing anyway, so there should be no cost to this. Only benefits when we pull the lower entries of the scope.

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
ClausHolbechArista: Let me experiment a bit. But this iterator seems quite effective.

ClausHolbechArista: I tried refactoring the code to this 
```python 
    def _ip(self, pool: str, prefixlen: int, subnet_offset: int, ip_offset: int) -> str:
        network = ipaddress.ip_network(pool, strict=False)
        prefixlen_diff = prefixlen - network.prefixlen
        subnets_in_pool = 2**prefixlen_diff
        if subnet_offset >= subnets_in_pool:
            raise AristaAvdError(
                f"Unable to get {subnet_offset+1} /{prefixlen} subnets from pool {pool}"
            )

        if network.version == 4:
            subnetsize = 2 ** (32 - prefixlen)
        else:
            subnetsize = 2 ** (128 - prefixlen)

        subnet = ipaddress.ip_network(
            (int(network.network_address) + subnet_offset * subnetsize, prefixlen)
        )
        if ip_offset >= subnet.num_addresses:
            raise AristaAvdError(
                f"Unable to get {ip_offset+1} hosts in subnet {subnet} taken from pool {pool}"
            )
        return subnet[ip_offset]
```
While it makes it simpler to read, it is also slightly slower for the very large inventory I am testing with. I would also expect it to be much slower for ipv6, since it calculates the large mask sizes. We could add a prefixlen indexed list with precomputed values. It would only be 128 items and a single lookup instead of calculating 2**128 etc.

ClausHolbechArista: Another attempt which seems to have ~ the same performance as the iterator:
```python
    def _ip(self, pool: str, prefixlen: int, subnet_offset: int, ip_offset: int) -> str:
            raise AristaAvdError(
                f"Unable to get {subnet_offset + 1} /{prefixlen} subnets from pool {pool}"
            )
        subnet = ipaddress.ip_network(
            (int(pool_network.network_address) + subnet_offset * subnet_size, prefixlen)
        )
        try:
            ip = subnet[ip_offset]
        except IndexError as e:
            raise AristaAvdError(
                f"Unable to get {ip_offset+1} hosts in subnet {subnet} taken from pool {pool}"
            ) from e
        return str(ip)
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
ClausHolbechArista: I have implemented a small helper function, but it is very limited how much we can save, since the template vars are slightly different for most of the functions. Let me know what you think

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
ClausHolbechArista: I am actually trying to do that by setting full imports first, standard libraries next and then the internal stuff

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
ClausHolbechArista: It is only locally significant for the instance of the class. Meaning the cache will be on instance level. Two instances of the class will _not_ share cache, but multiple requests for the same property on the same instance will be returned from cache instead of running the method.

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
ClausHolbechArista: In this function these two vars are required inputs, so we know they are set. The purpose here is 1:1 refactor.

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
ClausHolbechArista: I decided to enforce it by checking if they are a subclass of our default class, so if they only overload one or none of our methods, everything will still work. If they overload our methods with something bad, it is their own fault :)

ClausHolbechArista: I forgot to update the error message when I moved to this stricter enforcement. Have also updated on ip adddressing

ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
ClausHolbechArista: No we don't. Just trying to help someone who wants to change it.

ansible_collections/arista/avd/roles/eos_designs/defaults/main/default-templates.yml
ClausHolbechArista: Since the python lib can only be used from python (not jinja), we have to keep the templates that are used in jinja templates.
As we refactor, we can remove more of them.

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/__init__.py
ClausHolbechArista: submitted another PR with isort

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
```diff
@@ -0,0 +1,328 @@
...
+
+    @cached_property
+    def _loopback_ipv6_offset(self) -> int:
+        return get(self._hostvars, "switch.loopback_ipv6_offset", required=True)
+
+    @cached_property
+    def _vtep_loopback_ipv4_pool(self) -> str:
+        return get(self._hostvars, "switch.vtep_loopback_ipv4_pool", required=True)
+
+    def mlag_ibgp_peering_ip_primary(self, mlag_ibgp_peering_ipv4_pool: str) -> str:
```
gmuloc, line 94: So given that these are almost the same function rather than duplicating code I think it could be better to just add to these pairs of functions an input argutment being "role" (primary or secondary) and to return accordingly one or the other

This would reduce the code here as well as reducing the code in eos_designs_facts.py

```
if self.mlag_role == "primary":
  return self.avd_ip_address.mlag_ibgp_peering_ip_primary()
elif self.mlag_role == "secondary":
  return self.avd_ip_address.mlag_ibgp_peering_ip_secondary()
```

could just become
```
  return self.avd_ip_address.mlag_ibgp_peering_ip_primary(self.mlag_role)
```


The only thing it would really change is that now people would need to implement method with similar signatures when injecting their own.

ClausHolbechArista: We need to respect the existing split of templates. Otherwise it may not work with existing deployments using custom templates.

gmuloc: Are you talking about 

```
switch.ip_addressing.mlag_ibgp_peering_ip_primary
```

?

It can always be written
```
f"switch.ip_addressing.mlag_ibgp_peering_ip_{role}"
```

right?

ClausHolbechArista: if it was a string then yes. But it is a function call.

carlbuchmann (approved):
LGTM - Tested with customers repos

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/__init__.py
ClausHolbechArista: We agreed to keep it as is for now and we will revisit later.

gmuloc (approved):
LGTM

PR #2093, opened 2022-09-23
Doc: Document the path of the templates

pvinci-arista:
## Change Summary

Document that the templates are relative to the playbook. 

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


tgodaA (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/doc/node-types-v4.0.md
```diff
@@ -47,6 +47,7 @@ node_type_keys:
     mpls_lsr: < true | false | default -> false >
 
     # Optional | Override ip_addressing templates
+    # Path is relative to playbook
```
ClausHolbechArista: This is not completely correct. They can also be pointing to the builtin templates inside the collection. But since we will soon remove those in favor of python, we can leave this.

I suggest to add a section below this code block to describe the full background similar to https://avd.sh/en/stable/roles/eos_cli_config_gen/index.html#extensibility-with-custom-templates
Then refer to that section in this comment. Because it is not only relative to playbook, it must also be _below_ the playbook dir.

gmuloc: Maybe let's reference the ansible documentation as well: https://docs.ansible.com/ansible/latest/user_guide/playbook_pathing.html#resolving-local-relative-paths ?

ansible_collections/arista/avd/roles/eos_designs/doc/node-types-v4.0.md
```diff
@@ -59,6 +62,9 @@ node_type_keys:
       vtep_ip: <path to J2 template - default inherited from templates.ip_addressing.vtep_ip >
 
     # Optional | Override interface_descriptions templates
+    # Path is relative to playbook
+    # See: https://avd.sh/en/stable/roles/eos_cli_config_gen/index.html#extensibility-with-custom-templates
```
carlbuchmann: @pvinci-arista, thanks for taking this on and helping improve the documentation.
IMO we should explain here, and not just have a link to eos_cli_config_gen documentation!
It is ok in the documentation if we repeat ourselves a little!

carlbuchmann (changes requested):

PR #2131, opened 2022-10-01
Refactor(eos_cli_config_gen): Merge input validation with templating

ClausHolbechArista:
## Change Summary

Refactor(eos_cli_config_gen): Merge input validation with templating

## Component(s) name

- `arista.avd.eos_cli_config_gen`
- `arista.avd.validate_and_template` plugin

## Proposed changes

- Reverting recent changes to variable precedence and removing `arista.avd.include_vars`
- Removing `arista.avd.validate` from `eos_cli_config_gen` tasks
- Build new `arista.avd.validate_and_template` plugin and replace the builtin `template`
  - Purpose is to run variable conversion and validation as part of the same task, so we are 
    not subject to variable precedence of Ansible.

## How to test

No changes to molecule
Need retesting of customer projects who reported issues with 3.8.0-dev0.

### User Checklist

- N/A

### Repository Checklist


carlbuchmann (approved):
LGTM,
- Tested with two complex customer scenarios, which exhibited issues with current devel.

To do, add plugin documentation: https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/plugins/README.md

ClausHolbechArista:
@carlbuchmann I have updated plugin docs. Please review.

ansible_collections/arista/avd/plugins/README.md
```diff
@@ -590,3 +590,76 @@ The facts can be inspected in a file per device by running the `arista.avd.eos_d
...
+
+  # Run data conversion in either "warning", "info", "debug" or "disabled" mode | Optional, default: "debug"
+  # Conversion will perform type conversion of input variables as defined in the schema.
+  # Conversion is intended to help the user to identify minor issues with the input data, while still allowing the data to be validated.
+  # During conversion, messages will generated with information about the host(s) and key(s) which required conversion.
+  # conversion_mode:disabled means that conversion will not run.
+  # conversion_mode:warning will produce warning messages.
+  # conversion_mode:info will produce regular log messages.
+  # conversion_mode:debug will produce hidden messages viewable with -v.
+  conversion_mode: <str>
```
gmuloc, line 636: So I guess it is not going in the direction we want schemawise but I would find something like

```
converstion_mode: [disabled | warning | info | debug]
```

more readable than

```
converstion_mode: <str>
```

Not sure if this would make sense here..

ansible_collections/arista/avd/plugins/action/validate_and_template.py
```diff
@@ -0,0 +1,168 @@
...
+
+class ActionModule(ActionBase):
+    def run(self, tmp=None, task_vars=None):
+        if task_vars is None:
+            task_vars = {}
+
+        result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        # Validate Arguments
```
gmuloc, line 25: Maybe lets move the argument validation to a separate function as it is a lot of code for not much in the main run.

Any reason not to use the built-in argument spec? https://docs.ansible.com/ansible/latest/dev_guide/developing_program_flow_modules.html#argument-spec

ansible_collections/arista/avd/plugins/action/validate_and_template.py
```diff
@@ -0,0 +1,168 @@
...
+        if result_messages:
+            result["msg"] = " ".join(result_messages)
+
+        # Template to file
+        # Update result from Ansible "copy" operation (setting 'changed' flag accordingly)
+        result.update(self.template(task_vars))
+
+        return result
+
+    def convert_data(self, result_messages):
```
gmuloc, line 90: Any reason to pass the input `result_messages` rather than returning a list in this method?

ansible_collections/arista/avd/plugins/action/validate_and_template.py
```diff
@@ -0,0 +1,168 @@
...
+        # The returned values are a list of conversion errors
+        exceptions = list(self.avd_schema.convert(self.data))
+        conversion_counter = self.handle_exceptions(exceptions, self.conversion_mode)
+        # This is not yet effective, since the converter does not yield on succesful conversion.
+        # TODO: Get converter to return the data that was converted, so the user can see what needs to be updated.
+        if conversion_counter:
+            result_messages.append(f"{conversion_counter} data conversions done to conform to schema.")
+        return
+
+    def validate_data(self, result_messages):
```
gmuloc, line 107: similarly for this one we could return result_messages or empty and test on this info if this is valid or not

ansible_collections/arista/avd/plugins/action/validate_and_template.py
```diff
@@ -0,0 +1,168 @@
...
+        # avd_schema.validate returns a generator, which we need to run over to perform the actual validations.
+        # The returned values are a list of validation errors
+        exceptions = list(self.avd_schema.validate(self.data))
+        validation_counter = self.handle_exceptions(exceptions, self.validation_mode)
+        if validation_counter:
+            result_messages.append(f"{validation_counter} errors found during schema validation of input vars.")
+            return False
+        return True
+
+    def handle_exceptions(self, exceptions, mode):
```
gmuloc, line 121: Let's add some minimal docstrings

ansible_collections/arista/avd/plugins/action/validate_and_template.py
```diff
@@ -0,0 +1,168 @@
...
+            output = add_md_toc(output, skip_lines=self.md_toc_skip_lines)
+
+        write_file_result = self.write_file(output, task_vars)
+
+        # Return result with the result from the copy operation
+        return write_file_result
+
+    def write_file(self, content, task_vars):
+        # The write_file function is implementing the Ansible 'copy' action_module, to benefit from Ansible builtin functionality like 'changed'.
+        # Reuse task data
```
gmuloc:
```suggestion
        """
        This function implements the Ansible 'copy' action_module, to benefit from Ansible builtin functionality like 'changed'.
        Reuse task data
        """
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
```diff
@@ -27,38 +27,36 @@
...
-  # since it is responsible for upgrading input data to v4.0 models.
-  tags: [always, validate]
-  arista.avd.validate:
-    schema: "{{ lookup('ansible.builtin.file', role_schema_path) | from_yaml }}"
-    conversion_mode: "{{ avd_validate_conversion_mode }}"
-    validation_mode: "{{ avd_validate_validation_mode }}"
-  delegate_to: localhost
-
 - name: Generate eos intended configuration
   tags: [build, provision]
```
gmuloc, line 39: so now that the validate "which should not be skipped using AVD v3.x" is not having the `always` tags - could there be situation where people are going to skip it?

ansible_collections/arista/avd/plugins/action/validate_and_template.py
ClausHolbechArista: Because it does not work in our oldest supported versions of Ansible.

ansible_collections/arista/avd/plugins/README.md
ClausHolbechArista: Nobody is meant to use this :) At some point we should use our schema table format for this. For now I just produced this based on the module doc.

ansible_collections/arista/avd/plugins/action/validate_and_template.py
ClausHolbechArista: Because this is not always called in case of disabled mode, so we would have to use extend on the calling function. This is simpler IMO.

ansible_collections/arista/avd/plugins/action/validate_and_template.py
ClausHolbechArista: That would just add more logic to the calling code instead of just updating the original value.

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
ClausHolbechArista: Well we no longer care, since it is part of every module that needs it. We will still respect the `avd_validate_conversion_mode` so if people want to break it, they can :)

gmuloc (approved):
LGTM

PR #2218, opened 2022-10-24
Feat(eos_designs): Add support for Trident 2 devices (7050X2)

### LINKED ISSUE #2216: Add support for T2 devices (7050X2)
### Enhancement summary

Add support for Trident 2 devices.
Today, no support in AVD is present. These devices are very similar to T3, with the difference that "trident_forwarding_table_partition flexible" command is not supported.
Devices are still in support, and to keep a difference with T3, we agreed to add a new platform.

eos_designs

### Use case example

Allow 7050X2 as new platform:
```
l3leaf:
  defaults:
    platform: 7050X2
```

Add a new platform under default_platform_settings:
```
default_platform_settings:
  - platforms: ['7050X2']
    reload_delay:
      mlag: 300
      non_mlag: 330
    feature_support:
```

_No response_

### Additional context

_No response_

### Contributing Guide

xaviramon:
## Change Summary

Add new platform under default_platform_settings and set the right variables to support 7050X2 devices.

## Related Issue(s)

Fixes #2216 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add new platform:

## How to test
Molecule test included

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/mgmt_interface_platform_7050X2.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/mgmt_interface_platform_7050X2.cfg
@@ -1,0 +1,136 @@
+!RANCID-CONTENT-TYPE: arista
+!
+boot secret sha512 a153de6290ff1409257ade45f
+!
+daemon TerminAttr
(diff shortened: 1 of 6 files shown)


ClausHolbechArista:
The added platform is identical to the `default` platform, so it is not really needed. If you enter platform 7050X2 it will be mapped to the `default` platform, and get the settings you want.

xaviramon:
@ClausHolbechArista right but in the documentation we will have 'default' instead of '7050X2'. It is true that it is more cosmetic issue but if we have a specific platform it makes sense that it is documented as so.;

ClausHolbechArista:
> @ClausHolbechArista right but in the documentation we will have 'default' instead of '7050X2'. It is true that it is more cosmetic issue but if we have a specific platform it makes sense that it is documented as so.;

I would not expect `default` in the docs. It is just the `platform_settings` that would be taken from `default`. Still keep the `platform` var under the node to what you want.

PR #2258, opened 2022-11-03
Feat(eos_cli_config_gen): Add key_type for ntp.authentication_keys

### LINKED ISSUE #2211: ntp authentication_keys not applied with the right type
### Issue Summary

When using config_gen for ntp authentification, the type 7 key provided in the yml is applied as a plaintext key in the generated config. It look like the template is misisng the "7" type in the generated text.
**
image
**

will trigger the ntp [template](https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ntp.j2):  
but the jinja template is missing the type 7 setting, making the key viewed as a plaintext key instead of a type 7 by the switch.

image

eos_cli_config_gen

### How do you run AVD ?

Ansible AWX

### Steps to reproduce

```shell
In config_gen, configure NTP with authentification. make sure to set the key parameter and the authenticate to true.

Generate the config

You can obesrve that the configlet generated put the key as a plaintext key instead of a type 7, making the ntp not working properly
```

### Relevant log output

_No response_

### Contributing Guide

philippebureau:
## Change Summary

adding key_type under ntp in eos_cli_config_gen
To prevent impact on existing implementations, the current behaviour will be preserved until AVD 4.0.0 (see release notes)
Starting with AVD 4.0.0, if no key type is defined, type 7 will be default.

## Related Issue(s)

Fixes #2211 

## Component(s) name

## Proposed changes
Add a variable to define ntp encryption key type and define with type 7, 8a and no key for ntp authentication keys

example ( key 1 will use no key (old behaviour), key 2 type 7 and key 3 type 8a)
```yaml
  authentication_keys:
    - id: 1
      hash_algorithm: "md5"
      key: "7 044F0E151B"
    - id: 2
      hash_algorithm: "md5"
      key: "044F0E151B"
      key_type: 7
    - id: 3
      hash_algorithm: "sha1"
      key: "$BYk2Sjahe+D9T7uDgIItSA==$JTw5JOAPcYEo0O2hsvsxFQ==$C7wmpXOo"
      key_type: 8a
```

## How to test

tested without key_type set to confirm type 7 is used and test with other values

output in molecule:
```eos
ntp authentication-key 1 md5 7 044F0E151B
ntp authentication-key 2 md5 7 044F0E151B
ntp authentication-key 3 sha1 8a $BYk2Sjahe+D9T7uDgIItSA==$JTw5JOAPcYEo0O2hsvsxFQ==$C7wmpXOo

```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/docs/release-notes/3.x.x.md
+++ b/ansible_collections/arista/avd/docs/release-notes/3.x.x.md
@@ -21,6 +21,12 @@ No changes are required for existing deployments.
   - `node_type_keys.<key>.custom_templates_extra_vars` introduced in 3.7.0 has been removed. This is non-breaking since the templating environment
   used for custom templates now supports all host and group vars by default.
 
+#### Changes to eos_cli_config_gen role
+
+- **NTP authentication keys variables:**
+
+  - `authentication_key.key_type` introduced in 3.8.0 allow you to define the authentication key type.  If the key type is not defined, the previous behavior will be preserved.  The key type will be set to 7 as default starting in 4.0.0.
+
 ### Changes to requirements
 
 - AVD now requires ansible-core from **2.11.3** to **2.12.x** excluding **2.12.0** to **2.12.5**<br>
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
@@ -23,6 +23,6 @@ ntp:
       key: "044F0E151B"
     - id: 2
       hash_algorithm: "sha1"
-      key: "15060E1F10"
+      key: "$BYk2Sjahe+D9T7uDgIItSA==$JTw5JOAPcYEo0O2hsvsxFQ==$C7wmpXOo"
       key_type: 8a
   trusted_keys: "1-2"
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
@@ -20,9 +20,13 @@ ntp:
   authentication_keys:
     - id: 1
       hash_algorithm: "md5"
-      key: "044F0E151B"
+      key: "7 044F0E151B"
     - id: 2
+      hash_algorithm: "md5"
+      key: "044F0E151B"
+      key_type: 7
+    - id: 3
       hash_algorithm: "sha1"
       key: "$BYk2Sjahe+D9T7uDgIItSA==$JTw5JOAPcYEo0O2hsvsxFQ==$C7wmpXOo"
       key_type: 8a
-  trusted_keys: "1-2"
+  trusted_keys: "1-3"
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
@@ -24,4 +24,5 @@ ntp:
     - id: 2
       hash_algorithm: "sha1"
       key: "15060E1F10"
+      key_type: 8a
   trusted_keys: "1-2"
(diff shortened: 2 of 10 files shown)


ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ntp.yml
```diff
@@ -24,4 +24,5 @@ ntp:
     - id: 2
       hash_algorithm: "sha1"
       key: "15060E1F10"
+      key_type: 8a
```
ClausHolbechArista, line 31: please correct the keys to be correct for the type.

philippebureau: fixed invalid key

ClausHolbechArista (commented):
While I agree this is a fix since the documentation is misleading, this is also a breaking change for ppl who have worked around the issue by using a clear-text password or by inserting `7 ` before their password.
@carlbuchmann 
Maybe worth postponing to 4.x or at the very least call it out in more detail in release-notes?
Please add a description to draft release-notes as part of this PR.
Alternatively we could remove the default 7 in this PR, and only add the type if it is specifically set. Then the behavior would be 1:1 for existing deployments. We can then add the default in 4.0.

philippebureau:
updated the ntp.j2 template with @ClausHolbechArista inputs

ansible_collections/arista/avd/roles/eos_cli_config_gen/README.md
```diff
@@ -1993,7 +1993,8 @@ ntp:
   authentication_keys:
     - id: < key_identifier | 1-65534 >
       hash_algorithm: < md5 | sha1 >
-      key: "< type7_obfuscated_key >"
+      key: "< obfuscated_key >"
+      key_type: < 0 | 7 | 8a | default -> 7 >
```
carlbuchmann: The default 7 will be implemented in 4.0, so let's remove it from the current documentation.
```suggestion
      key_type: < 0 | 7 | 8a >
```

ClausHolbechArista: Maybe this is the right place to add a description
`# From AVD4.0 the value of key_type will be 7 by default`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ntp.j2
```diff
@@ -5,7 +5,12 @@
 {%         if authentication_key.id is arista.avd.defined and
               authentication_key.key is arista.avd.defined and
               authentication_key.hash_algorithm is arista.avd.defined %}
-ntp authentication-key {{ authentication_key.id }} {{ authentication_key.hash_algorithm }} {{ authentication_key.key }}
+{%             set ntp_auth_key_cli = "ntp authentication-key "  ~ authentication_key.id ~ " " ~ authentication_key.hash_algorithm %}
```
ClausHolbechArista: you have a double space before `~ authentication_key.id`

philippebureau: fixed

ClausHolbechArista (approved):

ansible_collections/arista/avd/docs/release-notes/3.x.x.md
```diff
@@ -21,6 +21,12 @@ No changes are required for existing deployments.
   - `node_type_keys.<key>.custom_templates_extra_vars` introduced in 3.7.0 has been removed. This is non-breaking since the templating environment
   used for custom templates now supports all host and group vars by default.
 
+#### Changes to eos_cli_config_gen role
+
+- **NTP authentication keys variables:**
+
+  - `authentication_key.key_type` introduced in 3.8.0 allow you to define the authentication key type.  If the key type is not defined, the previous behavior will be preserved.  The key type will be set to 7 as default starting in 4.0.0.
```
gmuloc, line 28: Probably worth stating what the previous behavior is in this otherwise as a common user I don't know where to look for it.

gmuloc (approved):
some small comment that can be ignored if you feel that's enough.

LGTM

ansible_collections/arista/avd/docs/release-notes/3.x.x.md
carlbuchmann: I think it's enough for this context - we will capture more info on this in 4.0 when behavior changes.

PR #2261, opened 2022-11-04
Fix(eos_validate_state): Always create directory for reports

chetryan:
## Change Summary

Currently there is a bug in `eos_validate_state` where if the task is called with any of the tags *and* the directory for the reports does not yet exists, it is not created. This results in the play failing when the vars need to be created in that directory. 

This Pul Requests adds the following line to always run the task to create the directory, irrelevant what tags are supplied at runtime. 
```yaml
tags:
  - always
```

## Related Issue(s)

No issues had been raised about this. 

## Component(s) name

## Proposed changes

This change will make sure the task `Create required output directories if not present` is always run, even when tags are used to just run subsets of tests. 

## How to test

1. Delete any directory previously used for reports (example: `inventory/reports`)
2. Run the playbook that uses `eos_validate_state` with some tags. Example:

```shell
ansible-playbook <playbook_name>.yml --tags interfaces_state -i <inventory>.yml
```

3. The directory where reports are save should be created.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
+++ b/ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
@@ -20,6 +20,8 @@
     - "{{ eos_validate_state_dir }}"
   delegate_to: localhost
   run_once: true
+  tags:
+    - always
 
 - name: Include device intended structure configuration variables
   include_vars: "{{ filename }}"


ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -20,6 +20,8 @@
     - "{{ eos_validate_state_dir }}"
   delegate_to: localhost
   run_once: true
+  tags:
+    - always
```
ClausHolbechArista, line 24: Maybe align with the reports task, so always and reports. So if run with skip-tags reports this is not run

ClausHolbechArista: nevermind.

ClausHolbechArista (approved):

tgodaA (approved):
LGTM

PR #2289, opened 2022-11-22
Feat(eos_designs): Add support for system mac address derived engine ID

### LINKED ISSUE #2180: Add system mac-address based localized snmp-server engineID
### Enhancement summary

The current AVD snmp server settings local engineID is produced with the following formula (method 1):

>  `(inventory_hostname ~ switch.mgmt_ip) | ansible.builtin.hash('sha1')`

However, Arista EOS produces a local snmp-server engineID based on the following formula (method 2):

> "f5717f" + str(mac_address).replace(":", "") + "00"

This feature request is to enhance snmp-server settings in eos_designs in order to produce localized engineID in the same way as Arista EOS, which will require the system mac-address as input.  The snmp-server settings must take into account that mac-address is optional, so we could check whether mac_address is defined; if true, use method 2, else fallback to use method 1.

Also, as `mac_address` is currently only used by `dhcp_provisioner`, this feature request will also seek to update the `eos_designs_facts.py `to include `mac_address` in `avd_switch_facts`

reference: http://www.iana.org/assignments/enterprise-numbers/enterprise-numbers - snmp engineID is computed using a derivative of the Private Enterprise Number (30065 decimal, 0x7517 hex)

> 

eos_designs, plugins

### Use case example

This feature would allow operators to optionally supply system mac address per node in order to generate local engineID using the same method as Arista EOS

update `eos_designs_facts.py`:

```
@@ -425,6 +425,10 @@
         return get(self._switch_data_combined, "mgmt_ip")
     
     @cached_property
+    def mac_address(self):
+        return get(self._switch_data_combined, "mac_address")
+
+    @cached_property
     def platform(self):
         return get(self._switch_data_combined, "platform")
```

update eos_designs/templates/base/snmp-server.j2:

```
@@ -2,8 +2,12 @@
 snmp_server:
 {% if snmp_settings is arista.avd.defined %}
 {%     if snmp_settings.compute_local_engineid is arista.avd.defined(true) %}
-{%         set local_engine_id = (inventory_hostname ~ switch.mgmt_ip) | ansible.builtin.hash('sha1') %}
-    engineid:
+{%         if switch.mac_address is arista.avd.defined %}
+{%             set local_engine_id = 'f5717f' ~ (switch.mac_address | regex_replace (":","")) ~ '00' %}
+{%         else %}
+{%             set local_engine_id = (inventory_hostname ~ switch.mgmt_ip) | ansible.builtin.hash('sha1') %}
+{%         endif %}
+    engine_ids:
       local: {{ local_engine_id }}
 {%     endif %}
 {%     if snmp_settings.contact is arista.avd.defined %}
```

We will also need to update related documentation and tests (mac-address expected format will be alphanumeric or colon separated alphanumeric )

alternative would be to simply consume existing (method 1) snmp-server generated local engineID

Also, a suggestion from other contributor:

The mac_address (and serial_number) keys make more sense in the inventory file, than distributed through the group_vars.  These are properties of the inventory host and do not change based on the device's role. 

An alternative to supporting `mac_address` in eos_designs_facts.py would be to support mac_address in the inventory file.

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

Allow to leverage system_mac_address to compute default engineID

## Related Issue(s)

Fixes #2180

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Enable a new option to generate the local Engine ID which is to derive the default one from the system MAC address (available in show version) using the following formula `f5717f + system_mac_address + 00`

The implementation introduces one new fabric-topology variable which is `system_mac_address` (different from the `mac_address` one currently leveraged by the DHCP role)

```
 ---
  <node_type_key>:
    defaults:
      # Define vars for all nodes of this type
    node_groups:
      <node group name>:
      # Vars related to all nodes part of this group
        nodes:
          <node inventory hostname>:
            # Vars defined per node
    nodes:
      <node inventory hostname>:
        # Vars defined per node

        # Unique identifier per node_type | Required.
        id: < integer >

        # Node management IP address | Optional.
        mgmt_ip: < IPv4_address/Mask >

        # The following variable SHOULD be set to the MAC address
        # available in show version
        # NOTE: the mac_address variable used in dhcp_provisioner role is
        # different from this variable
        # Expected format: xx:xx:xx:xx:xx:xx
        system_mac_address: < str >
```

and add one new option under `snmp_settings` in `eos_designs`:

```
snmp_settings:
  contact: < contact_info >
  location: < boolean | default -> false > # Formatted as: {{ fabric_name }} {{ dc_name }} {{ pod_name }} {{ switch_rack }} {{ inventory_hostname }}
  # Generate a local engineId for SNMP by hashing via SHA1 the string
  # generated via the concatenation of the hostname plus the management IP.
  # {{ inventory_hostname }} + {{ switch.mgmt_ip }}
  # This SHOULD NOT be used with `default_engineid_from_system_mac`
  # `default_engineid_form_system_mac` will take precedence
   compute_local_engineid: < boolean | default -> false >
  # Generate the switch default engine id for AVD usage
  # To use this, system_mac_address MUST be set for the device
  # The formula is f5717f + system_mac_address + 00
  # This SHOULD NOT be used with compute_local_engineid
+  default_engineid_from_system_mac: < boolean | default -> false >
```

## How to test

molecule test (made the previous test for SNMP AUTOGEN engine ID more atomic than they were at the same time)

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
```diff
@@ -78,7 +78,14 @@ snmp_settings:
   # Generate a local engineId for SNMP by hashing via SHA1 the string
   # generated via the concatenation of the hostname plus the management IP.
   # {{ inventory_hostname }} + {{ switch.mgmt_ip }}
+  # This SHOULD NOT be used with `default_engineid_from_system_mac`
+  # `default_engineid_form_system_mac` will take precedence
```
ClausHolbechArista:
```suggestion
  # `default_engineid_from_system_mac` will take precedence
```

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
```diff
@@ -78,7 +78,14 @@ snmp_settings:
   # Generate a local engineId for SNMP by hashing via SHA1 the string
   # generated via the concatenation of the hostname plus the management IP.
   # {{ inventory_hostname }} + {{ switch.mgmt_ip }}
+  # This SHOULD NOT be used with `default_engineid_from_system_mac`
```
ClausHolbechArista: Could we simplify this by creating a `compute_local_engineid_source: < "hostname_and_ip" | "system_mac" | default -> "hostname_and_ip" >`
Then we don't have "competing" vars, but instead offer customization and future enhancements.

ClausHolbechArista: It would also reduce the amount of doc for other vars we would need to update. Ex. check line 89.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
gmuloc: there should not be a default value but yeah we can do this - by default it should stay disabled I think

Let me rework it with your suggestion!

gmuloc: ok following our chat - yes new variable with a default_value ;)

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -132,6 +132,13 @@ defaults <- node_group <- node_group.node <- node
 
       # Node management IP address | Optional.
       mgmt_ip: < IPv4_address/Mask >
+
+      # The following variable SHOULD be set to the MAC address
+      # available in show version
+      # NOTE: the mac_address variable used in dhcp_provisioner role is
+      # different from this variable
+      # Expected format: xx:xx:xx:xx:xx:xx
+      system_mac_address: < str >
```
ClausHolbechArista: we could support setting this as a regular inventory variable too.

gmuloc: Rebased and pushed - system_mac_address now available at both places and test added to check precedence - please give feedback on the way it is documented.

PS: had to rebase and fix some broken import for errors

cf: c6bc86c

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
```diff
@@ -71,14 +71,27 @@ name_servers:
   - < IPv4_address_1 >
   - < IPv4_address_2 >
 
+# System setinngs | Optional
+# `system_mac_address` at this level has less precedence
+# than at the fabric-topology level
+# Expected format: "xx:xx:xx:xx:xx:xx"
+system_mac_address: < str >
```
ClausHolbechArista:
```suggestion
# System Mac Address | Optional
# Set to the same MAC addres as available in "show version" on the device.
# "system_mac_address" can also be set under "Fabric Topology".
# If both are set, the setting under "Fabric Topology" takes precedence.
system_mac_address: < "xx:xx:xx:xx:xx:xx" >
```

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -132,6 +132,15 @@ defaults <- node_group <- node_group.node <- node
...
       mgmt_ip: < IPv4_address/Mask >
+
+      # The following variable SHOULD be set to the MAC address
+      # available in show version
+      # NOTE: the mac_address variable used in dhcp_provisioner role is
+      # different from this variable
+      # This topology variable has precedence over system_mac_address defined at
+      # the host variable level
+      # Expected format: "txx:xx:xx:xx:xx:xx"
+      system_mac_address: < str | default -> hostvar system_mac_address -> None >
```
ClausHolbechArista:
```suggestion
      # System Mac Address | Optional
      # Set to the same MAC addres as available in "show version" on the device.
      # NOTE: the "mac_address" variable used in dhcp_provisioner role is
      # different from this variable
      # "system_mac_address" can also be set directly as a hostvar.
      # If both are set, the setting under "Fabric Topology" takes precedence.
      system_mac_address: < "xx:xx:xx:xx:xx:xx" >
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
```diff
@@ -320,8 +328,24 @@ def snmp_server(self):
...
+
+        local_engine_id = None
+
+        if snmp_settings.get("compute_local_engineid") is True:
+            compute_source = get(snmp_settings, "compute_local_engineid_source", default="hostname_and_ip")
+            if compute_source == "hostname_and_ip":
+                local_engine_id = sha1(f"{self._hostname}{self._mgmt_ip}".encode("utf-8")).hexdigest()
+            elif compute_source == "system_mac":
+                if self._system_mac_address is None:
+                    raise AristaAvdError("default_engine_id_from_system_mac: true requires system_mac_address to be set!")
```
ClausHolbechArista: This should raise the missingvariable error.

gmuloc: Fixed in 4bcf83d

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
gmuloc: Fixed in 4bcf83d

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
gmuloc: Fixed in 4bcf83d

ClausHolbechArista (approved):

ansible_collections/arista/avd/roles/eos_designs/doc/fabric-topology.md
```diff
@@ -132,6 +132,14 @@ defaults <- node_group <- node_group.node <- node
 
       # Node management IP address | Optional.
       mgmt_ip: < IPv4_address/Mask >
+
+      # System Mac Address | Optional
+      # Set to the same MAC addres as available in "show version" on the device.
```
jonxstill:
```suggestion
      # Set to the same MAC address as available in "show version" on the device.
```

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
```diff
@@ -71,14 +71,27 @@ name_servers:
   - < IPv4_address_1 >
   - < IPv4_address_2 >
 
+# System Mac Address | Optional
+# Set to the same MAC addres as available in "show version" on the device.
```
jonxstill:
```suggestion
# Set to the same MAC address as available in "show version" on the device.
```

jonxstill (approved):
Couple of typos in docs but other than that LGTM.

ansible_collections/arista/avd/roles/eos_designs/doc/management-settings.md
gmuloc: Thanks! Fixed in 46cb480

PR #2303, opened 2022-11-30
Feat(eos_cli_config_gen): Add schema for management_api_gnmi

KonikaChaurasiya-GSLab:
### Contributor Checklist

- [x] Create schema fragment matching data model described in README.md and README_v4.0.md
  - README.md is most complete with all keys. README_v4.0 includes partial data models after conversion to lists.
  - Schema fragment path is `roles/eos_cli_config_gen/schemas/schema_fragments/<data_model_key>.schema.yml`.
  - Copy line 1-5 from another schema (comments and outer type:dict).
  - Refer to [schema documentation](https://avd.sh/en/devel/docs/input-variable-validation-BETA.html) for syntax and/or use YAML Lint plugin from Redhat in VSCode.
  - Use `convert_types` on value that could be mixed type or misinterpreted like integers and numeric strings.
- [x] If the data model has been converted from wildcard dicts:
  - Add `convert_types: ['dict']` to the schema.
  - Remove `convert_dicts` from the `templates/eos/<>.j2` and `templates/documentation/<>.j2` templates.
- [x] Run `molecule converge -s build_schemas_and_docs` to update schema and documentation.
- [x] Test by running `molecule converge -s eos_cli_config_gen` and verify no errors or changes to generated configs/docs.

### Reviewer Checklist

- Reviewer 1:
  - [ ] Verify that data model is fully covered in the described schema. Easiest by looking at the generated documentation.
  - [ ] Verify that `convert_dicts` has been removed from templates as applicable.
  - [ ] Verify no changes to configs/docs on any molecule scenario
  - [ ] Verify that CI pass

- Reviewer 2:
  - [ ] Verify that data model is fully covered in the described schema. Easiest by looking at the generated documentation.
  - [ ] Verify that `convert_dicts` has been removed from templates as applicable.
  - [ ] Verify no changes to configs/docs on any molecule scenario
  - [ ] Verify that CI pass

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/management_api_gnmi.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/management_api_gnmi.schema.yml
@@ -1,0 +1,62 @@
+# yaml-language-server: $schema=../../../../plugins/plugin_utils/schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  management_api_gnmi:
+    type: dict
+    keys:
+      provider:
+        type: str
+        default: eos-native
+      transport:
+        type: dict
+        keys:
+          grpc:
+            type: list
+            items:
+              type: dict
+              keys:
+                name:
+                  type: str
+                  description: Transport name
+                ssl_profile:
+                  type: str
+                  description: SSL profile name
+                vrf:
+                  type: str
+                  description: VRF name is optional
+                notification_timestamp:
+                  type: str
+                  valid_values: ["send-time", "last-change-time"]
+                  description: |
+                    Per the GNMI specification, the default timestamp field of a notification message is set to be
+                    the time at which the value of the underlying data source changes or when the reported event takes place.
+                    In order to facilitate integration in legacy environments oriented around polling style operations,
+                    an option to support overriding the timestamp field to the send-time is available from EOS 4.27.0F.
+                ip_access_group:
+                  type: str
+                  description: ACL name
+      enable_vrfs:
+        type: list
+        primary_key: name
+        convert_types:
+          - dict
+        description: |
+          Enable VRFs will be deprecated in AVD v4.0.
+          These should not be mixed with the new keys above.
+        items:
+          type: dict
+          keys:
+            name:
+              type: str
+              description: VRF name
+            access_group:
+              type: str
+              description: Standard IPv4 ACL name
+      octa:
+        type: dict
+        description: |
+          Octa key will be deprecated in AVD v4.0.
+          These should not be mixed with the new keys above.
+          Octa activates `eos-native` provider and it is the only provider currently supported by EOS.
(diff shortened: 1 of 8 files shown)


ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-gnmi.md
```diff
@@ -59,10 +59,10 @@ interface Management1
 ```eos
 !
 management api gnmi
-   transport grpc MGMT
+   transport grpc {'access_group': 'ACL-GNMI', 'name': 'MGMT'}
```
carlbuchmann: @KonikaChaurasiya-GSLab , this appears like an issue in the template. The output shouldn't change here.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/management-api-gnmi.j2
```diff
@@ -3,14 +3,14 @@
 !
 management api gnmi
 {%     for vrf in management_api_gnmi.enable_vrfs | arista.avd.natural_sort %}
-{%         if vrf == 'default' %}
+{%         if vrf.name == 'default' %}
    transport grpc default
 {%         else %}
    transport grpc {{ vrf }}
```
Shivani-chourasiya: Here it should be vrf.name

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/management_api_gnmi.schema.yml
```diff
@@ -0,0 +1,62 @@
...
+          type: dict
+          keys:
+            name:
+              type: str
+              description: VRF name
+            access_group:
+              type: str
+              description: Standard IPv4 ACL name
+      octa:
+        type: dict
```
gmuloc, line 58: Should not it be a boolean? - not sure how it would have been used before as our documentation is empty

```
{%     if management_api_gnmi.octa is defined %}
   provider eos-native
{%     endif %}
```

Though looking at molecule we are using it like this:

```
--management_api_gnmi:
    enable_vrfs:
      MGMT:
        access_group: ACL-GNMI
      MONITORING:
    octa:
```

so probably just the fact that the dict exist makes it work .. could it break existing implementations ? (we can always argue that it is how it was intended to be used ..)

@ClausHolbechArista

ClausHolbechArista: This is one of the problematic data models which is not well defined. It should have been a boolean, but it was created like a "dangling" key, which would also work if set with `null`, so it is difficult to provide a schema covering all variations. Normally we have made these keys into dicts and provided an empty dict in the molecule vars, but you are correct that it might be breaking here.
Let us sync up offline and discuss a solution.

ClausHolbechArista: So since the documented value is `null` and we basically ignore `null` values during schema validation, this will not block anything. I suggest to keep it as dict in the schema, and move on. This is deprecated for 4.0, so it will only be an issue if someone used another value than `null` or `dict` for 3.8+.

gmuloc (approved):
LGTM

ClausHolbechArista (approved):

PR #2307, opened 2022-12-01
Refactor: Jinja templating with caching

ClausHolbechArista:
## Change Summary

Jinja templating with caching

## Related Issue(s)

Improves runtime for jinja templating.

## Component(s) name

Everything that uses templating

## Proposed changes

- Give `searchpath` as argument instead of gleaning from vars
- Enable bytecode caching of compiled jinja2 templates. This avoids recompiling the templates on every run for every device.
- Use the `Templar` instance already created for action module instead of creating a new.
  This is done by loading new temporary context with:
    - updated vars
    - caching enabled

Considerations for later: 
- Create script that clears cache and compiles all templates again
- Allow j2cache to be stored in repo (so containers etc can take advantage of this)
  - Issues with eos_cli_config_gen custom templates, since they are included in jinja and 
    shares the same cache as our templates

## How to test

No changes to molecule.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
@@ -1,4 +1,4 @@
-from pathlib import Path, PurePath
+from pathlib import Path
 
 from ansible.module_utils._text import to_text
 from jinja2 import FileSystemBytecodeCache
@@ -36,7 +36,7 @@ def template(template_file, template_vars, templar, searchpath):
     j2template, dummy = loader._get_file_contents(template_file_path)
     j2template = to_text(j2template)
 
-    cache_dir = Path(PurePath(template_file_path).parent, ".j2cache")
+    cache_dir = Path(template_file_path).parent.joinpath(".j2cache")
     # Create cache_dir if not existing. We can asume the parent dir is there since we found the template file.
     cache_dir.mkdir(0o775, False, True)
 
--- a/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
@@ -37,7 +37,7 @@ def template(template_file, template_vars, templar, searchpath):
     j2template = to_text(j2template)
 
     cache_dir = Path(template_file_path).parent.joinpath(".j2cache")
-    # Create cache_dir if not existing. We can asume the parent dir is there since we found the template file.
+    # Create cache_dir if not existing. We can assume the parent dir is there since we found the template file.
     cache_dir.mkdir(0o775, False, True)
 
     bytecode_cache = FileSystemBytecodeCache(cache_dir)
--- a/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
@@ -1,4 +1,7 @@
+from pathlib import Path, PurePath
+
 from ansible.module_utils._text import to_text
+from jinja2 import FileSystemBytecodeCache
 
 
 def template(template_file, template_vars, templar, searchpath):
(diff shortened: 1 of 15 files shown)


ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
```diff
@@ -33,6 +36,16 @@ def template(template_file, template_vars, templar, searchpath):
     j2template, dummy = loader._get_file_contents(template_file_path)
     j2template = to_text(j2template)
 
-    templar.available_variables = template_vars
-    result = templar.template(j2template, convert_data=False, escape_backslashes=False)
+    cache_dir = Path(PurePath(template_file_path).parent, ".j2cache")
+    # Create cache_dir if not existing. We can asume the parent dir is there since we found the template file.
```
gmuloc:
```suggestion
    # Create cache_dir if not existing. We can assume the parent dir is there since we found the template file.
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
```diff
@@ -33,6 +36,16 @@ def template(template_file, template_vars, templar, searchpath):
     j2template, dummy = loader._get_file_contents(template_file_path)
     j2template = to_text(j2template)
 
-    templar.available_variables = template_vars
-    result = templar.template(j2template, convert_data=False, escape_backslashes=False)
+    cache_dir = Path(PurePath(template_file_path).parent, ".j2cache")
```
tgodaA: We can do it this way instead of creating the Path twice

```suggestion
    cache_dir = Path(template_file_path).parent.joinpath(".j2cache")
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
```diff
@@ -33,6 +36,16 @@ def template(template_file, template_vars, templar, searchpath):
     j2template, dummy = loader._get_file_contents(template_file_path)
     j2template = to_text(j2template)
 
-    templar.available_variables = template_vars
-    result = templar.template(j2template, convert_data=False, escape_backslashes=False)
+    cache_dir = Path(template_file_path).parent.joinpath(".j2cache")
+    # Create cache_dir if not existing. We can asume the parent dir is there since we found the template file.
+    cache_dir.mkdir(0o775, False, True)
+
+    bytecode_cache = FileSystemBytecodeCache(cache_dir)
```
gmuloc, line 44: Maybe we can try to define our own pattern for the cached files? https://jinja.palletsprojects.com/en/3.1.x/api/#jinja2.FileSystemBytecodeCache 

That could explain the names you were seeing (it is using the cache key) - otherwise this is the code: https://github.com/pallets/jinja/blob/52843b5cbf635b37a82ac0b6c901921a8ee076ff/src/jinja2/bccache.py#L154

but then the clear() method is implemented so we could imagine clearing everything and recompiling everything if a jinja2 template changes https://github.com/pallets/jinja/blob/52843b5cbf635b37a82ac0b6c901921a8ee076ff/src/jinja2/bccache.py#L315

ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
ClausHolbechArista: implemented in https://github.com/aristanetworks/ansible-avd/pull/2307/commits/34f55e3e7773eb18da19bba7acb4af280a8422ba

ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
ClausHolbechArista: implemented in https://github.com/aristanetworks/ansible-avd/pull/2307/commits/cf591d27e84102d0213d0a19ceb262b79b0dfcae

ansible_collections/arista/avd/plugins/plugin_utils/utils/template.py
ClausHolbechArista: I have tried all these options. Remember the template.py is run for every template file, so a `.clear()` in here would clear for every run, defeating the purpose of cache.
The names of the files make sense to be the checksum. It will ensure that the cache is auto-updated when something changes. We could write a more clever FileSystemBytecodeCache which would track on filenames+crc and remove the invalid variant, but it would need to have the full path in the name, to handle duplicate names (ex. `main.j2`)

gmuloc: Resolving this post our discussion - the conclusion is that actually the hashed filename is used for the cache filename and so given that we don't recreate many new file / change many file names, we won't need to handle this in a much clever fashion than "delete your cache folder and run again the playbook"

gmuloc (approved):
LGTM

carlbuchmann (approved):
LGTM, tested with customer inventories and noticed ~25% improvement when running the build playbook (eos_designs/eos_cli_config_gen)!

PR #2322, opened 2022-12-07
Refactor(eos_designs): Move structured_config for connected_endpoints to Python

ClausHolbechArista:
## Change Summary

Move structured_config for connected_endpoints to Python

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Add python module for generating structured_config for connected_endpoints
- Remove jinja2 templates for connected_endpoints
- Change default_templates to use python module

## How to test

Molecule artifacts updated for all scenarios.
Only reordering of vars in structured_config. No changes to final configs.

### User Checklist

- N/A

### Repository Checklist


carlbuchmann (approved):
LGTM, tested with customer inventories!

ghost (commented):
LGTM!

ghost (approved):

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+from ansible_collections.arista.avd.plugins.plugin_utils.merge import merge
+from ansible_collections.arista.avd.plugins.plugin_utils.utils import default, get, get_item
+from ansible_collections.arista.avd.roles.eos_designs.python_modules.interface_descriptions import AvdInterfaceDescriptions
+from ansible_collections.arista.avd.roles.eos_designs.python_modules.ip_addressing import AvdIpAddressing
+
+
+class UtilsMixin:
+    """
+    Mixin Class with internal functions.
+    Class should only be used as Mixin to a AvdStructuredConfig class
```
gmuloc: to another Mixin?

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+                connected_endpoint["adapters"] = [adapter for adapter in connected_endpoint["adapters"] if self._hostname in adapter.get("switches", [])]
+
+                for adapter_index, adapter in enumerate(connected_endpoint["adapters"]):
+                    if "profile" not in adapter:
+                        continue
+
+                    port_profile = get_item(self._merged_port_profiles, "profile", adapter["profile"], default={})
+                    adapter = merge(port_profile, adapter, list_merge="replace", destructive_merge=False)
+                    adapter.pop("profile")
+                    connected_endpoint["adapters"][adapter_index] = adapter
```
gmuloc: Wondering if this is the most efficient to first recreate a new list of adapters line 41 by filtering and then reiterating over all of them a second time to handle profile.

Probably best to build the list once and handle everything in one iteration through the list.

```suggestion
                filtered_adapters = []
                for adapter in connected_endpoint["adapters"]:
                    if self._hostname not in adapter.get("switches", [])]:
                        continue
                     
                    if "profile" in adapter:
                        port_profile = get_item(self._merged_port_profiles, "profile", adapter["profile"], default={})
                        adapter = merge(port_profile, adapter, list_merge="replace", destructive_merge=False)
                        adapter.pop("profile")
                    
                    filtered_adapters.append(adapter)
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+        """
+        filtered_network_ports = []
+        for network_port in get(self._hostvars, "network_ports", default=[]):
+            if not self._match_regexes(network_port.get("switches"), self._hostname):
+                continue
+
+            if "profile" in network_port:
+                port_profile = get_item(self._merged_port_profiles, "profile", network_port["profile"], default={})
+                network_port = merge(port_profile, network_port, list_merge="replace", destructive_merge=False)
+                network_port.pop("profile")
```
gmuloc: We could write some `_apply_profile` to dict method that will do it both for network_port and adapter to optimise.

Can be done later of course ;)

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+
+    @cached_property
+    def _filtered_network_ports(self) -> list:
+        """
+        Return list of endpoints defined under "network_ports"
+        which are connected to this switch.
+        """
+        filtered_network_ports = []
+        for network_port in get(self._hostvars, "network_ports", default=[]):
+            if not self._match_regexes(network_port.get("switches"), self._hostname):
```
gmuloc: can probably use Python `any` here https://docs.python.org/3/library/functions.html#

```suggestion
if any(re.match(rf"^{regex}$", self._hostname) for regex in network_port.get("switches")):
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+        Return list of merged "port_profiles" where "parent_profile" has been applied.
+        """
+        merged_port_profiles = []
+
+        for port_profile in self._port_profiles:
+            if "parent_profile" in port_profile:
+                parent_profile = get_item(self._port_profiles, "profile", port_profile["parent_profile"], default={})
+                port_profile = merge(parent_profile, port_profile, list_merge="replace", destructive_merge=False)
+                port_profile.pop("parent_profile")
+            merged_port_profiles.append(port_profile)
```
gmuloc: Note that this will make us "kind of supporting but not exactly" multiple level of `parent_profile` and the outcome would be dependent of the order of `self._port_profiles`. That could lead to some confusing error.

I guess we don't catch this in our tests today as it is claimed as not supported

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+            adapter.get("server_ports"),
+        )
+        short_esi = str(short_esi)
+        if short_esi.lower() == "auto":
+            esi_hash_switches = "".join(adapter["switches"][:2])
+            esi_hash_switch_ports = "".join(adapter["switch_ports"][:2])
+            esi_hash_endpoint_ports = "".join(endpoint_ports[:2])
+            esi_hash = sha256(
+                f"{hash_extra_value}{esi_hash_switches}{esi_hash_switch_ports}{esi_hash_endpoint_ports}{channel_group_id}".encode("UTF-8")
+            ).hexdigest()
```
gmuloc: https://docs.python.org/3/library/hashlib.html
```suggestion
        if short_esi.lower() == "auto":
            esi_hash = sha256()
            esi_hash.update("".join(adapter["switches"][:2]).encode("UTF-8"))
            esi_hash.update("".join(adapter["switch_ports"][:2]).encode("UTF-8"))
            esi_hash.update("".join(endpoint_ports[:2]).encode("UTF-8"))
            esi_hash = esi_hash.hexdigest()
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+            short_esi = re.sub(r"([0-9a-f]{4})", "\\1:", esi_hash)[:14]
+
+        if len(short_esi.split(":")) != 3:
+            return None
```
gmuloc: Are we sure of this? if I give a `short_esi` to this method with a wrong format I guess I would prefer to be told? I know this is the current behavior to silently ignore but ..

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+                "dont_preempt": adapter_ethernet_segment.get("dont_preempt"),
+            }
+
+        return evpn_ethernet_segment
+
+    def _get_adapter_link_tracking_groups(self, adapter: dict) -> dict | None:
+        """
+        Return link_tracking_groups for one adapter
+        """
+        if not (get(adapter, "link_tracking.enabled") is True and self._link_tracking_groups is not None):
```
gmuloc: written this way it allows to not do the `get` if the cached property is None anyway
```suggestion
        if self._link_tracking_groups is None or get(adapter, "link_tracking.enabled") is not True:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,287 @@
...
+            return None
+
+        return [
+            {
+                "name": get(adapter, "link_tracking.name", default=self._link_tracking_groups[0]["name"]),
```
gmuloc, line 265: Not to be fixed now as this is the current behavior but this look fishy - we could in theory have multiple `link_tracking_groups` according to our documentation - maybe let's add a TODO to revisit this?

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -0,0 +1,286 @@
...
+    @cached_property
+    def _ptp_profile(self) -> str | None:
+        return get(self._hostvars, "switch.ptp.profile")
+
+    @cached_property
+    def _ptp_profiles(self) -> list:
+        """
+        Return ptp_profiles or []
+        """
+        return get(self._hostvars, "ptp_profiles", default=[])
```
gmuloc, line 279: Hmm maybe I am not reviewing the latest version? let's align with underlay and others

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        monitor_session_configs = []
+        for connected_endpoint in self._filtered_connected_endpoints:
+            for adapter in connected_endpoint["adapters"]:
+                if "monitor_sessions" not in adapter:
+                    continue
+
+                # Monitor session on Port-channel interface
+                if get(adapter, "port_channel.mode") is not None:
+                    default_channel_group_id = int("".join(re.findall(r"\d", adapter["switch_ports"][0])))
```
gmuloc, line 76: Maybe let's make a util function for getting the `default_channel_group_id` as it is used as well in port-channel and ethernet and it will be good to have it in one place if we want to modify it later

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py
```diff
@@ -0,0 +1,115 @@
...
+                    channel_group_id = get(adapter, "port_channel.channel_id", default=default_channel_group_id)
+
+                    port_channel_interface_name = f"Port-Channel{channel_group_id}"
+                    for monitor_session in adapter["monitor_sessions"]:
+                        monitor_session["interface"] = port_channel_interface_name
```
gmuloc: so here we modify the `monitor_session` instance in place in the `adapter["monitor_sessions"]` - I guess this is ok?

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py
```diff
@@ -0,0 +1,115 @@
...
+                # Monitor session on Port-channel interface
+                if get(network_port, "port_channel.mode") is not None:
+                    default_channel_group_id = int("".join(re.findall(r"\d", ethernet_interface_name)))
+                    channel_group_id = get(network_port, "port_channel.channel_id", default=default_channel_group_id)
+
+                    port_channel_interface_name = f"Port-Channel{channel_group_id}"
+                    for monitor_session in adapter["monitor_sessions"]:
+                        monitor_session["interface"] = port_channel_interface_name
+                        monitor_session_configs.append(monitor_session)
+                    continue
```
gmuloc: Could make a oneliner of these using `extend` + `map` + `lambda` though not sure it is worth it for readability

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/ethernet_interfaces.py
```diff
@@ -0,0 +1,125 @@
...
+        Return structured config for ethernet_interfaces
+        """
+        ethernet_interfaces = {}
+        for connected_endpoint in self._filtered_connected_endpoints:
+            for adapter in connected_endpoint["adapters"]:
+                for node_index, node_name in enumerate(adapter["switches"]):
+                    if node_name != self._hostname:
+                        continue
+
+                    # TODO: verify that length of all lists are the same
```
gmuloc: should we implement it now?

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py
```diff
@@ -0,0 +1,172 @@
...
+
+            connected_endpoint = {
+                "name": network_port.get("description"),
+                "type": "network_port",
+            }
+            for ethernet_interface_name in range_expand(network_port["switch_ports"]):
+                # Override switches and switch_ports to only render for a single interface
+                # The blank extra switch is only inserted to work around port_channel validations
+                # This also means that port-channels defined with network_ports data model will be single-port per switch.
+                # Caveat: "short_esi: auto" and "designated_forwarder_algorithm: auto" will not work correctly.
```
gmuloc, line 59: Let's create an issue to fix it for 4.0?

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py
```diff
@@ -0,0 +1,172 @@
...
+                },
+            },
+        }
+
+        # EVPN A/A
+        if (
+            "short_esi" in subinterface
+            and (short_esi := self._get_short_esi(adapter, channel_group_id, short_esi=subinterface["short_esi"], hash_extra_value=subinterface["number"]))
+            is not None
+        ):
```
gmuloc:
```suggestion
            (short_esi := self._get_short_esi(adapter, channel_group_id, short_esi=subinterface.get("short_esi", None), hash_extra_value=subinterface["number"]))
            is not None
        ):
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: I don't think it would allow 4-tier inheritance like `parent-parent-profile -> parent-profile -> profile -> adapter`, but it _would_ allow `parent-a -> profile-a -> adapter` *and* `profile-a -> profile-b -> adapter`, so reusing a profile in both parent and child level. It can be confusing, but is not really contradicting our docs.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: The format feature of schemas should catch this - but not there yet.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: It is behaving like documented. I am not sure what else to do here?
```
        link_tracking:
          enabled: < true | false >
          # The default group name is taken from fabric variable of the switch, link_tracking.groups[0].name with default value being "LT_GROUP1".
          # Optional if default link_tracking settings are configured on the node.
          name: < tracking_group_name >```

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 10 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py:279]: We cannot do the same, since we need to profiles specified per adapter.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py (hunk 12 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py]: I was aware, but chose to ignore. We don't need to touch the monitor_sessions again, but it is not clean. Let me look again.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py (hunk 13 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py]: You could say that about ~all our loops. Let's not go overboard.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py (hunk 16 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py]: This is not the same, as it will call the function and return. And it adds to the very long condition. I will see how to make it readable.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py (hunk 13 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py]: modified to extend with list comprehension and dict()

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py (hunk 12 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py]: modified to dict() to avoid changing original

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py (hunk 11 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py:76]: I will opt to postpone this to a separate PR. We have the same function in many different modules - including facts, so I think it justifies a separate PR.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py (hunk 16 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py]: done

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py (hunk 15 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/port_channel_interfaces.py:59]: I don't think there is a real good solution here, unless we introduce a some new knobs.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 8 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: done

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 6 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: done a variation

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 7 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: added raise

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 4 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: I have updated the helper function. We can consider moving it to utils at some point when we broaden support for matching like this.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 2 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: closing as agreed. Not worth it for small amount of adapters per endpoint.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py [hunk 17]
```diff
@@ -0,0 +1,116 @@
...
+
+
+class MonitorSessionsMixin(UtilsMixin):
+    """
+    Mixin Class used to generate structured config for one key.
+    Class should only be used as Mixin to a AvdStructuredConfig class
+    """
+
+    @cached_property
+    def monitor_sessions(self):
```
Shivani-chourasiya: shall we add return type here, as we are following in all the methods in other files?

ClausHolbechArista: Yes! Thank you for catching this. I will update.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: fixed

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: Later :)

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: Ended up doing a variant of this any way, because we need to merge profiles before checking if device is part of "switches". to cover the case where "switches" is set on the profile

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/ethernet_interfaces.py
ClausHolbechArista: Done - during the filtering

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
gmuloc: ok so I just realised what confused me - the fact that we use `destructive_merge=False` and reuse the `port_profile` name in this line:

```
port_profile = merge(parent_profile, port_profile, list_merge="replace", destructive_merge=False)
```

Can we maybe call the variable differently - like `merged_port_profile` ? As this looks like we are changing in place the port profile and in the situation you describe:
`parent-a -> profile-a -> adapter` and `profile-a -> profile-b -> adapter`

The outcome could be different if the order in `_port_profiles` because if we update the profiles in place with merge 
1. `[parent-a, profile-a, profile-b]`
or
2. `[profile-b, profile-a, parent-a]`

In 1. the parent-a would be merged with parent-a first and then profile-b would be updated with the value of the merged profile-a

In 2. the profile-b would be updated with the values of unmerged profile-a

so it becomes order dependent.

Right now everything in the clarity of reading the code relies on this destructive_merge False flag.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
gmuloc: Indeed it is documented like this - and I don't think there is something to do in this PR but this looks not very robust as it means the order in the list ends up being important though maybe I am not familiar enough with a typical configuration for link_tracking.

Resolving this for now

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: good point. Renaming.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/monitor_sessions.py
ClausHolbechArista: done

gmuloc (approved):
LGTM

PR #2347, opened 2022-12-20
Feat(eos_designs): Add per MACVRF EVPN domain scope

mitchv85:
## Change Summary

Add an optional `evpn_l2_multi_domain` boolean parameter to enable the extension of VLANs into remote EVPN domains

TOI: https://www.arista.com/en/support/toi/eos-4-26-1f/14785-multi-domain-evpn-vxlan

The `evpn_l2_multi_domain` parameter can be defined at the following levels:

- Tenant
- VRF
- SVI/L2VLAN

It will default to a value of `True`, but this can be overridden as needed at the different levels defined above.

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Extend the Network Services data model, providing an optional `evpn_l2_multi_domain` boolean parameter at the following levels:

- Tenant
- VRF
- SVI/L2VLAN

```yaml
---

tenants:
  <TENANT NAME>:
    evpn_l2_multi_domain: < true | false | default -> true >
    vrfs:
      <VRF NAME>:
        evpn_l2_multi_domain: < true | false >
        svis:
          <VLAN ID>:
            evpn_l2_multi_domain: < true | false >

    l2vlans:
      <VLAN ID>:
            evpn_l2_multi_domain: < true | false >
```

Generated output if `evpn_l2_multi_domain: false`:

```
router bgp 100
   !
   vlan 10
      rd 10.231.0.2:10010
      route-target both 10010:10010
      redistribute learned
```

Generated output if `evpn_l2_multi_domain: true`:

```
router bgp 100
   !
   vlan 10
      rd 10.231.0.2:10010
      rd evpn domain remote 10.231.0.2:10010
      route-target both 10010:10010
      route-target import export evpn domain remote 10010:10010
      redistribute learned
```

## How to test

`molecule converge -s eos_designs_unit_tests`

### User Checklist

N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -291,8 +291,10 @@ def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:
             "struct_cfg": get(vlan, "bgp.structured_config"),
         }
         if self._evpn_gateway_vxlan_l2:
-            bgp_vlan["rd_evpn_domain"] = {"domain": "remote", "rd": vlan_rd}
-            bgp_vlan["route_targets"]["import_export_evpn_domains"] = [{"domain": "remote", "route_target": vlan_rt}]
+            vlan_evpn_domain_scope = get(vlan, "evpn_domain_scope")
```
ClausHolbechArista: please use `vlan.get("evpn_domain_scope")`. `get()` is only required for deeper lookups, or if we have a required value and need to produce an error. You might also want to add the default value here `vlan.get("evpn_domain_scope", "remote")`

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -291,8 +291,10 @@ def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:
...
+            if vlan_evpn_domain_scope == "global":
```
ClausHolbechArista: This could be collapsed to the if in line 293 like `if self._evpn_gateway_vxlan_l2 and vlan.get("evpn_domain_scope", "remote") == "remote"`

ClausHolbechArista (commented):
We need to add
- Documentation (update the network-services.md file)
- Molecule test case

Maybe we also need to check if this data model is what we want. Maybe a single boolean "evpn_export_to_remote_domain" (or something better).

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
mitchv85: Thanks for the tip! It has been changed to the vlan.get() model.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
mitchv85: Done!

mitchv85:
> We need to add
> 
> * Documentation (update the network-services.md file)
> * Molecule test case
> 
> Maybe we also need to check if this data model is what we want. Maybe a single boolean "evpn_export_to_remote_domain" (or something better).

I like this idea. I've changed the data model to a boolean parameter called `evpn_multi_domain`. It's optional, and will default to `false`, which is a change from the existing default behavior of making all MACVRFs mutli-domain when a node is an EVPN/VXLAN L2 GW.

ClausHolbechArista:
@carlbuchmann please review the documentation on this when you are back.

ClausHolbechArista (approved):

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -576,6 +587,11 @@ mac_address_table:
             # Nodes is optional. Default is all nodes where the VRF is defined.
             nodes: [ < node_1 >, < node_2> ]
 
+        # EVPN Multi-Domain | Optional
+        # Explicitly extend all VLAN-aware-bundles to remote EVPN domains
+        # to override setting of tenants.<tenant>.evpn_multi_domain (default -> true)
+        evpn_multi_domain: < true | false >
```
gmuloc:
```suggestion
        # to override setting of tenants.<tenant>.evpn_multi_domain
        evpn_multi_domain: < true | false | default -> true >
```

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -633,6 +649,11 @@ mac_address_table:
         # Extend this L2VLAN over VXLAN
         vxlan: < true | false | default -> true >
 
+        # EVPN Multi-Domain | Optional
+        # Extend this VLAN to remote EVPN domains
+        # Explicitly extend this VLAN to remote EVPN domains to override setting of tenants.<tenant>.evpn_multi_domain (default -> true)
+        evpn_multi_domain: < true | false >
```
gmuloc:
```suggestion
        # Explicitly extend this VLAN to remote EVPN domains to override setting of tenants.<tenant>.evpn_multi_domain
        evpn_multi_domain: < true | false | default -> true >
```

gmuloc (commented):
couple of quick comments

Question on the testing - it looks like the feature is only tested on VRF and not on VLANs - is it correct? Would it be possible to add a test for Vlan as well to prevent regression there in the future?

mitchv85:
> couple of quick comments
> 
> Question on the testing - it looks like the feature is only tested on VRF and not on VLANs - is it correct? Would it be possible to add a test for Vlan as well to prevent regression there in the future?

Are you referring to setting `evpn_multi_domain` to `False` for a VLAN in the inventory files used for the molecule test (`molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/`)? If so, we can certainly make this adjustment. Let me know!

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -373,6 +378,12 @@ mac_address_table:
             # Extend this SVI over VXLAN
             vxlan: < true | false | default -> true >
 
+            # EVPN Multi-Domain | Optional
+            # Extend this VLAN to remote EVPN domains
+            # Explicitly extend this VLAN to remote EVPN domains to override setting of tenants.<tenant>.evpn_multi_domain (default -> true)
+            # and tenants.<tenant>.vrfs.<vrf>.evpn_multi_domain
+            evpn_multi_domain: < true | false >
```
gmuloc:
```suggestion
            # Explicitly extend this VLAN to remote EVPN domains to override setting of tenants.<tenant>.evpn_multi_domain
            # and tenants.<tenant>.vrfs.<vrf>.evpn_multi_domain
            evpn_multi_domain: < true | false | default -> true >
```

gmuloc:
> > couple of quick comments
> > Question on the testing - it looks like the feature is only tested on VRF and not on VLANs - is it correct? Would it be possible to add a test for Vlan as well to prevent regression there in the future?
> 
> Are you referring to setting `evpn_multi_domain` to `False` for a VLAN in the inventory files used for the molecule test (`molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/`)? If so, we can certainly make this adjustment. Let me know!


Yeah that's what I mean adding a SVI test

for now I am not sure what is the expected config for a SVI if you have the new setting as follow:

```
global: True
vrf: False
svi: None
```

Looking at
`def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:` new line it seems that the VRF value would be ignored and the value for the vlans would end up being `False`

https://github.com/aristanetworks/ansible-avd/pull/2347/files#diff-91e209c523bb520a8c29f348c4bf3734e3c9e0323090b85f1549f6ec28b898d4R293

```
if self._evpn_gateway_vxlan_l2 and default(vlan.get("evpn_multi_domain"), tenant.get("evpn_multi_domain", True)) is True:
```

however the documentation states that it should override tenant AND vrf settings
```
# Explicitly extend this VLAN to remote EVPN domains to override setting of tenants.<tenant>.evpn_multi_domain (default -> true)
# and tenants.<tenant>.vrfs.<vrf>.evpn_multi_domain
```

Should it be something like: (though the `vrf` is not available in the function today)

```
if self._evpn_gateway_vxlan_l2 and default(vlan.get("evpn_multi_domain"), vrf.get("evpn_multi_domain"), tenant.get("evpn_multi_domain", True)) is True:
```

mitchv85:
> 

With the current changes...

- VLAN-Aware Bundle MACVRF model, the Tenant/VRF is where `evpn_multi_domain` can be set. Setting this value under the SVI will not have an impact on a VLAN that's a part of a VLAN-Aware Bundle.

- For VLAN-Based, setting the `evpn_multi_domain` parameter under the SVI/VLAN works as intended. This was tested by setting `evpn_vlan_aware_bundles: false` in `molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml`.

- Setting `evpn_multi_domain` parameter under the VRF in a VLAN-Based model will not have an effect since (as you pointed out) VRF is not available in the `_router_bgp_vlans_vlan` function today.

It doesn't appear that anything in DC1 leverages VLAN-Based MACVRFs for molecule testing today...Do we need a new `DC2` test?

It sounds like we're at a fork in the road here...perhaps adding `vrf` to the `_router_bgp_vlans_vlan` function is the right approach here so that there isn't any confusion related to the `evpn_multi_domain` parameter under `vrf` being ignored for the VLAN-Based MACVRF model. I'm not sure what the ripple effect is of this change, but I will give it a shot this evening!

Thanks for the feedback thus far!

mitchv85:
@gmuloc and @ClausHolbechArista 

I've added `vrf` to the `_router_bgp_vlans_vlan` function in `eos_designs/python_modules/network_services/router_bgp.py`

setting `evpn_multi_domain` parameter under `vrf` is now valid for both the VLAN-Based MACVRF model and VLAN-Aware Bundle MACVRF models.

The molecule test was adjusted to enable VLAN-Based MACVRF testing.

- `evpn_vlan_aware_bundles: false` was set in `molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml`
- `evpn_multi_domain` was set to `false` at the `tenant` level in `molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/Tenant_A.yml`
- `evpn_multi_domain` was set to `false` at the `vrf` level in `molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/Tenant_B.yml`
- `evpn_multi_domain` was set to `false` at the `vrf` level and `true` at the `svi` level in `molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/Tenant_C.yml`

Net result is that only VLAN 350 in Tenant C has a MACVRF with remote domain RD/RT values. This is the expected behavior.

Let me know what I missed! Once this is good, will update the documentation...

Thanks again!

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -290,7 +290,7 @@ def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:
             "eos_cli": get(vlan, "bgp.raw_eos_cli"),
             "struct_cfg": get(vlan, "bgp.structured_config"),
         }
-        if self._evpn_gateway_vxlan_l2:
+        if self._evpn_gateway_vxlan_l2 and default(vlan.get("evpn_multi_domain"), tenant.get("evpn_multi_domain"), vrf.get("evpn_multi_domain"), True) is True:
```
gmuloc: I would change the order to `vlan -> vrf -> tenant` rather than `vlan -> tenant -> vrf`?

ClausHolbechArista: +1

mitchv85:
Closed in error - re-opening PR...

mitchv85:
Converting this PR to draft

While the molecule tests were passing after adding `vrf` to the `_router_bgp_vlans_vlan` function in `eos_designs/python_modules/network_services/router_bgp.py`

A subsequent test using a customer build encountered an error `UnboundLocalError: local variable 'vrf' referenced before assignment` during the `Generate device configuration in structured format` step.

Need to do some more work before marking this PR as ready for review again...

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -245,13 +245,13 @@ def _router_bgp_vlans(self) -> dict | None:
...
                 for svi in vrf["svis"]:
-                    if (vlan := self._router_bgp_vlans_vlan(svi, tenant)) is not None:
+                    if (vlan := self._router_bgp_vlans_vlan(svi, tenant, vrf)) is not None:
                         vlan_id = int(svi["id"])
                         vlans[vlan_id] = vlan
 
             # L2 Vlans per Tenant
             for l2vlan in tenant["l2vlans"]:
-                if (vlan := self._router_bgp_vlans_vlan(l2vlan, tenant)) is not None:
+                if (vlan := self._router_bgp_vlans_vlan(l2vlan, tenant, vrf)) is not None:
```
tgodaA: Why the l2vlan is added under a VRF?

tgodaA: l2vlan is tied to a tenant but not a VRF.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -260,7 +260,7 @@ def _router_bgp_vlans(self) -> dict | None:
 
         return None
 
-    def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:
+    def _router_bgp_vlans_vlan(self, vlan, tenant, vrf) -> dict | None:
```
tgodaA, line 263:
```suggestion
    def _router_bgp_vlans_vlan(self, vlan, tenant, vrf=None) -> dict | None:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -260,7 +260,7 @@ def _router_bgp_vlans(self) -> dict | None:
...
         """
         Return structured config for one given vlan under router_bgp.vlans
         """
```
tgodaA, line 266:
```suggestion
        """
        if vrf is None:
            vrf = {}
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
tgodaA:
```suggestion
                if (vlan := self._router_bgp_vlans_vlan(l2vlan, tenant)) is not None:
```

mitchv85: Thanks again - this did the trick!

mitchv85:
Ready for review again. Thanks again for all the help @gmuloc @ClausHolbechArista @tgodaA !

ClausHolbechArista:
Since all this only affect the L2 gateway and vlan services, we should probably include `l2` in the key. We also have L3 gateway functionality, where you may want to limit VRFs/tenants independently of L2.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -6,7 +6,7 @@
 shutdown_interfaces_towards_undeployed_peers: true
 
 # Enable vlan aware bundles
-evpn_vlan_aware_bundles: true
+evpn_vlan_aware_bundles: false
```
ClausHolbechArista: please don't change existing molecule tests. This will change the coverage of all the existing tests based on the same file. I suggest you create a new separate test set similar to EVPN_MULTICAST_TESTS (only configure the stuff you need and with the least amount of devices required). Two L2 EVPN gateways, one with vlan-aware-bundle and one without should be enough in this case.

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -271,6 +274,10 @@ mac_address_table:
         # Non-selectively enabling or disabling redistribute ospf inside the VRF | Optional.
         redistribute_ospf: < true | false, Default -> true >
 
+        # Explicitly extend all VLANs/VLAN-Aware Bundles inside the VRF to remote EVPN domains.
+        # Overrides tenants.<tenant>.evpn_multi_domain ( default --> true ).
+        evpn_multi_domain: < true | false | default --> true >
```
ClausHolbechArista: This is not default true. This inherits from tenant, so it has no default

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -373,6 +380,11 @@ mac_address_table:
             # Extend this SVI over VXLAN
             vxlan: < true | false | default -> true >
 
+            # Explicitly extend this VLAN to remote EVPN domains.
+            # Overrides tenants.<tenant>.evpn_multi_domain ( default --> true ).
+            # Overrides tenants.<tenant>.vrf.<vrf>.evpn_multi_domain ( default --> true ).
```
ClausHolbechArista: This is not default true. This inherits from tenant, so it has no default

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -373,6 +380,11 @@ mac_address_table:
...
+            evpn_multi_domain: < true | false | default --> true >
```
ClausHolbechArista: This is not default true. This inherits from vrf, so it has no default

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
```diff
@@ -651,6 +663,10 @@ mac_address_table:
           source_address: < ipv4_address -> default ip address of Loopback0 >
           version: < 1, 2, 3 -> default 2 (EOS) >
 
+        # Explicitly extend this VLAN to remote EVPN domains.
+        # Overrides tenants.<tenant>.evpn_multi_domain ( default --> true ).
+        evpn_multi_domain: < true | false | default --> true >
```
ClausHolbechArista: This is not default true. This inherits from tenant, so it has no default

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -260,10 +260,12 @@ def _router_bgp_vlans(self) -> dict | None:
 
         return None
 
-    def _router_bgp_vlans_vlan(self, vlan, tenant) -> dict | None:
+    def _router_bgp_vlans_vlan(self, vlan, tenant, vrf=None) -> dict | None:
         """
         Return structured config for one given vlan under router_bgp.vlans
         """
+        if vrf is None:
```
ClausHolbechArista: instead of having None as default value and then setting {}, just add {} as the `vrf` paremeter when you configure l2vlans.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
mitchv85: Will do - I have changed this back to it's original setting of `true`. Thanks!

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
mitchv85: Thanks - resolved!

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
mitchv85: Resolved. Thanks!

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
mitchv85: Resolved

ansible_collections/arista/avd/roles/eos_designs/doc/network-services.md
mitchv85: Resolved!

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
mitchv85: Recommended changes have been made. Thanks!

mitchv85:
> Since all this only affect the L2 gateway and vlan services, we should probably include `l2` in the key. We also have L3 gateway functionality, where you may want to limit VRFs/tenants independently of L2.

I like this suggestion. Changed `evpn_multi_domain` parameter to `evpn_l2_multi_domain`

ClausHolbechArista (approved):

tgodaA (approved):
LGTM

PR #2349, opened 2022-12-22
Refactor: Input data conversion to support conversion messages

ClausHolbechArista:
## Change Summary

Refactor Input data conversion to support conversion messages

## Component(s) name

`arista.avd.validate_and_template`

## Proposed changes

- Move validation and conversion code from action plugin to common lib
- Refactor schema conversion to class
- Return conversion messages for every variable conversion performed

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
@@ -4,6 +4,7 @@ from typing import Generator
 
 from ansible_collections.arista.avd.plugins.filter.convert_dicts import convert_dicts
 from ansible_collections.arista.avd.plugins.plugin_utils.errors import AristaAvdError, AvdConversionWarning
+from ansible_collections.arista.avd.plugins.plugin_utils.schema.avdschema import AvdSchema
 from ansible_collections.arista.avd.plugins.plugin_utils.utils import get_all
 
 SCHEMA_TO_PY_TYPE_MAP = {
@@ -26,7 +27,7 @@ class AvdDataConverter:
     AvdDataConverter is used to convert AVD Data Types based on schema options.
     """
 
-    def __init__(self, avdschema):
+    def __init__(self, avdschema: AvdSchema):
         self._avdschema = avdschema
 
         self.converters = {
@@ -34,7 +35,11 @@ class AvdDataConverter:
             "keys": self.convert_keys,
         }
 
-    def convert_data(self, data, schema: dict = None, path: list = None) -> Generator:
+    def convert_data(self, data, schema: dict = None, path: list[str] = None) -> Generator:
+        """
+        Perform in-place conversion of data according to the provided schema.
+        Main entry function which is recursively called from the child functions performing the actual conversion of keys/items.
+        """
         if schema is None:
             # Get fully resolved schema (where all $ref has been expanded recursively)
             # Performs inplace update of the argument so we give an empty dict.
@@ -43,6 +48,7 @@ class AvdDataConverter:
             resolve_errors = self._avdschema.resolve(schema)
             for resolve_error in resolve_errors:
                 if isinstance(resolve_error, Exception):
+                    # TODO: Raise/yield multiple errors
                     raise AristaAvdError(resolve_error)
 
         if path is None:
@@ -56,7 +62,7 @@ class AvdDataConverter:
             # Converters will do inplace update of data. Any returns will be yielded conversion messages.
             yield from converter(schema[key], data, schema, path)
 
-    def convert_keys(self, keys: dict, data: dict, schema: dict, path: list):
+    def convert_keys(self, keys: dict, data: dict, schema: dict, path: list[str]):
         """
         This function performs conversion on each key with the relevant subschema
         """
@@ -96,7 +102,7 @@ class AvdDataConverter:
             # Dive in to child keys/schemas
             yield from self.convert_data(data[key], childschema, path + [key])
 
-    def convert_items(self, items: dict, data: list, schema: dict, path: list):
+    def convert_items(self, items: dict, data: list, schema: dict, path: list[str]):
         """
         This function performs conversion on each item with the items subschema
         """
@@ -111,7 +117,7 @@ class AvdDataConverter:
             # Dive in to child items/schema
             yield from self.convert_data(item, items, path + [f"[{index}]"])
 
-    def convert_types(self, convert_types: list, data: dict | list, index: str | int, schema: dict, path: list):
+    def convert_types(self, convert_types: list, data: dict | list, index: str | int, schema: dict, path: list[str]):
         """
         This function performs type conversion if necessary on a single data instance.
         It is invoked for child keys during "keys" conversion and for child items during
@@ -145,6 +151,7 @@ class AvdDataConverter:
                         data[index] = SIMPLE_CONVERTERS[schema_type](value)
                     except Exception:
                         # Ignore errors
+                        # TODO: Log message
                         return
 
                 elif convert_type in ["dict", "list"] and schema_type == "list" and "primary_key" in schema:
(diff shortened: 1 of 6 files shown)


ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
+    def convert_data(self, data, schema: dict = None, path: list = None) -> Generator:
+        if schema is None:
+            # Get fully resolved schema (where all $ref has been expanded recursively)
+            # Performs inplace update of the argument so we give an empty dict.
+            # By default it will resolve the full schema
+            schema = {}
+            resolve_errors = self._avdschema.resolve(schema)
+            for resolve_error in resolve_errors:
+                if isinstance(resolve_error, Exception):
+                    raise AristaAvdError(resolve_error)
```
gmuloc: So here if there are multiple exceptions, only the first one will be raised - is it something we would like to change (even in a future PR and in that case a TODO is enough here)

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
-    )
+    def __init__(self, avdschema):
+        self._avdschema = avdschema
+
+        self.converters = {
+            "items": self.convert_items,
+            "keys": self.convert_keys,
+        }
+
+    def convert_data(self, data, schema: dict = None, path: list = None) -> Generator:
```
gmuloc: Add a docstring

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
+
+        for key, converter in self.converters.items():
+            if key not in schema:
+                # Ignore keys not in schema
+                continue
+
+            # Converters will do inplace update of data. Any returns will be yielded conversion messages.
+            yield from converter(schema[key], data, schema, path)
+
+    def convert_keys(self, keys: dict, data: dict, schema: dict, path: list):
```
gmuloc: Probably worth saying what is the format for `path` - a list of strings that give a path in the schema if I get it right?


Missing return type

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
+        for key, childschema in keys.items():
+            if key not in data:
+                # Skip key since there is nothing to convert if the key is not set in data
+                continue
+
+            # Perform type conversion of the data for the child key if required based on "convert_types"
+            if "convert_types" in childschema:
+                yield from self.convert_types(childschema["convert_types"], data, key, childschema, path + [key])
+
+            yield from self.convert_data(data[key], childschema, path + [key])
```
gmuloc, line 85: Exact same code block as from line 87 to 97 - making let's use a helper function

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
+        # Prepare string for var path used in warning messages.
+        path_str = ".".join(path)
+
+        for convert_type in convert_types:
+            if isinstance(value, SCHEMA_TO_PY_TYPE_MAP.get(convert_type)):
+                if schema_type in SIMPLE_CONVERTERS:
+                    try:
+                        data[index] = SIMPLE_CONVERTERS[schema_type](value)
+                    except Exception:
+                        # Ignore errors
```
gmuloc, line 155: probably worth adding a log message at debug level if this happens

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -1,238 +1,166 @@
...
-                "str": jsonschema._types.is_string,
-                "bool": jsonschema._types.is_bool,
-                "list": jsonschema._types.is_array,
-                "int": jsonschema._types.is_integer,
-                "float": _is_float,
-            }
-        ),
-    )
+    def __init__(self, avdschema):
+        self._avdschema = avdschema
```
gmuloc, line 33: It looks like in this case we expect an AvdSchema object when in other similar classes (validator and schemaresolver) we only need a dict. Lets add the type hint here.

I am not quite sure why we need it to be different as it seems the only place we need this kind of object is if the schema is not given to convert_data we resolve the schema. Would it not be enough to instantiate the class with an already resolved schema?

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: Added TODOs since this code will be refactored (again) in the deprecation PR. We could add another error type we can yield for this case.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: There is no return type.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: Will be done in later PR - the one with deprecation messages.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: added TODO since some of this will change in the later PR.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: Added type hint. If we instatiate with resolved schema it just means we need to have more tooling in all the calling functions. We have the same as this in the documentation and schema converts.

ClausHolbechArista: Also this would allow for future constants loaded from the schema instance like `ignore keys`

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: done

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
ClausHolbechArista: Removed type hint again, since this class is imported into AvdSchema and we cannot have a circular import

ClausHolbechArista: I will add a TODO to refactor this to a resolved schema instead. We have several pending PRs in this area, and I think it needs to wait until all those have merged.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
gmuloc: technically there is always a return type but ok - https://stackoverflow.com/questions/42227288/pythons-pep-484-type-annotation-for-generator-expression

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
gmuloc: resolving but for the future you can have circular import for type hints: https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/

gmuloc (approved):
LGTM

PR #2351, opened 2022-12-22
Fix(eos_designs): Remove j2caching since it is unstable between versions

ClausHolbechArista:
## Change Summary

Fix(eos_designs): Remove j2caching since it is unstable and does not invalidate between different versions with incompatible bytecode.

## Related Issue(s)

Upgrading and downgrading ansible, python and various libraries may result in cached bytecode which is incompatible with other versions. 

When you later upgrade the environment you get weird errors from Jinja, like `sequence item 6: expected str instance, int found` or similar. 
Deleting the j2cache files solves this.

Until j2caching becomes more stable, we will have to avoid it.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Remove j2caching from template utils.

## How to test

No changes to output. But everything runs slower since all templates will be converted to bytecode for every device on every run.

### User Checklist

- N/A

### Repository Checklist


gmuloc (approved):
LGTM

carlbuchmann (approved):
LGTM!

PR #2391, opened 2023-01-10
Fix(eos_cli_config_gen): Add convert_types to router ospf area id schema

danieltudares:
## Change Summary

The `router_ospf > areas` expects the area to be input as a string (i.e. "0.0.0.2"), and if an user inputs the area number as an integer, it gets the following error message:

```
[WARNING]: b"[leaf1]: 'Validation Error: router_ospf.process_ids[0].areas[0].id': 2 is not of type 'str'"  
```

The EOS CLI will let users input the ospf area in both format, so to avoid this error and allow users to input the area as an integer or string, we need to check and convert from `int` to `str` if that's the case. 

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.eso_cli_config_gen`

## Proposed changes

This is the current `router_ospf` area schema: 

```yml
            areas:
              type: list
              primary_key: id
              convert_types:
                - dict
              items:
                type: dict
                keys:
                  id:
                    type: str
```
Proposed new schema: 

```yml
            areas:
              type: list
              primary_key: id
              convert_types:
                - dict
              items:
                type: dict
                keys:
                  id:
                    type: str
                    convert_types:
                      - int
```

## How to test

`molecule converge -s eos_cli_config_gen`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-ospf.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-ospf.md
@@ -244,7 +244,7 @@ interface Vlan24
 | Process ID | Area | Area Type | Filter Networks | Filter Prefix List | Additional Options |
 | ---------- | ---- | --------- | --------------- | ------------------ | ------------------ |
 | 200 | 0.0.0.2 | normal | 1.1.1.0/24, 2.2.2.0/24 | - |  |
-| 200 | 0.0.0.3 | normal | - | PL-OSPF-FILTERING |  |
+| 200 | 3 | normal | - | PL-OSPF-FILTERING |  |
 | 600 | 0.0.0.1 | normal | - | - |  |
 | 600 | 0.0.10.11 | stub | - | - | no-summary |
 | 600 | 0.0.20.20 | nssa | - | - |  |
@@ -309,7 +309,7 @@ router ospf 200 vrf ospf_zone
    router-id 192.168.254.1
    area 0.0.0.2 filter 1.1.1.0/24
    area 0.0.0.2 filter 2.2.2.0/24
-   area 0.0.0.3 filter prefix-list PL-OSPF-FILTERING
+   area 3 filter prefix-list PL-OSPF-FILTERING
    max-lsa 5
    timers lsa rx min interval 100
    default-information originate always
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-ospf.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-ospf.cfg
@@ -79,7 +79,7 @@ router ospf 200 vrf ospf_zone
    router-id 192.168.254.1
    area 0.0.0.2 filter 1.1.1.0/24
    area 0.0.0.2 filter 2.2.2.0/24
-   area 0.0.0.3 filter prefix-list PL-OSPF-FILTERING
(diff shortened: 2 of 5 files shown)


gmuloc (approved):
LGTM for the functionality though going through testing (vEOS - Software image version: 4.27.3F) the 3 is being converted to dotted notation
```
Leaf4(config-router-ospf)#   area 0.0.0.2 filter 1.1.1.0/24
Leaf4(config-router-ospf)#   area 0.0.0.2 filter 2.2.2.0/24
Leaf4(config-router-ospf)#   area 3 filter prefix-list PL-OSPF-FILTERING
Leaf4(config-router-ospf)#show active
router ospf 666
   area 0.0.0.2 filter 1.1.1.0/24
   area 0.0.0.2 filter 2.2.2.0/24
   area 0.0.0.3 filter prefix-list PL-OSPF-FILTERING
```

I mean this is not invalid configuration per se as it is supported by EOS as per documentation (https://www.arista.com/en/um-eos/eos-open-shortest-path-first-version-2#xx1152625):

> area_id Area number: 0 to 4294967295 or 0.0.0.0 to 255.255.255.255 running-config stores value in dotted decimal notation.

But what is rendered is not what will be showing on the running-config (though probably the user would know that already)

tgodaA (approved):
LGTM

ClausHolbechArista:
> LGTM for the functionality though going through testing (vEOS - Software image version: 4.27.3F) the 3 is being converted to dotted notation
> 
> ```
> Leaf4(config-router-ospf)#   area 0.0.0.2 filter 1.1.1.0/24
> Leaf4(config-router-ospf)#   area 0.0.0.2 filter 2.2.2.0/24
> Leaf4(config-router-ospf)#   area 3 filter prefix-list PL-OSPF-FILTERING
> Leaf4(config-router-ospf)#show active
> router ospf 666
>    area 0.0.0.2 filter 1.1.1.0/24
>    area 0.0.0.2 filter 2.2.2.0/24
>    area 0.0.0.3 filter prefix-list PL-OSPF-FILTERING
> ```
> 
> I mean this is not invalid configuration per se as it is supported by EOS as per documentation (https://www.arista.com/en/um-eos/eos-open-shortest-path-first-version-2#xx1152625):
> 
> > area_id Area number: 0 to 4294967295 or 0.0.0.0 to 255.255.255.255 running-config stores value in dotted decimal notation.
> 
> But what is rendered is not what will be showing on the running-config (though probably the user would know that already)

That behavior is EOS, and not something that has changed with this PR.

tgodaA:
Should we do something like, convert_value??

gmuloc:
I don't think we should convert the value as integer is technically a valid input value:

```
Leaf4(config-router-ospf)#area ?
  A.B.C.D         OSPF area-id in IP address format
  <0-4294967295>  OSPF area-id in decimal format
```

Was just pointing out that the transformation occurs on EOS.

ClausHolbechArista:
> I don't think we should convert the value as integer is technically a valid input value:
> 
> ```
> Leaf4(config-router-ospf)#area ?
>   A.B.C.D         OSPF area-id in IP address format
>   <0-4294967295>  OSPF area-id in decimal format
> ```
> 
> Was just pointing out that the transformation occurs on EOS.

Technically all inputs to EOS are strings. It is CLI after all.
We can not (and should not) have vars with multiple types, so in this case string covers both. For ease of use, we have decided to translate between types every time we have an `int` value _or_ if we have `str` with a potential nummeric value. This helps with some of the type ambiguity that comes with YAML inputs.

PR #2437, opened 2023-01-18
Refactor(plugins): Improvements of schema tools preparing for eos_designs schemas

ClausHolbechArista:
## Change Summary

Improvements of schema tools preparing for eos_designs schemas

## Proposed changes

- Add support for convert_types `str` -> `list` splitting on `,`
- Add full support for $ref including in combination with `dynamic_keys`

## How to test

Tested as part of PR for eos_designs_schema. This code is just extracted from that PR, to make it cleaner.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -394,6 +394,13 @@ def _rt_admin_subfield(self) -> str | None:
         if admin_subfield == "bgp_as":
             return self._bgp_as
 
+        try:
+            # Try to convert input var (str) to int
+            admin_subfield = int(admin_subfield)
+        except ValueError:
+            # Ignore if we could not convert
+            pass
+
```
gmuloc: should we add a TODO to do this in the schema later on?

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschema.py
```diff
@@ -204,11 +204,27 @@ def subschema(self, datapath: list, schema: dict = None):
...
         if schema["type"] == "list" and key in schema.get("items", {}).get("keys", []):
             return self.subschema(datapath[1:], schema["items"]["keys"][key])
 
         # Falling through here in case the schema is not covering the requested datapath
         raise AvdSchemaError(f"The datapath '{datapath}' could not be found in the schema")
+
+    def simple_resolve(self, schema):
+        # Get fully resolved schema (where all $ref has been expanded recursively)
+        # Performs inplace update of the argument so we give an empty dict.
+        # By default it will resolve the full schema
```
gmuloc:
```suggestion
    def simple_resolve(self, schema: dict):
        """
        Get fully resolved schema (where all $ref has been expanded recursively)
        Performs inplace update of the argument so we give an empty dict.
        By default it will resolve the full schema
        """
```

Maybe change the name to `get_resolved_schema`

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschema.py
```diff
@@ -204,11 +204,27 @@ def subschema(self, datapath: list, schema: dict = None):
...
+        resolved_schema = {}
+        resolve_errors = self.resolve(resolved_schema, schema)
+        for resolve_error in resolve_errors:
+            if isinstance(resolve_error, Exception):
```
gmuloc, line 228: add a TODO for future revisit here as we would raise only on the first error

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
ClausHolbechArista: This _is_ actually to prepare for it to be done in the schema. The schema will convert an input `int` to `str`, so here we are trying to convert back to detect pure numbers.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschema.py
ClausHolbechArista: done

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschema.py
ClausHolbechArista: done

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
gmuloc: so the schema will say `str` with `convert_type` set to `int` and here we will convert back to `int` and then add some validation that is not part of the schema but hidden in the code.
That's a bit unusual - I guess it is to support the special case `"bgp_as"` which would be the only valid string. Would it not be best to change the model in that ca?

ClausHolbechArista: An IP address is also a string. So most cases for RD and RT are actually strings.

ClausHolbechArista: I can replace all this with a regex check for ^[0-9]+$ to avoid converting back and forth.

ClausHolbechArista: simplified with regex match both here and in facts

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -394,8 +395,8 @@ def _rt_admin_subfield(self) -> str | None:
         if admin_subfield == "bgp_as":
             return self._bgp_as
 
-        if isinstance(admin_subfield, int) and admin_subfield >= 0 and admin_subfield <= 4294967295:
-            return str(admin_subfield)
+        if re_fullmatch(r"[0-9]+", str(admin_subfield)):
+            return admin_subfield
```
tgodaA, line 399: This won't match the condition: admin_subfield <= 4294967295. Are we removing that check or moving that task to schemas?

ClausHolbechArista: correct, we will relax that test.

tgodaA (approved):
LGTM

PR #2442, opened 2023-01-19
Fix(eos_designs): vtep_vvtep_ip doesn't generate any config

### LINKED ISSUE #2441: eos_designs: vtep_vvtep_ip doesn't work
### Issue Summary

The fabric variable vtep_vvtep_ip doesn't generate any config in AVD 3.8. This is because of missing logic in eos_designs_facts.py, switch.vtep_vvtep_ip is never set but it is referenced by the underlay/utils.py module. There is also an untriggered typo in underlay/prefix_lists.py which needs fixing.

At the moment this fails silently and doesn't generate the proper secondary vtep ip on l3 vteps.

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
add a vtep_vvtep_ip: <ip/mask> to your fabric variables and see how no secondary ip for lo1 is generated for l3 vteps.
```

### Relevant log output

_No response_

### Contributing Guide

emilarista:
## Change Summary

Fix issue with vtep_vvtep_ip not generating config.

## Related Issue(s)

Fixes #2441 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Implement switch.vtep_vvtep_ip property in eos_desings_facts.py module and fix typo in underlay/prefix_lists.py module.

## How to test
Add vtep_vvtep_ip to any inventory running AVD 3.8 and you will see the secondary IP on lo1 is not generated for l3 vteps. 

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
@@ -582,13 +582,6 @@ class EosDesignsFacts(AvdFacts):
         return {}
 
     @cached_property
-    def vtep_vvtep_ip(self):
-        """
-        switch.vtep_vvtep_ip set based on Global var vtep_vvtep_ip
-        """
-        return get(self._hostvars, "vtep_vvtep_ip")
-
-    @cached_property
     def mgmt_interface(self):
         """
         mgmt_interface is inherited from
--- a/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
@@ -582,6 +582,13 @@ class EosDesignsFacts(AvdFacts):
         return {}
 
     @cached_property
+    def vtep_vvtep_ip(self):
+        """
+        switch.vtep_vvtep_ip set based on Global var vtep_vvtep_ip
+        """
+        return get(self._hostvars, "vtep_vvtep_ip", default=None)
+
+    @cached_property
     def mgmt_interface(self):
         """
         mgmt_interface is inherited from
--- a/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
@@ -586,7 +586,7 @@ class EosDesignsFacts(AvdFacts):
         """
         switch.vtep_vvtep_ip set based on Global var vtep_vvtep_ip
         """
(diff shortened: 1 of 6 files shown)


ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
```diff
@@ -581,6 +581,13 @@ def platform_settings(self):
 
         return {}
 
+    @cached_property
```
gmuloc: Try maybe to add this closer to the `vtep_ip` one so they are kind of close by.
Wondering at the same time if it would be overkill to set it in the `avdipaddressing` module rather than here. Thoughts @ClausHolbechArista ?

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
```diff
@@ -581,6 +581,13 @@ def platform_settings(self):
...
+    def vtep_vvtep_ip(self):
+        """
+        switch.vtep_vvtep_ip set based on Global var vtep_vvtep_ip
+        """
+        return get(self._hostvars, "vtep_vvtep_ip", default=None)
```
gmuloc: None is already the default value returned by get so no need to add it here
```suggestion
        return get(self._hostvars, "vtep_vvtep_ip")
```

gmuloc (commented):
Thanks for catching this! One comment about the property - I think probably it can stay in here for now as it is a direct vars -> fact lookup but would prefer to get some other opinion

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
ClausHolbechArista: there is no algorithm here, just 1:1 fetching of the input variable. We don't even need it under switch and in facts. The plugins reading this value could just read from hostvars directly.

gmuloc: so @emilarista what @ClausHolbechArista is suggesting is to actually have this line:

```
return get(self._hostvars, "vtep_vvtep_ip")
```

in `ansible_collections/arista/avd/roles/eos_designs/python_modules/underlay/utils.py` and to avoid making it a fact at all.

carlbuchmann: resolved in 9859d0df0b4de2417633512031101fae198e89e8

carlbuchmann (approved):
LGTM!

ClausHolbechArista (approved):

PR #2450, opened 2023-01-20
Feat(eos_cli_config_gen): Extend aaa_accounting with options for dot1x

### LINKED ISSUE #2343: Feat(eos_cli_config_gen): extend aaa_accounting with options for dot1x
### Enhancement summary

aaa_accounting global configuration will be extended with missing options for dot1x accounting

eos_cli_config_gen

### Use case example

```
aaa accounting dot1x default start-stop group < group-name > logging
```

```
aaa_accounting:
  dot1x:
    default:
      type: < none | start-stop | stop-only >
      group: < group_name >
      logging: < true | false >
```

This can be accomplished today with cvp configlet or eos_cli:.

### Additional context

I would like to provide the implementation on this.

### Contributing Guide

dmilyt:
## Change Summary

Add aaa_accounting with options for dot1x

## Related Issue(s)

Fixes #2343

## Component(s) name

## Proposed changes

```
aaa_accounting:
  dot1x:
    default:
      type: start-stop
      group: ISE_NAC
```

## How to test

Molecule

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_cli_config_gen/README_v4.0.md
```diff
@@ -393,6 +393,10 @@ aaa_accounting:
     default:
       type: < none | start-stop | stop-only >
       group: < group_name >
+  dot1x:
+    default:
+      type: < none | start-stop | stop-only >
+      group: < group_name >
```
tgodaA: This document is optional. We no longer will be maintaining this one from 4.0. Because the documentation is auto generated.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/aaa-accounting.j2
```diff
@@ -23,6 +23,9 @@ aaa accounting exec default {{ aaa_accounting.exec.default.type }} group {{ aaa_
 {%     if aaa_accounting.system.default.type is arista.avd.defined and aaa_accounting.system.default.group is arista.avd.defined %}
 aaa accounting system default {{ aaa_accounting.system.default.type }} group {{ aaa_accounting.system.default.group }}
 {%     endif %}
+{%     if aaa_accounting.dot1x.default.type is arista.avd.defined and aaa_accounting.dot1x.default.group is arista.avd.defined %}
+aaa accounting dot1x default {{ aaa_accounting.dot1x.default.type }} group {{ aaa_accounting.dot1x.default.group }}
+{%     endif %}
```
tgodaA, line 28: Still supports more groups

image

It should be a list then

tgodaA: Just for example I added radius twice but it supports other custom groups as well.

ClausHolbechArista: We don't do that on any of the others (we should), so maybe let this one go for now, and we can add `groups` to all of them and deprecate the singular group.

dmilyt: I will try my best to implement the request from Tony.

kmueller68: It currently turned out to be a little difficult to implement it right now supporting multiple groups.

Since the others currently also do not support multiple groups and from perspective of time for our current campus project, would it please be possible to start with the provided implementation?

If you agree,
should we stay with this data model
```
aaa_accounting:
  dot1x:
    default:
      type: start-stop
      group: RADIUS
      multicast: true
      logging: true
 ```
or should we already go for an expandable one:
```
  dot1x:
    default:
      type: start-stop
      group:
        - name: RADIUS
          multicast: true
          logging: true
 ```

tgodaA: Use the last with minor modification:

```yaml
  dot1x:
    default:
      type: start-stop
      groups:
        - name: < group_name >
          multicast: < boolean >
          logging: < boolean >
```

ClausHolbechArista: This EOS CLI is a bit confusing, and the documentation has a couple of typos which does not help :)
`logging` is a "method" similar to `group abc`, so it is _not_ a property of a group.
For some reason EOS does not have a guard against repeating the same method multiple times, so that is why `logging` shows up after every group.
I do not know how the ordering of `logging` vs. `group abc` might affect the behavior, so until we know that, I would suggest that we skip expanding for now and just implement what we need (one group with `group` key similar to the rest of the models).
A future data model should probably be referring to `methods` which can then either be a group (with or without multicast) or logging.
```
dot1x:
  default:
    type: start-stop
    methods:
      - method: < logging | group >
        group: < group_name >   <-- Only relevant for method group
        multicast: < bool >     <-- Only relevant for method group
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/README_v4.0.md
```diff
@@ -393,6 +393,10 @@ aaa_accounting:
     default:
       type: < none | start-stop | stop-only >
       group: < group_name >
+  dot1x:
+    default:
+      type: < none | start-stop | stop-only >
```
ClausHolbechArista: `none` is not a valid value for this data model. Should be removed here and in the schema.

ClausHolbechArista (changes requested):

ClausHolbechArista (approved):

gmuloc (approved):
LGTM - probably should log an issue for moving to groups in the future so we track it
Thanks for the contribution!

gmuloc:
> LGTM - probably should log an issue for moving to groups in the future so we track it Thanks for the contribution!

https://github.com/aristanetworks/ansible-avd/issues/2521

PR #2485, opened 2023-02-03
Feat(eos_designs): Only require 'virtual_router_mac_address' when using VARP or anycast IP on SVIs

### LINKED ISSUE #2473: Unable to not use virtual-router mac-address
### Issue Summary

At the moment, the `virtual_router_mac_address` needs to be defined on L3 Leafs, even if the leafs do not use MLAG and VxLANs.

```
    # Virtual router mac address for anycast gateway | Required.
    virtual_router_mac_address: < mac address >
```

The impact is that for the cases where this command is not required, `custom_structured_configuration` needs to be used override this behaviour.

eos_designs

### How do you run AVD ?

Ansible CLI with AVD Runner

### Steps to reproduce

```shell
Trying to run eos_designs against a fabric with L3 Leafs with: 
- Missing the `virtual_router_mac_address` field
- or having `virtual_router_mac_address` be empty or set to None

will result in an error.
```

### Relevant log output

Tested on both AVD 3.7.0 and 3.8.0. 

Example output when `virtual_router_mac_address` is not defined with AVD 3.8.0 : 

```shell
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.arista.avd.plugins.plugin_utils.errors.errors.AristaAvdMissingVariableError: virtual_router_mac_address
```

### Contributing Guide

ClausHolbechArista:
## Change Summary

Only require 'virtual_router_mac_address' when using VARP or anycast IP on SVIs

## Related Issue(s)

Fixes #2473 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Remove static requirement for `virtual_router_mac_address`
- Update docs
- Add check for mac during rendering of SVIs and emit errors accordingly
- Update molecule group_vars under `eos_designs_unit_tests` to only set the virtual mac where needed.

## How to test

- Removed the mac from the unit test groups vars where not strictly required. Still works.
- Tested with removing the mac from some of the remaining unit test group vars and saw the errors being raised.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/ip_virtual_router_mac_address.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/ip_virtual_router_mac_address.py
@@ -2,8 +2,6 @@ from __future__ import annotations
 
 from functools import cached_property
 
-from ansible_collections.arista.avd.plugins.plugin_utils.utils import get
-
 from .utils import UtilsMixin
 
 
@@ -18,8 +16,7 @@ class IpVirtualRouterMacAddressMixin(UtilsMixin):
         """
         Return structured config for ip_virtual_router_mac_address
         """
-        if self._network_services_l2 and self._network_services_l3:
-            mac = get(self._hostvars, "switch.virtual_router_mac_address", required=True, org_key="virtual_router_mac_address")
+        if self._network_services_l2 and self._network_services_l3 and (mac := self._virtual_router_mac_address) is not None:
             return str(mac).lower()
 
         return None
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
@@ -46,6 +46,19 @@ class VlanInterfacesMixin(UtilsMixin):
         return None
 
     def _get_vlan_interface_config_for_svi(self, svi, vrf, tenant) -> dict:
+        def _check_virtual_router_mac_address(vlan_interface_config: dict, variables: list):
+            """
+            Helper function
+
+            Check if any variable in the list of variables is not None in vlan_interface_config
+            and if it is the case, raise an Exception if virtual_router_mac_address is None
+            """
+            if any(vlan_interface_config[var] for var in variables) and self._virtual_router_mac_address is None:
+                quoted_vars = [f"'{var}'" for var in variables]
+                raise AristaAvdMissingVariableError(
+                    f"'virtual_router_mac_address' must be set for node '{self._hostname}' when using {' or '.join(quoted_vars)} under 'svi'"
+                )
+
         vlan_interface_config = {
             "tenant": tenant["name"],
             "tags": svi.get("tags"),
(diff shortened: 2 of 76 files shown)


ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/ip_virtual_router_mac_address.py
```diff
@@ -18,8 +16,7 @@ def ip_virtual_router_mac_address(self) -> str | None:
         """
         Return structured config for ip_virtual_router_mac_address
         """
-        if self._network_services_l2 and self._network_services_l3:
-            mac = get(self._hostvars, "switch.virtual_router_mac_address", required=True, org_key="virtual_router_mac_address")
+        if self._network_services_l2 and self._network_services_l3 and (mac := self._virtual_router_mac_address) is not None:
```
tgodaA, line 19: What if the user has defined `virtual_router_mac_address: ''` empty string because I see that in older AVD versions.

ClausHolbechArista: We will add a string format rule to the schema for this key. To enforce mac address format.

carlbuchmann: Agreed to address this with schema formating rules.

gmuloc:
@ClausHolbechArista  - FYI pushed a PR on your branch: https://github.com/ClausHolbechArista/ansible-avd/pull/24

emilarista (approved):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
```diff
@@ -46,6 +46,17 @@ def vlan_interfaces(self) -> dict | None:
         return None
 
     def _get_vlan_interface_config_for_svi(self, svi, vrf, tenant) -> dict:
+        def _check_virtual_router_mac_address(vlan_interface_config: dict, variables: list):
+            """
+            Check if any variable in the list of variables is not None in vlan_interface_config
+            and if it is the case, raise an Exception if virtual_router_mac_address is None
+            """
+            if any(vlan_interface_config[var] for var in variables) and self._virtual_router_mac_address is None:
+                quoted_vars = [f"'{var}'" for var in variables]
```
tgodaA, line 55: !!! Optional

This won't create any intermediate list.
```suggestion
                quoted_vars = map(lambda var: f"'{var}'", variables)
```

tgodaA: Not an optimization for smaller lists but definitely good for bigger lists.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
```diff
@@ -46,6 +46,17 @@ def vlan_interfaces(self) -> dict | None:
         return None
 
     def _get_vlan_interface_config_for_svi(self, svi, vrf, tenant) -> dict:
+        def _check_virtual_router_mac_address(vlan_interface_config: dict, variables: list):
+            """
+            Check if any variable in the list of variables is not None in vlan_interface_config
+            and if it is the case, raise an Exception if virtual_router_mac_address is None
+            """
+            if any(vlan_interface_config[var] for var in variables) and self._virtual_router_mac_address is None:
```
tgodaA, line 54:
```suggestion
            if any(map(vlan_interface_config.get, variables)) and self._virtual_router_mac_address is None:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
ClausHolbechArista: Thank you for the suggestion.
I don't think it would be better in this case. It would just make it harder to read.
In general `map`+`lambda` will probably be slower that `iter(f"{var}" for var in variables)`.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vlan_interfaces.py
ClausHolbechArista: The current any(<list comprehension>) is more optimal, since the list comprehension is never compiled to a list, but just taken as an iterator.

tgodaA (approved):
LGTM

PR #2529, opened 2023-02-13
Fix(eos_designs): Correct range_expand behaviour with .0 4byte ASNs

jonxstill:
## Change Summary

Fix behaviour of range_expand when used with 32-bit ASNs with .0 as the 2nd pair of bytes.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Zero in the 2nd byte would be treated as false and would never be appended to the rest of the ASN.  Changed test to "is not None" to correct this behaviour.

## How to test
Molecule test cases added for 32-bit ASNs, including with .0 in the 2nd half.

### Repository Checklist


carlbuchmann (approved):
LGTM

tgodaA (approved):
LGTM

ClausHolbechArista (approved):
Thank you for fixing this.

PR #2624, opened 2023-03-15
Refactor(eos_designs): base python_module as per eos_cli_config_gen

mayurgs34:
Refactor(eos_designs): base python_module as per eos_cli_config_gen

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
@@ -122,7 +122,7 @@ class AvdStructuredConfig(AvdFacts):
         return static_routes
 
     @cached_property
-    def ipv6_static_routes(self):
+    def ipv6_static_routes(self) -> list | None:
         """
         ipv6_static_routes set based on ipv6_mgmt_gateway, ipv6_mgmt_destination_networks and mgmt_interface_vrf
         """
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
@@ -13,7 +13,7 @@ from ansible_collections.arista.avd.plugins.plugin_utils.utils import get
 
 class AvdStructuredConfig(AvdFacts):
     @cached_property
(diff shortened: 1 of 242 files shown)


ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
```diff
@@ -174,7 +174,7 @@ def ipv6_unicast_routing(self):
         return None
 
     @cached_property
-    def ip_routing_ipv6_interfaces(self):
+    def ip_routing_ipv6_interfaces(self) -> bool | None:
```
gmuloc, line 177: Thanks for taking the time to add these return types everywhere - could you also add it to router_multicast below line 186 and other you could have skipped so that things are aligned?

gmuloc (commented):
LGTM - thanks for taking the time to adding the return types to the functions in `base/` would you mind finishing adding the return types to the few methods you have missed so that we align everything? Thanks in advance

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
mayurgs34: Yes sure, Thanks.

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
```diff
@@ -13,65 +13,65 @@
 
 class AvdStructuredConfig(AvdFacts):
     @cached_property
-    def _mgmt_interface_vrf(self):
+    def _mgmt_interface_vrf(self) -> str:
```
gmuloc: For generic get, the output could be None as well - this applies to the next 7 functions omitting the platform one.

gmuloc (approved):
thanks a lot - added one comment if you can check and address it but otherwise LGTM

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
mayurgs34: I have added None to generic get, omitting the platform.

emilarista (approved):
LGTM!

PR #2616, opened 2023-03-10
Test(eos_cli_config_gen): Move testing of deprecated data models

ClausHolbechArista:
## Change Summary

Move testing of deprecated data models to `eos_cli_config_gen_deprecated_vars` scenario

PR #2630, opened 2023-03-16
Refactor(eos_cli_config_gen,eos_designs): Deprecate isis_af_defaults and address_family knobs

### LINKED ISSUE #1299: eos_cli_config_gen - Refactor `isis_af_defaults`
This implementation should be refactored into the separate address families. We should not do this kind of abstraction in eos_cli_config_gen. 

```yaml
 isis_af_defaults:
    - maximum-paths < Integer 1-64 >
    - fast-reroute ti-lfa mode link-protection < level-1 | level-2 (Optional, defaults to protect all levels) >
```

emilarista:
## Change Summary

Adds deprecation warnings for outdated isis_af_defaults and address_family knobs, moves old knobs to deprecation molecule scenario for eos_cli_config_gen and makes eos_designs output updated variables.

Requires #2616 

## Related Issue(s)

Fixes #1299 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Updated schemas with deprecation warnings, altered eos_designs python modules.

## How to test

Ran molecule, made sure no or only intended changes in intended config occur.

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/python_modules/underlay/router_isis.py
```diff
@@ -72,7 +71,7 @@ def router_isis(self) -> dict | None:
             # TODO - enabling IPv6 only in SR cases as per existing behavior
             # but this could probably be taken out
             if self._underlay_ipv6 is True:
-                router_isis["address_family"].append("ipv6 unicast")
+                router_isis["address_family_ipv6"] = router_isis["address_family_ipv4"]
```
gmuloc: I think it would be better here to just rewrite the ipv6 and ipv4 blocks separately in case we want to enable ipv6 only.

Additionally, here your are copying over the dict object and if we want to make them different at some point it will have unexpected effects - notice how when I add "ADD" to `ipv6` it is also added to `ipv4` because this is the same dict (I know for now it does not matter but we should really have two different objects):

```python
>>> d1 = {"ipv4": {"ffr": True, "blah": 42}}
>>> d1["ipv6"] = d1["ipv4"]
>>> id(d1["ipv6"])
4345528640
>>> id(d1["ipv4"])
4345528640
>>> d1
{'ipv4': {'ffr': True, 'blah': 42}, 'ipv6': {'ffr': True, 'blah': 42}}
>>> d1["ipv6"]["ADD"] = 666
>>> d1
{'ipv4': {'ffr': True, 'blah': 42, 'ADD': 666}, 'ipv6': {'ffr': True, 'blah': 42, 'ADD': 666}}
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/underlay/router_isis.py
```diff
@@ -57,9 +56,9 @@ def router_isis(self) -> dict | None:
             }
         ti_lfa_protection = get(self._hostvars, "isis_ti_lfa.protection")
         if ti_lfa_protection == "link":
-            router_isis["isis_af_defaults"].append("fast-reroute ti-lfa mode link-protection")
+            router_isis["address_family_ipv4"].update({"fast_reroute_ti_lfa": {"mode": "link-protection"}})
```
gmuloc: don't use update for only one key
```suggestion
            router_isis["address_family_ipv4"]["fast_reroute_ti_lfa"] = {"mode": "link-protection"}

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/underlay/router_isis.py (hunk 1 above)
Comment (emilarista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/underlay/router_isis.py]: I updated the module to build the v6 AF independently, thanks!

File: ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/router_isis.schema.yml [hunk 3]
```diff
@@ -56,13 +56,21 @@ keys:
             - "ipv4 unicast"
             - "ipv6 unicast"
           description: Address Family
+          deprecation:
+            warning: true
+            new_key: address_family_ipv4/address_family_ipv6
```
ClausHolbechArista: Same elsewhere 
```suggestion
            new_key: address_family_ipv4.enabled or address_family_ipv6.enabled
```

ClausHolbechArista: We should add a new `enabled` key under each address family, so we never have a situation where we can have a "dangling" key. The issue is in YAML, if you just want to have the address family activated but no specific config below, you would write:

```yaml
router_isis:
  address_family_ipv4:
```

which YAML interprets as `address_family_ipv4: null` so when we do our `arista.avd.defined` check, it fails.

I know that requiring this `enabled` flag is breaking, but now is the time for breaking changes :)

emilarista: Cool, added enabled keys to isis AFs in the module!

carlbuchmann (changes requested):
@emilarista thank you for this PR!

carlbuchmann (approved):
LGTM

gmuloc (approved):
LGTM - added a test in `eos_designs_unit_test` to trigger ISIS TI LFA node protection from eos_designs as it was never exercised (from what became apparent in this deprecation PR ;) )

PR #2639, opened 2023-03-17
Feat(eos_cli_config_gen): Enable redistribution of leaked (static, connected, bgp) routes into OSPF

### LINKED ISSUE #2497: Feat(eos_cli_config_gen): Enable redistribution of leaked (static and connected) routes into OSPF
### Enhancement summary

Enable the ability to redistribute routes from the leak agent.

Additional details can be found at: https://www.arista.com/en/support/toi/eos-4-24-1f/14504-redistribution-of-leaked-routes-into-igps

eos_cli_config_gen

### Use case example

Need to announce leaked routes via OSPF.

Specify new keys for `include_leaked` in the existing router_ospf structure.
``` 
router_ospf:
  process_ids:
    - id: < process_id >
      redistribute:
        static:
          include leaked: < true | false>
          route_map: < route_map_name >
        connected:
          include leaked: < true | false>
          route_map: < route_map_name >
```

The new key would render the needed option in the following line:
```
redistribute (static|connected) [ include leaked ] [ route-map <route-map-name> ]
```

OSPF does not seem to have a `eos_cli ` key.

### Additional context

_No response_

### Contributing Guide

carl-baillargeon:
## Change Summary

Enables the ability to redistribute leaked (static, connected, bgp) routes into OSPF.

## Related Issue(s)

Fixes #2497

## Component(s) name

## Proposed changes
```
router_ospf:
  process_ids:
    - id: < process_id >
      redistribute:
        static:
          include_leaked: < true | false>
          route_map: < route_map_name >
        connected:
          include_leaked: < true | false>
          route_map: < route_map_name >
       bgp:
          include_leaked: < true | false>
          route_map: < route_map_name >         
```

## How to test
Tested with molecule :
`molecule converge --scenario-name eos_cli_config_gen -- --limit router-ospf`
`molecule converge --scenario-name eos_cli_config_gen_v4.0 -- --limit router-ospf`

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
```diff
@@ -5,31 +5,31 @@ keys:
...
-      for that child.\n  - In the case of `max_parallel_uplinks` > 1 the `default_downlink_interfaces`
-      are mapped with consecutive downlinks per child ID.\n  - Example for `max_parallel_uplinks:
-      2`, downlink interfaces will be mapped as `[ <downlink1 to leaf-id1>, <downlink2
-      to leaf-id1>, <downlink1 to leaf-id2>, <downlink2 to leaf-id2> ...]`\n- Please
-      note that no default interfaces are defined in AVD itself. You will need to
-      create your own based on the example below.\n\nExample:\n\n```yaml\ndefault_interfaces:\n
-      \ - types: [ spine, l3leaf ]\n    platforms: [ \"7050[SC]X3\", vEOS.*, default
-      ]\n    uplink_interfaces: [ Ethernet49-54/1 ]\n    mlag_interfaces: [ Ethernet55-56/1
-      ]\n    downlink_interfaces: [ Ethernet1-32/1 ]\n```\n"
+    description: "- Set default uplink, downlink, and MLAG interfaces, which will\
```
gmuloc: This changes are due to the fact that you are not running using `libyaml: true` on ansible - try to install ansible with libyaml and re running the molecule

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-ospf.yml
```diff
@@ -70,10 +70,13 @@ router_ospf:
       redistribute:
         static:
           route_map: rm-ospf-static
+          include_leaked: true
         connected:
           route_map: rm-ospf-connected
+          include_leaked: true
```
gmuloc, line 76: can we also test include_leaked true without route-map? seems to be accepted by EOS as well
```
Leaf1(config-router-ospf)#redistribute bgp include leaked
Leaf1(config-router-ospf)#show active
router ospf 666
   redistribute bgp include leaked
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/eos_cli_config_gen.schema.yml
```diff
@@ -379,9 +379,9 @@ keys:
     documentation_options:
       filename: data_model/Aliases
     type: str
-    description: "Multi-line string with one or more alias commands.\n\nExample:\n\n```yaml\naliases:
-      |\n  alias wr copy running-config startup-config\n  alias siib show ip interface
-      brief\n```"
+    description: "Multi-line string with one or more alias commands.\n\nExample:\n\
+      \n```yaml\naliases: |\n  alias wr copy running-config startup-config\n  alias\
```
gmuloc: this change is due to the absence of libyaml while running molecule

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
```diff
@@ -88,22 +88,37 @@
...
 {%         for process_id in router_ospf.process_ids | arista.avd.natural_sort('id') %}
 {%             if process_id.redistribute is arista.avd.defined %}
 {%                 set source_protocols = [] %}
 {%                 if process_id.redistribute.connected is defined %}
-{%                     do source_protocols.append(('connected', process_id.redistribute.connected.route_map | arista.avd.default('-'))) %}
+{%                     if process_id.redistribute.connected.include_leaked is arista.avd.defined(true) %}
+{%                         set include_leaked = 'enabled' %}
+{%                     else %}
+{%                         set include_leaked = 'disabled' %}
+{%                     endif %}
```
gmuloc, line 101:
```suggestion
{%                     set include_leaked = 'enabled'  if process_id.redistribute.connected.include_leaked is arista.avd.defined(true) else 'disabled' %}
```

gmuloc (changes requested):
Thanks for this PR - looks like you need to:
1. install ansible with libyaml (`ansible --version` to check) so that you don't get the `\` new line changes in molecule
2. re-run pre-commit - it failed in CI on the schema.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
carl-baillargeon: Thanks for the review. I did it this way for consistency with the current template. Should I refactor the whole template with your suggested approach ?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
```diff
@@ -88,22 +88,37 @@
...
 #### Router OSPF Router Redistribution
 
-| Process ID | Source Protocol | Route Map |
-| ---------- | --------------- | --------- |
+| Process ID | Source Protocol | Include Leaked | Route Map |
+| ---------- | --------------- | -------------- | --------- |
 {%         for process_id in router_ospf.process_ids | arista.avd.natural_sort('id') %}
 {%             if process_id.redistribute is arista.avd.defined %}
 {%                 set source_protocols = [] %}
 {%                 if process_id.redistribute.connected is defined %}
```
gmuloc, line 96: Actually can we change this line to `arista.avd.defined` as well? (I know it is not part of your PR but well this should be refactored so let's take the opportunity)

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
gmuloc: Just focus on the part you are adding for now.

As a side note, I was almost suggesting to refactor the different `if` statements for `connected`, `static` and `bgp` in a for loop as this template looks quite bloated and creating a tuple to then access the various element with `[0]`,... does not look very optimal.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/router-ospf.j2
```diff
@@ -120,20 +120,29 @@ router ospf {{ process_id.id }}
 {%     endif %}
 {%     if process_id.redistribute.static is defined %}
```
gmuloc, line 121: This could / should also be replace with
```suggestion
{%     if process_id.redistribute.static is arista.avd.defined %}
```

carl-baillargeon: As discussed, `is defined` must be kept for now since `static:` can be `null`.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
carl-baillargeon: As discussed, `is defined` must be kept for now since `connected:` can be `null`.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/router-ospf.j2
ClausHolbechArista: I really don't like the loops, when the data models are dicts in the schema. They are kept as dicts because there are different options under each (in EOS).

ClausHolbechArista: Re the suggestion from @gmuloc, we try to avoid that style everywhere in Jinja. It will probably also trigger errors from j2lint.

gmuloc: not for these 3 regarding the loop but ok for the other not implemented options so "for future cases" let's keep it like this.
```
Leaf1(config-router-ospf)#show cli commands | grep redistribute
[no|default] redistribute ( ( rip | aggregate ) ) [ route-map MAPNAME ]
[no|default] redistribute ( ( static | connected | bgp ) ) [ include leaked ] [ route-map MAPNAME ]
[no|default] redistribute isis [ include leaked ] [ level-1|level-2|level-1-2 ] [ route-map MAPNAME ]
[no|default] redistribute ospf ( ( instance [ include leaked ] ) | leaked ) [ match ( internal | external | nssa-external ) ] [ route-map MAPNAME ]
```

Regarding the suggestion - we write five lines that could be one 3 times so that makes the templates much longer and less easy to maintain. j2lint does not error on this syntax (if there were some operators in the line it could eventually complain about some spaces but it is ok with this and it is good as it is very valid syntax).

All that being said - let's leave it like this. Going to resolve my comment

gmuloc (approved):

carlbuchmann (approved):
LGTM

PR #2645, opened 2023-03-20
Feat(eos_designs): Add support for serial_number

gmuloc:
## Change Summary

* Support to define serial number as hostvar or fabric topology var, the latter having precedence. For now only used for documentation purposes in the fabric documentation.

## Related Issue(s)

None

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

in management_settings:

```yaml
# Serial Number | Optional
# Set to the Serial Number of the device
# "serial_number" can also be set under "Fabric Topology".
# If both are set, the setting under "Fabric Topology" takes precedence.
serial_number: < string >
```

In fabric topology:

```yaml
<node_type_key>:
  [...]
  nodes:
    <node inventory hostname>:
      [...]
      # "system_mac_address" can also be set directly as a hostvar.
      # If both are set, the setting under "Fabric Topology" takes precedence.
      system_mac_address: < "xx:xx:xx:xx:xx:xx" >
      # Serial Number | Optional
      # Set to the Serial Number of the device
      # "serial_number" can also be set directly as a hostvar.
      # If both are set, the setting under "Fabric Topology" takes precedence.
      serial_number: < string >
```

## How to test

molecule tests added

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/fabric/DC1_FABRIC-documentation.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/fabric/DC1_FABRIC-documentation.md
@@ -15,18 +15,18 @@
 
 ## Fabric Switches and Management IP
 
-| POD | Type | Node | Management IP | Platform | Provisioned in CloudVision |
-| --- | ---- | ---- | ------------- | -------- | -------------------------- |
-| DC1_FABRIC | leaf | LEAF1A | 172.100.100.103/24 | cEOSLab | Provisioned |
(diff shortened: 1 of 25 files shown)


ClausHolbechArista (approved):
Maybe add something to the documentation explaining how/where the serial_number is used?
For now only documentation, but later we could add it to deploy_cvp and maybe the password filter?

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #2660, opened 2023-03-27
Fix(eos_designs): Configure "ip routing ipv6 interface vrf X" for RFC5549

### LINKED ISSUE #2646: RFC5549 IPv4 routing in overlay broken due to missing config
### Issue Summary

The configuration would work fine for underlay due to `ip routing ipv6 interfaces` being present. For overlay the switch generates the correct BGP update however doesn't know how to route traffic without the knob in the overlay.

`ip routing ipv6 interfaces  vrf transit`

Is mandatory so that the routing can be done for the VRF using iBGP over the peer-link. Otherwise traffic will be black-holed in case of uplink convergence.

The above config in turn should remove the config from running-config:

`ip routing vrf transit`

To enable:

```
# Enable RFC 5549(iBGP) i.e. IPv6 Unnumbered for MLAG iBGP connections.
# Requires "underlay_rfc5549: true"
overlay_mlag_rfc5549: < true | false | default -> false >
```

If this would be tweaked for the VRF template that would solve the problem.

```
`    @cached_property
    def ip_routing_ipv6_interfaces(self) -> bool | None:
        """
        ip_routing_ipv6_interfaces set based on underlay_rfc5549 variable
        """
        if get(self._hostvars, "underlay_rfc5549") is True:
            return True
        return None`
```

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
If you configure RFC 5549 and you have MLAG iBGP peers in VRFs this will not generate the config `ip routing ipv6 interfaces  vrf transit` which is required for routing of IPv4 packets with IPv6 next-hops.

This works well for the underlay when using RFC5549.

```

### Relevant log output

_No response_

### Contributing Guide

gmuloc:
## Change Summary

Renders correctly the `ip routing ipv6 interfaces vrf X` config for RFC5549 VRFs as reported in the linked issue.
Add support for it in `eos_cli_config_gen` and fix documentation and a couple of issues.

## Related Issue(s)

Fixes #2646 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

* Add configuration of `ip routing ipv6 interfaces vrf X` knob to eos_cli_config_gen
* Adjust `eos_cli_config_gen` documentation templates to show the information for `ip routing ipv6 interfaces`
* Fix a bug where AVD renders `ipv6 unicast-routing vrf default` which is converted to `ipv6 unicast-routing`
* Stop rendering `ip routing` when `ip routing ipv6 interfaces` is configured in eos_designs
* Render `ip routing ipv6 interfaces vrf X` in VRF where it is required based on the underlay/overlat RFC5549 variables from `eos_designs`

## How to test

molecule scenario and particularly the evpn_underlay_rfc5549_overlay_ebgp one

### Repository Checklist


gmuloc:
@onurgashi please take a look at this as a starting point

ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF1A.md
```diff
@@ -1384,10 +1384,10 @@ service routing protocols model multi-agent
 
 #### IP Routing Summary
 
-| VRF | Routing Enabled |
-| --- | --------------- |
-| default | True |
-| MGMT | false |
+| VRF | Routing Enabled | IPv6 Interfaces |
```
ClausHolbechArista: This feature will be used in 0.5% of deployments (give or take :). I don't think we should place it as a column in this table. You could add `(ipv6 interfaces)` after the boolean.

ansible_collections/arista/avd/examples/single-dc-l3ls/intended/structured_configs/dc1-leaf1b.yml
```diff
@@ -180,10 +180,10 @@ vrfs:
   ip_routing: false
 - name: VRF10
   tenant: TENANT1
-  ip_routing: true
+  ip_routing_ipv6_interfaces: true
```
ClausHolbechArista: I think some trigger is too sensitive. The examples do not use rfc5549.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/vrfs.py
```diff
@@ -45,8 +45,14 @@ def vrfs(self) -> list | None:
                 new_vrf = {
                     "name": vrf_name,
                     "tenant": tenant["name"],
-                    "ip_routing": True,
                 }
+
+                # MLAG IBGP Peering VLANs per VRF
+                if self._mlag_ibgp_peering_enabled(vrf, tenant):
```
ClausHolbechArista: This is not the correct condition. You should be looking for `overlay_mlag_rfc5549` (as well)

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/ip-routing.md
```diff
@@ -47,16 +47,15 @@ interface Management1
 
 | VRF | Routing Enabled |
 | --- | --------------- |
-| default | False |
-| default | true |
-| TEST1 | true |
-| TEST2 | false |
+| default | True (ipv6 interfaces) |
+| default | True |
```
ClausHolbechArista: Two lines here? Let's try and clean up. I understand why, but we should fix this.

ansible_collections/arista/avd/molecule/evpn_underlay_rfc5549_overlay_ebgp/intended/configs/DC1-SPINE1.cfg
```diff
@@ -86,8 +86,6 @@ interface Management1
    no shutdown
    vrf MGMT
    ip address 192.168.200.101/24
-!
-ip routing
```
ClausHolbechArista, line 90: are we sure about the ordering here?

gmuloc:
```
!
ip routing ipv6 interfaces
no ip icmp redirect
ip routing vrf mgmt
ip routing ipv6 interfaces  vrf toto
ip routing vrf zozo
!
ipv6 unicast-routing
```

Ordering for future reference

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-routing.j2
```diff
@@ -7,13 +7,26 @@
...
-{%         if vrf.ip_routing is arista.avd.defined(true) %}
-| {{ vrf.name }} | true |
-{%         elif vrf.ip_routing is arista.avd.defined(false) %}
-| {{ vrf.name }} | false |
+{%     set default_ip_routing = ip_routing | arista.avd.default(false) %}
+{%     set default_ip_routing_ipv6_interfaces = ip_routing_ipv6_interfaces | arista.avd.default(false) %}
+{# Handle the case where VRF default exists #}
+{%     set vrf_default = vrfs | arista.avd.default([]) | selectattr("name", "arista.avd.defined", "default") | first %}
+{%     if vrf_default is arista.avd.defined %}
+{%         set default_ip_routing = (vrf_default.ip_routing or default_ip_routing) %}
```
ClausHolbechArista: no variable protection on `vrf_default.ip_routing`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-routing.j2
```diff
@@ -7,13 +7,26 @@
...
-| default | {{ ip_routing | arista.avd.default(false) }} |
-{%     for vrf in vrfs | arista.avd.natural_sort('name') %}
-{%         if vrf.ip_routing is arista.avd.defined(true) %}
-| {{ vrf.name }} | true |
-{%         elif vrf.ip_routing is arista.avd.defined(false) %}
-| {{ vrf.name }} | false |
+{%     set default_ip_routing = ip_routing | arista.avd.default(false) %}
+{%     set default_ip_routing_ipv6_interfaces = ip_routing_ipv6_interfaces | arista.avd.default(false) %}
+{# Handle the case where VRF default exists #}
+{%     set vrf_default = vrfs | arista.avd.default([]) | selectattr("name", "arista.avd.defined", "default") | first %}
```
ClausHolbechArista: All this logic is handled in the documentation template, so you can set ip routing for default vrf either with the global keys or with `vrf:default`. Unfortunately that is not how it is implemented in the configuration. So we will not _configure_ ip routing for vrf default, if it is only set under the vrfs.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-routing.j2
```diff
@@ -7,13 +7,26 @@
...
+{%     if vrf_default is arista.avd.defined %}
+{%         set default_ip_routing = (vrf_default.ip_routing or default_ip_routing) %}
+{%         if default_ip_routing_ipv6_interfaces or (vrf_default.ip_routing_ipv6_interfaces | arista.avd.default(false)) %}
+{%             set default_ip_routing = "True (ipv6 interfaces)" %}
 {%         endif %}
+{# No VRF "default" defined - checking if default_ip_routing_ipv6_interfaces is True #}
```
ClausHolbechArista: I think this could be simplified by just setting `default_vrf_xxx` vars from vrf with defaults and then do the or check afterwards.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-routing.j2
gmuloc: so the point of this comment is to implement this in the eos template as well right?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/ip-routing.j2
gmuloc: good point - the only thing is that I am playing with booleans and strings but I was not completely pleased with the logic so will rearrange- thanks for feedback

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #2672, opened 2023-03-30
Refactor(plugins): Optimize schema tooling

ClausHolbechArista:
## Change Summary

Optimize schema tooling

## Component(s) name

`arista.avd.plugins`

## Proposed changes

- Remove redundant validation of schemas.
- Remove support for adhoc schemas (not used in our plugins)
- Remove `is_valid` on `AvdSchema` class (not used in our plugins)

Only risk is if someone is using our tooling outside of AVD, don't load the schema first, and is not providing a valid schema.

I see huge speed improvements after adding removing the redundant validations.

## How to test

No changes expected anywhere - except much better runtime.

When comparing `eos_designs_units_tests`:

```sh
#devel
real    4m19,687s
user    30m52,599s
sys     0m55,736s

#pr2672
real    3m23,947s
user    20m11,712s
sys     0m54,875s
```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
@@ -32,9 +32,12 @@ class AvdDataConverter:
     def __init__(self, avdschema):
         self._avdschema = avdschema
 
+        # We run through all the regular keys first, to ensure that all data has been converted
+        # in case some of it is referenced in "dynamic_keys" below
         self.converters = {
             "items": self.convert_items,
             "keys": self.convert_keys,
+            "dynamic_keys": self.convert_dynamic_keys,
             "deprecation": self.deprecation,
         }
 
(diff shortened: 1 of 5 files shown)


gmuloc (approved):
LGTM - was wondering reagrding clearing the cached resolved schema if it would not be better to
1/ do it at the beginning of the functions rather than the end
2/ the __dict__.pop method (the python doc recommends deleting the attribute to clear the cache: https://docs.python.org/3/library/functools.html#functools.cached_property).

ClausHolbechArista:
> LGTM - was wondering reagrding clearing the cached resolved schema if it would not be better to 
> 1/ do it at the beginning of the functions rather than the end 

The method could fail, not changing the loaded schema, but I guess we can survive re-resolving in that rare case. I will move it.

> 2/ the **dict**.pop method (the python doc recommends deleting the attribute to clear the cache: https://docs.python.org/3/library/functools.html#functools.cached_property).

Lazy I guess. I save checking if the key is there. Pop will do exactly the same, check if key is there, and delete it.

carlbuchmann (approved):
LGTM

PR #2690, opened 2023-04-02
Feat(eos_cli_config_gen): Add support for PoE configurations

### LINKED ISSUE #2663: Feat(eos_cli_config_gen): Global poe configuration options
### Enhancement summary

Add global poe options to configure the switch behaviour for poe on reboot or interface shutdown:

eos_cli_config_gen

### Use case example

Global configuration of the switch behaviour regarding poe supply for all interfaces:

```
poe
   reboot action maintain
   interface shutdown action power-off
```

```
poe:
  interface_shutdown_action: < maintain | power-off>
  reboot:
    action: < maintain | power-off>
```

This can be accomplished today with cvp configlet or eos_cli:.

### Additional context

_No response_

### Contributing Guide

### LINKED ISSUE #2683: Interface PoE commands
### Enhancement summary

Add interface level poe commands to configure common campus interface configurations

eos_cli_config_gen

### Use case example

Interface level poe commands to address many of the commands found in this [EOS TOI](https://www.arista.com/en/support/toi/eos-4-26-1f/14784-power-over-ethernet-poe).

A data model that resembles the below:

```
ethernet_interfaces:
  - name: <Interface-Name>
    poe:
      disabled: < true | false >
      priority: < critical | high | medium | low >
      reboot_action: < maintain | power-off >
      shutdown_action: < maintain | power-off >
      link_down_action: < maintain | power-off >
      limit:
        class: < 0-8 >
        watts: < 0-90 >
        fixed: < true | false >
      negotiation_lldp: < true | false >
      legacy_detect: < true | false >
```

This can be accomplished today with eos_cli.

### Additional context

_No response_

### Contributing Guide

jrecchia1029:
## Change Summary

Add support for global PoE configurations
Add support for PoE configurations to the ethernet interface data model

## Related Issue(s)

Fixes #2663 
Fixes #2683

## Component(s) name

## Proposed changes

Add a `poe` dictionary at the root level for global poe options
Add a `poe` dictionary under the `ethernet_interfaces` key

```
poe:
  reboot:
    action:  < maintain | power-off >
  interface_shutdown:
    action:  < maintain | power-off >
```

```
ethernet_interfaces:
  - name: <interface-name>
    poe:
      disabled: < true | false >
      priority: < critical | high | medium | low >
      reboot:
        action: < maintain | power-off >
      shutdown:
        action: < maintain | power-off >
      link_down:
        action: < maintain | power-off >
        power_off_delay: < 1 - 86400 >
      limit:
        class: < 0-8 >
        watts: < 1-90 >
        fixed: < true | false >
      negotiation_lldp: < true | false >
      legacy_detect: < true | false >
```

## How to test

Added onto existing molecule tests

Testing environment as described in the [Contributing to AVD](https://www.avd.sh/en/stable/docs/contribution/setup-environment.html) section of documentation

### User Checklist

- N/A

### Repository Checklist


gmuloc:
For reference:

```
Leaf1(config-if-Et3)#show cli commands | grep poe
[no|default] poe [disabled]
[no|default] poe legacy detect
[no|default] poe limit ( ( class CLASS_NUM ) | ( NUM_WATTS watts ) ) [ fixed ]
[no|default] poe link down action ( maintain | ( power-off [ NUM_SECONDS seconds ] ) )
[no|default] poe negotiation lldp [disabled]
[no|default] poe pairset ( ( alternative ( a | b ) ) | 4-pair )
[no|default] poe priority PRIORITY
[no|default] poe reboot action ( maintain | power-off )
[no|default] poe shutdown action ( maintain | power-off )
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -463,6 +463,53 @@ keys:
...
+              description: Optional field used to disable PoE on a POE capable port. PoE is enabled on all ports that support it by default in EOS.
+            priority:
+              type: str
+              valid_values: ["critical", "high", "medium", "low"]
+              description: Optional field used to prioritize a port's power in the event that one of the switch's power supplies loses power
+            reboot_action:
+              type: str
+              valid_values: ["maintain", "power-off"]
+              description: Optional field used to set the PoE power behavior for a PoE port when the system is rebooted
+            link_down_action:
```
gmuloc: for link-down action only there is a timing option in second as well

```
#show cli commands | grep poe
[...]
[no|default] poe link down action ( maintain | ( power-off [ NUM_SECONDS seconds ] ) )
[...]
[no|default] poe reboot action ( maintain | power-off )
[no|default] poe shutdown action ( maintain | power-off )
```

not saying it should be implemented directly but wondering if the model should not be:
```
link_down:
  action: [maintain, power-off]
  delay: <1-86400> # with a comment that it works only for power-off
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -463,6 +463,53 @@ keys:
...
+                  min: 1
+                  max: 90
+                  convert_types:
+                  - str
+                fixed:
+                  type: bool
+                  description: Optional field used to ignore hardware classification
+            negotiation_lldp:
+              type: bool
+              description: Optional field used to disable to prevent port from negotiating power with powered devices over LLDP. Enabled by default in EOS.
```
gmuloc: Not sure I understand this bit:

```
used to disable to prevent port from negotiating power with powered devices over LLDP
```

did you mean to write:

```
used to prevent port from negotiating power with powered devices over LLDP
```

Also you don't necessarily need to add `Optional field` for each description. There is a required key word that will show up in the documentation: (cf the required column for instance here: https://avd.sh/en/devel/roles/eos_cli_config_gen/docs/data_model/DHCP%20Relay.html)

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -463,6 +463,53 @@ keys:
...
+              description: Optional field used to set the PoE power behavior for a PoE port when the port goes down
+            shutdown_action:
+              type: str
+              valid_values: ["maintain", "power-off"]
+              description: Optional field used to set the PoE power behavior for a PoE port when the port is admin down
+            limit:
+              type: dict
+              description: Optional field used to override the hardware-negotiated power limit using either wattage or a power class.
+              keys:
+                class:
```
gmuloc, line 512: For classes I a not sure if this makes sense to document that it will be rendered as `classX` where `X` is the value in the CLI.

gmuloc: Additionally on my vEOS when using `classX` the show output is converted to a Watt variable:

```
Leaf1(config-if-Et3)#poe limit class class7
Leaf1(config-if-Et3)#show active
interface Ethernet3
   poe limit 75.00 watts
Leaf1(config-if-Et3)#poe limit class class4
Leaf1(config-if-Et3)#show active
interface Ethernet3
   poe limit 30.00 watts
```

I guess we would still want to have the usage of class in AVD but this means there will be some diff between the generated config and the output of show run. @ClausHolbechArista not sure how we generally handle this.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
ClausHolbechArista: If it is only for power off, then rename to `power_off_delay`

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
ClausHolbechArista: We can perform the same mapping in AVD, so take `class` attributes, but output the same as EOS does in the generated config. We could then let a `watts` value take precedence?

gmuloc: probably need to make sure the values are the same on every HW. Looking at our doc: https://www.arista.com/en/um-eos/eos-ethernet-ports#xx1313737

```
Class 0 = 15.4 W
Class 1 = 4 W
Class 2 = 7 W
Class 3 = 15.4 W
Class 4 = 30 W
Class 5 = 45 W
Class 6 = 60 W
```

VS on my vEOS device I have the same value + class7 (75W) and class8 (90W)

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
jrecchia1029: Okay perfect.  And how you wrote it is how I meant to write it.  Copy and paste error on my part.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
jrecchia1029: If we change the data model for `link_down`, I feel like we should change it for `reboot` and `shutdown` as well for consistency purposes.  And I think that would be fine.

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/inventory/host_vars/ethernet-interfaces.yml
```diff
@@ -416,3 +416,33 @@ ethernet_interfaces:
         source_interface: Loopback0
         vrf: MGMT
       - ip_helper: 10.10.64.151
+
```
ClausHolbechArista: Please move these changes to the `eos_cli_config_gen` scenario instead of 4.0. It will solve the conflict at the same time :)

jrecchia1029:
Not sure why the pre-commit validation check fails.  The error message doesn't seem to align with what I see in the file where the issue is detected.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -484,6 +484,43 @@ interface {{ ethernet_interface.name }}
...
+   {{ poe_link_down_action_cli }}
+{%         endif %}
+{%         if ethernet_interface.poe.shutdown.action is arista.avd.defined %}
+   poe shutdown action {{ ethernet_interface.poe.shutdown.action }}
+{%         endif %}
+{%         if ethernet_interface.poe.disabled is arista.avd.defined(true) %}
+   poe disabled
+{%         endif %}
+{%         if ethernet_interface.poe.limit.class is arista.avd.defined %}
+{%             set poe_class_map = dict({0: "15.40", 1: "4.00", 2: "7.00", 3: "15.40", 4: "30.00", 5: "45.00", 6: "60.00", 7: "75.00", 8: "90.00"}) %}
```
ClausHolbechArista: great way of doing it. But I suggest to use an upper-case variable, to signal that it is a constant, and then set it at the top of the file. Then you avoid setting it for every loop. Also I would except the naked dict would be good enough `set poe_class_map = {0: " ...`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -484,6 +484,43 @@ interface {{ ethernet_interface.name }}
...
+{%             set poe_limit_cli = 'poe limit ' ~ poe_class_map[ethernet_interface.poe.limit.class] ~ ' watts' %}
+{%         elif ethernet_interface.poe.limit.watts is arista.avd.defined %}
+{%             set poe_limit_cli = 'poe limit ' ~ "%.2f" | format(ethernet_interface.poe.limit.watts | float) ~ ' watts' %}
```
ClausHolbechArista: you are converting to float twice. Once with the filter and second with the format string. I think just doing `'poe limit ' ~ ethernet_interface.poe.limit.watts | float ~ ' watts'` should do it.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -484,6 +484,43 @@ interface {{ ethernet_interface.name }}
...
+{%             endif %}
+   {{ poe_link_down_action_cli }}
+{%         endif %}
+{%         if ethernet_interface.poe.shutdown.action is arista.avd.defined %}
+   poe shutdown action {{ ethernet_interface.poe.shutdown.action }}
+{%         endif %}
+{%         if ethernet_interface.poe.disabled is arista.avd.defined(true) %}
+   poe disabled
+{%         endif %}
+{%         if ethernet_interface.poe.limit.class is arista.avd.defined %}
```
ClausHolbechArista: Insert an outer if to check for `ethernet_interface.poe.limit is arista.avd.defined`. Then you know that `poe_limit_cli` will be set when you print out or add `fixed`, so you can skip those ifs. It will also ensure that all the code will be skipped if not set.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -463,6 +463,71 @@ keys:
...
+                  - float
+                fixed:
+                  type: bool
+                  description: Set to ignore hardware classification
+            negotiation_lldp:
+              type: bool
+              description: Disable to prevent port from negotiating power with powered devices over LLDP. Enabled by default in EOS.
+            legacy_detect:
+              type: bool
+              description: Allow a subset of legacy devices to work with the PoE switch.  Disabled by default in EOS because it can cause false positive detections.
```
ClausHolbechArista:
```suggestion
              description: Allow a subset of legacy devices to work with the PoE switch. Disabled by default in EOS because it can cause false positive detections.
```

ClausHolbechArista (commented):
- Please add a poe table under ethernet-interfaces documentation.
  - instead of having everything as columns, something like power_off_delay can be added to the "Link Down Action" column like `Power-off (delayed 42 seconds)`
  - Similar for fixed
- I don't know this area of config well enough, so please confirm that all the EOS defaults are not something a user could change with global configuration. Ex. default enabled or disabled. If something can be changed on a global level, we need to output configuration for both cases of a boolean, so the user can override in both directions. Like print `poe` if `disabled: false`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
jrecchia1029: Attempted your suggested solution and it looks like it fall short.  When converting `ethernet_interface.poe.limit.watts` from a string type to a float, if the the float value is specified to 1 decimal place or less, this value will print out to only 1 decimal place (i.e. When `45` is converted to a float, it displays `45.0` when printed).  This is an issue because EOS displays the watts value to 2 decimal places.  Therefore, we need to first convert our string value to a float value (`ethernet_interface.poe.limit.watts | float`) and then we can use the f-string expression `"%.2f"` (which only works on floats and will error out on strings) to format the float value to 2 decimal places.  So I believe we need the original expression `"%.2f" | format(ethernet_interface.poe.limit.watts | float)`

ClausHolbechArista: ok makes sense then. I will resolve this comment.

jrecchia1029:
> * Please add a poe table under ethernet-interfaces documentation.
>   
>   * instead of having everything as columns, something like power_off_delay can be added to the "Link Down Action" column like `Power-off (delayed 42 seconds)`
>   * Similar for fixed
> * I don't know this area of config well enough, so please confirm that all the EOS defaults are not something a user could change with global configuration. Ex. default enabled or disabled. If something can be changed on a global level, we need to output configuration for both cases of a boolean, so the user can override in both directions. Like print `poe` if `disabled: false`

So we want this directly under the ethernet interfaces documentation?  It doesn't warrant its own section?  Also, the poe global parameters are `reboot action`, `shutdown action`, and `lldp negotation`.

ClausHolbechArista:
> > * Please add a poe table under ethernet-interfaces documentation.
> >   
> >   * instead of having everything as columns, something like power_off_delay can be added to the "Link Down Action" column like `Power-off (delayed 42 seconds)`
> >   * Similar for fixed
> > * I don't know this area of config well enough, so please confirm that all the EOS defaults are not something a user could change with global configuration. Ex. default enabled or disabled. If something can be changed on a global level, we need to output configuration for both cases of a boolean, so the user can override in both directions. Like print `poe` if `disabled: false`
> 
> So we want this directly under the ethernet interfaces documentation? It doesn't warrant its own section? Also, the poe global parameters are `reboot action`, `shutdown action`, and `lldp negotation`.

Good idea to create a new section where global knobs and interfaces are handled. See sflow or ospf for examples.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -463,6 +463,71 @@ keys:
...
+              keys:
+                action:
+                  type: str
+                  valid_values: ["maintain", "power-off"]
+                  description: PoE action for interface
+            limit:
+              type: dict
+              description: Override the hardware-negotiated power limit using either wattage or a power class.
+              keys:
+                class:
```
gmuloc, line 512: Maybe add a description in class to indicate that AVD takes care of the conversion :) Thanks for implementing this!

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/poe.schema.yml
```diff
@@ -0,0 +1,30 @@
...
+  poe:
+    documentation_options:
+      filename: data_model/POE
+    type: dict
+    keys:
+      reboot:
+        type: dict
+        description: Set the global level PoE power behavior for PoE ports when the system is rebooted
+        keys:
+          action:
```
gmuloc, line 15: Given that we are implementing defaults in the template - we should add a `default: <value>` in the schema so that it is rendered in the documentation

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,65 @@
...
+{%     if ethernet_interfaces_poe | length > 0 %}
+
+#### PoE Interfaces
+
+{%         if ethernet_interfaces_poe | length > 0 %}
+| Interface | PoE Enabled | Priority | Limit | Reboot Action | Link Down Action | Shutdown Action | LLDP Negotiation | Legacy Detection |
+| --------- | --------- | --------- | ----------- | ----------- | ----------- | ----------- | --------- | --------- |
+{%             for ethernet_interface in ethernet_interfaces_poe | arista.avd.natural_sort('name') %}
+{%                 set poe = ethernet_interface.poe %}
+{%                 set enabled = not poe.disabled | arista.avd.default(true) %}
```
gmuloc: Hmm is it working as expected ?  I think this will give a default of false as it is understood as

```
{%                 set enabled = not (poe.disabled | arista.avd.default(true)) %}
```

But you would rather want (not sure if it works)

```
{%                 set enabled = (not poe.disabled) | arista.avd.default(true) %}
```

as by default poe is enabled right?

cf order of precedence for operators in Jinja

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/ethernet-interfaces.yml
```diff
@@ -635,3 +635,48 @@ ethernet_interfaces:
...
+        class: 4
+        fixed: false
+      negotiation_lldp: false
+
+  Ethernet58:
+    description: interface_with_poe_commands_and_limit_in_class
+    mode: access
+    vlans: 20
+    poe:
+      disabled: true
```
gmuloc: probably worth adding a test where disabled is not configured to test documentation behavior

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/poe.j2
```diff
@@ -0,0 +1,18 @@
...
+{% if print_poe %}
+!
+poe
+{%     if poe.reboot.action is arista.avd.defined('maintain') %}
+   reboot action {{ poe.reboot.action }}
+{%     endif %}
+{%     if poe.interface_shutdown.action is arista.avd.defined('power-off') %}
+   interface shutdown action {{ poe.interface_shutdown.action }}
+{%     endif %}
+{% endif %}
```
gmuloc: This logic should to the same as yours

```suggestion
{% if poe.reboot.action is arista.avd.defined('maintain') or  poe.interface_shutdown.action is arista.avd.defined('power-off') %}
!
poe
{%     if poe.reboot.action is arista.avd.defined('maintain') %}
   reboot action {{ poe.reboot.action }}
{%     endif %}
{%     if poe.interface_shutdown.action is arista.avd.defined('power-off') %}
   interface shutdown action {{ poe.interface_shutdown.action }}
{%     endif %}
{% endif %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
jrecchia1029: You're right.  Good catch! I think we'll need to implement another way still since the `(not poe.disabled)` part of the line will error out when `disabled` is not a defined key

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/poe.schema.yml
jrecchia1029: Would this create an issue when it comes to non-poe switches though?  Basically, would a non-poe-capable switch still have POE EOS CLI generated for it?

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
jrecchia1029: Yes. Good call!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
jrecchia1029: Looks like `set enabled = not (poe.disabled | arista.avd.default(false))` gets the job done

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/poe.schema.yml
jrecchia1029: I take it back. Looks like there's no issue with setting defaults for the poe values.  However, I believe I'm setting these defaults for documentation purposes on line 3 of [ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2](https://github.com/aristanetworks/ansible-avd/pull/2690/files/206c0e7021927e3f4a19f7f4197934d9e0b95fa2#diff-474c47fe975239f18b2e733ee14dd541bd27a6dd719bbe354a519715a4f85ad3).

Not saying this is the best approach but just want to make sure you're aware of it.

gmuloc: Ah yeah so basically in documentation it will states that this is the default from EOS but in the config it does not show up. I did not notice indeed. Maybe in that case add in the description that by default on EOS this is what is there as you did in some other places. Thanks for the comment!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
gmuloc: I agree that this does the job eventhough it is quite confusing  to read (but it is Jinja's fault :) )

jrecchia1029:
> Not sure why the pre-commit validation check fails. The error message doesn't seem to align with what I see in the file where the issue is detected.

@ClausHolbechArista @gmuloc any idea what the issue could be? All of the pre-commit checks pass in my local environment.

ansible_collections/arista/avd/molecule/eos_cli_config_gen_v4.0/documentation/devices/ethernet-interfaces.md
```diff
@@ -0,0 +1,519 @@
+# ethernet-interfaces
```
gmuloc: Hmmm eos_cli_config_v4.0 should be gone .. I will take care of that

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,65 @@
...
+{%         set poe_global_defaults.reboot_action = poe.reboot.action | arista.avd.default(poe_global_defaults.reboot_action) %}
+{%         set poe_global_defaults.interface_shutdown_action = poe.interface_shutdown.action | arista.avd.default(poe_global_defaults.interface_shutdown_action) %}
+{#         get lldp transmit power via mdi #}
+{%         if lldp.tlvs is arista.avd.defined %}
+{%             for tlv in lldp.tlvs if tlv.name == 'power-via-mdi' %}
+{%                 set poe_global_defaults.lldp_negotiate = tlv.transmit %}
+{%             endfor %}
+{%         endif %}
+| {{ poe_global_defaults.reboot_action }} | {{ poe_global_defaults.interface_shutdown_action }} | {{ poe_global_defaults.lldp_negotiate }} |
+{%     endif %}
```
gmuloc, line 31: I am now wondering if we should not render this all the time given that your are providing default values for the three categories and in that case I would change line 3 to be

```
{% set poe_global_defaults = namespace(reboot_action="power-off (default)", interface_shutdown_action="maintain (default)", lldp_negotiate="True (default)") %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,65 @@
...
+{% set ethernet_interfaces_poe = [] %}
+{% for ethernet_interface in ethernet_interfaces | arista.avd.natural_sort('name') %}
+{%     if ethernet_interface.poe is arista.avd.defined %}
+{%         do ethernet_interfaces_poe.append(ethernet_interface) %}
+{%     endif %}
+{% endfor %}
+{% if (poe is arista.avd.defined) or (ethernet_interfaces_poe | length > 0) %}
+## Power Over Ethernet (PoE)
+
+### PoE Summary
```
gmuloc, line 14: Looking at the documentation again we should add a  `### PoE Configuration` as for LLDP - so essentially rendering the global poe template (probably not worth rendering the interface level ones and so the category could be PoE Global Configuration rendered only if `poe` is defined)

```
    ### LLDP Device Configuration

    ```eos
    {%     include 'eos/lldp.j2' %}
    ```
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
jrecchia1029: Yea that makes total sense.  Good thinking.

jrecchia1029: But actually, how do we account for non-poe platforms now?  Unless, there's a way to check if a particular device rendering cli is poe-capable, I think it's best to not include the POE global documentation.

gmuloc: you also make a good point - actually I agree with you and lets include it only when there is actually something to configure. Thanks for your latest change.
@ClausHolbechArista what do you think of the current status?

ClausHolbechArista: In `eos_cli_config_gen` we try to avoid documenting EOS defaults. Otherwise we would have a lot of "stale" documentation. I suggest to only document what the user has defined. Never interpret EOS defaults at all, unless it is absolutely required for either valid configuration or documentation clarity. We normally just set `-` as the value, if it is unset by the user.

jrecchia1029: In that case, the only default I think that should be defined is poe `enabled` or `disabled`

gmuloc (approved):
LGTM - thanks @jrecchia1029 for the latest changes!

carlbuchmann (approved):
LGTM, great job @jrecchia1029

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,72 @@
...
+{% set POE_CLASS_MAP = {0: "15.40", 1: "4.00", 2: "7.00", 3: "15.40", 4: "30.00", 5: "45.00", 6: "60.00", 7: "75.00", 8: "90.00"} %}
+{% set poe_global_defaults = namespace(reboot_action="-", interface_shutdown_action="-", lldp_negotiate="-") %}
+{% set ethernet_interfaces_poe = [] %}
+{% for ethernet_interface in ethernet_interfaces | arista.avd.natural_sort('name') %}
+{%     if ethernet_interface.poe is arista.avd.defined %}
+{%         do ethernet_interfaces_poe.append(ethernet_interface) %}
+{%     endif %}
+{% endfor %}
+{% if (poe is arista.avd.defined) or (ethernet_interfaces_poe | length > 0) %}
+## Power Over Ethernet (PoE)
```
ClausHolbechArista, line 12:
```suggestion

## Power Over Ethernet (PoE)
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,72 @@
...
+{%                 set poe_global_defaults.lldp_negotiate = tlv.transmit %}
+{%             endfor %}
+{%         endif %}
+| {{ poe_global_defaults.reboot_action }} | {{ poe_global_defaults.interface_shutdown_action }} | {{ poe_global_defaults.lldp_negotiate }} |
+{%     endif %}
+{%     if ethernet_interfaces_poe | length > 0 %}
+
+#### PoE Interfaces
+
+{%         if ethernet_interfaces_poe | length > 0 %}
```
ClausHolbechArista: Remove this `if`. It is the same as in line 31. Of course there is more to it than removing this single line :) Deindentation of the block below and removing the endif
```suggestion
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/poe.j2
```diff
@@ -0,0 +1,72 @@
...
+{%                 set shutdown_action = poe.shutdown.action | arista.avd.default(poe_global_defaults.interface_shutdown_action) %}
+{%                 set negotiation_lldp = poe.negotiation_lldp | arista.avd.default(poe_global_defaults.lldp_negotiate) %}
+{%                 set legacy_detect = poe.legacy_detect | arista.avd.default("-") %}
+| {{ ethernet_interface.name }} | {{ enabled }} | {{ priority }} | {{ limit }} | {{ reboot_action }} | {{ link_down_action }} | {{ shutdown_action }} | {{ negotiation_lldp }} | {{ legacy_detect }} |
+{%             endfor %}
+{%         endif %}
+{%     endif %}
+{% endif %}
+{% if poe is arista.avd.defined %}
+### PoE Device Configuration
```
ClausHolbechArista, line 67:
```suggestion

### PoE Device Configuration
```

ClausHolbechArista (changes requested):

ClausHolbechArista (approved):

PR #2710, opened 2023-04-07
Fix(eos_designs): Allow to return a single host subnet when IPv4/32 or IPv6/128

### LINKED ISSUE #2705: Fix _ip in AvdIpAddressing to be able to allocate one /32 out of a /32 (or /128 out of a /128)
### Issue Summary

```python
import ipaddress

def _ip(pool: str, prefixlen: int, subnet_offset: int, ip_offset: int) -> str:
    print(pool_network, prefixlen_diff, subnet_size)
    print(subnet_offset +1)
    print(subnet_size)
    print(pool_network.num_addresses)

        raise Exception(f"Unable to get {subnet_offset + 1} /{prefixlen} subnets from pool {pool}")
    subnet = ipaddress.ip_network((int(pool_network.network_address) + subnet_offset * subnet_size, prefixlen))
    try:
        ip = subnet[ip_offset]
    except IndexError as e:
        raise Exception(f"Unable to get {ip_offset+1} hosts in subnet {subnet} taken from pool {pool}") from e
    return str(ip)

if __name__ == "__main__":
    _ip("10.10.0.1/32", 32, 1, 0)
```

This will fail because the `_ip` function skips the "network address" 

_No response_

### How do you run AVD ?

None

### Steps to reproduce

_No response_

### Relevant log output

_No response_

### Contributing Guide

gmuloc:
**CAUTION:** The code makes offset and subnet offset not being used. Probably should document this somewhere?

## Related Issue(s)

Fixes #2705

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

When the pool subnet size is a /32 and the required subnet is a /32 for IPv4, the current code is failing as it skips the `netwok_address` in the pool which is the only available. (same for /128 for IPv6). This code allows to not skip the network address in this scenario and returns the only available /32 or /128.

## How to test

molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/loopback_pool_one_host.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/loopback_pool_one_host.yml
@@ -6,6 +6,6 @@ l3leaf:
     loopback_pool_one_host:
       id: 1
       bgp_as: 65000
-      loopback_ipv4_pool: 42.42.42.42/32
+      loopback_ipv4_pool: 192.168.42.42/32
       vtep_loopback_ipv4_pool: 1.1.1.1/32
-      loopback_ipv6_pool: dead:beef::cafe/128
+      loopback_ipv6_pool: 2001:db8::cafe/128
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/loopback_pool_one_host.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/loopback_pool_one_host.yml
@@ -1,0 +1,11 @@
+# Minimum config to only test /32 IPv4 and /128 IPv6 loopbacks pool
+type: l3leaf
+
+l3leaf:
+  nodes:
+    loopback_pool_one_host:
+      id: 1
+      bgp_as: 65000
+      loopback_ipv4_pool: 42.42.42.42/32
+      vtep_loopback_ipv4_pool: 1.1.1.1/32
+      loopback_ipv6_pool: dead:beef::cafe/128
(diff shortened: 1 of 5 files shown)


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/loopback_pool_one_host.yml
```diff
@@ -0,0 +1,11 @@
+# Minimum config to only test /32 IPv4 and /128 IPv6 loopbacks pool
+type: l3leaf
+
+l3leaf:
+  nodes:
+    loopback_pool_one_host:
+      id: 1
+      bgp_as: 65000
+      loopback_ipv4_pool: 42.42.42.42/32
```
ClausHolbechArista: please use RFC1918 addresses.

ClausHolbechArista (commented):
Both subnet_offset and ip_offset are 0-based. So you are asking for the second /32 in the pool. When I call with 0,0 it works. So if we need to change something, I think we should modify the calling function somehow.

gmuloc:
> Both subnet_offset and ip_offset are 0-based. So you are asking for the second /32 in the pool. When I call with 0,0 it works. So if we need to change something, I think we should modify the calling function somehow.

so in this case it means modifying all the calling functions, in which we ignore the offset then

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -249,5 +264,9 @@ def vtep_ip(self) -> str:
                 loopback_ipv4_offset=self._loopback_ipv4_offset,
             )
 
-        offset = self._id + self._loopback_ipv4_offset
+        if self._is_host_subnet(self._vtep_loopback_ipv4_pool):
```
ClausHolbechArista, line 267: Since this is behavioral change, we should update the description on the pools covered by this change.
You could just have entered `-1` as the offset value.

gmuloc: not sure what you mean about `-1` it would not work right if id is 10 ? the value has to be 0 for it to work as you pointed out

ClausHolbechArista: Right, but what I fail to understand is why do we want to break the rules of the pool? I think the right solution is to implement new knobs to allow specific IPs everywhere where we have pools. I have the same need on uplinks, so I can implement if you like.

gmuloc:
Closed in favor of https://github.com/aristanetworks/ansible-avd/issues/2719

PR #2561, opened 2023-02-22
Feat(eos_config_deploy_cvp): Add support for device_inventory_mode

ClausHolbechArista:
## Change Summary

Add support for device_inventory_mode
## Related Issue(s)

Fixes #2380 

## Component(s) name

`arista.avd.eos_config_deploy_cvp`

## Proposed changes

Adding new options:
- **`device_inventory_mode`**: `strict` / `loose`. Define how missing devices are handled. "loose" will ignore missing devices. "strict" will fail on any missing device. Default: `strict`. **NOTE**: This option requires `arista.cvp` collection version >= 3.8.0.

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #2718, opened 2023-04-11
Feat(eos_config_deploy_cvp): Option for deploying using serial number as identifier

### LINKED ISSUE #2472: Feature request - Add support for SN based host lookup in CVP
### Enhancement summary

Under the hood, eos_config_deploy_cvp uses modules like [cv_device_v3](https://cvp.avd.sh/en/stable/docs/modules/cv_device_v3.rst) which support a search key option to locate devices in CVP inventory by hostname, FQDN, or serial number (default is hostname). There is no way today to tell eos_config_deploy_cvp to lookup devices by serial number instead of hostname, so this feature request would be to

1. Add a knob to eos_config_deploy_cvp to change search key
2. Add a serial number key to either hosts in inventory (similar to the is_deployed key) or eos_design to be used for this lookup 

eos_designs, eos_config_deploy_cvp

### Use case example

Customer wants to rename devices in fabric and has already deployed them in production - changing hostnames in Ansible inventory and re-running AVD pipeline will result in it not matching devices in CVP, because Ansible will now have the new hostnames for nodes, and CVP will have the old ones; lookup by hostname fails. To make this work today, hostnames have to be changed on the devices either by CLI or Ansible playbook (either to eAPI directly or via ansible-cvp) so that CVP picks up the new hostnames, and then the AVD pipeline can be ran.

Implementing this change allows for a pure-AVD way to update device hostnames

1. Add a knob to eos_config_deploy_cvp to pass-through a search-key selection to cv_device_v3 and similar modules in ansible-cvp.
2. Add a field to map a serial number to a device in Ansible inventory - multiple locations where this could work.

See non-AVD playbook workflow described in use case above.

### Additional context

_No response_

### Contributing Guide

ClausHolbechArista:
## Change Summary

Option for deploying using serial number as identifier

## Related Issue(s)

Fixes #2472

## Component(s) name

`arista.avd.eos config deploy cvp`

## Proposed changes

- Add passthrough of more options to `arista.cvp.v3_*` modules - cherrypicked from #2561
- Implement improvements in `eos_config_deploy_cvp` and document an example.

## How to test

Tested locally with and without serial number using internal testbed.

### User Checklist

- N/A

### Repository Checklist


noredistribution (commented):
LGTM! tested on CVP 2023.1.0 and CVaaS as well

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/README.md
```diff
@@ -133,6 +136,51 @@ Here is an overview with the key configured in the YAML inventory:
...
+To use serial number as the key, each device must have a hostvar `serial_number` set. This can be set as hostvars in the inventory like:
+
+```yaml
+  DC1_BL1:
+    hosts:
+      DC1-BL1A:
+        serial_number: ABCDEF0123456789
+```
+
+If using `arista.avd.eos_designs`, it is also possible to set `serial_number` under the fabric topology definitions like `l3leaf`.
```
gmuloc: Probably need to reword this as node_type or maybe add an example?

gmuloc (approved):
LGTM from doc point of view - feature has been tested by @noredistribution and @ClausHolbechArista so not giving it another run.

PR #2720, opened 2023-04-11
CI: Add first draft for schema molecule coverage

gmuloc:
## Change Summary

Add a pre-commit hook to compute schema coverage of molecule tests

## Component(s) name

`ci`

## Proposed changes

* remaining work:
    * eos_designs
    * document code

## How to test

run molecule

### Repository Checklist

### DIFF
--- a/.pre_commit_scripts/current_coverage.yml
+++ b/.pre_commit_scripts/current_coverage.yml
@@ -1,2 +1,2 @@
 eos_cli_config_gen: 91.88761377127028
-eos_designs: 9.393939393939394
+eos_designs: 6.548100242522231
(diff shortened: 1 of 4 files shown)


.pre_commit_scripts/current_coverage.yml
ClausHolbechArista, line 1: Adding this as a file is not very friendly for contributors not running pre-commit locally. They can do most other things manually in some way or another, but this one requires pre-commit to run. So every time you change a molecule scenario you have to rerun this.
Maybe we can run this in CI on `devel` first, before checking out PR branch, and then run it again, submitting a comment on the PR with the before and after.

gmuloc: they can run it using the python script as well - if we have to we can make a makefile entry .. :) 
waiting to push to get feedback is not really better

gmuloc: this is the exact same as for the schemas hook actually

ClausHolbechArista:
This will be covered (ha) by pytest "coverage" over time, so I think we can skip this.

PR #2745, opened 2023-04-19
Fix: Description key not considered with connected endpoints

### LINKED ISSUE #2744: Description key not considered with connected endpoints
### Issue Summary

The documentation shows:
```
        # Interface descriptions Description | Optional
        description: <description>
```

however it is not honored, the endpoint name is used.
https://avd.sh/en/stable/roles/eos_designs/doc/connected-endpoints.html

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
Molecule test added in eos_designs_unit_tests
```

### Relevant log output

_No response_

### Contributing Guide

pvinci-arista:
## Change Summary

description key under connected endpoints not used.

## Related Issue(s)

Fixes #2744

## Component(s) name

`arista.avd.eos_designs`

## How to test

molecule test

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/ethernet_interfaces.py
```diff
@@ -75,7 +75,7 @@ def _get_ethernet_interface_cfg(self, adapter: dict, node_index: int, connected_
             "peer_interface": peer_interface,
             "peer_type": connected_endpoint["type"],
             "port_profile": adapter.get("profile"),
-            "description": self.shared_utils.interface_descriptions.connected_endpoints_ethernet_interfaces(peer, peer_interface),
+            "description": adapter.get("description") or self.shared_utils.interface_descriptions.connected_endpoints_ethernet_interfaces(peer, peer_interface),
```
gmuloc:
```suggestion
            "description": default(adapter.get("description"), self.shared_utils.interface_descriptions.connected_endpoints_ethernet_interfaces(peer, peer_interface)),
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/connected_endpoints.yml
```diff
@@ -0,0 +1,35 @@
...
+servers:
+  OLD_SW-1/4:
+    adapters:
+      - switches: [connected_endpoints]
+        switch_ports: [Ethernet4]
+        profile: desktop_phone-1234-2345
+        description: VVxx:abcd123
+        native_vlan: 2020
+        raw_eos_cli: |-
+          !! VoIP/PC
```
gmuloc: can you limit even further the test?

gmuloc (commented):
Great catch - could you also please implement the same for Port-channels?

ClausHolbechArista:
It was implemented as part of network_ports, where we want a more generic description, but makes perfect sense to include here as well. I would prefer to move the logic to the description module, so anyone customizing those keep full control. Similar to how it is done with port_channel_description.

gmuloc:
> It was implemented as part of network_ports, where we want a more generic description, but makes perfect sense to include here as well. I would prefer to move the logic to the description module, so anyone customizing those keep full control. Similar to how it is done with port_channel_description.

On this topic @ClausHolbechArista - should port-channel description also take into account this special description at the adapter level as currently it is doing `<adapter_description>_<oprtional_port_channel_description>`

ClausHolbechArista:
> > It was implemented as part of network_ports, where we want a more generic description, but makes perfect sense to include here as well. I would prefer to move the logic to the description module, so anyone customizing those keep full control. Similar to how it is done with port_channel_description.
> 
> On this topic @ClausHolbechArista - should port-channel description also take into account this special description at the adapter level as currently it is doing `<adapter_description>_<oprtional_port_channel_description>`

That would be good

pvinci-arista:
> I would prefer to move the logic to the description module, so anyone customizing those keep full control. Similar to how it is done with port_channel_description.

If I follow, won't we need to modify the method signature in the base class and all the subclasses?

gmuloc:
Hi @pvinci-arista so basically the suggestion is to add as an optional keyword argument the adapter.description key so that you don't break possible subclasses (though for now nobody has been using this functionality as it has not been officially released anyway).

Let's take the opportunity in this PR to also document better how the descriptions are built!

Let me know if you need some help

gmuloc (approved):

carlbuchmann (changes requested):
The fix looks good - As mentioned in one of the comments, we should add an explanation of how the descriptions are built for physical ports and port-channels.

pvinci-arista:
> we should add an explanation of how the descriptions are built for physical ports and port-channels.

Docstrings were added to the methods.  Do you want something additional?  See, as an example: https://github.com/aristanetworks/ansible-avd/pull/2745/files#diff-bbfc03125089c9310eb2bebb44079e7849bbf1e5031bae6df3024d280c0f306dR94

carlbuchmann:
> > we should add an explanation of how the descriptions are built for physical ports and port-channels.
> 
> Docstrings were added to the methods. Do you want something additional? See, as an example: https://github.com/aristanetworks/ansible-avd/pull/2745/files#diff-bbfc03125089c9310eb2bebb44079e7849bbf1e5031bae6df3024d280c0f306dR94

We have added addition information in schema/documentation.

carlbuchmann (approved):
LGTM

PR #2758, opened 2023-04-24
Fix(eos_designs): connected endpoints interface mode valid values

carlbuchmann:
## Change Summary

Added `trunk phone` to valid values for adapter_config.mode


pvinci-arista (commented):
LGTM

gmuloc (commented):
@carlbuchmann do we need a test to catch this in `eos_designs_unit_tests` - sounds like we did not get any error about it?

gmuloc (approved):

Issue #2311, opened 2022-12-01
Feat(eos_designs): Enable sFlow at fabric level


xaviramon:
### Enhancement summary

We would like a fabric-wise command the enables sflow on all fabric interfaces.

We would like a similar variable like PTP:
```yaml
# Enable PTP on all infrastructure links | Optional
uplink_ptp:
  enable: < boolean | default -> false >
  ```
But allowing sflow variables conveniently.

### Which component of AVD is impacted

eos_designs

### Use case example

Allow sflow monitoring at a fabric level reducing the amount of interfaces involved in sflow and the number of packets sent to the collector.

### Describe the solution you would like

Control of basic sflow parameters at a fabric level, without including host-facing interfaces.

### Describe alternatives you have considered

using eos_cli variable, which is not autocontained.

### Contributing Guide

- [X] I agree to follow this project's Code of Conduct

PR #2767, opened 2023-04-28
Feat(eos_cli_config_gen): Generate sFlow egress commands

xaviramon:
## Change Summary

Knob to enable unmodified egress sflow on the CLI.

## Related Issue(s)

To be used in #2311 

## Component(s) name

## Proposed changes
Add a new keys to the data model:
```
sflow:
  interface:
    egress:
      unmodified_enable_default: <bool>
      enable_default: <bool>
```
## How to test
Molecule scenario added

CLI testing for command:
```
ld005(config)#sh run | i sflow
sflow interface disable default
sflow interface egress unmodified enable default
sflow run
```
This is a platform-dependent command, but adding no check following the component philosophy.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
@@ -102,14 +102,10 @@ keys:
           egress:
             type: dict
             keys:
-              unmodified:
-                type: dict
-                keys:
-                  enable:
-                    type: dict
-                    keys:
-                      default:
-                        type: bool
+              unmodified_enable_default:
+                type: bool
+              enable_default:
+                type: bool
       run:
         type: bool
       hardware_acceleration:
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
@@ -102,15 +102,14 @@ keys:
           egress:
             type: dict
             keys:
-              unmodified_enable_default:
-                type: bool
-                description: |
-                  Enable unmodified egress sFlow by default.
-                  Platform dependent feature.
               enable_default:
                 type: bool
                 description: |
                   Enable egress sFlow by default.
+              unmodified:
+                type: bool
+                description: |
+                  Enable egress sFlow unmodified.
                   Platform dependent feature.
       run:
         type: bool
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
@@ -99,6 +99,17 @@ keys:
             keys:
               default:
                 type: bool
+          egress:
+            type: dict
+            keys:
+              unmodified:
+                type: dict
+                keys:
+                  enable:
+                    type: dict
+                    keys:
+                      default:
+                        type: bool
       run:
         type: bool
       hardware_acceleration:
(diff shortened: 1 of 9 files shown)


gmuloc (commented):
Very open comments here!

For reference:

```
Leaf1(config)#show cli commands | grep "sflow interface"
[no|default] sflow interface disable default
[no|default] sflow interface egress enable default
[no|default] sflow interface egress unmodified enable default
```

Behavior on EOS 4.27.3F

```
Leaf1(config)#default sflow interface egress enable default
Leaf1(config)#show run | i sflow
<EMPTY>
Leaf1(config)#sflow interface egress enable default
Leaf1(config)#show run | i sflow
sflow interface egress unmodified enable default

Leaf1(config)#default sflow interface egress enable default
Leaf1(config)#sflow interface egress un  enable default
Leaf1(config)#show run | i sflow
sflow interface egress unmodified enable default
```

So not sure if there is a difference between enable_default and unmodified_default

Wondering if this could impact the model?

Could this make sense? (though seems like always enable_default is needed)
```
          egress:
            type: dict
            keys:
              unmodified:
                type: bool
              enable_default:
                type: bool
```

xaviramon:
Egress sFlow is a platform dependent feature and for this reason and to distinguish different behaviours, the commands used to activate it are different. 

For SAND platforms sflow will sample all defined packets:
```
lp032(config)#sflow interface egress ?
  enable  Set default for egress sFlow to Enabled on interfaces

lp032(config)#sflow interface egress un
  unmodified  not supported on this hardware platform

lp032#sh ver
Arista DCS-7280QR-C36-F
```

For STRATA platforms sflow will sample defined packets ingress of an interface wich are exiting a defined interface, for this the keywork "unmodified" is used:
```
ld003(config)#sflow interface egress ?
  unmodified  Unmodified egress sFlow

ld003(config)#sflow interface egress en
  enable  not supported on this hardware platform

ld003(config)#sflow interface egress
ld003#sh ver
Arista DCS-7050CX3-32S-F
```

I am also proposing a new data model:
```
sflow:
  interface:
    egress:
      unmodified_enable_default: <bool>
      enable_default: <bool>
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/sflow.schema.yml
```diff
@@ -99,6 +99,13 @@ keys:
             keys:
               default:
                 type: bool
+          egress:
+            type: dict
+            keys:
+              unmodified_enable_default:
+                type: bool
+              enable_default:
+                type: bool
```
gmuloc: Let's document the platform dependency of each key in descriptions

ClausHolbechArista: Just say that they are platform dependent like we do elsewhere. Don't describe which platforms, since that will change over time.

gmuloc (commented):
so the last thing I have as a comment is that it is possible to render both on the same device which is not valid.

@ClausHolbechArista Wondering if we should for now in the template enforce that only one should be set?


Otherwise the PR content is good for me and I will approve it

ClausHolbechArista:
> so the last thing I have as a comment is that it is possible to render both on the same device which is not valid.
> 
> @ClausHolbechArista Wondering if we should for now in the template enforce that only one should be set?
> 
> Otherwise the PR content is good for me and I will approve it

We can implement them as mutually exclusive (use `elif`) instead, and then document that they are mutually exclusive and the correct one depends on the platform.

ClausHolbechArista:
> > so the last thing I have as a comment is that it is possible to render both on the same device which is not valid.
> > @ClausHolbechArista Wondering if we should for now in the template enforce that only one should be set?
> > Otherwise the PR content is good for me and I will approve it
> 
> We can implement them as mutually exclusive (use `elif`) instead, and then document that they are mutually exclusive and the correct one depends on the platform.

With that in mind, actually your suggestion @gmuloc of having `unmodified: <bool>` would be more solid.

ClausHolbechArista (approved):

haylinmoore (approved):
LGTM

PR #2770, opened 2023-04-28
Feat(eos_cli_config_gen): Add ECN Support

chetryan:
## Change Summary

Added support for ECN configuration in uc-tx-queues and tx-queues is qos profiles.
Added support for ECN count command on the ethernet interfaces. This is required on some platforms where the count command on the qos profiles does not apply to the interfaces.

## Related Issue(s)

No issues raised.
- This PR will add to mainline some features that are in use by customers right now by means of custom jinja templates

## Component(s) name

## Proposed changes

- Update the `eos_cli_config_gen` schema to support `random-detect ecn` for tx-queues and uc-tx-queues
- Update template to generate the configuration for tx-queues and uc-tx-queues
- Update templates to allow the use of tx-queue and uc-tx-queue commands within the ethernet port

## How to test

- Run the molecule scenario `eos_cli_config_gen`. The `qos.cfg` should include the commands to configure ecn on tx and uc-tx queues as well as an example setting `ecn count` on an ethernet interface.
- 
### User Checklist

- Verify that the generated config does not have typos
- Verify that the way the schema is defined makes sense for someone using ECN for the first time
- Verify that the Documentation is clear enough.

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
```diff
@@ -325,6 +340,11 @@ QOS Profile: **no_qos_trust**
 | ----------- | ------------ | ----- | ---------- | ------------------ |
 | 3 | 4 | disabled | - | - |
 
+**ECN Configuration**
```
ClausHolbechArista: Please add some condition to avoid printing the empty table.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -78,6 +78,39 @@ keys:
...
+              random_detect:
+                type: dict
+                keys:
+                  ecn:
+                    type: dict
+                    keys:
+                      count:
+                        type: bool
+                        description: |
+                          Turn on random-detect ecn count
```
ClausHolbechArista:
```suggestion
                          Enable counter for random-detect ECNs
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -78,6 +78,39 @@ keys:
                     description: |
                       Supported options are platform dependent
                       Example: "< rate > kbps", "1-100 percent", "< rate > pps"
+              random_detect:
+                type: dict
+                keys:
+                  ecn:
```
ClausHolbechArista, line 86:
```suggestion
                  ecn:
                    description: Explicit Congestion Notification
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -78,6 +78,39 @@ keys:
...
+                      threshold:
+                        type: dict
+                        keys:
+                          units:
+                            type: str
+                            valid_values: ["segments", "bytes", "kbytes", "mbytes"]
+                          min:
+                            type: int
+                            description: |
+                              Set the random-detect ecn minimum-threshold
```
ClausHolbechArista: No need to repeat the key names etc so just try to help the user fill in expected values. Like for various timers we say "timeout in seconds" etc.
Same for the other keys in this area.

```suggestion
                              Minimum threshold in ???unit???
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -78,6 +78,39 @@ keys:
...
+                        type: bool
+                        description: |
+                          Turn on random-detect ecn count
+                      threshold:
+                        type: dict
+                        keys:
+                          units:
+                            type: str
+                            valid_values: ["segments", "bytes", "kbytes", "mbytes"]
+                          min:
```
ClausHolbechArista, line 103: Please look into improving the schema with `min` and `max` values if we know them for all keys in this area.
Also to catch some corner cases with YAML input and inline jinja, please add
```yaml
convert_types:
  - str
```
under all `type: int` keys.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
```diff
@@ -61,6 +61,47 @@ QOS Profile: **{{ profile.name }}**
 {%                 endfor %}
 {%             endif %}
 {%         endif %}
+
```
ClausHolbechArista, line 84: So _above_ the blank line, you want to insert some logic that parse the queues looking for ecn key. Only if you find any, you include the table.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
```diff
@@ -61,6 +61,47 @@ QOS Profile: **{{ profile.name }}**
...
+{%                 set max = tx_queue.random_detect.ecn.threshold.max | arista.avd.default("-") %}
+{%                 set prob = tx_queue.random_detect.ecn.threshold.max_probability | arista.avd.default("-") %}
+{%                 set units = tx_queue.random_detect.ecn.threshold.units | arista.avd.default("") %}
+| {{ tx_queue.id }} | {{ type }} | {{ ecn }} | {{ min }}  {{ units }} | {{ max }} {{ units }} | {{ prob }} |
+{%             endfor %}
+{%         endif %}
+{%         if profile.uc_tx_queues is arista.avd.defined %}
+{%             for uc_tx_queue in profile.uc_tx_queues | arista.avd.natural_sort('id') %}
+{%                 set type = "Unicast" %}
+{%                 if uc_tx_queue.random_detect.ecn.threshold is arista.avd.defined %}
```
ClausHolbechArista: Detecting on the presence of a dict, is _not_ a good way to detect if it is enabled or not. If it is possible to just configure `random-detect ecn` with nothing else set, we should add an `enabled: <bool>` key under the dict and detect on that here.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tx-queues.j2
```diff
@@ -0,0 +1,43 @@
...
+{%         else %}
+{%             set max_probability = "" %}
+{%         endif %}
+{%         if tx_queue.random_detect.ecn.threshold.weight is arista.avd.defined %}
+{%             set ecn_weight = " weight " ~ tx_queue.random_detect.ecn.threshold.weight %}
+{%         else %}
+{%             set ecn_weight = "" %}
+{%         endif %}
+      random-detect ecn{{ min_threshold }}{{ max_threshold }}{{ max_probability }}{{ ecn_weight }}
+{%     endif %}
```
ClausHolbechArista: First see my connect on using the presence of the dict alone under the doc template.
Next, please look at other templates, on how we add CLI tokens bit by bit to _one_ variable instead of creating multiple like here.
Ex. (please adjust of course)
```j2
{% if feature.enabled is arista.avd.defined(true) %}
{%     set feature_cli = "basic command" %} <-- notice not space after the command
{%     if feature.knob1 is arista.avd.defined %}
{%         set feature_cli = feature_cli ~ " knob1 " ~ feature.knob1 %} <-- notice the leading space before the keyword, and no trailing space.
{%     endif %}
... more knobs ...
      {{ feature_cli }} <--- Just printing the CLI var which may or may not contain extra tokens.
{% endif %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/qos-profiles.j2
```diff
@@ -22,34 +22,10 @@ qos profile {{ profile.name }}
...
-{%         elif tx_queue.bandwidth_guaranteed_percent is arista.avd.defined %}
-      bandwidth guaranteed percent {{ tx_queue.bandwidth_guaranteed_percent }}
-{%         endif %}
-{%         if tx_queue.priority is arista.avd.defined %}
-      {{ tx_queue.priority }}
-{%         endif %}
-{%         if tx_queue.shape.rate is arista.avd.defined %}
-      shape rate {{ tx_queue.shape.rate }}
-{%         endif %}
+{%         include 'eos/tx-queues.j2' %}
```
ClausHolbechArista: Adding these extra sub-sub templates to enable a bit of reuse will make it more complex to understand the templates later. @carlbuchmann @gmuloc where do we set the bar here? All the other features under the qos profiles could also be shared, and same for many other areas of eos_cli_config_gen.
IMO we should stay conservative on this.

No matter what we need to clearly name the files so we can identify the hierarchy. Maybe these should be `ethernet-interface-tx-queues.j2`

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
ClausHolbechArista: After looking through the templates, I know understand that this is your own key - not part of the CLI as such. Is is a requirement that all values are using the same unit or would some users want different units?

chetryan: The EOS command accepts an integer value and a unit for each of minimum-threshold and maximum-threshold. There are multiple units supported,  but I could not find any documentation whether the same command (for the same queue) can have mixed units. So here I am assuming that all values for 1 queue will be using the same unit.

chetryan: I tested on a `DCS-7050DX4-32S-F` and the units used in the same command/line need to match. So I believe we have the right setup here, where only 1 unit can be specified for a single queue.

ClausHolbechArista: OK, thank you for testing and confirming. Then just make the description `Minimum threshold` etc. and we keep the units like you have it.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
ClausHolbechArista: So because the units can be different, we cannot set a min/max. But still add the convert_types.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/qos-profiles.j2
chetryan: Happy to rename these, but wouldn't that mean that in the `qos-profiles.j2` we would be making reference to the `ethernet-interface-tx-queues.j2` ? I named it `tx-queues.j2` to make it independent of whether it was in the profile or the ethernet level.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
chetryan: Added logic to skip any ECN Configuration tables if there are no tx queues or if none of them have any ECN config.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
chetryan: I set the max to the absolute maximum if the smallest unit is used with the platform that has the largest buffers. Shall I keep that or remove it altogether ? 
We know the min is 1, so I can leave it there.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
chetryan: Added logic here to check if queues have any ecn thresholds defined. If there are none found, the generation of the table is skipped.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
chetryan: After going thought the EOS documentation and testing on `DCS-7050DX4-32S-F` I found that `random-detect ecn` on it's own is not a valid command. I am thus marking 'min', 'max' and 'units' as required in the schema, and testing on the presence of `units` instead of using an added `enabled` bool.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/tx-queues.j2
chetryan: Thank you for pointing this out! I agree it is more elegant. Refactored both `tx-queues.j2` and `uc-tx-queues.j2`

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
ClausHolbechArista: Maybe it will be better to remove the limits since it might be misleading if someone thinks they can use the max no matter which unit.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/qos-profiles.j2
ClausHolbechArista: I understand your concern on naming, but I would rather have qos-profiles use the ethernet-interface... variant than confusing "tx-queues" with some global key. In general I would prefer not to have this reuse at all, but I understand why you do it.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -78,6 +78,64 @@ keys:
...
+                              Set the random-detect ecn maximum-threshold
+                          max_probability:
+                            type: int
+                            min: 1
+                            max: 100
+                            convert_types:
+                              - str
+                            description: |
+                              Set the random-detect ecn max-mark-probability.
+                              The value should be between 1 and 100.
```
ClausHolbechArista: Avoid the "set the ...". Also try to write a description that is more clear text and helping the user if the key itself is not self-explanatory.
If the description is =~ the key, just remove the description.
Also the min/max are written in the docs, so you don't have to repeat in the description.
```suggestion
                              Random-detect ECN maximum mark probability.
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
```diff
@@ -61,6 +61,52 @@ QOS Profile: **{{ profile.name }}**
 {%                 endfor %}
 {%             endif %}
 {%         endif %}
+
```
ClausHolbechArista: Move this blank line below the `if ns.ecn_table` to avoid double blank lines (check the molecule outputs before and after).

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos.j2
```diff
@@ -10,6 +10,10 @@ QOS rewrite DSCP: **enabled**
 {%     else %}
 QOS rewrite DSCP: **disabled**
 {%     endif %}
+{%     if qos.random_detect.ecn.allow is arista.avd.defined %}
+
```
ClausHolbechArista, line 24: This file is a mess with blank lines (not your fault :))
Remove the blank in line 7, add one before each of the two rewrite. That way we avoid a double blank.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -683,6 +683,16 @@ interface {{ ethernet_interface.name }}
 {%     if ethernet_interface.traffic_policy.output is arista.avd.defined %}
    traffic-policy output {{ ethernet_interface.traffic_policy.output }}
 {%     endif %}
+{%     if ethernet_interface.tx_queues is arista.avd.defined %}
```
ClausHolbechArista: you can remove this outer `if` since `arista.avd.natural_sort` handles it gracefully and just returns an empty list.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
chetryan: Removed the max limit but left the minimum.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/qos-profiles.j2
chetryan: Understood, renames these templates in the latest commit.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
chetryan: That's good to know - thank you! Removed the outer if.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -111,6 +165,60 @@ keys:
...
+                      threshold:
+                        type: dict
+                        keys:
+                          units:
+                            required: true
+                            type: str
+                            valid_values: ["segments", "bytes", "kbytes", "mbytes"]
+                            description: |
+                              Units to be used for the threshold values.
+                              This should be one of segments, byte, kbytes, mbytes.
```
ClausHolbechArista: Avoid repeating the valid values.
Also should this be singular `unit` since it we set only one "unit" to be used on the other values?

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -111,6 +165,60 @@ keys:
...
+                          min:
+                            type: int
+                            required: true
+                            min: 1
+                            convert_types:
+                              - str
+                            description: |
+                              Random-detect ecn minimum-threshold
```
ClausHolbechArista: Use upper case ECN in all descriptions

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos.schema.yml
```diff
@@ -31,3 +31,20 @@ keys:
...
+          Global random-detect settings
+        keys:
+          ecn:
+            type: dict
+            description:
+              Global ECN Configuration
+            keys:
+              allow:
+                type: str
+                valid_values: ["non-ect", "non-ect chip-based"]
```
ClausHolbechArista: It seems wrong to have a str value that always have to contain "non-ect".
Maybe change to dict with booleans if there are no other values?
```yaml
keys:
  allow_non_ect:
    type: dict
      keys:
        enabled:
          type: <bool>
        chip_based:
          type: <bool>
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
chetryan: Agreed.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
chetryan: Agreed

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos.schema.yml
chetryan: My doubts on this point relate to the EOS documentation, as it only mentions "non-ect chip-based" but on some platforms that I tested on, the command accepted was only "non-ect". I thought that having it as a str would allow us to easily update this for future platforms if there are differences. Happy to discuss this further.

chetryan: Discussed in chat and agreed it is best to use the dictionary structure.

ClausHolbechArista (dismissed):

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+      bandwidth guaranteed percent {{ tx_queue.bandwidth_guaranteed_percent }}
+{% endif %}
+{% if tx_queue.priority is arista.avd.defined %}
+      {{ tx_queue.priority }}
+{% endif %}
+{% if tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ tx_queue.shape.rate }}
+{% endif %}
+{% if tx_queue.random_detect is arista.avd.defined %}
+{%     if tx_queue.random_detect.ecn.count is arista.avd.defined(True) %}
```
tgodaA: Matching the same style

```suggestion
{%     if tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+      random-detect ecn count
+{%     endif %}
+{%     if tx_queue.random_detect.ecn.threshold.min is arista.avd.defined %}
+{%         set units = tx_queue.random_detect.ecn.threshold.units %}
+{%         set ecn_command = "random-detect ecn" %}
+{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ tx_queue.random_detect.ecn.threshold.min ~ " " ~ units  %}
```
tgodaA:
```suggestion
{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ tx_queue.random_detect.ecn.threshold.min ~ " " ~ units %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+      shape rate {{ uc_tx_queue.shape.rate }}
+{% endif %}
+{% if uc_tx_queue.random_detect is arista.avd.defined %}
+{%     if uc_tx_queue.random_detect.ecn.count is arista.avd.defined(True) %}
+      random-detect ecn count
+{%     endif %}
+{%     if uc_tx_queue.random_detect.ecn.threshold.min is arista.avd.defined %}
+{%         set units = uc_tx_queue.random_detect.ecn.threshold.units %}
+{%         set ecn_command = "random-detect ecn" %}
+{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ uc_tx_queue.random_detect.ecn.threshold.min ~ " " ~ units  %}
```
tgodaA:
```suggestion
{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ uc_tx_queue.random_detect.ecn.threshold.min ~ " " ~ units %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+      bandwidth guaranteed percent {{ uc_tx_queue.bandwidth_guaranteed_percent }}
+{% endif %}
+{% if uc_tx_queue.priority is arista.avd.defined %}
+      {{ uc_tx_queue.priority }}
+{% endif %}
+{% if uc_tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ uc_tx_queue.shape.rate }}
+{% endif %}
+{% if uc_tx_queue.random_detect is arista.avd.defined %}
+{%     if uc_tx_queue.random_detect.ecn.count is arista.avd.defined(True) %}
```
tgodaA:
```suggestion
{%     if uc_tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interfaces.j2
```diff
@@ -676,6 +676,14 @@ interface {{ ethernet_interface.name }}
...
    traffic-policy output {{ ethernet_interface.traffic_policy.output }}
 {%     endif %}
+{%     for tx_queue in ethernet_interface.tx_queues | arista.avd.natural_sort('id') %}
+{%         include 'eos/ethernet-interface-tx-queues.j2' %}
+{%     endfor %}
+{%     if ethernet_interface.uc_tx_queues is arista.avd.defined %}
+{%         for uc_tx_queue in ethernet_interface.uc_tx_queues | arista.avd.natural_sort('id') %}
+{%             include 'eos/ethernet-interface-uc-tx-queues.j2' %}
+{%         endfor %}
+{%     endif %}
```
tgodaA: I believe the natural_sort checks if the var is defined or not.

```suggestion
{%     for uc_tx_queue in ethernet_interface.uc_tx_queues | arista.avd.natural_sort('id') %}
{%         include 'eos/ethernet-interface-uc-tx-queues.j2' %}
{%     endfor %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+{% endif %}
+{% if tx_queue.random_detect is arista.avd.defined %}
+{%     if tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
+      random-detect ecn count
+{%     endif %}
+{%     if tx_queue.random_detect.ecn.threshold.min is arista.avd.defined %}
+{%         set units = tx_queue.random_detect.ecn.threshold.units %}
+{%         set ecn_command = "random-detect ecn" %}
+{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ tx_queue.random_detect.ecn.threshold.min ~ " " ~ units %}
+{%         set ecn_command = ecn_command ~ " maximum-threshold " ~ tx_queue.random_detect.ecn.threshold.max ~ " " ~ units %}
```
tgodaA, line 33: There is no defined test added for this.

tgodaA: tx_queue.random_detect.ecn.threshold.max

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+{% endif %}
+{% if uc_tx_queue.random_detect is arista.avd.defined %}
+{%     if uc_tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
+      random-detect ecn count
+{%     endif %}
+{%     if uc_tx_queue.random_detect.ecn.threshold.min is arista.avd.defined %}
+{%         set units = uc_tx_queue.random_detect.ecn.threshold.units %}
+{%         set ecn_command = "random-detect ecn" %}
+{%         set ecn_command = ecn_command ~ " minimum-threshold " ~ uc_tx_queue.random_detect.ecn.threshold.min ~ " " ~ units %}
+{%         set ecn_command = ecn_command ~ " maximum-threshold " ~ uc_tx_queue.random_detect.ecn.threshold.max ~ " " ~ units %}
```
tgodaA, line 33: No defined test added for **max**

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
tgodaA: Based on the documentation, as the min, max, and units are required. We should use "{%     if tx_queue.random_detect.ecn.threshold is arista.avd.defined %}"

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
tgodaA: same comment as tx-queues

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+{% if tx_queue.bandwidth_percent is arista.avd.defined %}
+      bandwidth percent {{ tx_queue.bandwidth_percent }}
+{% elif tx_queue.bandwidth_guaranteed_percent is arista.avd.defined %}
+      bandwidth guaranteed percent {{ tx_queue.bandwidth_guaranteed_percent }}
+{% endif %}
+{% if tx_queue.priority is arista.avd.defined %}
+      {{ tx_queue.priority }}
+{% endif %}
+{% if tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ tx_queue.shape.rate }}
```
tgodaA, line 23: I don't see schema for this. Let me know if it's there in other location.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,37 @@
...
+      !! {{ comment_line }}
+{%     endfor %}
+{% endif %}
+{% if tx_queue.bandwidth_percent is arista.avd.defined %}
+      bandwidth percent {{ tx_queue.bandwidth_percent }}
+{% elif tx_queue.bandwidth_guaranteed_percent is arista.avd.defined %}
+      bandwidth guaranteed percent {{ tx_queue.bandwidth_guaranteed_percent }}
+{% endif %}
+{% if tx_queue.priority is arista.avd.defined %}
+      {{ tx_queue.priority }}
```
tgodaA, line 20: Just a string?

Why not **priority {{ tx_queue.priority }}**. I am not sure about command line.

tgodaA: @chetryan

ClausHolbechArista: The schema says:
```yaml
              priority:
                type: str
                valid_values: ["priority strict", "no priority"]
```
so the template is correct. Also this is not new. It was just moved from the old templates.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
ClausHolbechArista: It is covered in the qos_profiles schema, but maybe not in ethernet-interfaces schema.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
ClausHolbechArista: Good catch. Today we only create warnings on schema violations, so we have to catch it here for a while longer.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
chetryan: As Claus mentioned, I am keeping this as it exists at the moment for backward compatibility.

tgodaA: Agreed. I don't have idea of the command line. So I commented.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
chetryan: Agreed, and added a test for the max too. Should be fixe in the latest commit.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -979,6 +979,149 @@ keys:
...
+                              Set the random-detect ecn maximum-threshold
+                          max_probability:
+                            type: int
+                            min: 1
+                            max: 100
+                            convert_types:
+                              - str
+                            description: |
+                              Set the random-detect ecn max-mark-probability.
+                              The value should be between 1 and 100.
```
carlbuchmann:
```suggestion
```
This information is already contained in min,max and doesn't need to be repeated. This should be corrected through the proposed schema. Also for valid values etc.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/eos_cli_config_gen.schema.yml
```diff
@@ -2269,6 +2269,176 @@ keys:
...
+                            convert_types:
+                            - str
+                            description: 'Set the random-detect ecn max-mark-probability.
+
+                              The value should be between 1 and 100.
+
+                              '
+                          weight:
+                            type: int
+                            min: 1
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/eos_cli_config_gen.schema.yml
```diff
@@ -6778,6 +6972,70 @@ keys:
...
+                            max: 100
+                            convert_types:
+                            - str
+                            description: 'Random-detect ECN maximum mark probability.
+
+                              '
+                          weight:
+                            type: int
+                            min: 1
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/eos_cli_config_gen.schema.yml
```diff
@@ -6819,6 +7077,68 @@ keys:
...
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -979,6 +979,149 @@ keys:
...
+                          weight:
+                            type: int
+                            min: 1
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -979,6 +979,149 @@ keys:
...
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -81,6 +81,60 @@ keys:
...
+                            type: int
+                            min: 1
+                            max: 100
+                            convert_types:
+                              - str
+                            description: |
+                              Random-detect ECN maximum mark probability.
+                          weight:
+                            type: int
+                            min: 1
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
```diff
@@ -117,6 +171,59 @@ keys:
...
```
bjmeuer: the min for the weight could be between 0 and 15, see cli:

ld002(config-s-test-qos-profile-NET-INFRA-txq-4)# random-detect ecn minimum-threshold 256 kbytes maximum-threshold 512 kbytes weight ?
  <0-15>  Value of ECN weight

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/qos_profiles.schema.yml
chetryan: Agreed - will be channging all references.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/ethernet_interfaces.schema.yml
chetryan: Agreed - removing such references in descriptions.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
```diff
@@ -410,6 +428,17 @@ QOS Profile: **uc_mc_queues_test**
 | 2 | Multicast | 10 | priority strict | - | Test strict priority |
 | 4 | Multicast | 10 | - | - | Test guaranteed percent |
 
+**ECN Configuration**
+
+| TX queue | Type | ECN Enabled | Min Threshold | Max Threshold | Max Mark Probability |
+| -------- | ---- | ----------- | ------------ | ------------- | -------------------- |
+| 1 | Unicast | False | -  | -  | - |
+| 2 | Unicast | False | 320 kbytes | 320 kbytes | 90 |
+| 4 | Unicast | False | 320 segments | 320 segments | - |
```
ClausHolbechArista: Is it correct here that ECN is not enabled?

ClausHolbechArista: If this is the counter, we should change the heading to `ECN Counter Enabled` to make it clear. Also in eos_cli_config_gen we normally show `-` if the setting is unset.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,42 @@
...
+{% endif %}
+{% if tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ tx_queue.shape.rate }}
+{% endif %}
+{% if tx_queue.random_detect is arista.avd.defined %}
+{%     if tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
+      random-detect ecn count
+{%     endif %}
+{%     if tx_queue.random_detect.ecn.threshold.min is arista.avd.defined and tx_queue.random_detect.ecn.threshold.max is arista.avd.defined %}
+{%         set units = tx_queue.random_detect.ecn.threshold.units %}
```
ClausHolbechArista, line 30: Please add a defined check for `units` in the `if` above.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-uc-tx-queues.j2
```diff
@@ -0,0 +1,42 @@
...
+{% endif %}
+{% if uc_tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ uc_tx_queue.shape.rate }}
+{% endif %}
+{% if uc_tx_queue.random_detect is arista.avd.defined %}
+{%     if uc_tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
+      random-detect ecn count
+{%     endif %}
+{%     if uc_tx_queue.random_detect.ecn.threshold.min is arista.avd.defined and uc_tx_queue.random_detect.ecn.threshold.max is arista.avd.defined %}
+{%         set units = uc_tx_queue.random_detect.ecn.threshold.units %}
```
ClausHolbechArista, line 30: Please add a defined check for `units` in the `if` above.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
chetryan: This was a bug in the qos_profiles.j2. However, I agree that this column does not add a lot of information and it can be removed.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
```diff
@@ -360,6 +370,14 @@ QOS Profile: **test**
 | 2 | All | 10 | priority strict | - | - |
 | 4 | All | 10 | - | - | - |
 
+**ECN Configuration**
+
+| TX queue | Type | Min Threshold | Max Threshold | Max Mark Probability |
+| -------- | ---- | ----------- | ------------ | ------------- | -------------------- |
```
ClausHolbechArista: Something looks off here

ansible_collections/arista/avd/roles/eos_cli_config_gen/meta/main.yml
ClausHolbechArista, line 1: Please bring this back.

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/meta/main.yml
ClausHolbechArista, line 1: This is the wrong location, so you probably dragged it by mistake from the parent folder.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/qos-profiles.j2
```diff
@@ -69,6 +69,50 @@ QOS Profile: **{{ profile.name }}**
...
+{%             if tx_queue.random_detect.ecn.threshold.units is arista.avd.defined() %}
+{%                 set ns.ecn_table = true %}
+{%             endif %}
+{%         endfor %}
+{%         if ns.ecn_table %}
+
+**ECN Configuration**
+
+| TX queue | Type | Min Threshold | Max Threshold | Max Mark Probability |
+| -------- | ---- | ----------- | ------------ | ------------- | -------------------- |
```
ClausHolbechArista:
```suggestion
| TX queue | Type | Min Threshold | Max Threshold | Max Mark Probability |
| -------- | ---- | ------------- | ------------- | -------------------- |
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/ethernet-interface-tx-queues.j2
```diff
@@ -0,0 +1,42 @@
...
+      {{ tx_queue.priority }}
+{% endif %}
+{% if tx_queue.shape.rate is arista.avd.defined %}
+      shape rate {{ tx_queue.shape.rate }}
+{% endif %}
+{% if tx_queue.random_detect is arista.avd.defined %}
+{%     if tx_queue.random_detect.ecn.count is arista.avd.defined(true) %}
+      random-detect ecn count
+{%     endif %}
+{%     if tx_queue.random_detect.ecn.threshold.min is arista.avd.defined and tx_queue.random_detect.ecn.threshold.max is arista.avd.defined and tx_queue.random_detect.ecn.threshold.units %}
```
ClausHolbechArista:
```suggestion
{%     if tx_queue.random_detect.ecn.threshold.min is arista.avd.defined and tx_queue.random_detect.ecn.threshold.max is arista.avd.defined and tx_queue.random_detect.ecn.threshold.units is arista.avd.defined %}
```

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/qos.md
```diff
@@ -410,6 +428,17 @@ QOS Profile: **uc_mc_queues_test**
 | 2 | Multicast | 10 | priority strict | - | Test strict priority |
 | 4 | Multicast | 10 | - | - | Test guaranteed percent |
 
+**ECN Configuration**
+
+| TX queue | Type | Min Threshold | Max Threshold | Max Mark Probability |
+| -------- | ---- | ----------- | ------------ | ------------- | -------------------- |
```
ClausHolbechArista:
```suggestion
| TX queue | Type | Min Threshold | Max Threshold | Max Mark Probability |
| -------- | ---- | ------------- | ------------- | -------------------- |
```

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #2800, opened 2023-05-08
Fix(eos_designs): Change authentication method and token path for on-prem token auth

### LINKED ISSUE #2700: Fixes needed for token auth implemented in #2685
one quick comment here, `token-secure` method is only supported on CVaaS, whereas on `on-prem` it has to be `token`
they are essentially the same mechanism but `token-secure` will only work on CVaaS and `token` will only work on on-prem, so we will have to make sure to use `"method": "token"` unless CVaaS is used in which case it's `"method": "token-secure"

_Originally posted by @noredistribution in https://github.com/aristanetworks/ansible-avd/issues/2685#issuecomment-1496484031_

Also the default token file should be different between them. For cvaas the current cv-onboarding-token is correct. For onprem it should be `/tmp/token`

ClausHolbechArista:
## Change Summary

Change authentication method and token path for on-prem token auth

## Related Issue(s)

Fixes #2700 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Fix defaults according to best practices. Not breaking since feature was implemented during 4.0.0 development.

## How to test

Updated molecule

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/cvp_ingestauth_key.schema.yml
```diff
@@ -10,5 +10,7 @@ keys:
     type: str
     description: |-
       On-premise CVP ingest auth key. If set, TerminAttr will be configured with key-based authentication for on-premise CVP.
-      If not set, TerminAttr will be configured with certificate based authentication using token-secure onboarding.
+      If not set, TerminAttr will be configured with certificate based authentication:
+      - On-premise using token onboarding. Default token path is '/tmp/token'.
+      - CVaaS using token-secure onboarding. Default token path is '/tmp/cv-onboarding-token'.
```
gmuloc, line 15: we should take th eopportunity of this PR to fix the description for `cvp_instance_ips:`


```
description: 'You can either provide a list of IPs to target on-premise CloudVision
      cluster or
      use DNS name for your CloudVision as a Service instance. If you have both on-prem
      and
      CVaaS defined, only on-prem is going to be configured.
      '
```

Our code actually will use CVaaS only if `arista.io` is insisde the domain name. Otherwise we use on-prem.

Also I am not quite sure that if we have both on-prem and CVaaS defined, only on-prem is configured looking at the code. Looking at the for loop it looks to me that depending on the order of the ip addresses mixing arista.io and not, the outcome can be different

ClausHolbechArista: I just took from the old description. Let me check the code as well and update accordingly.

ClausHolbechArista: Ah sorry I didn't see you were pointing to an unrelated schema. That will be touched on in the schema PR for management. Let's not bring it in here.

gmuloc (approved):

ClausHolbechArista:
got ok from @noredistribution offline. So I will merge this.

PR #2804, opened 2023-05-08
Feat(eos_cli_config_gen): Add trust and chain certificate

### LINKED ISSUE #2470: Add Trust command for management security ssl profiles
### Enhancement summary

Support the Trust command under management_security.ssl_profiles

eos_cli_config_gen

### Use case example

Below use case:

image

```yaml
ssl_profiles:
  - name: < profile name >
     trust_certificates:
       - list of certificate names with options available
```

image

eos_cli

### Additional context

_No response_

### Contributing Guide

tgodaA:
## Change Summary

Support trust and chain certificate commands under management security ssl profiles.

## Related Issue(s)

Fixes #2470 

## Component(s) name

## Proposed changes

```yaml
management_security:
  ssl_profiles:
    - name: <name>
      trust_certificate:
        certificates: [<>]
        requirement:
          enabled: bool
          basic_constraint_ca: bool
          hostname_fqdn: bool
        policy:
          enabled: bool
          expiry_date_ignore: bool
        system: bool
      chain_certificate:
        certificates: [<>]
        requirement:
          enabled: bool
          basic_constraint_ca: bool
          include_root_ca: bool
```

## How to test

```molecule converge -s eos_cli_config_gen -- --limit management-security```

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
@@ -7,6 +7,10 @@
 - [Management Security](#management-security)
   - [Management Security Summary](#management-security-summary)
   - [Management Security SSL Profiles](#management-security-ssl-profiles)
+  - [SSL profile test1-chain-cert Certificates Summary](#ssl-profile-test1-chain-cert-certificates-summary)
+  - [SSL profile test1-trust-cert Certificates Summary](#ssl-profile-test1-trust-cert-certificates-summary)
+  - [SSL profile test2-chain-cert Certificates Summary](#ssl-profile-test2-chain-cert-certificates-summary)
+  - [SSL profile test2-trust-cert Certificates Summary](#ssl-profile-test2-trust-cert-certificates-summary)
   - [Management Security Configuration](#management-security-configuration)
 
 ## Management
@@ -50,15 +54,43 @@ interface Management1
 
 ### Management Security SSL Profiles
 
-| SSL Profile Name | TLS protocol accepted | Certificate filename | Key filename | Cipher List | Trust Certificate | Chain Certificate |
-| ---------------- | --------------------- | -------------------- | ------------ | ----------- | ----------------- | ----------------- |
-| certificate-profile | - | eAPI.crt | eAPI.key | - | - | - |
-| cipher-list-profile | - | - | - | ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384 | - | - |
-| test-chain-cert | - | - | - | - | - | test-chain-cert1.crt, test-chain-cert2.crt, requirement include root-ca |
-| test-trust-cert | - | - | - | - | test-trust.crt, requirement, system | - |
-| tls-single-version-profile-as-float | 1.0 | - | - | - | - | - |
-| tls-single-version-profile-as-string | 1.1 | - | - | - | - | - |
-| tls-versions-profile | 1.0 1.1 | - | - | - | - | - |
+| SSL Profile Name | TLS protocol accepted | Certificate filename | Key filename | Cipher List |
+| ---------------- | --------------------- | -------------------- | ------------ | ----------- |
+| certificate-profile | - | eAPI.crt | eAPI.key | - |
+| cipher-list-profile | - | - | - | ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-SHA384 |
+| test1-chain-cert | - | - | - | - |
+| test1-trust-cert | - | - | - | - |
+| test2-chain-cert | - | - | - | - |
+| test2-trust-cert | - | - | - | - |
+| tls-single-version-profile-as-float | 1.0 | - | - | - |
+| tls-single-version-profile-as-string | 1.1 | - | - | - |
+| tls-versions-profile | 1.0 1.1 | - | - | - |
+
+### SSL profile test1-chain-cert Certificates Summary
+
+
+| Chain Certificates | Requirement |
+| ------------------ | ----------- |
+| test-chain-cert1.crt, test-chain-cert2.crt | Enabled |
+
+### SSL profile test1-trust-cert Certificates Summary
+
+| Trust Certificates | Requirement | Policy | System |
+| ------------------ | ----------- | ------ | ------ |
+| test-trust1.crt, test-trust2.crt | Enabled | Enabled | Enabled |
+
+### SSL profile test2-chain-cert Certificates Summary
+
+
+| Chain Certificates | Requirement |
+| ------------------ | ----------- |
+| - | Basic Constraint CA, Root CA Included |
+
+### SSL profile test2-trust-cert Certificates Summary
+
+| Trust Certificates | Requirement | Policy | System |
+| ------------------ | ----------- | ------ | ------ |
+| - | Basic Constraint CA, Hostname must be FQDN | Ignore Expiry Date | - |
 
 ### Management Security Configuration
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-security.yml [hunk 1]
```diff
@@ -18,3 +18,13 @@ management_security:
...
+    - name: test-trust-cert
+      trust_certificates:
+        - test-trust.crt
+        - requirement
+        - system
+    - name: test-chain-cert
+      chain_certificates:
+        - test-chain-cert1.crt
+        - test-chain-cert2.crt
+        - requirement include root-ca
```
ClausHolbechArista: This seems like a hack. Are these proper CLI options? If so we should write the proper data model for them. Same for requirement & system above

gmuloc:
For reference:

```
Leaf3(config-mgmt-sec-ssl-profile-DUMMY)#show cli commands
APPEND [ COMMENT ]
[no|default] certificate CERT_NAME key KEY_NAME
[no|default] certificate common-name username regexp { <REGEX> }
[no|default] certificate policy expiry-date ignore
[no|default] certificate requirement extended-key-usage
[no|default] certificate requirement hostname match
chain certificate CERT_NAME
(no|default) chain certificate CERT_NAME
[no|default] chain certificate requirement include root-ca
[no|default] cipher-list CIPHERS
[no|default] comment
crl CERT_NAME
(no|default) crl CERT_NAME
[no|default] crl policy expiry-date ignore
[no|default] diffie-hellman parameters DHPARAMS
do CMD
exit
[no|default] fips restrictions
[no|default] peer certificate requirement hostname match subject-alternative-name [ common-name ]
show active
show active all [ detail ]
show comment
[no|default] tls versions [ add | remove ] { 1.0 | 1.1 | 1.2 }
trust certificate CERT_NAME
(no|default) trust certificate CERT_NAME
[no|default] trust certificate policy expiry-date ignore
[no|default] trust certificate requirement hostname fqdn
[no|default] (trust|chain) certificate requirement basic-constraint ca true
```

```
Leaf3(config-mgmt-sec-ssl-profile-DUMMY)#show active
management security
   ssl profile DUMMY
      chain certificate TITI
      chain certificate TOTO
      chain certificate requirement basic-constraint ca true
      chain certificate requirement include root-ca
     
```

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/management-security.yml
gmuloc: +1 with @ClausHolbechArista we should separate the list of certificate from the `requirement` ... options.

Given that `requirement basic-constraint ca true` can be in both chain and trust I think it would make more sense to use a `$defs` for this one

tgodaA: Updated it.

ClausHolbechArista:
We need to work more on the schema here.

There is no need for `enabled` under the `requirement` key, since it should never be configured alone. I tested that:
```sh
bko016#show run sec ssl
management security
   ssl profile test
      trust certificate requirement
      trust certificate requirement basic-constraint ca true
bko016#show management security ss profile test
   Profile       State      Additional Info                         
------------- ------------- ----------------------------------------
   test          invalid    Certificate 'requirement' does not exist
```
Since the command takes any word as certificate, it also catch the "requirement" word when given alone.

ClausHolbechArista:
> @ClausHolbechArista, I thought you would disagree with the `certificates: ['cert1.crt', 'requirement', 'system']` So I developed the schema for all though I didn't like the enabled key. Also, when the enabled is used, the command order is changed, which is different from the EOS.
> 
> If you agree, I will remove all enabled keys.

My point is that there is no such option as `requirement` alone, so we should not have a data model for it.
I did not check the full data model in detail, so you need to go through the output from Guillaume above and match it up agains the data model.

tgodaA:
Just the system command exists
image

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/management_security.schema.yml
```diff
@@ -44,6 +44,48 @@ keys:
...
+                      - test2.crt
+                  items:
+                    type: str
+                requirement:
+                  type: dict
+                  keys:
+                    basic_constraint_ca:
+                      type: bool
+                    hostname_fqdn:
+                      type: bool
```
chetryan, line 63:
```suggestion
                      type: bool
                      description: |
                              Enforce hostname to be FQDN without wildcard.
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/management_security.schema.yml
```diff
@@ -44,6 +44,48 @@ keys:
...
+                policy_expiry_date_ignore:
+                  type: bool
+                system:
+                  type: bool
```
chetryan, line 69:
```suggestion
                  type: bool
                  description: |
                           Use system-supplied trust certificates.
```

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
```diff
@@ -54,10 +58,40 @@ interface Management1
...
+| test1-chain-cert | - | - | - | - |
+| test1-trust-cert | - | - | - | - |
+| test2-chain-cert | - | - | - | - |
+| test2-trust-cert | - | - | - | - |
 | tls-single-version-profile-as-float | 1.0 | - | - | - |
 | tls-single-version-profile-as-string | 1.1 | - | - | - |
 | tls-versions-profile | 1.0 1.1 | - | - | - |
 
+### SSL profile test1-chain-cert Certificates Summary
+
```
chetryan, line 70: Can we fix this to be 1 newline instead of 2 to be consistent with the rest?

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-security.j2
```diff
@@ -25,8 +25,73 @@
...
+{%             if tmp_cert | length > 0 %}
+{%                 do tmp_cert.update({'name': ssl_profile.name}) %}
+{%                 do ssl_profiles_certs.append(tmp_cert) %}
+{%             endif %}
+{%         endfor %}
+{%         for ssl_profile in ssl_profiles_certs | arista.avd.natural_sort('name') %}
+
+### SSL profile {{ ssl_profile.name }} Certificates Summary
+
+{%             if ssl_profile.trust_certificate is arista.avd.defined %}
```
chetryan: I propose to move the newline after the if to avoid 2 newlines if no trust_ceritificates are present.
```suggestion
{%             if ssl_profile.trust_certificate is arista.avd.defined %}

```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-security.j2
```diff
@@ -25,8 +25,73 @@
...
+{%                     if ssl_profile.trust_certificate.requirement.basic_constraint_ca is arista.avd.defined(true) %}
+{%                         do tmp_requirement.append('Basic Constraint CA') %}
+{%                     endif %}
+{%                     if ssl_profile.trust_certificate.requirement.hostname_fqdn is arista.avd.defined(true) %}
+{%                         do tmp_requirement.append('Hostname must be FQDN') %}
+{%                     endif %}
+{%                     if tmp_requirement | length > 0 %}
+{%                         set requirement = tmp_requirement | join(', ') %}
+{%                     endif %}
+{%                 endif %}
```
chetryan, line 63: Could we not use the same logic as `trust_certs` to avoid setting and using if statements? 

```suggestion
{%                 endif %}
{%                 set requirement = tmp_requirement  | arista.avd.default('-') |  join(', ') %}
```
This would also make the line 51 no longer required. If this logic makes sense, it could be used for the policy set on line 64 as well.

chetryan (commented):
Some comments about improving Markdown and schema documentation, but other than that it LGTM.
I even tested the commands out on `DCS-7050DX4-32S-F` running `4.30.1F` and all were accepted without issue.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-security.md
tgodaA: Good catch!

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/management-security.j2
tgodaA: Yes, but I just moved everything under the if condition `ssl_profile.trust_certificate.requirement` so, if it's not defined then tmp_requirement is not defined. Your suggestion reduces the code to one-liner but I did this way because I don't want the `tmp_requirement  | arista.avd.default('-')` being invoked everytime even when the requirement is not defined.

chetryan (approved):
LGTM

ClausHolbechArista (approved):

PR #2870, opened 2023-05-19
Fix(plugins): Fix minor schema tooling issues

ClausHolbechArista:
Fix minor schema tooling issues:
- Fix YAML documentation for rows without keys or items
- Remove keys with "deprecation.removed: true" from YAML documentation
- Raise during schema conversion if schema is invalid


ClausHolbechArista:
Documentation for various keys updated in this PR reveals that the schema are missing `items` or `keys` definitions. This should be fixed.

PR #2888, opened 2023-05-24
Fix(eos_designs): Remove EVPN related config if VRF 'default' is not EVPN enabled

ClausHolbechArista:
## Change Summary

Remove EVPN related config if VRF 'default' is not EVPN enabled

## Related Issue(s)

For EVPN networks carrying VRF 'default' the best practice is to use the overlay (EVPN l3vni) for services, but keep loopbacks in the underlay. To avoid the services from being routed in the underlay, AVD implements route-maps and prefix-lists when VRF 'default' is defined on an EVPN VTEP.
It is possible to disable EVPN under the VRF configuration by setting `address_families:[]` in which case the VRF should _not_ be added to EVPN configuration, and the route-maps and prefix-lists should not be created at all.
Before this fix, the route-maps and prefix-lists were still configured even if the VRF was not configured for EVPN. The VRF was also configured with a VNI under the VXLAN interface.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Detect if VRF 'default' has `evpn` address-family enabled (true by default)
- Only configure the underlay filtering route-map/prefix-lists if EVPN is enabled for VRF 'default'
- Only configure VNI for VRFs where EVPN is enabled _and_ the overlay is EVPN-VXLAN (changed from pure EVPN check - which would also be true with EVPN-MPLS)

## How to test

Added a molecule test case for this corner case.
No other changes to molecule outputs.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn-vtep-with-default-vrf-not-evpn.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn-vtep-with-default-vrf-not-evpn.yml
@@ -1,4 +1,4 @@
-# Testing an EVPN VTEP where thed default VRF is _not_ EVPN enabled.
+# Testing an EVPN VTEP where the default VRF is _not_ EVPN enabled.
 # Make sure there is no route-map added to the underlay peer group filtering the evpn prefixes.
 # Also make sure the VRF is not configured under interface Vxlan 1
 # Minimum config to only test the specific feature.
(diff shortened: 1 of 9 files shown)


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn-vtep-with-default-vrf-not-evpn.yml
```diff
@@ -0,0 +1,28 @@
+# Testing an EVPN VTEP where thed default VRF is _not_ EVPN enabled.
```
carlbuchmann:
```suggestion
# Testing an EVPN VTEP where the default VRF is _not_ EVPN enabled.
```

carlbuchmann (approved):
LGTM

PR #2894, opened 2023-05-25
CI: Hardcode version of galaxy-importer

ClausHolbechArista:
CI: Hardcode version of galaxy-importer

We don't see the same errors in our pipeline as we do when importing to Ansible galaxy
This is most likely because of the ancient version of galaxy-importer used on galaxy.ansible.com.


carlbuchmann (approved):
LGTM. Reviewed action results and consistent with what we see on ansible-galaxy

PR #2897, opened 2023-05-26
Feat(eos_designs): Add schema for queue_monitor_length

### LINKED ISSUE #2852: Add schema for `queue_monitor_length`
```yaml
queue_monitor_length:
  log: 5
  notifying: true
  # Note "notifying" should only be rendered on 7280R since default
  # platform_settings for default has queue_monitor_length_notify_support: false
```

ClausHolbechArista:
## Change Summary

Add schema for queue_monitor_length

## Related Issue(s)

Fixes #2852 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Add schema for queue monitor length
- Enhance support through eos_designs by not parsing the inputs, but just relaying it (after filtering out `notifying` where needed)

## How to test

Covered by existing molecule scenario `evpn_underlay_ebgp_overlay_ebgp`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/queue_monitor_length.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/queue_monitor_length.yml
@@ -1,5 +1,4 @@
 queue_monitor_length:
-  # "enabled: true" will be required in AVD4.0. In AVD3.x default is true as long as queue_monitor_length is defined and not None
   enabled: true
   default_thresholds:
     low: 10
--- a/ansible_collections/arista/avd/molecule/evpn_underlay_ebgp_overlay_ebgp/inventory/group_vars/AVD_LAB.yml
+++ b/ansible_collections/arista/avd/molecule/evpn_underlay_ebgp_overlay_ebgp/inventory/group_vars/AVD_LAB.yml
@@ -43,7 +43,8 @@ ntp:
       vrf: "{{ mgmt_interface_vrf | arista.avd.default('MGMT') }}"
 
 queue_monitor_length:
-  log: 5
+  enabled: true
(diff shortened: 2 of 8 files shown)


jrecchia1029:
LGTM.  Just curious as to why `notifying` has its own description in eos designs. Is it simply to indicate that this key will only be configured if it is supported on the platform for which config is being generated?

xaviramon:
I see we can configure `notifying` and if we set it up as false in a device which is not supporting this feature, it will generate a broken config. Is this something we should cover here?
```
queue_monitor_length:
  enabled: true
  notifying: false
```
  For a platform `default` generates this command:
```
  queue-monitor length
  no queue-monitor length notifying
```
Which is not supported:
```
ptt013(config)#no queue-monitor length not
% Unavailable command (not supported on this hardware platform)
```

ClausHolbechArista:
> I see we can configure `notifying` and if we set it up as false in a device which is not supporting this feature, it will generate a broken config. Is this something we should cover here?

Yes thank you for testing thoroughly. I will make sure not to have the key at all if not supported.

ClausHolbechArista:
> LGTM. Just curious as to why `notifying` has its own description in eos designs. Is it simply to indicate that this key will only be configured if it is supported on the platform for which config is being generated?

In `eos_cli_config_gen` we render what is given to us, and we have no idea if the command is supported or not, so we cannot specify that in the description.
In `eos_designs` we reuse most of the schema, except for this specific key, where we can use our `platform_settings` to control whether the key is supported or not.

ClausHolbechArista:
@xaviramon please test again. I have now removed the key completely when not supported. Thanks

xaviramon (approved):
LGTM!

carlbuchmann (approved):
LGTM!

PR #2900, opened 2023-05-26
Fix(eos_designs): Ensure deterministic behavior when defining the same VRF in multiple Tenants

### LINKED ISSUE #2812: Provide deterministic support for defining the same VRF in multiple tenants
### Enhancement summary

In AVD versions < 4.0.0 it was possible to define the same VRF in multiple tenants. Since the output of `eos_designs` where mostly dictionaries, it was a matter of ordering and "chance" on which of those VRF definitions got rendered. In some cases all of them got combined - ex. for `svis` or `static_routes`, while other settings where overridden by the "last" edition of the VRF - ex. for `bgp_peers`.

In AVD 4.0.0 there is a check for duplicate VRFs preventing the same VRF from being defined across multiple tenants. This will block usage of AVD 4.0.0 in certain customer scenarios unless we add this feature.

eos_designs

### Use case example

Defining a VRF for shared services across multiple "customer" tenants.

This issue is a request to provide deterministic support for defining the same VRF across multiple tenants.
- For list items like `svis`, `l3_interfaces`, `bgp_peers` etc we should combine all of them.
  - If the merged lists contain duplicate items, they will be caught and errors will be raised 
    by duplicate checks for each feature.
  - This means it is _not_ supported to define the same entity (`svi`, `l3_interface` etc) 
    across different tenants for the same VRF.
- For regular keys like `vrf_vni` or `description` they should be combined
  - If a key is only set in one of the VRF instances, it will be accepted
  - If a key is defined in multiple VRF instances with the same value, it will be accepted
  - If a key is defined in multiple VRF instances with _different_ values, an error will be raised.

_No response_

### Additional context

_No response_

### Contributing Guide

ClausHolbechArista:
## Change Summary

Ensure deterministic behavior when defining the same VRF in multiple Tenants

## Related Issue(s)

Fixes #2812
Partial fix of #2749 (The remaining duplication checks should be modified to use this new function)

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

VRFs will only be configured on a node if any of the underlying objects like `svis` or `l3_interfaces` apply to the node.

It is recommended to only define a VRF in one Tenant. If the same VRF name is used across multiple tenants and those tenants
are accepted by `filter.tenants` on the node, any object set under the duplicate VRFs must either be unique or be an exact match.

## How to test

Added both positive and negative test cases to molecule scenarios.

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista:
Need to change the rules to not enforce all vars to be the same, since they may not matter for the node(s) having multiple tenants

emilarista (approved):
LGTM provided CI passes!

xaviramon:
If I understand this PR right, this should be failing or throwing an error and it is not (`mac_vrf_vni_base` should be the same):
```
  TENANT_B:
    enable_mlag_ibgp_peering_vrfs: false
    mac_vrf_vni_base: 20000
    vrfs:
      PLATFORM_B:
        vrf_id: 1002
        svis:
          20:
            name: "REDES_B_PLATFORM"
            tags: [ POD1, POD2 ]
            ip_address_virtual: 10.10.20.1/24
            enabled: true
  TENANT_C:
    enable_mlag_ibgp_peering_vrfs: false
    mac_vrf_vni_base: 30000
    vrfs:
      PLATFORM_B:
        vrf_id: 1002
        svis:
          30:
            name: "REDES_C_PLATFORM"
            tags: [ POD1, POD2 ]
            ip_address_virtual: 10.10.30.1/24
            enabled: true
```

ClausHolbechArista:
@xaviramon I forgot to change the PR text after changing the merge policy. We tried to implement what I wrote here first, but it did not cover the use case we needed, so we reimplemented it with another policy:

```text
It is recommended to only define a VRF in one Tenant. If the same VRF name is used across multiple tenants and those tenants
are accepted by `filter.tenants` on the node, any object set under the duplicate VRFs must either be unique or be an exact match.
```

The result of this change is to only error out if the different keys actually affect the output configuration, by creating two different configs.

You would see an error if you added a copy of SVI 20 on the second tenant.

carlbuchmann (approved):
LGTM

Issue #1047, opened 2021-06-21
`eos_validate_state` should not rely on `eos_designs` variable `type`


ClausHolbechArista:
## Issue Type

- [x] Bug Report
- [ ] EOS Configuration rendering error
- [ ] Documentation Issue Report

## Summary

In `eos_validate_state` the template `generate_vars_for_testing.j2` tests for `hostvars[node].type == 'l3leaf'`.
This variable is only set when using eos_designs, so we should not rely on it for tests.

Also with AVD 3.0.0 we will make the `type` value customizable, so we should not rely on that, to decide on reachability tests.

## Component Name 
`arista.avd.validate_state`

First new version will be `arista.avd.validate_state_preview` and only when it is complete and ready, it will replace `arista.avd.validate_state`

## Steps to reproduce

```yaml

```

### Expected results

```shell
$
```

### Actual results

```shell

```

ClausHolbechArista:
`eos_validate_state` should only rely on `eos_cli_config_gen` data model.

xaviramon:
I can work on this one.

There are some assumptions that we need to do in order to solve this bug.

This is partially solved by PR#1360 https://github.com/aristanetworks/ansible-avd/pull/1360/files

If we assume all devices should be able to ping the VXLAN source interface and the Router-ID defined for all devices, that should be ok. If we need to add more intelligence, we will need to discuss.

PR #2907, opened 2023-05-29
Feat(eos_validate_state): Add ANTA integration to eos_validate_state role

carl-baillargeon:
## Change Summary

Integration of ANTA Python framework to eos_validate_state role.

## Tasks List

- [X] (hardware) Validate environment (power supplies status)
- [X] (hardware) Validate environment (fan status).
- [X] (hardware) Validate environment (temperature).
- [X] (hardware) Validate transceivers manufacturer.
- [X] (NTP) Validate NTP status.
- [X] (interface_state) Validate Ethernet interfaces admin and operational status.
- [X] (interface_state) Validate Port-Channel interfaces admin and operational status.
- [X] (interface_state) Validate VLAN interfaces admin and operational status.
- [X] (interface_state) Validate VXLAN interfaces admin and operational status.
- [X] (interface_state) Validate Loopback interfaces admin and operational status.
- [X] (lldp_topology_fqdn) Validate LLDP topology when there is a domain name configured.
- [X] (lldp_topology_no_fqdn) Validate LLDP topology when there is no domain name configured.
- [X] (MLAG) Validate MLAG status.
- [X] (ip_reachability) Validate IP reachability (on directly connected interfaces).
- [X] (loopback_reachability) Validate loopback reachability (between devices).
- [X] (bgp_check) Validate ArBGP is configured and operating.
- [X] (bgp_check) Validate IP BGP and BGP EVPN sessions state.
- [X] (reload_cause) Validate last reload cause. (Optional)
   - Change of behavior as now there is a new reload cause that is valid: https://github.com/arista-netdevops-community/anta/blob/main/anta/tests/system.py#L72C1-L75C11
- [X] (routing_table) Validate remote Lo0 addresses and remote Lo1 addresses are in the routing table (based on devices type).
- [x] schema -> probably not needed as part of this PR except if we want to add the 3 `accepted_` values - later discussion
- [x] make eos_validate_state agnostic of eos_designs -> by adding knobs in structured_config where it is required #1047 - will be addressed after this PR 
- [x] discuss mode strict behavior (vs mode loose) - not implemented for anta mode
- [x] Usage of display (ansible future proofing) ? Do we need all of them that exist today or are some present only because of the impossibility to remove "skipped" messages on Ansible -> make it a logger
- [ ] Release tag for anta - with CACHE
- [x] Use check flag in ansible
- [x] module documentation -> DONE need review - the module is considered as preview
- [ ] eos_validate_state preview validation -  make the "feature" a preview for first release
    - [ ] add warning to state that loose mode is ignored
- [ ] testing current eos_validate_state and anta based eos_validate_test and make sure there is no change - document the changes for the tests

## Postponed Tasks List
- [ ] fix loopback_reachability tags support when running anta -> PR for existing validate state + rebase.
- [ ] Use anta variable instead of tag for running eos_validate_state.
- [ ] Remove even further the ties to ansible in the plugins_utils
- [ ] Hardware tests are now grouped together in the final report. Will need to add a functionality in ANTA to get more specific results per test.
- [ ] Adding  unit tests for module and Handler

## Component(s) name

## Proposed changes
- `eos_validate_state_runner.py` is an action plugin to run the ANTA framework. It calls the `get_anta_results` function to grab the test results.
- `get_anta_results.py` is responsible of creating the ANTA tests catalog and the other required objects to run ANTA. 
- Connections to the devices are made with the regular Ansible `httpapi` connection plugin. 
- Results from ANTA are returned as JSON and registered as the `anta_results` variable in Ansible.
- ANTA related tasks have been added to `eos_validate_state` `tasks/main.yml` and tagged `anta` to keep the legacy `eos_validate_state` working.
- A new task has been added to `tasks/reports.yml` to generate ANTA results, leveraging `yaml_templates_to_facts` plugin and `template/generate_anta_report_results.j2` template specific to ANTA results but keeping the same format.
- The following Python classes have been created to replace the existing eos_validate_state tests:
```python 
__all__ = [
    "AvdTestP2PIPReachability",
    "AvdTestLoopback0Reachability",
    "AvdTestLLDPTopology",
    "AvdTestInbandReachability",
    "AvdTestHardware",
    "AvdTestMLAG",
    "AvdTestNTP",
    "AvdTestReloadCause",
    "AvdTestInterfacesState",
    "AvdTestRoutingTable",
    "AvdTestBGP",
]
```

## How to test
- Install the main branch of anta:
`pip install git+https://github.com/arista-netdevops-community/anta.git`
- `eos_validate_state` can be used as usual, with `--tags anta` if you want to run the role with ANTA: 
`ansible-playbook playbooks/fabric-validate.yaml --tags anta`
__Not providing the `anta` tag will run the regular `eos_validate_state`__
- Legacy Ansible tags are also supported if you want to run/skip tests:
`ansible-playbook playbooks/fabric-validate.yaml --tags anta,routing_table`

The variable `skipped_tests` can now be used for running/skipping tests:
```
skipped_tests:
    AvdTestHardware:
```
You can also decide to skip specific subtests (ANTA test) for more granularity:
```
skipped_tests:
    AvdTestHardware:
        - VerifyTransceiversTemperature
```
### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
```diff
@@ -0,0 +1,24 @@
...
+{# ANTA Results #}
+{%     if copy_hostvars[node].anta_results is arista.avd.defined %}
+{%         for anta_result in copy_hostvars[node].anta_results.results %}
+{%             set test_id.value = test_id.value + 1 %}
+- id: {{ test_id.value }}
+  device: "{{ anta_result.name }}"
+  test_category: "{{ anta_result.test_category }}"
+  test_description: "{{ anta_result.test_description }}"
+  test: "{{ anta_result.test }}"
+{%             if anta_result.result is arista.avd.defined("success") %}
```
carl-baillargeon, line 16: Need to add validation for "skipped" and "failed" test results.

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
carl-baillargeon, line 1: - Need to handle the "skipped" and "error" test results.
- Need to handle the list items returned by ANTA : `anta.result.messages` and `anta_result.test_category`.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
```diff
@@ -0,0 +1,104 @@
...
+from typing import Any, Dict, List, Tuple
+
+from ansible.plugins.action import ActionBase
+from anta.loader import parse_catalog
+
+TestTuple = Tuple[str, Dict[str, str]]
+Catalog = Dict[str, List[Dict[str, Any]]]
+
+
+AVD_MAPPING: Dict[str, List[TestTuple]] = {
```
carl-baillargeon, line 14: This is where we map Ansible task_vars to ANTA different test cases, including the right parameters as inputs. Another mapping example would be if `banners` key exists in task_vars, run the relevant banner test cases of ANTA with `banners.motd` and `banners.login` as inputs.

This mapping could get quite large the more tests requiring inputs we add in ANTA. It is also a manual process, meaning if we add a new test in ANTA, we need to update the mapping if needed.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+from typing import Any, Dict, List, Tuple
```
ClausHolbechArista, line 5: Please `from __future__ import annotations` and use the regular `dict, list, tuple` for typing.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _get_test_name(test_name: str) -> str:
+    """Extracts the test name from a fully qualified test name."""
+    return test_name.rsplit(".", 1)[1]
+
+
+def _update_test_parameters(value: Any, test_parameters: Dict[str, str]) -> Dict[str, Any]:
+    """Updates the test parameters based on the provided structured configuration values."""
+    return {param_key: value.get(structured_config_key) for param_key, structured_config_key in test_parameters.items()}
```
ClausHolbechArista, line 31: Maybe you could use our `get()` util to support deeper nested keys? Check it out in `plugins/plugin_utils/utils/get.py`

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
```diff
@@ -0,0 +1,104 @@
...
+TestTuple = Tuple[str, Dict[str, str]]
+Catalog = Dict[str, List[Dict[str, Any]]]
+
+
+AVD_MAPPING: Dict[str, List[TestTuple]] = {
+    "management_api_http": [("anta.tests.security.VerifyAPIHttpStatus", {}), ("anta.tests.security.VerifyAPIHttpsSSL", {"profile": "https_ssl_profile"})]
+}
+
+
+def _get_package_name(test_name: str) -> str:
```
ClausHolbechArista, line 19: please move all your functions and constants into a lib you can import from plugin_utils. That will help us reuse the code in `pyavd`.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
```diff
@@ -0,0 +1,104 @@
...
+
+def _build_new_catalog(task_vars: Dict[str, Any], test_mapping: Dict[str, List[TestTuple]]) -> Catalog:
+    """
+    Constructs package tests from the structured configuration and the test mapping.
+
+    Returns:
+        A dictionary containing package tests.
+    """
+    new_catalog: Catalog = {}
+    for key, value in task_vars.items():
```
ClausHolbechArista, line 42: task vars can be _very_ large, so I would suggest to flip this to loop over test_mappings and check if the key is there.

ClausHolbechArista: Thinking more about this, maybe it would be cleaner to build a small function/class for each of the ANTA tests which accepts the structured config as input, and extracts the stuff we need. I think it would be simpler to maintain over time.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,50 @@
...
+
+        result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        device_params = {
+            "name": task_vars["inventory_hostname"],
+            "host": task_vars["ansible_host"],
+            "username": task_vars["ansible_user"],
+            "password": task_vars["ansible_httpapi_password"],
+            "port": 443,
```
ClausHolbechArista, line 25: this is also configurable

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,50 @@
...
+            task_vars = {}
+
+        result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        device_params = {
+            "name": task_vars["inventory_hostname"],
+            "host": task_vars["ansible_host"],
+            "username": task_vars["ansible_user"],
+            "password": task_vars["ansible_httpapi_password"],
```
ClausHolbechArista, line 24: I think you want to take these from the _connection object and not the input vars, since they need to be de-vaulted etc.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,50 @@
...
+            "session": self._connection.httpapi,
+        }
+
+        try:
+            anta_device = InventoryDevice(**device_params)
+            tests_catalog = task_vars["anta_catalog"]["anta_catalog"]
+            results = ResultManager()
+
+            for test_class, test_params in tests_catalog:
+                test_instance = test_class(device=anta_device, from_ansible=True)
```
ClausHolbechArista, line 35: What does this `from_ansible` do? Maybe add some more descriptions in the code to help the reader.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,50 @@
...
+                for index, command in enumerate(test_instance.instance_commands):
+                    response = anta_device.session.send_request(command.command, version=command.version, output=command.ofmt)
```
ClausHolbechArista, line 37: So with this you basically moved the device interaction out of ANTA and loop over them here. It is fine for now, but I think we want it to be done/maintained inside ANTA instead, but using a passed interface to httpapi.

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
ClausHolbechArista: You could add it as another template to the regular "Generate Results" invocation of `yaml_templates_to_facts`. Inside the yaml template (or python if you refactor) you can then check for the tag and only run if the anta tag is set.

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -50,6 +50,39 @@
...
+    name: input_catalog
+  delegate_to: localhost
+  when: input_catalog_path is defined
+  run_once: true
+  tags:
+    - anta
+
+- name: Generate ANTA catalog
+  arista.avd.anta_catalog_builder:
+    input_catalog: "{{ input_catalog | default(None) }}"
```
ClausHolbechArista, line 65: Moving large data blobs into Ansible will only slow things down. I think we should just pass the path to the action, and then we can load the input_catalog in there. Alternatively use a lookup plugin to load inline in this parameter.

Also collapsing the `anta_catalog_builder` and `anta_runnner` would save us from getting all the catalog vars set in Ansible. You could save the catalog still in a file for testing/dryrun

ClausHolbechArista:
```suggestion
    input_catalog: "{{ (lookup('ansible.builtin.file', input_catalog_path) | from_yaml) if input_catalog_path is arista.avd.defined }}"
```

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
carl-baillargeon: Yeah I had in mind to create a dynamic catalog depending on the task_vars. If we decide to extract the stuff we need in the ANTA tests directly, we will have to create the test catalog manually.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
carl-baillargeon: Catalog handling has been merged in `anta_runner.py`.

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
carl-baillargeon: 10/4

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
carl-baillargeon: Nice little tool! I'm now using it in ANTA :)

ansible_collections/arista/avd/plugins/action/anta_catalog_builder.py
carl-baillargeon: 10/4

ansible_collections/arista/avd/plugins/action/anta_runner.py
carl-baillargeon: More descriptions have been added.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,84 @@
...
+
+            for test_class, test_params in tests_catalog:
+                # sync is a new attribute to run ANTA tests synchronously.
+                test_instance = test_class(device=anta_device, sync=True)
+
+                # Tests requiring input parameters will extract them from task_vars.
+                # If ALL parameters are not found in task_vars, tests will run with the parameters from the input_catalog file.
+                # In ANTA, tests requiring parameters are "skipped" if any of the parameters are not provided.
+                if hasattr(test_class, "extract_parameters"):
+                    extracted_params = test_class.extract_parameters(task_vars)
```
ClausHolbechArista, line 64: I don't like that we store the mapping of AVD variables to ANTA test arguments somewhere in ANTA. It will break if you upgrade on or the other out of sync.
Please store the mapping in AVD test-by-test and keep the extraction locally in AVD.
ANTA should just take it's own parameters per test, and not have any AVD specific code.

ClausHolbechArista (commented):
It is hard to review when most of the actual things happening are in ANTA, and from this it seems like a highly customized version of ANTA.
As mentioned we should try to keep ANTA "neutral" and perform all input/output conversions locally in AVD.

carl-baillargeon:
> It is hard to review when most of the actual things happening are in ANTA, and from this it seems like a highly customized version of ANTA. As mentioned we should try to keep ANTA "neutral" and perform all input/output conversions locally in AVD.

Actually ANTA is now a lot more flexible since it can accept any session connection object to create the appropriate subclass of `InventoryDevice`. `InventoryDeviceHttpApi` is just a subclass created with the Ansible `HttpApi` interface. Tests can now also be ran synchronously or asynchronously within ANTA depending on how you initiate the tests. All the rest of ANTA functionality (CLI functions) stays the same since it is using his own `InventoryDeviceAioeapi`. In the future, we could even create other subclasses to support different connectors if needed.
```
class InventoryDevice(ABC, BaseModel):
    """
    Abstract class for InventoryDevice.

    Attributes:
        name (str): Device name.
        session (Any): JSONRPC session.
        hw_model (str): HW name gathered during device initialization.
    """

    name: str
    session: Any
    hw_model: str = DEFAULT_HW_MODEL

    @abstractmethod
    def collect(self, command: AntaTestCommand) -> Any:
        """
        Prepares and sends request(s) to device.

        Args:
            command (AntaTestCommand): Command to execute on the device.

        Returns:
            Any: The command that was executed, including its output data.
        """
```

The input stays the same (with a catalog) but I can move back the mapping of AVD variables in AVD no problem. As for the output, it is already being handled in the action plugin.

Let me know what you think!

ClausHolbechArista:
`Actually ANTA is now a lot more flexible since it can accept any session connection object to create the appropriate subclass of InventoryDevice. InventoryDeviceHttpApi is just a subclass created with the Ansible HttpApi interface.`

Wouldn't it be more appropriate to store the subclass in AVD or do you imagine other Ansible integrations could leverage this?

carl-baillargeon:
> `Actually ANTA is now a lot more flexible since it can accept any session connection object to create the appropriate subclass of InventoryDevice. InventoryDeviceHttpApi is just a subclass created with the Ansible HttpApi interface.`
> 
> Wouldn't it be more appropriate to store the subclass in AVD or do you imagine other Ansible integrations could leverage this?

I think we should leave it in ANTA with the other models. We can then easily see which connection object ANTA supports and it could be leveraged outside of AVD for generic tests that don't require the structured config. It's more flexible that way.

ansible_collections/arista/avd/plugins/action/anta_runner.py
carl-baillargeon: I created a small function for the mapping.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/update_catalog.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+    The mapping (dict) variable inside this function needs to be updated if new test cases requiring AVD input parameters are added in ANTA.
+
+    Ex:
+      mapping: dict = {
+        'anta.tests.management': [{'VerifyBannerMOTD': {'banner': get(task_vars, 'banners.motd')}}]
+      }
+    """
+    mapping: dict = {"anta.tests.security": [{"VerifyAPIHttpsSSL": {"profile": get(task_vars, "management_api_http.https_ssl_profile")}}]}
```
ClausHolbechArista, line 25: I guess you would want to loop over the input_catalog and perform the mapping for each test in there - and only if they are present. Maybe having a constant with dict of tests containing the key paths to map in.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,80 @@
...
+            # Creating the catalog from the Ansible task
+            catalog_path = self._task.args.get("catalog_path")
+            with open(catalog_path, "r", encoding="UTF-8") as file:
+                input_catalog = safe_load(file)
+
+            # Raises error if the provided ANTA test catalog is empty
+            if not input_catalog:
+                raise AnsibleActionFail(f"No tests were provided in the catalog: {catalog_path}")
+
+            # In ANTA, tests requiring parameters are "skipped" if they are not provided or not found in task_vars
```
ClausHolbechArista, line 50: I would prefer not to rely on this behavior, but instead raise an error if a test is set in the input_catalog, but we cannot find the required variables for it.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,80 @@
...
+            task_vars = {}
+
+        result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        play_context = self._play_context
+        connection = self._connection.httpapi
+
+        # Need to set the right become_context to the _connection.httpapi object
+        connection.set_become(play_context)
```
ClausHolbechArista, line 30: Since you moved the connection integration into ANTA, this should also be covered in there. Otherwise someone running anta outside of AVD would be missing this.

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,74 @@
...
+
+            # Creating the ANTA ResultManager object
+            manager = ResultManager()
+
+            # Run tests
+            results = asyncio.run(run_tests(manager=manager, device=device, tests=tests))
+
+            # Format the data properly for the eos_validate_state report
+            for item in results:
+                item["test_category"] = ", ".join([category.upper() if len(category) <= 5 else category.title() for category in item["test_category"]])
```
ClausHolbechArista, line 64: This seems a bit hacky. I mean the <= 5 would easily break in the future. Maybe we need to have a constant with the mapping of categories? Or at list a list of the acronyms and then run `title` on the rest?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/models.py
```diff
@@ -0,0 +1,64 @@
...
+from ansible.errors import AnsibleError
+from anta.models import AntaTestCommand
+
+
+class InventoryDeviceHttpApi:
+    """
+    Custom implementation of an ANTA InventoryDevice with Ansible HttpApi connection session object.
+    """
+
+    def __init__(self, session):
```
ClausHolbechArista, line 15: Maybe add a type hint on the session object and maybe even do a `isinstance` check since we expect it to have certain methods below.

ClausHolbechArista (commented):
I like the structure of this now. We will have to build something to handle the default catalog while still allowing the user to override the full catalog or just disable individual tests from the defaults. Also for 4.x we need to map in the ansible_run_tags to enable/disable tests in that default catalog.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/runner.py
```diff
@@ -0,0 +1,33 @@
...
+
+import asyncio
+import json
+
+from anta.result_manager import ResultManager, TestResult
+
+from ansible_collections.arista.avd.plugins.plugin_utils.eos_validate_state_utils.models import InventoryDeviceHttpApi
+
+
+async def run_tests(manager: ResultManager, device: InventoryDeviceHttpApi, tests: list[tuple[callable[..., TestResult], dict]]) -> dict:
```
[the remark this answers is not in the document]
carl-baillargeon: Right this is the current runner in ANTA:

```
async def main(
    manager: ResultManager,
    inventory: AntaInventory,
    tests: List[Tuple[Callable[..., TestResult], Dict[Any, Any]]],
    tags: Optional[List[str]] = None,
    established_only: bool = True,
) -> None:
```

It takes the whole ANTA Inventory as input.

carl-baillargeon:
> I like the structure of this now. We will have to build something to handle the default catalog while still allowing the user to override the full catalog or just disable individual tests from the defaults. Also for 4.x we need to map in the ansible_run_tags to enable/disable tests in that default catalog.

Not sure how we should handle the tags... Should we have 1 tag per test or 1 per category should be enough?

ClausHolbechArista:
> > I like the structure of this now. We will have to build something to handle the default catalog while still allowing the user to override the full catalog or just disable individual tests from the defaults. Also for 4.x we need to map in the ansible_run_tags to enable/disable tests in that default catalog.
> 
> @ClausHolbechArista Not sure how we should handle the tags... Should we have 1 tag per test or 1 per category should be enough?

Since your implementation does not break any existing functionality we can do the tag part in a separate PR. The goal for that would be to replace the current ansible based tests 1:1.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/runner.py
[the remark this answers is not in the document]
ClausHolbechArista: Well we could run with that with a single host in the inventory?

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,74 @@
...
+            catalog_path = self._task.args.get("catalog_path")
+            with open(catalog_path, "r", encoding="UTF-8") as file:
+                input_catalog = safe_load(file)
+
+            # Raises error if the provided ANTA test catalog is empty
+            if not input_catalog:
+                raise AnsibleError(f"The provided test catalog is empty: {catalog_path}")
+
+            # Update the input catalog with AVD structured_config
+            updated_catalog = update_catalog(input_catalog, task_vars)
```
mtache, line 51: This did not worked with my inventory file that contains multiple module levels:
```
anta.tests.routing:
  generic:
    - VerifyRoutingProtocolModel:
        model: multi-agent
    - VerifyRoutingTableSize:
        minimum: 2
        maximum: 20
    - VerifyBFD:
  bgp:
    - VerifyBGPIPv4UnicastState:
    - VerifyBGPIPv4UnicastCount:
        number: 2
        template_params:
          - vrf: default
    - VerifyBGPIPv6UnicastState:
    - VerifyBGPEVPNState:
    - VerifyBGPEVPNCount:
        number: 2
    - VerifyBGPRTCState:
    - VerifyBGPRTCCount:
        number: 2
  ospf:
    - VerifyOSPFNeighborState:
    - VerifyOSPFNeighborCount:
        number: 3
```

I would suggest to first parse the catalog with the `anta.loader.parse_catalog()` function then update the return structure with AVD-provided test parameters. The `update_catalog()` function (maybe rename it then) would be simpler.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/runner.py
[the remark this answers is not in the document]
mtache: Here are the PRs:
https://github.com/arista-netdevops-community/anta/pull/220
https://github.com/arista-netdevops-community/anta/pull/224
Doc is coming :)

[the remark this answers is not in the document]
carl-baillargeon: Thanks @mtache I will work on that!

ansible_collections/arista/avd/roles/eos_validate_state/python_modules/base/avdbasetest.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
```
gmuloc, line 1: Move this to plugin_utils

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/models.py
ClausHolbechArista, line 1: Let's call this file `ansible-eos-device.py` since we would not want this for pyavd

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .catalog import Catalog
```
ClausHolbechArista: I think you should include all the things you include from external libs here. And then actually include them from the package level instead of the underlying files.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
```diff
@@ -0,0 +1,104 @@
+from __future__ import annotations
+
+from functools import cached_property
+
+from ansible.errors import AnsibleError
```
ClausHolbechArista, line 5: Please avoid using AnsibleError here. You can catch the errors in the action plugin and reraise as AnsibleActionError.

ansible_collections/arista/avd/roles/eos_validate_state/python_modules/base/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .avdbasetest import AvdBaseTest
```
ClausHolbechArista, line 1: Move this base class to plugin_utils

ansible_collections/arista/avd/roles/eos_validate_state/python_modules/ip_reachability/avdipreachabilitytest.py
```diff
@@ -0,0 +1,112 @@
...
+        for node in groups[fabric_name]:
+            if (loopback_interfaces := get(self.task_vars, f"hostvars.{node}.loopback_interfaces")) is None:
+                self.display.error(f"{node}: 'loopback_interfaces' is missing from structured_config.")
+                continue
+
+            for loopback_interface in loopback_interfaces:
+                if loopback_interface["name"] == "Loopback0":
+                    loopback_mappings.append((node, format(ipaddress.ip_interface(loopback_interface["ip_address"]).ip)))
+
+        if node_type == "l3leaf":
```
ClausHolbechArista, line 98: This should be refactored at some point. We have an open issue on that. We should not read any vars from eos_designs like `type`, `fabric_name` etc. and we should also not use Ansible's `groups`.

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -32,123 +32,153 @@
...
+    anta_logging: "{{ anta_logging }}"
+    skipped_tests: "{{ skipped_tests }}"
+  register: anta_results
+  tags:
+    - anta
+    - never
 
-- name: Display device platform and release information
-  delegate_to: localhost
+- name: Display ANTA results
```
ClausHolbechArista, line 50: This should be removed or at least have the anta tag on it.

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -32,123 +32,153 @@
...
+    - name: Remove generated vars for testing
+      ansible.builtin.file:
+        path: "{{ eos_validate_state_dir }}/validate_state_vars.yml"
+        state: absent
+      delegate_to: localhost
+      run_once: true
+      tags:
+        - always
+
+  # Legacy eos_validate_state block will NOT run in combinaison with ANTA
```
ClausHolbechArista, line 176:
```suggestion
  # Legacy eos_validate_state block will NOT run in combination with ANTA
```

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
```diff
@@ -0,0 +1,32 @@
+{# j2lint: disable=V1 #}
```
ClausHolbechArista, line 1: Why?

ansible_collections/arista/avd/requirements.txt
```diff
@@ -9,3 +9,4 @@ PyYAML>=5.4.1
 md-toc>=7.1.0
 deepmerge>=1.1.0
 cryptography>=38.0.4
+anta==0.7.1
```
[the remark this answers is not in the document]
ClausHolbechArista: @gmuloc says no...

ansible_collections/arista/avd/plugins/action/anta_runner.py
```diff
@@ -0,0 +1,35 @@
...
+        ansible_connection = self._connection
+        anta_logging = self._task.args.get("anta_logging")
+        skipped_tests = self._task.args.get("skipped_tests")
+
+        try:
+            anta_results = get_anta_results(vars=dict(task_vars), connection=ansible_connection, logging=anta_logging, skipped_tests=skipped_tests)
+        except Exception as error:
+            raise AnsibleActionFail(message=str(error)) from error
+
+        result["changed"] = True
```
ClausHolbechArista, line 28: Actually nothing changes on the devices. So this should probably be False or unset.

ansible_collections/arista/avd/requirements.txt
```diff
@@ -9,3 +9,4 @@ PyYAML>=5.4.1
 md-toc>=7.1.0
 deepmerge>=1.1.0
 cryptography>=38.0.4
+anta==0.7.2
```
ClausHolbechArista, line 12:
```suggestion
anta==0.7.2 # Pinned to specific version until anta reached 1.0.0 and follows semver.
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/external_libs/__init__.py
```diff
@@ -0,0 +1,40 @@
...
+    HAS_ANTA = True
+except ImportError as imp_exc:
+    HAS_ANTA = False
+    ANTA_IMP_ERR = imp_exc
+
+# Ansible HTTPAPI import
+try:
+    from ansible_collections.ansible.netcommon.plugins.connection.httpapi import Connection
+
+    HAS_HTTPAPI = True
```
gmuloc, line 24: check if this is needed for HTTPAPI

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,103 @@
...
+from ansible.plugins.action import display
+
+from .external_libs import HAS_ANTA, HAS_HTTPAPI
+
+if HAS_ANTA and HAS_HTTPAPI:
+    from ansible_collections.arista.avd.plugins.plugin_utils.eos_validate_state_utils.ansible_eos_device import AnsibleEOSDevice
+    from ansible_collections.arista.avd.plugins.plugin_utils.eos_validate_state_utils.catalog import Catalog
+    from ansible_collections.arista.avd.plugins.plugin_utils.eos_validate_state_utils.constants import ACRONYM_CATEGORIES, AVD_TEST_CLASSES
+
+    from .external_libs import AntaInventory, Connection, ResultManager, anta_runner, parse_catalog, setup_logging
```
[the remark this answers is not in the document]
gmuloc: addressed in  de04df3

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/constants.py
```diff
@@ -0,0 +1,44 @@
...
+from ansible_collections.arista.avd.roles.eos_validate_state.python_modules.hardware import AvdTestHardware
+from ansible_collections.arista.avd.roles.eos_validate_state.python_modules.mlag import AvdTestMLAG
+from ansible_collections.arista.avd.roles.eos_validate_state.python_modules.system import AvdTestNTP, AvdTestReloadCause
+
+ACRONYM_CATEGORIES: set[str] = {"aaa", "mlag", "snmp", "bgp", "ospf", "vxlan", "stp", "igmp", "ip"}
+"""
+A set of network protocol or feature acronyms that should be represented in uppercase in the eos_validate_state report.
+"""
+
+AVD_TEST_CLASSES = {
```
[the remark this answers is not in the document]
gmuloc: addressed in https://github.com/aristanetworks/ansible-avd/commit/de04df313f6991308683742ffa0528b81d5d8ec9

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
gmuloc: addressed in https://github.com/aristanetworks/ansible-avd/commit/de04df313f6991308683742ffa0528b81d5d8ec9

ansible_collections/arista/avd/requirements.txt
gmuloc: addressed in https://github.com/aristanetworks/ansible-avd/commit/de04df313f6991308683742ffa0528b81d5d8ec9

our verify-requirement does not understand comments after the line so I put it above. will need to log an issue for this probably

ansible_collections/arista/avd/roles/eos_validate_state/templates/generate_anta_report_results.j2
```diff
@@ -0,0 +1,37 @@
...
+ Copyright (c) 2023 Arista Networks, Inc.
+ Use of this source code is governed by the Apache License 2.0
+ that can be found in the LICENSE file.
+#}
+{# j2lint: disable=V1 #}
+{# AVD State Validation Report #}
+{% if "anta" in ansible_run_tags %}
+eos_validate_state_report:
+{%     set test_id = namespace(value=0) %}
+{%     set copy_hostvars = namespace(hostvars) %}
```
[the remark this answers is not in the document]
gmuloc: addressed in 36cfa34

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/external_libs/__init__.py
gmuloc: removed

ansible_collections/arista/avd/requirements.txt
```diff
@@ -9,3 +9,5 @@ PyYAML>=5.4.1
 md-toc>=7.1.0
 deepmerge>=1.1.0
 cryptography>=38.0.4
+# Pinned to specific version until anta reaches 1.0.0 and follows semver.
+anta==0.7.2
```
gmuloc, line 13: will need to to be changed BEFORE we merge this PR to anta 0.8.0 once it is released with all the newer tests added

ansible_collections/arista/avd/roles/eos_validate_state/python_modules/tests/avdtesthardware.py
```diff
@@ -0,0 +1,46 @@
...
+            test_definition (dict): ANTA test definition.
+        """
+        anta_tests = [
+            # Single test in ANTA covering 2 current eos_validate_state tests
+            {"VerifyEnvironmentCooling": {"states": get(self.task_vars, "validation_role.pwr_supply_states", ["ok"])}},
+            {
+                "VerifyEnvironmentPower": {
+                    "states": get(self.task_vars, "validation_role.fan_states", ["ok"]),
+                    "result_overwrite": {"categories": self.categories, "description": "Power supply status", "custom_field": "Power supply"},
+                }
```
gmuloc, line 35: Looks like power supply and cooling tests have the wrong "states being picked up"

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/anta_runner.md
```diff
@@ -0,0 +1,34 @@
+<!--
```
ClausHolbechArista, line 1: This file should be removed.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        self.device_name = device_name
+        self.hostvars = hostvars
+        self.skipped_tests = skipped_tests
+
+        self.catalog = self._input_catalog(custom_catalog, custom_catalog_path) or self._default_catalog()
+
+    def _input_catalog(self, custom_catalog, custom_catalog_path) -> dict:
+        """
+        NOTE: Not being used by the anta_runner plugin yet. Only the default_catalog is used.
```
ClausHolbechArista, line 46: this note says it is not being used, but 3 lines above it is called in __init__. What do we want?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
```diff
@@ -0,0 +1,132 @@
...
+        Utility function to merge catalogs according to the ANTA catalog structure.
+
+        Args:
+            *args (dict): Catalogs to merge.
+
+        Returns:
+            dict: The merged catalog.
+        """
+        merged_catalog = {}
+        for catalog in args:
```
ClausHolbechArista, line 126: I think `deepmerge.always_merger(merged_catalog, *args, recursive=True)` could do this.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
+# Copyright (c) 2023 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+import logging
+from asyncio import run
+from json import loads
+
+from ansible.plugins.connection import ConnectionBase
```
ClausHolbechArista, line 10: Can we avoid importing ansible specific things here?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
...
+
+from ansible_collections.arista.avd.plugins.plugin_utils.errors import AristaAvdError
```
ClausHolbechArista, line 12: please use relative imports

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
...
+
+    HAS_YAML = True
+except ImportError:
+    HAS_YAML = False
+
+
+# https://ttl255.com/yaml-anchors-and-aliases-and-how-to-disable-them/
+class NoAliasDumper(yaml.SafeDumper):
+    def ignore_aliases(self, data):
+        return True
```
ClausHolbechArista, line 39: Should we move this to utils so we can use it in more places?
We aware that you may need more tricks to dump the data, if it comes from Ansible input vars, since Ansible wraps in their own unicode classes. Maybe consider taking the dumper as a kwarg as well, so you can import AnsibleDumper instead in the action plugin and use that here.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
...
+    catalog = Catalog(device_name, hostvars, skipped_tests=skipped_tests)
+
+    # NOTE: For now we only parse the default catalog
+    tests = parse_catalog(catalog.catalog)
+
+    if save_catalog_name is not None:
+        dumpable_ = {}
+        for test in tests:
+            test_class = test[0]
+            test_params = test[1]
```
ClausHolbechArista, line 127:
```suggestion
        for test_class, test_params in tests:
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
...
+        # Run
+        run(anta_runner(manager, inventory, tests))
+
+    # Save the results
+    results = loads(manager.get_results(output_format="json"))
+
+    # Format the data properly for the eos_validate_state report
+    for item in results:
+        item["categories"] = ", ".join(
+            [" ".join([word.upper() if word.lower() in ACRONYM_CATEGORIES else word.title() for word in category.split()]) for category in item["categories"]]
```
ClausHolbechArista, line 154: Coudl we build some of these lists first to make it more readable?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
```diff
@@ -0,0 +1,174 @@
...
+        )
+
+        item["messages"] = "\n".join(item["messages"]).replace(",", "-")
+
+    return results
+
+
+def _create_dry_run_report(device_name: str, tests: list[tuple], manager: ResultManager) -> None:
+    """ """
```
ClausHolbechArista, line 163: ?

ansible_collections/arista/avd/tests/unit/plugins/filter/test_range_expand.py
```diff
@@ -9,7 +9,36 @@
 
 from ansible_collections.arista.avd.plugins.filter.range_expand import AnsibleFilterError, FilterModule, range_expand
 
-RANGE_TO_EXPAND_INVALID_VALUES = [True, {"key": "value"}, 33]
+RANGE_TO_EXPAND_INVALID_VALUES = [
```
ClausHolbechArista, line 12: Unrelated changes

ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/eos_validate_state_utils/test_catalog.py
```diff
@@ -0,0 +1,13 @@
...
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+# import pytest
+
+
+class TestCatalog:
+    def test(self):
+        pass
```
ClausHolbechArista, line 13: Add a TODO here, so we can search for it.

ClausHolbechArista (changes requested):
Looking good, but a few comments.

ansible_collections/arista/avd/molecule/eos_validate_state/inventory/group_vars/FABRIC.yml
```diff
@@ -0,0 +1,91 @@
...
+  - types: [ l2leaf ]
+    platforms: [ vEOS-lab ]
+    uplink_interfaces: [ Ethernet1-2 ]
+
+# L3 Edge port definitions. This can be any port in the entire Fabric, where IP interfaces are defined.
+l3_edge:
+  # Define a new IP pool that will be used to assign IP addresses to L3 Edge interfaces.
+  p2p_links_ip_pools:
+    - name: DCI_IP_pool
+      ipv4_pool: 172.100.100.0/24
```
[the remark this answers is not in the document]
gmuloc: This scenario is a verbatim copy of our example `dual-dc-l3ls`
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/examples/dual-dc-l3ls/group_vars/FABRIC.yml#L75
do you want me to fix it there as well?

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -0,0 +1,102 @@
...
+        hostvars = task_vars["hostvars"]
+
+        try:
+            anta_results = get_anta_results(
+                device_name=hostname,
+                hostvars=hostvars,
+                connection=ansible_connection,
+                anta_logging=anta_logging,
+                skipped_tests=skipped_tests,
+                ansible_tags=ansible_tags,
```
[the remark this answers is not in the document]
gmuloc: The idea is to not support this later on - this is on backward compatibility for ansible tags to behave like existing eos_validate_state but in the future, only `skipped_tests` should be used.
That's the reason why I made this variable called "ansible_tags" but happy to rename it tags instead
What do you think?

ansible_collections/arista/avd/molecule/eos_validate_state/reports/FABRIC-state.md
```diff
@@ -0,0 +1,790 @@
+
```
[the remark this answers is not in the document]
gmuloc: This is auto generated by validate state and we have not touched this in this PR - I just removed it but this will impact existing user of eos_validate_state

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -165,6 +165,11 @@ def range_expand(range_to_expand):
                     def expand_subinterfaces(interface_string):
                         result = []
                         if last_subinterface is not None:
+                            if first_subinterface > last_subinterface:
```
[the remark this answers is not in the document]
gmuloc: Doh!

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
gmuloc: it is used in `init` but you cannot pass these argument to the function from ansible_module

gmuloc: I guess the issue here is that I renamed anta_runner in eos_validate_state_runner and so this comment is very obscure

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
gmuloc: Hmmm - that is probably the case that deepmerge has something for us but there is no `recursive` keyword in always_merger. Will look into it

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
gmuloc: not really because we want an AnsibleEOSDevice class and we need to pass it the Connection from the action plugin to instantiate it - though we can import it only on TYPE_CHECKING for this specific file but we cannot avoid this for the Ansible part.

That being said when moving to pyavd, we will need to modify this function to use AsyncEOSDevice from ANTA as we won't have the Ansible connection to leverage for standalone pyavd so we will need to find a way to give another type of input and not the Connection for this. (or make it optional).

Probably worth discussing on a call

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
gmuloc: This is using the same import as `eos_designs_shared_utils` and `eos_designs_facts` plugins - should we change all the plugins imports?
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/utils.py#L9
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts/eos_designs_facts.py#L9

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
gmuloc: So for now even with Ansible we don't need the AnsibleDumper from what I have seen but it is quite possible that we will need it so I have moved the NoAliasDumper to utils.

I have taken the opportunity to move the dumping logic into the Catalog object which makes much more sense so thanks for this comment!

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/get_anta_results.py
gmuloc: what's the question? This is used to generate a report when using the dry_run options with all tests that would be run.
(check mode in ansible) - we have discussed this before.

Is it about the docstring?

ansible_collections/arista/avd/tests/unit/plugins/filter/test_range_expand.py
gmuloc: doh! (bis)

ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/eos_validate_state_utils/test_catalog.py
gmuloc: Fixed in [5eebda8](https://github.com/aristanetworks/ansible-avd/pull/2907/commits/5eebda8f72510128bffa100a3018a8c3b446eb51)

ansible_collections/arista/avd/tests/unit/plugins/filter/test_range_expand.py
gmuloc: Fixed in [733156c](https://github.com/aristanetworks/ansible-avd/pull/2907/commits/733156cdbc2ccc64a5a8145a7b42a6d989994c92)

ansible_collections/arista/avd/plugins/filter/range_expand.py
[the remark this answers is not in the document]
gmuloc: Fixed in [733156c](https://github.com/aristanetworks/ansible-avd/pull/2907/commits/733156cdbc2ccc64a5a8145a7b42a6d989994c92)

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -0,0 +1,102 @@
...
+        ansible_connection = self._connection
+
+        # Handle logging
+        setup_module_logging(hostname, result)
+
+        # Get task arguments
+        anta_logging = new_module_args.get("anta_logging", "WARNING")
+        skipped_tests = new_module_args.get("skipped_tests", {})
+
+        save_catalog = new_module_args.get("anta_save_catalog", False)
```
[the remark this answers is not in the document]
gmuloc: Addressed in [5eebda8](https://github.com/aristanetworks/ansible-avd/pull/2907/commits/5eebda8f72510128bffa100a3018a8c3b446eb51)

PR #2946, opened 2023-06-01
Feat(eos_designs): Compact MLAG allocations #2903

### LINKED ISSUE #2903: Feature: Add more compact MLAG ip addressing scheme
### Enhancement summary

Currently the MLAG peer link IPs are allocated with a /31 but the current algorithm skips the next /31 because of possible Node-ID collisions.

eos_designs

### Use case example

More effective use of MLAG peer IP pools

This feature would introduce an alternate algorithm which requires one odd and one even ID in every MLAG group, and then use only the odd ID for selecting the /31 from the pool. This should ensure that we get consistent allocations when the odd ID is either primary or secondary MLAG peer. (Today we only use the ID of the primary)

_No response_

### Additional context

_No response_

### Contributing Guide

haylinmoore:
## Change Summary

Add a compact MLAG boolean flag for nodes which forces MLAG pairs to have an odd and even ID allowing more for compact numbering. Currently the only generation algorithm will skip every other /31 causing extra space to be wasted.

## Related Issue(s)
Fixes #2903

## Component(s) name
`arista.avd.eos-designs`

## Proposed changes
* Add an option `mlag_compact_addressing` to the eos-designs
* In avdipaddressing.py check if `mlag_compact_addressing` is set to `True` then
    - Verify MLAG pair has a node with an even ID and an odd ID
    - Calculate Offset using the odd ID
    - Run AvdIpAddressing._ip() in a way that skips multiply by the subnet_size, which previously was causing the wasted /31

## How to test
Add molecule tests, currently waiting for Carl to finish 4.0.0 molecule migration

### Repository Checklist


pvinci-arista:
How can we document that all mlag node_types for a given address pool need to be configured consistently with `mlag_compact_addressing` to prevent an address collision.

 Can we add a molecule test to cover this code?

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -18,13 +18,13 @@ class AvdIpAddressing(AvdFacts, UtilsMixin):
...
 
-    def _ip(self, pool: str, prefixlen: int, subnet_offset: int, ip_offset: int) -> str:
+    def _ip(self, pool: str, prefixlen: int, subnet_offset: int, ip_offset: int, add_subnet_space: bool = True) -> str:
         pool_network = ipaddress.ip_network(pool, strict=False)
         prefixlen_diff = prefixlen - pool_network.prefixlen
         subnet_size = (int(pool_network.hostmask) + 1) >> prefixlen_diff
         if (subnet_offset + 1) * subnet_size > pool_network.num_addresses:
             raise AristaAvdError(f"Unable to get {subnet_offset + 1} /{prefixlen} subnets from pool {pool}")
-        subnet = ipaddress.ip_network((int(pool_network.network_address) + subnet_offset * subnet_size, prefixlen))
+        subnet = ipaddress.ip_network((int(pool_network.network_address) + subnet_offset * (subnet_size if add_subnet_space else 1), prefixlen))
```
ClausHolbechArista: I don't understand why you need to change this. We should not need to modify this neutral function for this corner case

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -35,6 +35,33 @@ def _template(self, template_path, **kwargs):
...
+
+        # Verify a mix of odd and even IDs
+        if (self._mlag_primary_id % 2) + (self._mlag_secondary_id % 2) != 1:
+            raise AristaAvdError(f"MLAG compact addressing mode requires all MLAG pairs to have a single odd and even ID")
+
+        odd_id = self._mlag_primary_id
+        if odd_id % 2 == 0:
+            odd_id = self._mlag_secondary_id
+        
+        return odd_id - 1
```
ClausHolbechArista: Instead of your change to the base _ip function, you should `return (odd_id - 1) / 2`.
So:
ID: 1 -> offset 0
ID: 3 -> offset 1
etc.

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
```diff
@@ -4605,6 +4605,12 @@ $defs:
               '
             type: str
             format: ipv4_cidr
+          mlag_compact_addressing:
+            description: |
+              Generate MLAG ip addresses based on the odd ID of an ID pair, this a /31 per allocation
+              Requires MLAG pairs have nodes with an odd and even ID
+            default: false
+            type: bool
```
ClausHolbechArista: I would prefer to build something more generic, since we will have other algorithms going forward (like reusing the same subnet everywhere).
```yaml
ip_addressing:
  mlag:
    algorithm: < 'first_id' | 'odd_id' | default -> 'first_id' >
```
Make sure to describe the two in the description for `algorithm` including the downside of skipping every other subnet for first_id (in favor of flexibility in IDs) and the more restrictive ID scheme for 'odd_id'

.gitignore
```diff
@@ -97,6 +97,9 @@ ENV/
 env.bak/
 venv.bak/
 
+# Pipfile
```
ClausHolbechArista: Why do we need this?

ClausHolbechArista (changes requested):
Please add a test case to `eos_designs_unit_tests` using the new ID method.

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
haylinmoore: That did not solve it when I tried that

ClausHolbechArista: We should just tell `_ip()` which subnet we want of a certain size. Not hack it to behave differently.

haylinmoore: Turns out it was a weird behavior with returning a float 1.0 vs int 1, that broke it when I tried to just divide.

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
haylinmoore: Added

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -35,6 +35,33 @@ def _template(self, template_path, **kwargs):
         template_vars = ChainMap(kwargs, self._hostvars)
         return self.shared_utils.template_var(template_path, template_vars)
 
+    def _mlag_ip(self, pool: str, ip_offset: int) -> str:
+        addressing_alg = self._ipaddress_mlag_algorithm
+
+        if addressing_alg == "odd_id":
```
ClausHolbechArista:
```suggestion
        if self._ipaddress_mlag_algorithm == "odd_id":
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -35,6 +35,33 @@ def _template(self, template_path, **kwargs):
...
+
+        # Verify a mix of odd and even IDs
+        if (self._mlag_primary_id % 2) + (self._mlag_secondary_id % 2) != 1:
+            raise AristaAvdError("MLAG compact addressing mode requires all MLAG pairs to have a single odd and even ID")
+
+        odd_id = self._mlag_primary_id
+        if odd_id % 2 == 0:
+            odd_id = self._mlag_secondary_id
+
+        return int(((odd_id - 1) / 2))
```
ClausHolbechArista:
```suggestion
        return int((odd_id - 1) / 2)
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/avdipaddressing.py
```diff
@@ -35,6 +35,33 @@ def _template(self, template_path, **kwargs):
...
+
+    def _mlag_odd_id_based_offset(self) -> int:
+        """
+        Return the subnet offset for an MLAG pair based on odd id
+
+        Requires a pair of odd and even IDs
+        """
+
+        # Verify a mix of odd and even IDs
+        if (self._mlag_primary_id % 2) + (self._mlag_secondary_id % 2) != 1:
```
ClausHolbechArista:
```suggestion
        if (self._mlag_primary_id % 2) == (self._mlag_secondary_id % 2):
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/ip_addressing.schema.yml
```diff
@@ -0,0 +1,24 @@
...
+          algorithm:
+            type: str
+            convert_to_lower_case: true
+            default: "first_id"
+            description: |
+              - This variable defines the Multi-chassis Link Aggregation (MLAG) algorithm used.
+              - The following MLAG algorithms are supported:
+                - first_id
+                - odd_id
+              - By default, the 'first_id' algorithm is used.
```
ClausHolbechArista: - No need to mention the default value in the description, since there is already a field for that above.
- No need for `convert_to_lower_case` since we only accept specific values and there is no history here.
- Please add a description for each option in the list you mention. Especially we need to mention the restriction on IDs when using `odd_id`

ClausHolbechArista (changes requested):
A few comments and then please look into adding a molecule test under the `eos_designs_unit_tests` scenario.

ClausHolbechArista (changes requested):
A few comments and then please look into adding a molecule test under the `eos_designs_unit_tests` scenario.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/ip_addressing.schema.yml
```diff
@@ -0,0 +1,24 @@
+# yaml-language-server: $schema=../../../../plugins/plugin_utils/schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  ip_addressing:
```
ClausHolbechArista: Change this to `fabric_ip_addressing`

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/mlag.py
```diff
@@ -61,6 +61,10 @@ def mlag_peer_ipv4_pool(self: SharedUtils) -> str:
     def mlag_peer_l3_ipv4_pool(self: SharedUtils) -> str:
         return get(self.switch_data_combined, "mlag_peer_l3_ipv4_pool", required=True)
 
+    @cached_property
```
ClausHolbechArista: I don't think we need this method any longer

ansible_collections/arista/avd/roles/eos_designs/docs/Input Variables.md
```diff
@@ -0,0 +1,4458 @@
+---
+search:
+  boost: 2
```
ClausHolbechArista: If you rebase from latest devel this file should not be there any longer. Otherwise try to delete it locally and rerun pre-commit.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_node_type.schema.yml
```diff
@@ -637,6 +637,12 @@ $defs:
             documentation_options:
               table: node-type-l2-mlag-configuration
             type: str
+          mlag_compact_addressing:
```
ClausHolbechArista: Please remove this again, since we use the fabric_ip_addressing construct instead

.gitignore
ClausHolbechArista: this should be removed

ClausHolbechArista (approved):

emilarista (approved):
Tested and reviewed, LGTM!

PR #2950, opened 2023-06-01
Fix(eos_designs): eBGP rfc5549 creates invalid configuration for MLAG scenarios

### LINKED ISSUE #2906: IBGP neighbors could be established for MLAG peers in each VRF when we set ‘overlay_mlag_rfc5549: true’
I just found some strange things when i tried to use ‘underlay_rfc5549: true’ and ‘overlay_mlag_rfc5549: true’.

Firstly, when i tried to set underlay_rfc5549 to true only, the overlay ibgp peers for mlag peers in each vrf are trying to use ipv4 addresses to establish ibgp neighbors (it will take an address from ‘mlag_peer_l3_ipv4_pool’), but in AF ipv4, it assigned command “neighbor <MLAG_PEER_GROUP> next-hop address-family ipv6 originate” (because of underlay mlag ibgp peers in vrf default) to only accept the updates which next-hop is ipv6 address, then, imported/networked routes will be dropped between each mlag-peer (error msg on the devices in syslog: because a local nexthop was not configured for afi/safi ipv6/unicast). So I couldn't see the routes from mlag peer device only (peer establishment is OK).

Secondly, If I set the var ‘overlay_mlag_rfc5549: true’ to deal with the issue above. Yes, it will change the overlay p2p mlag peer established in different vrfs via IPv6 link-local addr. But when I deployed the configs into devices, I found there are no bgp peers established in each VRFs anymore. I guess that’s because ipv6 unicast-routing is not enabled in each vrfs (there is only one command for vrf default: ‘ipv6 unicast-routing’). When I configured those commands ‘ipv6 unicast-routing vrf xxx’ manually, bgp peers are back in each VRFs. I think this/ese command/s should be generated automatically to support this variable parameter? For this point, i just had a quick workaround that we could modify "arista/avd/roles/eos_designs/python_modules/network_services/vrfs.py":

```
class VrfsMixin(UtilsMixin):
    <! --- omit some outputs  --- !> 
    def vrfs(self) -> dict | None:
        <! --- omit some outputs  --- !> 
        for tenant in self._filtered_tenants:
            for vrf in tenant["vrfs"]:
                <! --- omit some outputs  --- !> 
                if self._overlay_mlag_rfc5549 and self._mlag_ibgp_peering_enabled(vrf, tenant):
                    new_vrf["ip_routing_ipv6_interfaces"] = True
                    new_vrf["ipv6_routing"] = True
                else:
                    new_vrf["ip_routing"] = True
                <! --- omit some outputs  --- !> 
        if vrfs:
            return vrfs
```
It could generate the ipv6 routing vrf xxx when it has the corresponding vrfs and mlag overlay peers.

carlbuchmann:
## Change Summary

Fix issues in the eBGP rfc5549 design with MLAG configuration scenario as described in #2906 

## Related Issue(s)

Fixes #2906 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

If underlay_rfc5549 is True and overlay_mlag_rfc5549 is True configure:

```eos
ipv6 unicast-routing vrf <vrf_name>
```

If underlay_rfc5549 is True and overlay_mlag_rfc5549 is False configure:

```eos
router_bgp <asn>
   vrf <vrf_name>
      address-family ipv4
         no neighbor <mlag-peer-ip> next-hop address-family ipv6
```

## How to test

Run Molecule scenario: `molecule converge -s evpn_underlay_rfc5549_overlay_ebgp`

### User Checklist

- [x] Update test case for evpn_underlay_rfc5549_overlay_ebgp
- [x] Fix underlay_rfc5549 is True and overlay_mlag_rfc5549 is True
- [x] Add eos_cli_config_gen knob for router_bgp.vrfs<vrf>.address_family_ipv4."no neighbor <ip_address> next-hop address-family ipv6"
- [x] Fix If underlay_rfc5549 is True and overlay_mlag_rfc5549 is False

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/Input Variables.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/Input Variables.md
@@ -5158,6 +5158,10 @@ Set Link Aggregation Control Protocol (LACP) parameters.
     | [<samp>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;activate</samp>](## "router_bgp.vrfs.[].address_family_ipv4.neighbors.[].activate") | Boolean |  |  |  |  |
     | [<samp>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;route_map_in</samp>](## "router_bgp.vrfs.[].address_family_ipv4.neighbors.[].route_map_in") | String |  |  |  | Inbound route-map name |
     | [<samp>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;route_map_out</samp>](## "router_bgp.vrfs.[].address_family_ipv4.neighbors.[].route_map_out") | String |  |  |  | Outbound route-map name |
+    | [<samp>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;next_hop</samp>](## "router_bgp.vrfs.[].address_family_ipv4.neighbors.[].next_hop") | Dictionary |  |  |  |  |
+    | [<samp>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;address_family_ipv6</samp>](## "router_bgp.vrfs.[].address_family_ipv4.neighbors.[].next_hop.address_family_ipv6") | Dictionary |  |  |  |  |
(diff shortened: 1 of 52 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/Input Variables.md
```diff
@@ -5725,6 +5729,10 @@ Set Link Aggregation Control Protocol (LACP) parameters.
                 activate: <bool>
                 route_map_in: <str>
                 route_map_out: <str>
+                next_hop:
```
carlbuchmann, line 5732: @ClausHolbechArista 
Let me know if you agree with the proposed schema before I continue.
I kept the same schema as we have on the peer group and validated options on the command line.
The following forms are accepted at the CLI:
```
router bgp 65000
   vrf test1
      address-family ipv4
         neighbor 10.255.251.3 next-hop address-family ipv6
   vrf test2
      address-family ipv4
         neighbor 10.255.251.3 next-hop address-family ipv6 originate
   vrf test3
      address-family ipv4
         no neighbor 10.255.251.3 next-hop address-family ipv6
```

ClausHolbechArista: agreed. But I think we probably need to remove peer-groups from the vrf context completely (or at least most of it) since EOS does not accept that (it automatically moved the config to the global address family context).
Not related to this directly, so separate PR.

ClausHolbechArista:
Tested in lab where the issue was recreated, and these fixes work for both cases.

ClausHolbechArista (approved):

carlbuchmann:
@ClausHolbechArista thank you - LGTM, added unit tests for eos_cli_config_gen

onurgashi:
LGTM! Thanks for fixing this.

onurgashi (approved):

PR #2981, opened 2023-06-13
Doc: Remove wide margins from left and right menus

ClausHolbechArista:
## Change Summary

Remove wide margins from left and right menus on documentation generated on avd.sh and avd.arista.com


carlbuchmann (approved):
LGTM

PR #2997, opened 2023-06-21
Fix(eos_cli_config_gen): Force domain_identifier to be a string

haylinmoore:
## Change Summary
Currently domain_identifier accepts both str and int due to it's convert_type value. When parsing the YAML, pyyaml reads `domain_identifier: 65000:0` as a sexagesimal float converting it to 3900000. We then accept that due to the convert_type making it "3900000" causing the domain_identifier output to be `domain identifier 3900000` in the config. This change forces domain_identifier to be a str in the YAML because otherwise users could run into this and be quite confused

## Component(s) name

## Proposed changes
Remove the convert_types from relevant domain_identifiers
```
          domain_identifier:
            type: str
            convert_types:
            - int
```
to
```
          domain_identifier:
            type: str
```
## How to test
Updating the relevant molecules to reflect the correction functionality
### Repository Checklist


carlbuchmann (approved):
LGTM

ClausHolbechArista (approved):
Thank you for fixing this

PR #3021, opened 2023-07-05
Feat: Add parenthesis syntax for interface ranges

### LINKED ISSUE #2768: Add parenthesis syntax for interface ranges
### Enhancement summary

Some platforms have these kind of breakout layout

```
network_ports:
  - switches:
      - MY-AWESOME-SWITCHES
    switch_ports:
      # Special ports
      # Ethernet 26, 28, 30 and 32 have only 1 and 3
      # Ethernet 33 and 35 have from 1-8
      - Ethernet9-25/1-4
      - Ethernet26/1
      - Ethernet26/3
      - Ethernet27/1-4
      - Ethernet28/1
      - Ethernet28/3
      - Ethernet29/1-4
      - Ethernet30/1
      - Ethernet30/3
      - Ethernet31/1-4
      - Ethernet32/1
      - Ethernet32/3
      - Ethernet33/1-8
      - Ethernet34/1-4
      - Ethernet35/1-8
      - Ethernet36/1-4
 ```

Adding parenthesis would help reduce the snippet above to something like

```
network_ports:
  - switches:
      - MY-AWESOME-SWITCHES
    switch_ports:
      # Special ports
      # Ethernet 26, 28, 30 and 32 have only 1 and 3
      # Ethernet 33 and 35 have from 1-8
      Ethernet(9-25,27,29,31,34)/1-4
      Ethernet(26,28,30,32)/(1,3)
      Ethernet(33,35)/1-8
```

plugins

### Use case example

cf description

parenthesis for ranges - cf description

current functionality is working fine

### Additional context

_No response_

### Contributing Guide

### PR DESCRIPTION
**Pull request by Vibhu-gslab**
## Change Summary

Add parenthesis syntax for interface ranges

## Related Issue(s)

Fixes #2768 

## Component(s) name

`arista.avd.range_expand`

## Proposed changes
Same as the description of issue.

### User Checklist

- N/A

### Repository Checklist

### CODE REVIEW

**Review by gmuloc — COMMENTED** [review]
Can you please add some tests to the PR and make sure that your code can handle bogus syntaxes:

```
- Eth(1,2
- Eth(1,2)/3,4)
- Eth(1,2/3,4)
- ...
```

File: ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py [hunk 1]
```diff
@@ -50,7 +50,7 @@ def test_get_configlet(self, DATA):
 
     def test_get_configlet_invalid_source(self):
         output = get_configlet()
-        assert output == {}
+        assert not output
```
Vibhu-gslab: Correcting pre-commit warning

Vibhu-gslab:
Please let me know if any other testcase needs to get added.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -25,6 +25,22 @@
...
     "65100.0-2,65200.1-2",
     "1-2.0-1",
+    "Ethernet(1)",
+    "Ethernet(1,2,3)",
+    "Ethernet(1)/(4,5)",
+    "(1,2)",
+    "1,(2,3)/5",
+    "Eth(10-14)",
+    "Eth(101-103)/(5,7)",
+    "Eth(1,2",
```
gmuloc: Surely this should fail?

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -25,6 +25,22 @@
...
+    "Eth(1,2)/3,4)",
```
gmuloc: is this valid syntax? - what are we trying to do here? We should not accept extra parenthesis right?

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,6 +10,25 @@
 from ansible.errors import AnsibleFilterError
 
 
+def extend_range(one_range):
+    regex = r"(\(\s*(?:\d*)(?:(?:\s*,\s*\d+)*(?:\s*,?\s*\d+\-\d+)*)*\s*\))|([0-9]+\-[0-9]+)"
```
gmuloc: Please document what you are doing here like it used to be so that people can follow what the regex is doing

gmuloc (changes requested):

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Yes, this is invalid syntax for parentheses "()" or range "-".
As discussed with @ClausHolbechArista , our requirement was to only expand if it falls under "()" or "-" otherwise let it pass through, since we want to add more expand parameters later.
So in this case since the syntax does not matches with "()" or "-" it is going to pass through with the output of prefix + split(",")
which would be "Eth(1,2" --> ["Eth(1", "Eth2"]

I understand this would not be the expected output from the function we could discuss about errors over call if you like :)

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Same is the case with this one.. 
Since the syntax is invalid it will pass through and return prefix + split(",") for the invalid part and expand for the valid part which is the first half of the input "Eth(1,2)/3"
So, for "Eth(1,2)/3,4)" --> ["Eth1/3", "Eth2/3", "Eth4)"]

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: My apologies, Will do :)

Vibhu-gslab: Done!

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
ClausHolbechArista: I don't follow this logic.
The `(1,2)` should be the only expansion here, since it is the only expression matching the syntax.
So I would expect `["Eth1/3,4)", "Eth2/3,4)"]`.
If we wish, we can indeed decide to implement checks for the unmatched parentheses.

ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
ClausHolbechArista: Isn't this too "loose"? None, False, "", [], {} would all pass this assertion.

ClausHolbechArista: This change is unrelated to this PR, so please remove it again.

ClausHolbechArista (commented):
I still feel like this code is too specific on expected syntaxes, instead of being a string parser expanding each element on it's own, and calling itself recursively.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Any "," outside the parentheses are used for splitting the string.
For Ex. 
If input is "Eth2/3,4" it will return ["Eth2/3", "Eth4"] since we take the prefix of previous port if not given.
in similar fashion "Eth(1,2)/3, 4)" will return ["Eth1/3", "Eth2/3", "Eth4)"]

Vibhu-gslab:
> I still feel like this code is too specific on expected syntaxes, instead of being a string parser expanding each element on it's own, and calling itself recursively.

May be we can discuss it over call that would give more clear idea on the missing part.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
ClausHolbechArista: You are correct.

ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
Vibhu-gslab: Removed it :)

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,128 +10,69 @@
...
+
+        if re.search(r",(?![^(]*\))", one_range):
+            # Split comma outside parentheses
+            result.extend(range_expand(re.split(r",(?![^(]*\))", one_range), prefix))
+        elif parentheses_sr := re.search(r"(\(\s*(?:\d+)(?:(?:\s*,\s*\d+)*(?:,?(?:\d+)?\-\d+)*)*\s*\))", one_range):
+            # Expand parentheses
+            result.extend(range_expand(expand_parentheses(parentheses_sr, one_range), prefix))
+        elif hypen_sr := re.search(r"([0-9]+\-[0-9]+)", one_range):
+            # Expand hypen
+            result.extend(range_expand(expand_hypen(hypen_sr, one_range), prefix))
```
ClausHolbechArista:
```suggestion
        elif hyphen_sr := re.search(r"([0-9]+\-[0-9]+)", one_range):
            # Expand hyphen
            result.extend(range_expand(expand_hyphen(hyphen_sr, one_range), prefix))
```

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,128 +10,69 @@
...
+        else:
+            if one_range:
+                result.extend([one_range])
+            else:
+                result.extend(one_range)
```
ClausHolbechArista: I don't understand the logic here.

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,128 +10,69 @@
...
+        raise AnsibleFilterError(f"value must be of type list or str, got {type(one_range)}")
 
     result = []
 
     # If we got a list, unpack it and run this function recursively
-    if isinstance(range_to_expand, list):
-        for r in range_to_expand:
-            result.extend(range_expand(r))
+    if isinstance(one_range, list):
+        for r in one_range:
```
ClausHolbechArista: please avoid abbreviated vars, since it makes it harder to read the code. But also avoid reserved words when you use full words :)

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,128 +10,69 @@
...
+        prefix_group = sr.groups()
+        if prefix_group[0]:
+            prefix = prefix_group[0]
+        else:
+            one_range = prefix + one_range
+
+        if re.search(r",(?![^(]*\))", one_range):
+            # Split comma outside parentheses
+            result.extend(range_expand(re.split(r",(?![^(]*\))", one_range), prefix))
+        elif parentheses_sr := re.search(r"(\(\s*(?:\d+)(?:(?:\s*,\s*\d+)*(?:,?(?:\d+)?\-\d+)*)*\s*\))", one_range):
```
ClausHolbechArista: Please move all regex to constants with descriptions on what you match (and don't match).

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -10,128 +10,69 @@
 from ansible.errors import AnsibleFilterError
 
 
-def range_expand(range_to_expand):
-    if not (isinstance(range_to_expand, list) or isinstance(range_to_expand, str)):
-        raise AnsibleFilterError(f"value must be of type list or str, got {type(range_to_expand)}")
+def expand_parentheses(range_sr, one_range):
```
ClausHolbechArista: I think we should move all the code to `plugin_utils/utils/range_expand.py` and then you can import from there to the filter. After moving, you should add proper type hints on all arguments and return type for functions. Please also add docstrings to explain the purpose of each function.

ansible_collections/arista/avd/plugins/filter/range_expand.py
```diff
@@ -9,129 +9,65 @@
 
 from ansible.errors import AnsibleFilterError
 
+from ansible_collections.arista.avd.plugins.plugin_utils.utils.range_expand import expand_hyphen, expand_parentheses
 
-def range_expand(range_to_expand):
+
+def range_expand(range_to_expand, prefix=""):
```
ClausHolbechArista: Please also move this function to the plugin_utils and add type hints. There might be some imports in pyavd you need to update.

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
```diff
@@ -0,0 +1,61 @@
+def expand_parentheses(range_sr: object, range_to_expand: str) -> list:
```
ClausHolbechArista: Please import the correct type (`import Match from re`) and set as type hint, since that will make it easier to read the code below.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -25,6 +25,24 @@
...
+    ["(1-5)", "(2,3,4)", "14-16"],
+    ["Ethernet 1-2/(3-5)", "Eth3-5/(3,4,5)", "7-8/(3,4,5)"],
+    ["Et (3,4,5).1", "Et4/(3,4,5).1", "Et5.200", "Et(4-6).200"],
+    "(65000).1",
+    ["Pt(600, 7000)/2.1", "Pt( 600, 7000)/2.1", "Pt(600, 7000 )/2.1"],
+    "eth1-5,(7,9,11-13)",
+    "eth(7,9,11-13),21,26",
 ]
 
 EXPECTED_RESULT_VALID_VALUES = [
```
ClausHolbechArista: Can you refactor the constants here to be a list of tuples, where the value is the first element and the expected result is the second? That will make it easier to pair them up and ensure we have the logic we want.
Thanks

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: So, shall I remove range_expand() from filters or call it inside filters by importing from plugin_utils?

ClausHolbechArista: yes please

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: changed it.

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: Fixed

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: Fixed

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: Done

ansible_collections/arista/avd/plugins/filter/range_expand.py
Vibhu-gslab: Done

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
Vibhu-gslab: Fixed

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Done :)

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
```diff
@@ -0,0 +1,147 @@
...
+    replaced_list = []
+    if groups[0]:
+        # To replace the values of range, Ex. 4-6
+        range_start, range_end = groups[0].split("-")
+        for port_num in range(int(range_start), int(range_end) + 1):
+            replaced_list.append(range_to_expand.replace(range_to_expand[start_index:end_index], str(port_num), 1))
+    return replaced_list
+
+
+def range_expand_utils(range_to_expand: str or list, prefix: str = "") -> list:
```
ClausHolbechArista:
```suggestion
def range_expand_utils(range_to_expand: str | list, prefix: str = "") -> list:
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
```diff
@@ -0,0 +1,147 @@
...
+        #                                         (?:\d+)                                                matches 1 or more numbers Ex. 1, 21
+        #                                                   (?:\s*,\s*\d+)*                              matches 0 or more apperance of "," and one
+        #                                                                                                             or more numbers Ex. ,2,23
+        #                                                                  (?:,?\d+\-\d+)*               matches 0 or more apperance of "," and one or more
+        #                                                                                                             numbers with range (-) Ex. ,2-4
+        #                                         (?:\d+)(?:(?:\s*,\s*\d+)*(?:,?\d+\-\d+)*)*             matches values similar to 1 or 1,2,3 or 1,2-4
+        #                                   (\(\s*(?:\d+)(?:(?:\s*,\s*\d+)*(?:,?\d+\-\d+)*)*\s*\))       matches values similar to 1 or 1,2,3 or 1,2-4
```
ClausHolbechArista: This may be missing an update compared to line 121

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -8,43 +8,95 @@
...
+    ("Ethernet(1)/(4,5)", ["Ethernet1/4", "Ethernet1/5"]),
+    ("(1,2)", ["1", "2"]),
+    ("1,(2,3)/5", ["1", "2/5", "3/5"]),
+    ("Eth(10-14)", ["Eth10", "Eth11", "Eth12", "Eth13", "Eth14"]),
+    ("Eth(101-103)/(5,7)", ["Eth101/5", "Eth102/5", "Eth103/5", "Eth101/7", "Eth102/7", "Eth103/7"]),
+    (
+        "Eth(1,2",
+        ["Eth(1", "Eth2"],
+    ),
+    # ("Eth(1,2)/3,4)", ["Eth1/3", "Eth2/3", "Eth4)"]), # Exception need to fix regex_commas_outside_parentheses regex in range_expand
```
ClausHolbechArista: Has this been addressed?

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -8,43 +8,95 @@
...
+            "Eth4/2/1/4.300",
+            "Eth4/3/1/4.300",
+            "Eth4/2/32/4.200",
+            "Eth4/2/32/4.201",
+            "Eth4/2/32/4.202",
+            "Eth4/3/32/4.200",
+            "Eth4/3/32/4.201",
+            "Eth4/3/32/4.202",
+            "Eth4/2/32/4.300",
+            "Eth4/3/32/4.300",
```
ClausHolbechArista: The ordering here is wrong, so we might still have to fix some things.
`Eth4/x/x/x.300` should not come later than the others. We need to expand from left-to-right to make it intuitive. (except commas outside, which should be first like now).

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
```diff
@@ -8,43 +8,95 @@
...
+            "Eth5/4",
+            "Eth3/5",
+            "Eth4/5",
+            "Eth5/5",
+            "7/3",
+            "8/3",
+            "7/4",
+            "8/4",
+            "7/5",
+            "8/5",
```
ClausHolbechArista: The ordering here is wrong. It should be
7/3, 7/4, 7/5, 8/3, 8/4, 8/5

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
Vibhu-gslab: pipeline was not working for me, I have tried it.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: I have not found the solution for this yet, so commented the testcase to notify.

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Fixed as discussed

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
Vibhu-gslab: Fixed

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
Vibhu-gslab: Updated

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
ClausHolbechArista: You should upgrade your python version. But you should also add `from __future__ import annotations` at the top of the file (below license)

ansible_collections/arista/avd/tests/unit/filters/test_range_expand.py
ClausHolbechArista: We should change these invalid tests to use { and }

Vibhu-gslab: Done

ansible_collections/arista/avd/plugins/plugin_utils/utils/range_expand.py
Vibhu-gslab: Done

carlbuchmann:
replaced by #3244

PR #3027, opened 2023-07-07
Refactor(eos_designs): Deprecating port_channel.short_esi under connected_endpoints

ghost:
## Change Summary

Deprecating port_channel.short_esi under connected_endpoints

## Related Issue(s)

Fixes #[86](https://github.com/aristanetworks/avd-internal/issues/86)

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


gmuloc (changes requested):
@Shivani-chourasiya / @carlbuchmann 
So the issue states that the key would be deprecated in 4.0 (which has not been done yet since there is no deprecation)

And then it should be removed in 5.0.0

However the PR right now targets devel which will be used for 4.2.0,...
So in my opinion you cannot set `removed: true` now

You should only deprecate it and output a deprecation warning and when we are working on releasing 5.0 we will need a new PR to remove the key (and implement the tests you have here)..

So for now the PR should only focus on outputing a deprecation warning when the key is used and when we are closer to release 5.0 we should remove it only.

Shivani-chourasiya:
> @Shivani-chourasiya / @carlbuchmann So the issue states that the key would be deprecated in 4.0 (which has not been done yet since there is no deprecation)
> 
> And then it should be removed in 5.0.0
> 
> However the PR right now targets devel which will be used for 4.2.0,... So in my opinion you cannot set `removed: true` now
> 
> You should only deprecate it and output a deprecation warning and when we are working on releasing 5.0 we will need a new PR to remove the key (and implement the tests you have here)..
> 
> So for now the PR should only focus on outputing a deprecation warning when the key is used and when we are closer to release 5.0 we should remove it only.

Sure @gmuloc, will update it

Shivani-chourasiya:
> @Shivani-chourasiya / @carlbuchmann So the issue states that the key would be deprecated in 4.0 (which has not been done yet since there is no deprecation)
> 
> And then it should be removed in 5.0.0
> 
> However the PR right now targets devel which will be used for 4.2.0,... So in my opinion you cannot set `removed: true` now
> 
> You should only deprecate it and output a deprecation warning and when we are working on releasing 5.0 we will need a new PR to remove the key (and implement the tests you have here)..
> 
> So for now the PR should only focus on outputing a deprecation warning when the key is used and when we are closer to release 5.0 we should remove it only.

@gmuloc if I remove `removed: true` from schema, it won't be raising error on using `port_channel.short_esi`. Shall I remove the test-case from eos_designs_negative_unit_tests for now?

gmuloc:
> > @Shivani-chourasiya / @carlbuchmann So the issue states that the key would be deprecated in 4.0 (which has not been done yet since there is no deprecation)
> > And then it should be removed in 5.0.0
> > However the PR right now targets devel which will be used for 4.2.0,... So in my opinion you cannot set `removed: true` now
> > You should only deprecate it and output a deprecation warning and when we are working on releasing 5.0 we will need a new PR to remove the key (and implement the tests you have here)..
> > So for now the PR should only focus on outputing a deprecation warning when the key is used and when we are closer to release 5.0 we should remove it only.
> 
> @gmuloc if I remove `removed: true` from schema, it won't be raising error on using `port_channel.short_esi`. Shall I remove the test-case from eos_designs_negative_unit_tests for now?

@Shivani-chourasiya so what you should do is the following:

1/ Deprecate the key without removing it
2/ Update all the molecule tests in the eos_designs repo to use the new syntax (maybe they already do)
3/ Add a molecule scenario `eos_designs_deprecated_vars` like the one we have for eos_cli_config_gen https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule/eos_cli_config_gen_deprecated_vars where you can test for deprecated vars so we make sure we don't break support until we remove them

Let me know if you need help

Shivani-chourasiya:
> > > @Shivani-chourasiya / @carlbuchmann So the issue states that the key would be deprecated in 4.0 (which has not been done yet since there is no deprecation)
> > > And then it should be removed in 5.0.0
> > > However the PR right now targets devel which will be used for 4.2.0,... So in my opinion you cannot set `removed: true` now
> > > You should only deprecate it and output a deprecation warning and when we are working on releasing 5.0 we will need a new PR to remove the key (and implement the tests you have here)..
> > > So for now the PR should only focus on outputing a deprecation warning when the key is used and when we are closer to release 5.0 we should remove it only.
> > 
> > 
> > @gmuloc if I remove `removed: true` from schema, it won't be raising error on using `port_channel.short_esi`. Shall I remove the test-case from eos_designs_negative_unit_tests for now?
> 
> @Shivani-chourasiya so what you should do is the following:
> 
> 1/ Deprecate the key without removing it 2/ Update all the molecule tests in the eos_designs repo to use the new syntax (maybe they already do) 3/ Add a molecule scenario `eos_designs_deprecated_vars` like the one we have for eos_cli_config_gen https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule/eos_cli_config_gen_deprecated_vars where you can test for deprecated vars so we make sure we don't break support until we remove them
> 
> Let me know if you need help

Hi @gmuloc, I have added the molecule scenario with a test-case for port_channel.short_esi.
Once this is reviewed I will add it for cvp_instance_ip PR as well

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_adapter_config.schema.yml
```diff
@@ -441,7 +441,10 @@ $defs:
               new_key: short_esi
           short_esi:
             type: str
-            description: In format xxxx:xxxx:xxxx or "auto".
```
gmuloc, line 444: keep the description for now

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -94,9 +94,7 @@ def _get_short_esi(self, adapter: dict, channel_group_id: int, short_esi: str =
 
         # short_esi is only set when called from sub-interface port-channels.
         if short_esi is None:
-            # Setting short_esi under port_channel will be deprecated in AVD4.0
-            port_channel_short_esi = get(adapter, "port_channel.short_esi")
-            if (short_esi := get(adapter, "ethernet_segment.short_esi", default=port_channel_short_esi)) is None:
```
gmuloc: don't you need to keep this code to have the deprecated value still work? You can update the comment to state it will be removed in 5.0

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/create.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Configure local folders
-  hosts: EOS_DESIGNS_FAILURES
+  hosts: EOS_DESIGNS_DEPRECATED_VARS
```
gmuloc: Why s this changed? this is the negative_unit scenario so this should not change

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/structured_configs/port-channel-short-esi.yml
```diff
@@ -0,0 +1,33 @@
...
+  shutdown: false
+  type: port-channel-member
+  channel_group:
+    id: 10
+    mode: active
+port_channel_interfaces:
+- name: Port-Channel10
+  description: server01_ESI_PortChanne1
+  type: switched
+  shutdown: false
```
gmuloc: Where is the esi ? There should be an ethernet_segment popping up here. You should fix the python code. The deprecated vars MUST still generate the same config as before with only a warning to the user on the upcoming removal (which you have already done in schema)

ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/inventory/group_vars/SITE1_CONNECTED_ENDPOINTS.yml
```diff
@@ -17,10 +17,11 @@ cpes:
...
           - SITE1-LER2
+        ethernet_segment:
+          short_esi: 0303:0202:0101
         port_channel:
           description: EVPN-A-A-PortChannel
           mode: active
-          short_esi: 0303:0202:0101
           subinterfaces:
             - number: 111
               short_esi: 0303:0202:0111
```
gmuloc, line 27: is this still supported? (short_esi under subinterfaces ?)

If so it should probably be tacked in this PR as well!

gmuloc (changes requested):
so a couple of things to fix still in there.

The issue does state to deprecate `port_channel.short_esi` and this is taking the right direction. I am now wondering if anything should be done for the `port_channel.subinterface.short_esi`. Technically we still want to be able to set different ethernet_segment.identifier per port-channel subinterfaces so I think this should stay but would like to get the feedback of other reviewers.

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/create.yml
Shivani-chourasiya: oops! my mistake

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/structured_configs/port-channel-short-esi.yml
Shivani-chourasiya: so this was because of the condition inside `_get_short_esi` method-
 ```
if len(set(adapter["switches"])) < 2:
            # Only configure ESI for multi-homing.
            return None
```
https://github.com/aristanetworks/ansible-avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py#L91

I was giving only one switch in the input
Fixed it now and also fixed the python code

File: ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_adapter_config.schema.yml (hunk 1 above)
Comment (Shivani-chourasiya) [reply ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_adapter_config.schema.yml:444]: fixed

File: ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/inventory/group_vars/SITE1_CONNECTED_ENDPOINTS.yml (hunk 5 above)
Comment (Shivani-chourasiya) [reply ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/inventory/group_vars/SITE1_CONNECTED_ENDPOINTS.yml:27]: I am not sure about it, I didn't see anything in comments/docs regarding its deprecation.
For now, I have added `subinterfaces.short_esi` inside eos_designs_deprecated_vars input data-model to see the output.

Please let me know if anything else needs to be updated for subinterfaces.short_esi

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py (hunk 2 above)
Comment (Shivani-chourasiya) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py]: fixed

File: ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/create.yml (hunk 3 above)
Comment (Shivani-chourasiya) [reply ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/create.yml]: Fixed

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py [hunk 6]
```diff
@@ -92,9 +92,10 @@ def _get_short_esi(self, adapter: dict, channel_group_id: int, short_esi: str =
             # Only configure ESI for multi-homing.
             return None
 
-        # short_esi is only set when called from sub-interface port-channels.
         if short_esi is None:
-            if (short_esi := get(adapter, "ethernet_segment.short_esi")) is None:
+            # Setting short_esi under port_channel will be deprecated in AVD5.0
```
gmuloc:
```suggestion
            # Setting short_esi under port_channel will be removed in AVD 5.0
```

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
```diff
@@ -7,9 +7,9 @@ servers:
   - name: server01_ESI
     rack: RackC
     adapters:
-      - endpoint_ports: [ Eth1 ]
-        switch_ports: [ Ethernet10 ]
-        switches: [ port-channel-short-esi ]
+      - endpoint_ports: [ Eth1, Eth2]
+        switch_ports: [ Ethernet10, Ethernet10]
+        switches: [ port-channel-short-esi, Leaf1]
```
gmuloc, line 17: Why did you need to add the second element in the list here? (I am not sure I understand yet but maybe there is a good reason)

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/configs/port-channel-short-esi.cfg
```diff
@@ -16,12 +16,28 @@ vrf instance MGMT
...
+      route-target import 03:03:02:02:01:11
+!
+interface Port-Channel10.222
+   vlan id 222
+   encapsulation vlan
+      client dot1q 222 network client
+   evpn ethernet-segment
+      identifier 0000:0000:0303:0202:0222
+      route-target import 03:03:02:02:02:22
+!
```
gmuloc: @ClausHolbechArista - I am not 100% sure here - is this output correct from an EOS point of view?

gmuloc (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
```diff
@@ -0,0 +1,22 @@
...
+        profile: TENANT_A_B
+        port_channel:
+          description: PortChanne1
+          mode: active
+          short_esi: '0303:0202:0101'
+          subinterfaces:
+            - number: 111
+              short_esi: 0303:0202:0111
+            - number: 222
+              short_esi: 0303:0202:0222
```
ClausHolbechArista: These are not deprecated yet, so they should not be in the `deprecated_vars` scenario.

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
```diff
@@ -0,0 +1,22 @@
+type: l2leaf
+l2leaf:
+  nodes:
+    - name: port-channel-short-esi
+
+servers:
+  - name: server01_ESI
+    rack: RackC
+    adapters:
```
ClausHolbechArista, line 14: Add another adapter with `short_esi` _and_ `ethernet_segment.short_esi` set, to still test that `ethernet_segment.short_esi` overwrites the other one. (Those tests got removed from the regular molecule when you removed the deprecated variable)

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
Shivani-chourasiya: Its because of the condition I have mentioned in the comment above -
https://github.com/aristanetworks/ansible-avd/pull/3027#discussion_r1266312690

ansible_collections/arista/avd/molecule/eos_designs-mpls-isis-sr-ldp/inventory/group_vars/SITE1_CONNECTED_ENDPOINTS.yml
ClausHolbechArista: `subinterfaces.short_esi` is _not_ deprecated (yet)

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: I am not sure why we need code changes here. We should only be deprecating.

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
Shivani-chourasiya: Added.

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
Shivani-chourasiya: removed it

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
Shivani-chourasiya: Fixed

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
ClausHolbechArista: Lets add an error message instead of silently ignoring esi for single-homed devices.

gmuloc (approved):
LGTM - thanks

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/port-channel-short-esi.yml
```diff
@@ -0,0 +1,27 @@
+type: l2leaf
```
ClausHolbechArista: Connected endpoints ESI should only matter for EVPN VTEPs, so please change the test to l3leaf instead.
But this also reveals that our code for l2leaf is configuring ESI even though we should not. (Opening seperate issue on that.)

ClausHolbechArista (approved):

PR #2955, opened 2023-06-05
Fix(eos_designs): Ignore "overlay_routing_protocol_address_family: ipv6" on l2leaf

ClausHolbechArista:
## Change Summary

Fix Ignore `overlay_routing_protocol_address_family: ipv6` on l2leaf

## Related Issue(s)

Fixes #2954 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Add `underlay_router: true` condition for rendering overlay facts
- Add `underlay_router: true` condition for rendering overlay peering address
- Add test coverage first recreating the issue and then verifying the fix.

## How to test

- Added molecule coverage and no side effects

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #3043, opened 2023-07-14
Fix(eos_designs): Ignore "overlay_routing_protocol_address_family: ipv6" on l2leaf (#2955)

Vibhu-gslab:
## Change Summary

Cherry-pick PR #2955

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
@@ -121,8 +121,10 @@ interface Vlan4091
    no autostate
    ip address 10.255.252.14/31
 !
-ip routing
+ip routing ipv6 interfaces
 no ip routing vrf MGMT
+!
+ipv6 unicast-routing
 !
 mlag configuration
    domain-id DC1_L2LEAF1
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
@@ -121,10 +121,8 @@ interface Vlan4091
    no autostate
    ip address 10.255.252.14/31
 !
-ip routing ipv6 interfaces
+ip routing
 no ip routing vrf MGMT
-!
-ipv6 unicast-routing
 !
 mlag configuration
    domain-id DC1_L2LEAF1
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
@@ -2,7 +2,6 @@
 # Test that having underlay_ipv6 will not require
 # loopback_ipv6_pool for devices that are not underlay_router
 underlay_ipv6: true
-underlay_rfc5549: true
 
 # Test that having "overlay_routing_protocol_address_family: ipv6"
 # will not raise an error for devices that are not underlay_router (issue/2954)
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
@@ -2,6 +2,7 @@
 # Test that having underlay_ipv6 will not require
 # loopback_ipv6_pool for devices that are not underlay_router
 underlay_ipv6: true
+underlay_rfc5549: true
 
 # Test that having "overlay_routing_protocol_address_family: ipv6"
 # will not raise an error for devices that are not underlay_router (issue/2954)
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
@@ -2,3 +2,7 @@
 # Test that having underlay_ipv6 will not require
 # loopback_ipv6_pool for devices that are not underlay_router
 underlay_ipv6: true
+
+# Test that having "overlay_routing_protocol_address_family: ipv6"
+# will not raise an error for devices that are not underlay_router (issue/2954)
+overlay_routing_protocol_address_family: ipv6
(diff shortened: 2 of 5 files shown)


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
```diff
@@ -121,9 +121,11 @@ interface Vlan4091
    no autostate
    ip address 10.255.252.14/31
 !
-ip routing
+ip routing ipv6 interfaces
```
ClausHolbechArista: I don't think these changes are correct. Please compare with the `devel` files.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
```diff
@@ -121,9 +121,11 @@ interface Vlan4091
...
 no ip routing vrf MGMT
 !
+ipv6 unicast-routing
```
ClausHolbechArista: I don't think these changes are correct. Please compare with the `devel` files.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
```diff
@@ -1941,34 +1941,47 @@ def _overlay_vpn_ipv6(self) -> bool:
         )
 
     @cached_property
-    def overlay(self) -> dict:
+    def overlay_peering_address(self) -> str | None:
```
ClausHolbechArista: This should not be exposed in the generated facts, so it should be prefixed with underscore.

```suggestion
    def _overlay_peering_address(self) -> str | None:
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
```diff
@@ -1941,34 +1941,47 @@ def _overlay_vpn_ipv6(self) -> bool:
...
 
+        if self.underlay_router is True:
+            overlay_dict.update(
+                {
+                    "peering_address": self.overlay_peering_address,
+                    "evpn_mpls": evpn_mpls,
+                }
+            )
+
+        return overlay_dict
```
ClausHolbechArista: I don't think you need this change. The entire structure of `overlay` has changed between the versions, so it is not possible to apply the same logic.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
```diff
@@ -2,3 +2,8 @@
 # Test that having underlay_ipv6 will not require
 # loopback_ipv6_pool for devices that are not underlay_router
 underlay_ipv6: true
+underlay_rfc5549: true
```
ClausHolbechArista: This line is not in the cherry_picked PR, but maybe this actually reveals another issue with l2 leaf and underlay_rfc5549, since you see the ipv6_unicast_routing and ip_routing_ipv6_interfaces being set.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
Vibhu-gslab: Fixed!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-L2LEAF1A.cfg
Vibhu-gslab: Fixed!

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
Vibhu-gslab: Fixed!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/DC1-L2LEAF1A.yml
Vibhu-gslab: Removed the line.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
Vibhu-gslab: We have kept this almost same, just updating the "overlay_dict" with latest devel changes, in the end we are not returning "None" (as it is in latest devel), we are returning the same dict what was there, just removing "peering_address" and "evpn_mpls" from it (as it is in latest devel).

Would you like us to change this back to how it was in 3.8?

ClausHolbechArista: Yes I don't think these changes apply for the cherry pick.

Vibhu-gslab: Ok.. In that case shall I remove [this](https://github.com/aristanetworks/ansible-avd/pull/3043/files#diff-ef9c24ca7d287ca54b0ee4167c4958d47845c503f9ac5db2cb228060144c4915R8) line as well?
otherwise we will get an error similar to :

```
"Unexpected failure during module execution: 'overlay_routing_protocol_address_family: ipv6' is only supported in combination with 'underlay_ipv6: True' and 'underlay_rfc5549: True'"
```

Vibhu-gslab: Fixed!

carlbuchmann (dismissed):
LGTM

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_facts.py
```diff
@@ -1830,7 +1830,7 @@ def mlag_peer_mgmt_ip(self):
         return None
 
     @cached_property
-    def overlay_routing_protocol_address_family(self):
+    def _overlay_routing_protocol_address_family(self):
```
ClausHolbechArista: Please revert this change, since it was already there without the underscore, meaning it ends up as `switch.overlay_routing_protocol_address_family` which may be used elsewhere in 3.8.

Vibhu-gslab: If we remove underscore "_" then we get the following error in eos_designs_unit_tests molecule:

```
  An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.arista.avd.plugins.plugin_utils.errors.errors.AristaAvdError: 'overlay_routing_protocol_address_family: ipv6' is only supported in combination with 'underlay_ipv6: True' and 'underlay_rfc5549: True'
  ```
  
  In order to fix the error we modified it, is there any other way to do it?

ClausHolbechArista: Yeah because changing this will break that check overall even for other device types.

Vibhu-gslab: Done :)

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM based on CI results!

PR #3046, opened 2023-07-18
Fix(eos_config_deploy_cvp): device_filter is not behaving correctly if input is a string

gmuloc:
## Change Summary

In AVD 4.0, the move to use v3 of CVP collection by default broke the integration with inventory_to_container module in some cases.

With v1 tasks, if a string was given as `device_filter`, it was converted to the list `[<string>]` by the v1 tasks. This is not the case anymore.

This leads to some misleading result in particular with the empty string device_filter `""`

Additionally, while this has been transparent to the user, if a string like "DC1" was given as device_filter, the `inventory_to_container` action was actually comparing the hostnames to `"D"`, `"C"` and `"1"`, potentially leading to bad filtering. For instance if you had DC1 and DC2 devices and wanted to generate CVP config only for DC1 it would have failed and selecting also the device from DC2 because they have "D" or "C" in their name.

## Related Issue(s)

Reported internally

## Component(s) name

## Proposed changes

Add a step in the `inventory_to_container` action plugin to convert a string `"a"` to a list `["a"]` by default

## How to test

Added a molecule scenario.
First commit shows output before the fix, second commit shows after the fix the difference (notice the parentContainers are back and so are devices - yeah!)

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_config_deploy_cvp/converge.yml
+++ b/ansible_collections/arista/avd/molecule/eos_config_deploy_cvp/converge.yml
@@ -31,3 +31,24 @@
         execute_tasks: false
         state: present
         cv_collection: v3
+
+- name: "Converge -- Build Cloudvision Configuration - empty filter as a string"
+  hosts: CVP
+  gather_facts: false
+  connection: local
+  vars:
+    root_dir: '{{ playbook_dir }}'
+  tasks:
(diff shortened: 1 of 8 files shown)


philippebureau (approved):
Tested
LGTM

tgodaA (approved):
LGTM

PR #3075, opened 2023-08-03
Feat(eos_designs): EVPN vlan-aware-bundle option for l2vlan 

### LINKED ISSUE #3065: eos_designs: EVPN vlan-aware-bundle option for l2vlan
### Enhancement summary

To be able to use vlan-aware-bundles and having different names for l2vlans a new option for l2vlans should be introduced. With this every l2vlan could have a speaking name for the purpose of the VLAN and can be bundled into a EVPN vlan-aware-bundle to simplify the configuration.

eos_designs

### Use case example

I want to define l2vlans with a speaking name which is different for ever VLAN, but want to combine some of those l2vlans into one EVPN vlan-aware-bundle. The combination into vlan-aware-bundles keeps the configration simple and smaller. 

Introduce a new option for l2vlans, for example "evpn_vlan_bundle". All l2vlans where the same "evpn_vlan_bundle" is defined should be combined into a vlan-aware-bundle in case the global option for vlan-aware-bundles is enabled.

```
    l2vlans:
      - id: 10
        name: vmotion
        evpn_vlan_bundle: VmBundle
      - id: 20
        name: VmBackup   
        evpn_vlan_bundle: VmBundle
      - id: 30
        name: storageDate
        evpn_vlan_bundle: StorageBundle
      - id: 40
        name: storageBackup   
        evpn_vlan_bundle: StorageBundle
```

_No response_

### Additional context

_No response_

### Contributing Guide

bjmeuer:
## Change Summary

Add a structure to define vlan-aware bundles and an option to reference this definition from l2vlans. 

## Related Issue(s)

Fixes #3065 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Code checks if a l2vlan references a vlan-aware bundle with the option evpn_vlan_bundle and if so it will inherit the settings from the referenced vlan-aware bundle definition. If none is defined it falls back to the vlan name as vlan-aware-bundle name.

## How to test
```
---
  - name: FABRIC
    vlan_aware_bundle_number_base: 200
    mac_vrf_vni_base: 10000
    l2vlans:
      - id: 50
        name: ipmi
        tags: ['ipmi']
        evpn_vlan_bundle: bundle1
      - id: 60
        name: ipmi
        tags: ['fabric']
        evpn_vlan_bundle: bundle1
      - id: 70
        name: vmotion
        tags: ['ipmi']
        evpn_vlan_bundle: bundle2
      - id: 80
        name: vstorage
        tags: ['fabric']
        evpn_vlan_bundle: bundle2

evpn_vlan_bundles:
  - name: bundle1
    id: 1
    bgp:
      raw_eos_cli: |
        vlan add 1000
  - name: bundle2
    id: 2
    rt_override: 111
    rd_override: 1.1.1.1:222
```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_l2vlan_bundle_definition.yml
ClausHolbechArista, line 1: Please use the minimum amount of config to test only what you need.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/hosts.yml
```diff
@@ -23,6 +23,7 @@ all:
             mgmt_interface_dualstack:
             default_overlay_protocol_cvx:
             default_overlay_protocol_her:
+            evpn_l2vlan_bundle_definition:
```
ClausHolbechArista: Let's just call it `evpn_vlan_bundle` so it resembles the key we are testing.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -399,8 +399,41 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
                     )
 
             # L2 Vlans per Tenant
+            # If multiple L2 Vlans share the same evpn_vlan_bundle name, they will be part of the same vlan-aware-bundle
+            l2vlans = list(tenant["l2vlans"])
+            l2vlans_no_bundle_definition = list()
```
ClausHolbechArista:
```suggestion
            l2vlans_no_bundle_definition = []
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -399,8 +399,41 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
                     )
 
             # L2 Vlans per Tenant
+            # If multiple L2 Vlans share the same evpn_vlan_bundle name, they will be part of the same vlan-aware-bundle
+            l2vlans = list(tenant["l2vlans"])
```
ClausHolbechArista: I believe the data has been normalized in `_filtered_tenants` so we don't need to make it a list here.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -399,8 +399,41 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                    bundle["name"] = str(bundle_name)
+                    bundle.update(values["bundle_details"])
+                    append_if_not_duplicate(
+                        list_of_dicts=bundles,
+                        primary_key="name",
+                        new_dict=bundle,
+                        context="BGP VLAN-Aware Bundles defined under network services",
+                        context_keys=["name"],
+                    )            
+
```
ClausHolbechArista: Instead of adding all this extra logic which will be a bit "expensive", I think we should just create a small function we can give to the `groupby` below instead of the static key `"name"`. I believe `groupby` supports a callable as `key` so you can write the function to get `evpn_vlan_bundle` key if present, or fallback to the `name` key. This way we get less code duplication, and it will run much more efficiently.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
bjmeuer: removed

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/hosts.yml
bjmeuer: done

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_l2vlan_bundle_definition.yml
bjmeuer: updated

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_network_services.schema.yml
```diff
@@ -938,6 +938,12 @@ $defs:
                 type: bool
                 default: true
                 description: Extend this L2VLAN over VXLAN.
+              evpn_vlan_bundle:
+                type: str
+                description: |
+                  Define a vlan-aware-bundle name which should be used instead of the vlan name. If several vlans have the same evpn_vlan_bundle defined they will be grouped together.
+                  The RD and RT will be used from the lowest VLAN id defined in the same evpn_vlan_bundle.
+                  To use this option the global evpn_vlan_aware_bundles option must be set to true.
```
carlbuchmann: We should probably catch this in the code, and add a negative unit test for this.

bjmeuer: Not sure I get what you mean, what do you want to catch in the code?

ClausHolbechArista: If someone sets the evpn bundle name but the global evpn_vlan_aware_bundles is not set, we should error out in the code.

bjmeuer: I added a check in the code and also a negative test, can you have a look?

carlbuchmann: thank you that looks good!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/evpn_vlan_bundle.cfg
```diff
@@ -0,0 +1,102 @@
...
+   !
+   vlan-aware-bundle bundle1
+      rd 10.10.0.3:10050
+      route-target both 10050:10050
+      redistribute learned
+      vlan 50,60
+   !
+   vlan-aware-bundle storage_prod
+      rd 10.10.0.3:10010
+      route-target both 10010:10010
```
carlbuchmann, line 101: With this strategy of creating the rd and rt automatically derived from the lowest VLAN, do we not risk service disruption, for example, if I add VLAN 9 to the bundle and cause the rd/rt values to change?

carlbuchmann: I think we could introduce a key called `evpn_vlan_bundle_id`, where the id is leveraged for the rd/rt admin subfield
example with default overlay_rt/rd_type.
evpn_vlan_bundle_id: 500
   rd 10.10.0.3:500
   route-target both 500:500

ClausHolbechArista: Agreed, it is a caveat we already have today if you use the same name on multiple l2vlans.
It would not make sense to set this id under each l2vlan. Maybe we need to build a completely new data structure to cover this; `l2vlan_aware_bundles` and then defining the vlans and even allowing vlan ranges there (we have a request for that).

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -384,6 +402,12 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
             return None
 
         if not self._evpn_vlan_aware_bundles:
+            first_evpn_vlan_bundle = self._get_first_evpn_vlan_bundle
+            if first_evpn_vlan_bundle is not None:
+                raise AristaAvdMissingVariableError(
+                    "Option 'evpn_vlan_aware_bundles' must be set to 'true' to be able to use 'evpn_vlan_bundle' option in l2vlans. "
+                    f"First occurence of 'evpn_vlan_bundle' seen for l2vlan {first_evpn_vlan_bundle['id']} in Tenant '{first_evpn_vlan_bundle['tenant']}'."
+                )
             return None
```
ClausHolbechArista: @carlbuchmann This check will affect all AVD runs, since it happens even if the feature is disabled. It loops through all l2vlans and checks for a key, just to create the error. I think we need to rethink this, and maybe accept that we just ignore the key for now, and later we can have the schema check these kind of things.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/evpn_vlan_bundle.cfg
bjmeuer: also thought of introducing an option for specifically setting the rt/rd, but I did follow the current way of implementation. Today when the vlan bundle is based on the vlan name it is the same, the rt / rd is based on the lowest vlan id and might change if you introduce a new vlan with the same name and lower id.

bjmeuer: Nevertheless I think it would make sense to at least provide an option to statically set the rd / rt for the vlan-aware-bundles. Should I implement it in this PR or better create a separate issue and PR?

ClausHolbechArista: It would lead to conflicts if we allow setting it under the vlan itself, since you could have the same name but different ids of different vlans. So I think we need to build another data model. Let me discuss with Carl and get back.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
carlbuchmann: agreed

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure_evpn_vlan_bundle.yml
```diff
@@ -0,0 +1,48 @@
...
+        name: ipmi
+        tags: ['fabric']
+        evpn_vlan_bundle: bundle2
+
+evpn_vlan_bundles:
+  - name: bundle1
+    id: 1
+
+expected_error_message: >-
+  The 'evpn_vlan_bundle' of the l2vlans must be defined in the global 'evpn_vlan_bundles' setting. First occurence seen for l2vlan 60 in Tenant 'FABRIC' and evpn_vlan_bundle 'bundle2'.
```
ClausHolbechArista: Please use `common` instead of `global`

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -154,12 +154,12 @@ AVD provides the capability to customize your node types, supporting a variety o
     ```yaml
     node_type_keys:
 
-      - key: spine
+    - key: spine
```
ClausHolbechArista: These changes break stuff. I am not sure you intended to change all these things in this file, so please revert what was done by mistake.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -381,6 +382,12 @@ def _router_bgp_vlans_vlan(self, vlan, tenant, vrf) -> dict | None:
     def _evpn_vlan_aware_bundles(self) -> bool:
         return get(self._hostvars, "evpn_vlan_aware_bundles", default=False)
 
+    def _get_vlan_aware_bundle_name(self, vlan: dict) -> str:
```
ClausHolbechArista: This could also return `None` so please add `| None` to the return type.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -406,20 +413,37 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                            bundle["route_targets"] = {"both": [self.get_vlan_rt(evpn_vlan_bundle, tenant)]}
+                            if (eos_cli := get(evpn_vlan_bundle, "bgp.raw_eos_cli")) is not None:
+                                bundle["eos_cli"] = eos_cli
+                        else:
+                            raise AristaAvdMissingVariableError(
+                                "The 'evpn_vlan_bundle' of the l2vlans must be defined in the global 'evpn_vlan_bundles' setting. First occurence seen for"
+                                f" l2vlan {l2vlans[0]['id']} in Tenant '{l2vlans[0]['tenant']}' and evpn_vlan_bundle '{l2vlans[0]['evpn_vlan_bundle']}'."
+                            )
+
+                append_if_not_duplicate(
```
ClausHolbechArista, line 469: I think you deindented this line by mistake, since now it will also run in case the `if` in line 421 fails.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -1140,6 +1068,14 @@ Note: structured configuration is not merged recursively and will be taken direc
 roles/eos_designs/docs/tables/svi-profiles.md
 --8<--
 
+### EVPN VLAN aware bundles settings
+
+Optional VLAN aware bundles to share common settings for l2vlans which are supposed to use the same vlan-aware-bundle.
+
+--8<--
+roles/eos_designs/docs/tables/evpn_vlan_bundles.md
```
ClausHolbechArista: Please use hyphens in the table names instead of underscore.

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure_evpn_vlan_bundle.yml
bjmeuer: replaced

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
bjmeuer: these changes were not intended, maybe one of my plugins removed spaces and I accidentially saved them... I copied the lines from the file in the dev branch, hope it is all sorted now.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
bjmeuer: changed

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
bjmeuer: done

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
bjmeuer: changed

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
```diff
@@ -0,0 +1,64 @@
...
+            A EVPN vlan-aware-bundle provide a construct to group L2 VLANs and define common settings.
+        id:
+          type: int
+          required: true
+          convert_types:
+            - str
+          min: 1
+          max: 1024
+          description: |
+            "id" may also be used for vlan-aware-bundle RD/RT ID and therefore should not overlap with l2vlan IDs which are not part of this bundle.
```
ClausHolbechArista:
```suggestion
            "id" may be used for vlan-aware-bundle RD/RT ID so it should not overlap with l2vlan IDs which are not part of this bundle.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
```diff
@@ -0,0 +1,64 @@
...
+    primary_key: name
+    convert_types:
+      - dict
+    items:
+      type: dict
+      keys:
+        name:
+          type: str
+          description: |
+            Specify a EVPN vlan-aware-bundle name.
```
ClausHolbechArista:
```suggestion
            Specify an EVPN vlan-aware-bundle name.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
```diff
@@ -0,0 +1,64 @@
...
+            A EVPN vlan-aware-bundle provide a construct to group L2 VLANs and define common settings.
```
ClausHolbechArista:
```suggestion
            EVPN vlan-aware-bundles group L2 VLANs and define common settings.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
```diff
@@ -0,0 +1,64 @@
...
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../../plugins/plugin_utils/schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  evpn_vlan_bundles:
+    documentation_options:
+      table: evpn_vlan_bundles
```
ClausHolbechArista: Use hyphens in table names

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
```diff
@@ -0,0 +1,64 @@
...
+          description: |
+            Specify a EVPN vlan-aware-bundle name.
+            A EVPN vlan-aware-bundle provide a construct to group L2 VLANs and define common settings.
+        id:
+          type: int
+          required: true
+          convert_types:
+            - str
+          min: 1
+          max: 1024
```
ClausHolbechArista: Remind me why we limit the id?

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_network_services.schema.yml
```diff
@@ -966,6 +966,11 @@ $defs:
                 type: bool
                 default: true
                 description: Extend this L2VLAN over VXLAN.
+              evpn_vlan_bundle:
+                type: str
+                description: |
+                  Name of the evpn_vlan_bundles to inherit configuration.
+                  To use this option the global "evpn_vlan_aware_bundles" option must be set to true.
```
ClausHolbechArista:
```suggestion
                description: |
                  Name of a bundle defined under 'evpn_vlan_bundles' to inherit configuration.
                  To use this option the global "evpn_vlan_aware_bundles" option must be set to true.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/evpn_vlan_bundles.schema.yml
bjmeuer: I think there is no real need to limit it, let me remove the limits

ClausHolbechArista:
Please update the PR description to reflect the latest state. Makes it easier for someone else to review.

bjmeuer (commented):
looks good to me, tested with data from my customer as well and it looks good.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -406,20 +413,63 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
             # L2 Vlans per Tenant
-            # If multiple L2 Vlans share the same name, they will be part of the same bundle
-            for bundle_name, l2vlans in groupby(tenant["l2vlans"], "name"):
+            # If multiple L2 Vlans share the same evpn_vlan_bundle name, they will be part of the same vlan-aware-bundle else they use the vlan name as bundle
+            sorted_vlan_list = sorted(tenant["l2vlans"], key=self._get_vlan_aware_bundle_name)
+            bundle_groups = itertools_groupby(sorted_vlan_list, self._get_vlan_aware_bundle_name)
+            for bundle_name, l2vlans in bundle_groups:
                 l2vlans = list(l2vlans)
-                if (bundle := self._router_bgp_vlans_vlan(l2vlans[0], tenant, vrf={})) is not None:
+                if "evpn_vlan_bundle" in l2vlans[0]:
```
gmuloc: so here we check on the first l2vlans - what if, unfortunate but possible, the first vlan after sort is one which name would be the `vlan.get("name")` and further vlans in the list come from `vlan.get("evpn_vlan_bundle")` ? Seems that we could end up taking the wrong branch


Need to error out if not all VLANs have "evpn_vlan_bundle"

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
```diff
@@ -839,6 +839,71 @@ keys:
     description: Enable vlan aware bundles for EVPN MAC-VRF.
     type: bool
     default: false
+  evpn_vlan_bundles:
+    documentation_options:
+      table: evpn-vlan-bundles
+    type: list
+    primary_key: name
+    convert_types:
+    - dict
```
gmuloc: We don't need to convert for newer schema

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
bjmeuer: We use a groupby in line 418 to group the vlans by the result of `vlan.get("evpn_vlan_bundle")` or `vlan.get("name")` which means all vlans in that group either have` evpn_vlan_bundle` or they don't. This is why we can just check in the first element of the group

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
bjmeuer: removed

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
ClausHolbechArista: The point is that since we use `or` a vlan without any "evpn_vlan_bundle" might still end up in the pile of vlans using the new "evpn_vlan_bundle" feature, if it shares the name of such a bundle. But in that case the settings would be taken from the "evpn_vlan_bundles" instead of the vlan, which might confuse the user.
So what Guillaume pointed out is that we might have to change this logic a bit, to check for "evpn_vlan_bundle" in all the vlans, so it is "all or nothing", and if not, raise an error.

ClausHolbechArista: An alternative solution could be to prefix the name from `evpn_vlan_bundle` with `_` so they end up in different groups, and then remove the prefix again before adding to the bundle list. Our duplicate detection would then catch the overlapping names.

gmuloc: Hi @bjmeuer - I appreciate my comment was a bit light just to add to what Claus said - if for some unfortunate reason someone chooses `name` for a `l2vlan` that is clashing with you bundle name (I know my example looks a but simple) - then the order of the l2vlans will have an impact with current code (because `l2vlans[0]` won't have an `evpn_vlan_bundle` but WILL have been grouped indeed with other vlans, because of its name

```
---
  - name: FABRIC
    vlan_aware_bundle_number_base: 200
    mac_vrf_vni_base: 10000
    l2vlans:
      - id: 50
        name: bundle1 <<<<<<<<<<<<<<<<<<<<<<<<<< unfortunate vlan name without evpn_vlan_bundle
        tags: ['ipmi']
      - id: 60
        name: ipmi
        tags: ['fabric']
        evpn_vlan_bundle: bundle1

evpn_vlan_bundles:
  - name: bundle1
    id: 1
    bgp:
      raw_eos_cli: |
        vlan add 1000
```

bjmeuer: change to adding a temporary "_" like Claus suggested

gmuloc: the last change does not solve the issue if the vlan name is `_bundle1`

bjmeuer: ok, think if I temporary prefix the vlan name as well with a different character this should work right?

ClausHolbechArista: I have implemented a different fix for this, which is more well defined instead of hacking prefixes. See https://github.com/aristanetworks/ansible-avd/pull/3075/commits/78f322c85e8d5151c9c0f3f30a63e5f50a358649

bjmeuer: This code is so much better. I was focusing on that prefix thing too much so I didn't thought of that...

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-duplicate-evpn-vlan-bundle-name.yml
```diff
@@ -0,0 +1,50 @@
...
+        evpn_vlan_bundle: bundle1
+
+evpn_vlan_bundles:
+  - name: bundle1
+    id: 1
+
+expected_error_message: >-
+  Found duplicate objects with conflicting data while generating configuration for BGP VLAN-Aware Bundles defined under network services.
+  {'name': 'bundle1', 'rd': '10.10.0.3:10010', 'route_targets': {'both': ['10010:10010']}, 'tenant': 'StorageProd', 'vlan': '10'}
+  conflicts with {'name': 'bundle1', 'rd': '10.10.0.3:1', 'route_targets': {'both': ['1:1']}, 'tenant': None, 'vlan': '50,60'}.
```
gmuloc, line 50: would be good to mention one is an `evpn_vlan_bundle` and the other a `l2vlan` for it to be very helpful

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-missing-evpn-vlan-bundle.yml
```diff
@@ -0,0 +1,48 @@
...
+tenants:
+  - name: StorageProd
+    mac_vrf_vni_base: 10000
+    l2vlans:
+      - id: 10
+        name: storage_prod
+        tags: ['storageProd']
+      - id: 20
+        name: storage_prod
+        tags: ['storageProd']
```
gmuloc: In favor of removing what is not needed in there in order to avoid maintaining things we don't need
```suggestion
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -406,20 +415,64 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                if is_evpn_vlan_bundle:
+                    # If "evpn_vlan_bundle" in l2vlan bundle group, we use the same logic for rd/rt as VRFs
+                    # using the settings under the given evpn_vlan_bundle only.
+
+                    # check if the referred name exists in the global evpn_vlan_bundles
+                    if (evpn_vlan_bundle := get_item(self._hostvars["evpn_vlan_bundles"], "name", bundle_name)) is None:
+                        raise AristaAvdMissingVariableError(
+                            "The 'evpn_vlan_bundle' of the l2vlans must be defined in the common 'evpn_vlan_bundles' setting. First occurence seen for l2vlan"
+                            f" {l2vlans[0]['id']} in Tenant '{l2vlans[0]['tenant']}' and evpn_vlan_bundle '{l2vlans[0]['evpn_vlan_bundle'].replace('_','')}'."
+                        )
```
gmuloc: No need for the replace anymore after the refactor
```suggestion
                            f" {l2vlans[0]['id']} in Tenant '{l2vlans[0]['tenant']}' and evpn_vlan_bundle '{l2vlans[0]['evpn_vlan_bundle']}'."
                        )
```

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-duplicate-evpn-vlan-bundle-name.yml
gmuloc: It is not possible today in the existing duplication of context so maybe adding to the context string what is the most probable cause (an l2vlan name clashing with an `evpn_vlan_bundle` name

gmuloc (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-duplicate-evpn-vlan-bundle-name.yml
bjmeuer: What do you think about:

_Found duplicate objects with conflicting data while generating configuration for BGP VLAN-Aware Bundles defined under network services. A common reason is a l2vlan name is overlapping with a evpn_vlan_bundle name._

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-duplicate-evpn-vlan-bundle-name.yml
```diff
@@ -0,0 +1,50 @@
...
+        name: ipmi
+        tags: ['fabric']
+        evpn_vlan_bundle: bundle1
+
+evpn_vlan_bundles:
+  - name: bundle1
+    id: 1
+
+expected_error_message: >-
+  Found duplicate objects with conflicting data while generating configuration for BGP VLAN-Aware Bundles defined under network services. A common reason is a l2vlan name is overlapping with a evpn_vlan_bundle name.
```
ClausHolbechArista:
```suggestion
  Found duplicate objects with conflicting data while generating configuration for BGP VLAN-Aware Bundles defined under network services. A common reason is that an 'l2vlan' name overlaps with an 'evpn_vlan_bundle' name.
```

ClausHolbechArista (approved):
Only found some wording in the error message that could be improved.
Approving since everything else looks great.

gmuloc (approved):
Once the CI finishes (after the commit I just pushed) - LGTM
Thanks @bjmeuer for the work!

PR #3078, opened 2023-08-08
Feat(eos_designs): Add sflow_settings

ClausHolbechArista:
## Change Summary

Add sflow_settings

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

```yaml
sflow_settings:
  destinations:
    - destination: <str>
      port: <int>
      vrf: <str>
  vrfs:
    - name: <str>
      source_inteface: <str>
```

## How to test

Added to existing sFlow molecule tests.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+keys:
+  sflow_settings:
+    documentation_options:
+      table: management-sflow-settings
+    type: dict
+    description: |-
+      sFlow settings.
+      The sFlow process will only be configured if any interface is enabled for sFlow.
+      For source-interfaces see "source_interfaces.sflow"
+      For default enabling of sFlow for varous interface types across the fabric see "fabric_sflow"
```
jonxstill:
```suggestion
      For default enabling of sFlow for various interface types across the fabric see "fabric_sflow"
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+            port:
+              type: int
+              description: Port number
+              convert_types:
+                - str
+            vrf:
+              type: str
+              description: |-
+                VRF Name.
+                Can be used in combination with "use_mgmt_interface_vrf" and "use_inband_mgmt_vrf" to configure the destination under multiple VRFs.
```
jonxstill:
```suggestion
                Can be used in combination with `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf` to configure the destination under multiple VRFs.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+              convert_types:
+                - int
+            use_mgmt_interface_vrf:
+              type: bool
+              description:
+                Configure the destination under the VRF set with "mgmt_interface_vrf".
```
jonxstill:
```suggestion
                Configure the destination under the VRF set with `mgmt_interface_vrf`.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+                Can be used in combination with "vrf" and "use_inband_mgmt_vrf" to configure the destination under multiple VRFs.
```
jonxstill:
```suggestion
                Can be used in combination with `vrf` and `use_inband_mgmt_vrf` to configure the destination under multiple VRFs.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+            use_inband_mgmt_vrf:
+              type: bool
+              description:
+                Configure the destination under the VRF set with "inband_mgmt_vrf".
```
jonxstill:
```suggestion
                Configure the destination under the VRF set with `inband_mgmt_vrf`.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+                Ignored if inband management is not configured for the device.
+                Can be used in combination with "vrf" and "use_mgmt_interface_vrf" to configure the destination under multiple VRFs.
```
jonxstill:
```suggestion
                Can be used in combination with `vrf` and `use_mgmt_interface_vrf` to configure the destination under multiple VRFs.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_sflow.schema.yml
```diff
@@ -8,9 +8,11 @@ type: dict
 keys:
   fabric_sflow:
     documentation_options:
-      table: fabric-settings
+      table: management-sflow-settings
     type: dict
-    description: Fabric-wide sFlow settings.
+    description: |-
+      Default enabling of sFlow for varous interface types across the fabric.
```
jonxstill:
```suggestion
      Default enabling of sFlow for various interface types across the fabric.
```

jonxstill (changes requested):
A few minor docs/schema changes.  Tested this against a few different scenarios - works well IMO.

jonxstill (dismissed):
LGTM!

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
```diff
@@ -2294,6 +2323,50 @@ keys:
...
+              description: sFlow destination IP address.
+            port:
+              type: int
+              description: Port number
+              convert_types:
+              - str
+            vrf:
+              required: true
+              type: str
+              description: "The value of `vrf` will be interpreted according to these
```
pmerlo2022: I'm assuming the reference in the description to configuring NTP server(s) actually meant the sFlow destination?

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,47 @@
...
+        type: list
+        items:
+          type: dict
+          keys:
+            destination:
+              type: str
+              required: true
+              description: |-
+                sFlow destination IP address.
+            port:
```
pmerlo2022, line 27: Is there any value to putting a value restriction here since it represents a port number? (e.g., 1 to 65535?)

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,47 @@
...
+      destinations:
+        type: list
+        items:
+          type: dict
+          keys:
+            destination:
+              type: str
+              required: true
+              description: |-
+                sFlow destination IP address.
```
pmerlo2022: Since this is a string does it make sense to specify this could be an IPv4 or IPv6 address? Or even an fqdn?

pmerlo2022 (commented):
First time reviewing an AVD PR ... tested the changes and they work as advertised. Just a couple of comments on the docs and schema.

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
ClausHolbechArista: Good catch! Too much copy/paste. Fixed.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
ClausHolbechArista: Correct. I have updated to "name or IP address" like we have elsewhere.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
ClausHolbechArista: Yes let me add that. Thanks.

ClausHolbechArista:
@jonxstill @pmerlo2022 thank you for your reviews. I have updated the data model a bit since Jon reviewed, to be simpler for VRF. And for both of you I just added support for the new `default_mgmt_method` so it respects that if you don't set the VRF.

pmerlo2022 (commented):
The changes generally look good to me - the only comment I have is that I still references in the documentation to NTP server(s) instead of sFlow destination.

ClausHolbechArista:
> The changes generally look good to me - the only comment I have is that I still references in the documentation to NTP server(s) instead of sFlow destination.

Fixed. Thank you for catching this (again)...

pmerlo2022 (approved):
The new updates look good to me

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,48 @@
...
+type: dict
+keys:
+  sflow_settings:
+    documentation_options:
+      table: management-sflow-settings
+    type: dict
+    description: |-
+      sFlow settings.
+      The sFlow process will only be configured if any interface is enabled for sFlow.
+      For default enabling of sFlow for various interface types across the fabric see "fabric_sflow".
```
gmuloc: in `fabric_sflow.schema.yml` we use backticks rather than double quotes but not sure we have a good guideline on this not consistency

```suggestion
      For default enabling of sFlow for various interface types across the fabric see `fabric_sflow`.
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
```diff
@@ -0,0 +1,48 @@
...
+            vrf:
+              type: str
+              description: |-
+                If not set, the VRF is automatically picked up from the global setting `default_mgmt_method`.
+                The value of `vrf` will be interpreted according to these rules:
+                - `use_mgmt_interface_vrf` will configure the sFlow destination under the VRF set with `mgmt_interface_vrf` and set the `mgmt_interface` as sFlow source-interface.
+                  An error will be raised if `mgmt_ip` or `ipv6_mgmt_ip` are not configured for the device.
+                - `use_inband_mgmt_vrf` will configure the sFlow destination under the VRF set with `inband_mgmt_vrf` and set the `inband_mgmt_interface` as sFlow source-interface.
+                  An error will be raised if inband management is not configured for the device.
+                - Any other string will be used directly as the VRF name but source interface must be set with `sflow.structured_config` as needed.
```
gmuloc:
```suggestion
                - Any other string will be used directly as the VRF name but source interface MUST be set with `sflow.structured_config` as needed.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/get_structured_config/get_structured_config.py
```diff
@@ -30,6 +31,11 @@
     AvdStructuredConfigNetworkServices,
     AvdStructuredConfigConnectedEndpoints,
     AvdStructuredConfigInbandManagement,
+    # The sFlow module must be rendered after others contributing interfaces,
+    # since it parses those interfaces for sFlow config.
+    AvdStructuredConfigSflow,
```
gmuloc: Not for today but I feel like we could have some better mechanism than relying on the order or rendered Classes - maybe some kind of `requires` or something like this to indicate the rendering of one specific class needs the rendering of another first (or maybe that's overkill if it is only for one ;) )

ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py
```diff
@@ -0,0 +1,132 @@
...
+        - source-interfaces based on source_interfaces.sflow
+        """
+        if not self._enable_sflow:
+            return None
+
+        destinations = get(self._hostvars, "sflow_settings.destinations")
+        if destinations is None:
+            # TODO:
+            # AVD5.0.0 raise an error if sflow is enabled on an interface but there are no destinations configured.
+            # This cannot be implemented today since it would be breaking for already released support for sflow on interfaces.
```
gmuloc: Feat(eos_designs)!: Raise an issue when Sflow enabled on an interface but no destination is configured

AVD5.0.0 raise an error if sflow is enabled on an interface but there are no destinations configured.
This cannot be implemented today since it would be breaking for already released support for sflow on interfaces.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -709,11 +709,16 @@ roles/eos_designs/docs/tables/management-settings.md
...
+
+--8<--
+roles/eos_designs/docs/tables/management-sflow-settings.md
+--8<--
+
 ### SNMP settings
 
 --8<--
 roles/eos_designs/docs/tables/management-snmp-settings.md
---8<--
```
gmuloc: Should this be removed?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS_LEAFS.yml
```diff
@@ -14,7 +14,7 @@ l3leaf:
     mlag_peer_l3_ipv4_pool: 10.254.1.96/27
     spanning_tree_priority: 4096
     spanning_tree_mode: mstp
-
+    virtual_router_mac_address: 00:11:22:33:44:55
```
gmuloc, line 17: not sure I understand this change?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS.yml
```diff
@@ -5,6 +5,21 @@ fabric_name: SFLOW_TESTS
...
+    - destination: 10.10.10.10
+      vrf: use_mgmt_interface_vrf
+    - destination: 10.10.10.11
+    - destination: 10.10.10.12
+      vrf: sflowvrf
+      port: 1234
+  structured_config:
+    vrfs:
+      - name: sflowvrf
+        source_interface: Loopback321
```
gmuloc: looking at this - would it not be easier to have the `structured_config` inside the destinations list to apply on the VRF directly? or simply adding source_interface at the destinations level? is there other use case for structured config here?

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
ClausHolbechArista: This is not an RFC :). I would rather use **bold** instead of upper-case?

ansible_collections/arista/avd/roles/eos_designs/python_modules/get_structured_config/get_structured_config.py
ClausHolbechArista: Yes this will have to happen as part of some major rework into smaller modules that have dependencies mapped out.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS.yml
ClausHolbechArista: For this specific config it could be done in other ways, but the whole point of structured_config is that we can put anything in there :). The area contains several things we don't support here (yet).

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS_LEAFS.yml
ClausHolbechArista: Because I add inband management using `inband_mgmt_subnet`, it will configure VARP, which in turn requires a vMAC.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: It should not. Rebase gone wrong.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -709,11 +709,16 @@ roles/eos_designs/docs/tables/management-settings.md
...
+### sFlow settings
+
+--8<--
+roles/eos_designs/docs/tables/management-sflow-settings.md
+--8<--
+
 ### SNMP settings
 
 --8<--
 roles/eos_designs/docs/tables/management-snmp-settings.md
```
ClausHolbechArista:
```suggestion
roles/eos_designs/docs/tables/management-snmp-settings.md
--8<--
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py
ClausHolbechArista: issue created. Thanks.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS.yml
jonxstill: This is going to be a *very* common config for anyone streaming flow data to CVP via TerminAttr.
```yaml
sflow_settings:
  destinations:
    - destination: 10.94.1.42
    - destination: 127.0.0.1
      vrf: default
  structured_config:
    source_interface: Loopback0
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml
jonxstill:
```suggestion
                - Any other string will be used directly as the VRF name but `source_interface` must be set with `sflow.structured_config` as needed.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py
```diff
@@ -0,0 +1,132 @@
...
+
+            else:
+                # Add destination with VRF field.
+                sflow_vrfs.setdefault(vrf, {}).setdefault("destinations", []).append(
+                    {
+                        "destination": destination.get("destination"),
+                        "port": destination.get("port"),
+                    }
+                )
+                sflow_vrfs[vrf]["source_interface"] = source_interface
```
jonxstill: Really not sure how I feel about relying on using `structured_config` to make the user set the `source_interface` for arbitrary VRFs.  For example I tried the following: 
```yaml
sflow_settings:
  destinations:
    - destination: 10.94.1.42
      vrf: MGMT
    - destination: 127.0.0.1
      vrf: default
  structured_config:
    source_interface: Loopback0
```
and got this error:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: UnboundLocalError: cannot access local variable 'source_interface' where it is not associated with a value
fatal: [dc1-leaf1b -> localhost]: FAILED! => {"changed": false, "msg": "cannot access local variable 'source_interface' where it is not associated with a value"}
```
It feels to me like it would be a nicer user experience to allow this to be overridden directly as part of this schema  and to raise a more informative error if this is mistakenly omitted.  The error above is not at all clear to the average user what they've done wrong.

ClausHolbechArista: Wow, I have not seen that error before. The issue with setting the source interface with a better model is that I then need a list of VRFs again and allow you to set source interface under each, but what if you also had destinations for that VRF using `use_mgmt_interface_vrf`. So I decided on this to keep the logic sane.
Your above example should really be using the `use_mgmt_interface_vrf` instead of static `MGMT`. Otherwise it will not add the source interface automatically.

Re the error, can you provide a bit more context, to the task where you see this?

ClausHolbechArista: Re TerminAttr, I plan a follow up PR with a single knob to enable streaming to terminattr. It will auto-create a dummy interface in the default VRF, if you don't have one (like an l2leaf).

jonxstill: Re: Terminattr - that sounds perfect.  And understand about the static MGMT issue - I think we just need to catch where someone has statically defined a VRF without a source-interface and give a nicer error.

Re: The error seen above, here's the full traceback:
```
The full traceback is:
  File "/Users/jonstill/src/ansible-avd/ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py", line 84, in run
    output = get_structured_config(
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jonstill/src/ansible-avd/ansible_collections/arista/avd/roles/eos_designs/python_modules/get_structured_config/get_structured_config.py", line 77, in get_structured_config
    results = eos_designs_module.render()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jonstill/src/ansible-avd/ansible_collections/arista/avd/plugins/plugin_utils/avdfacts.py", line 67, in render
    return {key: getattr(self, key) for key in self.keys() if getattr(self, key) is not None}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jonstill/src/ansible-avd/ansible_collections/arista/avd/plugins/plugin_utils/avdfacts.py", line 67, in <dictcomp>
    return {key: getattr(self, key) for key in self.keys() if getattr(self, key) is not None}
                                                              ^^^^^^^^^^^^^^^^^^
  File "/Users/jonstill/.pyenv/versions/3.11.4/lib/python3.11/functools.py", line 1001, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/Users/jonstill/src/ansible-avd/ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py", line 95, in sflow
    sflow_vrfs[vrf]["source_interface"] = source_interface
                                          ^^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'source_interface' where it is not associated with a value
fatal: [dc1-leaf1a -> localhost]: FAILED! => {
    "changed": false,
    "msg": "cannot access local variable 'source_interface' where it is not associated with a value"
}```

And the variables to trigger that are shown above - in this case, defined at the DC level.

File: ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml (hunk 12 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/sflow_settings.schema.yml]: Setting a source interface is not strictly required, so I will keep this as is.

File: ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS.yml (hunk 17 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/SFLOW_TESTS.yml]: I am planning a follow on PR with a single boolean knob to build the streaming via terminattr. But there are some complexity when it comes to L2 switches. We will need to create a loopback interface in the default VRF, and that interface must have an IP address. So the second we add that, the requests for customizing the assigned loopback number and IP will also arise :).

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py (hunk 19 above)
Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py]: @jonxstill I don't think source interface is required for sflow is it? If so we should indeed add a model to define arbitrary source interfaces and detect missing. If not I think we just need to fix this bug by ignoring.

Comment (ClausHolbechArista) [reply ansible_collections/arista/avd/roles/eos_designs/python_modules/sflow/avdstructuredconfig.py]: Finally fixed this. I decided to add a new `sflow_settings.vrfs` list to set/override source interface per VRF.

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/get_structured_config/get_structured_config.py [hunk 20]
```diff
@@ -33,6 +34,11 @@
     AvdStructuredConfigInbandManagement,
     # Metadata must be after anything else that can generate structured config, since CV tags can consume from structured config.
     AvdStructuredConfigMetadata,
+    # The Flows module must be rendered after others contributing interfaces,
+    # since it parses those interfaces for sFlow or flow tracking (ipfix) config.
+    AvdStructuredConfigFlows,
```
gmuloc, line 37: Maybe we want metadata after Sflow?

ClausHolbechArista: absolutely. Thanks.

ClausHolbechArista: done

gmuloc (approved):

PR #3085, opened 2023-08-11
Bump(pyavd): pre-release 4.2.0a6

ClausHolbechArista:
pyavd pre-release 4.2.0a6


ClausHolbechArista:
This PR updates the allowed scopes for our conventional commits to also allow `pyavd` as a scope. It will not work for this PR itself, so we can just ignore the failing check.

carlbuchmann (approved):
LGTM

PR #3087, opened 2023-08-11
Feat(eos_designs): Compact mlag vtep ip allocations

### LINKED ISSUE #3086: Compact MLAG IP Allocation for VTEP IP Address
### Enhancement summary

Currently the VTEP IP addresses for leaf pairs are allocated solely based on the id of the mlag primary switch to avoid allocating the same IP address to multiple VTEPs.  However, this allocation method wastes an IP address per MLAG pair.  If a user id's switches such that they account for 2 possible switches in a Leaf pair, they should be able to use the `odd_id` formula (which is already implemented for the MLAG local source interface IP address allocation).

eos_designs

### Use case example

Wanting to conserve IP addresses in a network so the VTEP IPs allocated aren't wasted by only allocating based on the `mlag_primary_id`.

Provide an option for users to use the `odd_id` formula when allocating VTEP IP Addresses.

_No response_

### Additional context

_No response_

### Contributing Guide

jrecchia1029:
## Change Summary

Provide the ability to allocate consecutive IP addresses to different VTEPs

## Related Issue(s)

Fixes #3086 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Added a key under fabric_ip_addressing for vtep allocations

```
fabric_ip_addressing:
    vtep:
        algorithm: <str>
```

## How to test

Updated hosts and group_vars for `FABRIC_IP_ADDRESSING` in `eos_designs_unit_tests` molecule test to test new vtep allocation method

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_ip_addressing.schema.yml
```diff
@@ -25,3 +25,19 @@ keys:
...
+        keys:
+          algorithm:
+            type: str
+            default: "first_id"
+            description: |
+              This variable defines the VTEP IP allocation algorithm used.
+              Each VTEP Source Interface will get a /32 IP address allocated from the relevant vtep_loopback_ipv4_pool via a calculated offset.
+              The offset is calculated using one of the following algorithms:
+                - first_id: `(mlag_primary_id - 1) * 2` where `mlag_primary_id` is the ID of the first node defined under the node_group.
+                  This allocation method will skip every other /31 subnet making it less space efficient than `odd_id`.
```
ClausHolbechArista, line 39: We need to cover both MLAG and non-MLAG cases here. Also the offset calculation is `mlag_primary_id + loopback_ipv4_offset` for MLAG and `id + loopback_ipv4_offset` not non-MLAG.
And we skip every other **/32 address** (only for MLAG case).

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_ip_addressing.schema.yml
```diff
@@ -25,3 +25,19 @@ keys:
...
+                - odd_id: `(odd_id - 1) / 2`. Requires the node_group to have a node with an odd ID and a node with an even ID.
```
ClausHolbechArista: The `odd_id` would not work in non-MLAG cases, and even if we ignored it in those cases, it could lead to overlapping IPs if you have:
- MLAG pair A - ID 2, 3 -> offset 1
- Stand-alone Leaf - ID 4 -> offset 4
- MLAG pair B - ID 9, 10 -> offset 4

ClausHolbechArista: We could enforce that stand-alone leaf switches must use an odd ID, but it would just lead to more waste on regular loopback pools and uplinks.

ClausHolbechArista (commented):
Not sure this is the right solution. See my comments on docs.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_ip_addressing.schema.yml
jrecchia1029: Yes you are definitely right.  I think the idea was that odd_id would not be the default but could be selected by a user so long as they understood that devices in the fabric should be ID'd under the assumption that each switch getting a VTEP IP is in an MLAG pair (even if not all leaf switches are actually in an MLAG pair).

So if you had 3 groups of leafs where group 1 consisted of 2 leafs, group 2 consisted of 1 leaf, and group 3 consisted of 2 leafs, the IDs we would assume the user would use are:

MLAG pair A - ID 1, 2 -> offset 0
Standalone Leaf - ID 3 -> offset 1
MLAG pair B - ID 5, 6 -> offset 2

Still, it looks like more updates need to be made around the `vtep_ip()` function for standalone leafs

jrecchia1029: I see what you're saying though that this will lead to more waste on regular loopback and uplink pools

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/utils.py
```diff
@@ -96,3 +100,28 @@ def _mlag_odd_id_based_offset(self: "AvdIpAddressing") -> int:
...
+            if (self._mlag_primary_id % 2) == (self._mlag_secondary_id % 2):
+                raise AristaAvdError("VTEP compact addressing mode requires all MLAG pairs to have a single odd and even ID")
+            odd_id = self._mlag_primary_id
+            if odd_id % 2 == 0:
+                odd_id = self._mlag_secondary_id
+        else:
+            # Verify the node id of the standalone switch is an odd number
+            if self._id % 2 == 0:
+                raise AristaAvdError(
+                    "The VTEP IP allocation algorithm is set to odd_id which requires any non-MLAG VTEP id to be set to an odd number but"
```
ClausHolbechArista:
```suggestion
                    "The VTEP IP allocation algorithm is set to odd_id which requires the id of any non-MLAG VTEP to be set to an odd number but"
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_ip_addressing.schema.yml
ClausHolbechArista: Please rework this documentation to describe both algorithms for both mlag and non-mlag devices. And make sure to include `loopback_ipv4_offset` in the shown algorithms. Also make sure to mention the caveats of each.

carlbuchmann:
As discussed, this can lead to other issues and challenges when building networks with MLAG and non-MLAG switches. The key requirement here is to allocate IP addresses for VTEP more efficiently, and we agreed on the best way to do this upcoming IP pool manager functionality.

PR #3136, opened 2023-09-11
Refactor(plugins): Generate schema documentation using pydantic

ClausHolbechArista:
Refactoring schema tooling to use pydantic data classes for generating documentation.


ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/build-schemas.yml
```diff
@@ -36,3 +36,17 @@
...
+  ansible.builtin.copy:
+    content: "{{ role_name | arista.avd.convert_schema(type='pydantic') }}"
+    dest: "{{ role_schema_pydantic_path }}"
+    mode: 0664
+
+- name: Run black formatting on generated Pydantic Class file
+  tags: [eos_designs]
+  run_once: true
+  ansible.builtin.shell:
+    cmd: "black {{ role_schema_pydantic_path }}"
```
gmuloc: Maybe we can find a way to include this in the arista.avd.convert_schema part ? I assume black entrypoint can be used somehow and this means black becomes part of our requirements anyway right?

ansible_collections/arista/avd/plugins/plugin_utils/schema/avd_meta_schema.json
```diff
@@ -272,13 +272,6 @@
                         "$defs": { "$ref": "#/$defs/$defs" }
                     },
                     "additionalProperties": false
-                },
-                {
-                    "required": [ "$ref" ],
```
gmuloc, line 277: this would be breaking right?

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschemaresolver.py
```diff
@@ -93,13 +98,21 @@ def __new__(cls, schema, store):
...
+                    "keys": _keys,
+                    "dynamic_keys": _dynamic_keys,
+                },
+            )
+        else:
+            ValidatorClass = jsonschema.validators.create(
+                meta_schema=store["avd_meta_schema"],
+                validators={
+                    "$ref": _ref,
+                },
```
gmuloc: not sure iI get why this need a specific case

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+
+class TableRowGenBase(ABC):
+    """
+    Base class to be used with schema pydantic models.
+    Provides the method "get_table_rows" to build documentation tables
+    """
+
+    def get_table_rows(
+        self,
+        schema: AvdSchemaInt | AvdSchemaBool | AvdSchemaStr | AvdSchemaList | AvdSchemaDict,
```
gmuloc: probably worth creatinga type `AvdSchemaType` that is this union

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+        """
+        descriptions = []
+        if self.schema.description:
+            descriptions.append(self.schema.description.replace("\n", "<br>"))
+        if deprecation := self.get_deprecation_description():
+            descriptions.append(deprecation)
+        return " ".join(descriptions) or None
+
+    def render_children(self) -> Generator:
+        """Noop for classes without children. Overload in subclasses  for dict and list."""
```
gmuloc:
```suggestion
        """Noop for classes without children. Overload in subclasses for dict and list."""
```

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+        path: list[str] | None = None,
+        render_table: str | None = None,
+        parent_schema: AvdSchemaInt | AvdSchemaBool | AvdSchemaStr | AvdSchemaList | AvdSchemaDict | None = None,
+        first_list_key: bool = False,
+        inherited_table: str | None = None,
+    ) -> Generator[TableRow]:
+        self.schema = schema
+        self.path = path or []
+        self.parent_schema = parent_schema
+        self.render_table = render_table
```
gmuloc: `target_table` ? 
`render_table` feels like it should be a boolean

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+                description=self.render_description(),
+            )
+
+        yield from self.render_children()
+
+    def set_table(self, inherited_table: str | None):
+        if self.schema.documentation_options is not None:
+            self.table = getattr(self.schema.documentation_options, "table", None) or inherited_table
+        else:
+            self.table = inherited_table
```
gmuloc:
```suggestion
        if self.schema.documentation_options is not None:
            self.table = getattr(self.schema.documentation_options, "table", None)
        self.table = self.table or inherited_table
```

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+        if self.first_list_key:
+            return i * (indentation_count - 2) + "-" + i
+
+        return i * indentation_count
+
+    def get_key(self) -> str:
+        """
+        Get key from path
+        """
+        return self.path[-1]
```
gmuloc: maybe a property - self.key?

schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,360 @@
...
+from abc import ABC
+from typing import Generator
+
+from pydantic import BaseModel
+
+from ..metaschema.meta_schema_model import AvdSchemaBool, AvdSchemaDict, AvdSchemaInt, AvdSchemaList, AvdSchemaStr
+from ..metaschema.resolvemodel import resolve_model
+
+
+def get_table_rows(
```
gmuloc: should not it be some function of a AvdSchemaType as well?

Like

```
AvdSchemaBool.get_table_rows()
```

schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,288 @@
...
+        "Example '#/keys/mykey' will resolve the schema for 'mykey' under the root dictionary of the current schema"
+    )
+
+    def __str__(self):
+        return self.root.__str__()
+
+
+class Deprecation(BaseModel):
+    warning: bool = True
+    "Emit deprecation warning if key is set"
```
gmuloc: should it be docstring instead `"""` for docymenation ?

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/build-schemas.yml
ClausHolbechArista: All this goes away. Please ignore everything outside of /schema

ansible_collections/arista/avd/plugins/plugin_utils/schema/avd_meta_schema.json
ClausHolbechArista: I am updating all schemas where we only have a ref, to also include the type.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschemaresolver.py
ClausHolbechArista: I think it was something around the new jsonschema version I tried to solve, but please ignore this. Only stuff in root dir / schema will be relevant.

schema/generate_docs/tablerowgen.py
ClausHolbechArista: Yeah thanks

schema/metaschema/meta_schema_model.py
ClausHolbechArista: fixing

schema/generate_docs/tablerowgen.py
ClausHolbechArista: I have gone back and forth on this a couple of times, and I decided to keep it outside the main schema class, since we will have many of these converters that will only be needed for development workflows, where the main classes may be needed for ongoing validation.
Table
Yaml
Pydantic
Jsonschema? (the builtin jsonschema generator from pydantic does not handle contrained types)

schema/generate_docs/tablerowgen.py
ClausHolbechArista: fixed

schema/generate_docs/tablerowgen.py
ClausHolbechArista: done

schema/generate_docs/tablerowgen.py
ClausHolbechArista: Normally I would say yes, but to keep it in the style of the other get_xxx I think it's ok to keep this.

python-avd/pyavd/schema/generate_docs/utils.py
```diff
@@ -0,0 +1,52 @@
...
+if TYPE_CHECKING:
+    from ..metaschema.meta_schema_model import AristaAvdSchema, AvdSchemaField
+
+
+def list_tables(schema: AristaAvdSchema) -> set[str]:
+    """
+    Returns list of tables to render.
+
+    - Checks every root key for it's own table, and if not defined it adds a default table name using the key.
+    - Scans every root key for decendant tables.
```
gmuloc: couple of typos
```suggestion
    - Checks every root key for its own table, and if not defined adds a default table name using the key.
    - Scans every root key for descendant tables.
```

When reading the two bullets I am not sure what this function does and also it looks like we could do withtout `list_tables` looking at the inner code

python-avd/pyavd/schema/metaschema/resolvemodel.py
```diff
@@ -0,0 +1,62 @@
...
+
+def merge_schema_from_ref(schema: dict) -> dict:
+    if "$ref" not in schema:
+        return schema
+
+    schema = deepcopy(schema)
+    ref = schema.pop("$ref")
+    ref_schema = merge_schema_from_ref(get_schema_from_ref(ref))
+    if ref_schema["type"] != schema["type"]:
+        raise ValueError(
```
gmuloc, line 30: just a note to check later in the process if keeping these generic python errors is ok or if we should add "pyavd centric exceptions"

ClausHolbechArista: added a todo

python-avd/pyavd/schema/generate_docs/utils.py
ClausHolbechArista: This was correct before I simplified everything. simplifying the docstring too :)

python-avd/pyavd/schema/generate_docs/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+# that can be found in the LICENSE file.
+
+"""
+Generate AvdSchema documentation.
+
+"mdtabsgen.py" generates a markdown file with mkdocs tabs for one given "table name":
+- Markdown table representing the schema using "tablegen.py":
+  - Generates table rows by recursively walking the schema with the relevant TableRowGen<type>
+- YAML code block representing the schema using "yamlgen.py"
+  - Generates YAML lines by recursively walking the schema with the relevant YamlLineGen<type>
```
gmuloc, line 12: Should this be included in the pyavd documentation "somewhere"? (For now we only have a md file under `ansible_collections/arista/avd/docs` so probably not the time

python-avd/pyavd/schema/generate_docs/mdtabsgen.py
```diff
@@ -0,0 +1,31 @@
...
+
+    - Table tab contains a markdown table.
+    - YAML tab contains a markdown code block with YAML.
+    """
+    return f"""\
+<!--
+  ~ Copyright (c) 2023 Arista Networks, Inc.
+  ~ Use of this source code is governed by the Apache License 2.0
+  ~ that can be found in the LICENSE file.
+  -->
```
gmuloc: Can we fetch the text from https://github.com/aristanetworks/ansible-avd/blob/devel/development/license-short.txt somehow in to make sure it is kept in sync?

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+    ) -> Generator[TableRow]:
+        """
+        Yields TableRow for this schema field if the field is part of the given target_table.
+
+        Recursively walks children if applicable (only for lists and dicts)
+        """
+        self.schema = schema
+        self.target_table = target_table
+
+        # The render_schema_field function contains all the logic to device wether this field should be part of the target_table or not.
```
gmuloc:
```suggestion
        # The render_schema_field function contains all the logic to decide whether this field should be part of the target_table or not.
```

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+
+            yield from self.render_children()
+
+    def get_indentation(self) -> str:
+        """
+        Indentation is two spaces for dicts and 4 spaces for lists (so the hyphen will be indented 2)
+        """
+        indentation_count = len(self.schema._path) * 2 - 2
+        if not self.schema._key:
+            # this is a flat list item so path is one shorter than for dict. So we add 2 to the indentation
```
gmuloc, line 88: Probably worth documenting somewhere what `_path` should look like with a couple of examples (leaving this comment here but probably not the best place for it)

python-avd/pyavd/schema/generate_docs/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        # Always render the field if it is the primary key in a list of dicts.
+        return True
+
+    # Keeping this check at the end, since _descendant_tables is a recursive function that can be "expensive"
+    if target_table in schema._descendant_tables:
+        # Always render the field if a descendant field has the target table
+        return True
+
+    # Do not render the key if none of the above match.
+    return False
```
gmuloc, line 39: This could be a single line with documentation and we could keep the expensive check for the end thanks to Python optimisation of boolean expressions
```
return <blah> or <blah> or <blah> ..
```

python-avd/pyavd/schema/generate_docs/utils.py
```diff
@@ -0,0 +1,46 @@
...
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from ..metaschema.meta_schema_model import AristaAvdSchema, AvdSchemaField
+
+
+def list_tables(schema: AristaAvdSchema) -> set[str]:
```
gmuloc: This is not used anywhere

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        Returns None or a markdown formatted colored string with the deprecation status.
+        """
+        if self.schema.deprecation is None:
+            return ""
+
+        if self.schema.deprecation.removed:
+            label = "removed"
+        else:
+            label = "deprecated"
```
gmuloc:
```suggestion
        label = "removed" if self.schema.deprecation.removed else "deprecated"
```

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+        """
+        descriptions = []
+        if self.schema.description:
+            descriptions.append(self.schema.description.replace("\n", "<br>"))
+        description = " ".join(descriptions)
+
+        if deprecation := self.get_deprecation_description():
+            description += deprecation
+
+        return description or None
```
gmuloc: Somehow it would make more sense for me to write it like this (get the full list in descriptions and then join everything)

```suggestion
        if deprecation := self.get_deprecation_description():
            descriptions.append(deprecation)
       
        return " ".join(descriptions) or None
```

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+
+        Only covers generic restrictions. Should be overridden in type specific subclasses.
+        """
+        restrictions = []
+        valid_values = []
+        if getattr(self.schema, "dynamic_valid_values", None):
+            valid_values.append(f"<value(s) of {self.schema.dynamic_valid_values}>")
+
+        if getattr(self.schema, "valid_values", None):
+            valid_values.extend(self.schema.valid_values)
```
gmuloc, line 229: Could we not use our `get` function ? 
```suggestion
        valid_values.extend(get(self.schema.valid_values, []))
```

python-avd/pyavd/schema/generate_docs/tablerowgen.py
```diff
@@ -0,0 +1,355 @@
...
+        Leverages common restrictions from base class.
+        """
+        restrictions = []
+        if self.schema.min is not None:
+            restrictions.append(f"Min: {self.schema.min}")
+        if self.schema.max is not None:
+            restrictions.append(f"Max: {self.schema.max}")
+
+        if base_restrictions := super().render_restrictions():
+            restrictions.extend(base_restrictions.split("<br>"))
```
gmuloc: we could probably avoid spliting to rejoin everything just after

something like

```
class TableRowGenBase:
  def render_restrictions(self) -> str | None:
    return "<br>".join(self._get_restrictions_list) or None

    def _render_restrictions(self) -> list[str]:
        """ This is what you should override in child classes"
        restrictions = []
        valid_values = []
        [...]
       
        return restrictions

[...]

class TableRowGenInt(TableRowGenBase):
    def _render_restrictions(self) -> list[str]:
        restrictions = []
        if self.schema.min is not None:
            restrictions.append(f"Min: {self.schema.min}")
        if self.schema.max is not None:
            restrictions.append(f"Max: {self.schema.max}")
        restrictions.extend(super()._render_restrictions()

        return restrictions
```

python-avd/pyavd/schema/generate_docs/yamlgen.py
```diff
@@ -0,0 +1,28 @@
...
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from ..metaschema.meta_schema_model import AristaAvdSchema
+
+
+def get_yaml(schema: AristaAvdSchema, target_table: str | None = None) -> str:
+    """
+    Returns one markdown codeblock with yaml either containing all keys of the given schema or only a subset if "table_name" is set.
```
gmuloc:
```suggestion
    Returns one markdown codeblock with yaml either containing all keys of the given schema or only a subset if "target_table" is set.
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def render_annotation_link(self, annotation_number: int) -> str:
+        """
+        Returns codeblock comment used for mkdocs codeblock annotations
+
+        Like: " # (123)!"
+        """
+        if self.annotation is None:
+            return ""
+        return f" # ({annotation_number})!"
```
gmuloc:
```suggestion
        return "" if self.annotation is None else f" # ({annotation_number})!"
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+
+if TYPE_CHECKING:
+    from ..metaschema.meta_schema_model import AvdSchemaField
+
+LEGACY_OUTPUT = True
+
+
+class YamlLine(BaseModel):
+    """
+    Dataclass for one table row.
```
gmuloc: Is it really a table row?

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+    Content of line is yaml formatted so it should be rendered in a code block.
+    The field may need an mkdocs code block annotation link to show large default values.
+    If so the content of the annotation is put into .annotation.
```
gmuloc:
```suggestion
    If so the content of the annotation is stored in the 'annotation' attribute.
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+
+    Since annotation numbers must be tracked across all yaml lines, it is up to the caller
+    to give us to annotation number at rendering time.
```
gmuloc:
```suggestion
    to provide the annotation number at rendering time.
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+
+            ```yaml
+            default: "one"
+            values: 2
+            ```
+        """
+        if self.annotation is None:
+            return ""
+        annotation_codeblock = f"```yaml\n{self.annotation}```"
+        return f"\n{annotation_number}. Default Value\n\n{indent(annotation_codeblock, '    ')}\n"
```
gmuloc, line 57: so I assume for now our only use case with annotations is `Default Value`? The docstring of the class says "The field may need an mkdocs code block annotation link to show large default values."

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+            return (i * (indentation_count - 2)) + "-" + i
+
+        return i * indentation_count
+
+    @property
+    def is_removed(self) -> bool:
+        if self.schema.deprecation and self.schema.deprecation.removed:
+            return True
+
+        return False
```
gmuloc:
```suggestion
        return self.schema.deprecation and self.schema.deprecation.removed:
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+        # TODO: Remove legacy output
+        if LEGACY_OUTPUT:
+            value = self.schema.type
+        else:
+            value = "; ".join(field for field in value_fields if field)
+
+        if self.schema._key:
+            key = f"{self.schema._key}: "
+        else:
+            key = ""
```
gmuloc:
```suggestion
        key = f"{self.schema._key}: " if self.schema._key else ""
```

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+        if self.schema.min is not None or self.schema.max is not None:
+            if self.schema.min is not None and self.schema.max is not None:
+                restrictions.append(f"{self.schema.min}-{self.schema.max}")
+            elif self.schema.min is not None:
+                restrictions.append(f">={self.schema.min}")
+            elif self.schema.max is not None:
+                restrictions.append(f"<={self.schema.max}")
+
+        if base_restrictions := super().get_restrictions():
+            restrictions.extend(base_restrictions.split("; "))
```
gmuloc: same comment as md for the split then join

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+        Leverages common restrictions from base class.
+        """
+        restrictions = []
+        if self.schema.min_length is not None or self.schema.max_length is not None:
+            if self.schema.min_length is not None and self.schema.max_length is not None:
+                restrictions.append(f"length {self.schema.min_length}-{self.schema.max_length}")
+            elif self.schema.min_length is not None:
+                restrictions.append(f"length >={self.schema.min_length}")
+            elif self.schema.max_length is not None:
+                restrictions.append(f"length <={self.schema.max_length}")
```
gmuloc, line 282: repeating with the Int stuff - maybe some helper method?

python-avd/pyavd/schema/generate_docs/yamllinegen.py
```diff
@@ -0,0 +1,412 @@
...
+            properties = ""
+
+        if not self.schema.keys or (self.schema.documentation_options and self.schema.documentation_options.hide_keys):
+            # Add <dict> when we don't generate yaml for child keys.
+            properties = f" <dict>{properties}"
+
+        if self.schema._key:
+            key = f"{self.schema._key}:"
+        else:
+            key = ""
```
gmuloc:
```suggestion
        key = f"{self.schema._key}:" if self.schema._key else ""
```

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+This module provides Pydantic models (classes) representing the meta-schema of the AVD Schema.
+
+Each variable in the schema is called a field, and for each type of field we have a corrosponding Pydantic model:
+- AvdSchemaInt
+- AvdSchemaBool
+- AvdSchemaStr
+- AvdSchemaList
+- AvdSchemaDict
+
+The alias "AvdSchemaField" is a union of of all the models above, and can be used as easy type hint for any field type.
```
gmuloc, line 27: Do you need this or could you simply use `AvdSchemaBaseModel` instead (though I agree the name may be better with Field)

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+
+All the type-specific Pydantic models inherit the common base class "AvdSchemaBaseModel", and have local overrides
+as needed. For example it is only "AvdSchemaList" and "AvdSchemaDict" that needs to parse child fields.
```
gmuloc:
```suggestion
as needed. For example, only "AvdSchemaList" and "AvdSchemaDict" need to parse child fields.
```

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+
+The overall schema is covered by the class "AristaAvdSchema" which inherits from a "AvdSchemaDict" since the root of the schema is a dict.
```
gmuloc:
```suggestion
The overall schema is covered by the class "AristaAvdSchema" which inherits from "AvdSchemaDict" since the root of the schema is a dict.
```

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+
+KEY_PATTERN = r"^[a-z][a-z0-9_]*$"
+"""Common pattern to match legal key strings"""
+DYNAMIC_KEY_PATTERN = r"^[a-z][a-z0-9_.]*$"
+"""Common pattern to match legal dynamic key strings"""
+KEY_PATTERN_WITH_UPPERCASE = r"^[a-zA-Z][a-zA-Z0-9_]*$"
+"""This is a temporary pattern allowing uppercase keys too"""
+
+
+class AvdSchemaBaseModel(BaseModel, ABC):
```
gmuloc, line 44: Maybe we can just rename this `AvdSchemaField` and keep the docstring to avoid confusion?

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+    # Signal to __init__ if the $ref in the schema should be resolved before initilizing the pydantic model.
+    _resolve_schema: ClassVar[bool] = True
+
+    def __init__(self, resolve_schema: bool | None = None, **data):
+        """
+        Overrides BaseModel.__init__(**data).
+        Takes a kwarg "resolve_schema" which controls if all subclasses of AvdSchemaBaseModel should expand any $ref in the input schema.
+
+        The $ref expansion _only_ covers this field.
+        Any $ref on child fields are expanded as they are initialized by Pydantic since they are based on this base class.
```
gmuloc, line 129: Is this still correct? I thought we discussed (and it seems to be implemented) the class attribute to be applied even on children?
It seems to contradict the line 112 comment

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+    def _table(self) -> str | None:
+        """
+        Return the name of the schema documentation table where this field should be included.
+
+        The table name can be derived from several sources depending on the position and configuration of the field.
+        In order the sources are:
+        - Statically defined under "documentation_options.table".
+        - Inherit from parent schema if available.
+        - None if this is the root dict.
+          This means that the first level of keys - called root keys - will not find a parent table.
```
gmuloc:
```suggestion
        - None if this is the root dict. This means that the first level of keys - called root keys - will not find a parent table.
```

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+        if self.documentation_options and self.documentation_options.table:
+            return self.documentation_options.table
+
+        # No local table, so use the _table from the parent_schema if available.
+        if self._parent_schema and self._parent_schema._table:
+            return self._parent_schema._table
+
+        # This should never happen, since only the root key should be without a parent_schema.
+        if len(self._path) != 1:
+            raise NotImplementedError("Something went wrong in _table", self._path)
```
gmuloc: probably a `ValueError`

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+
+        # This is a root key the default table is the key with hyphens and removing <,>
+        return self._key.replace("<", "").replace(">", "").replace("_", "-")
+
+    @cached_property
+    def _path(self) -> list[str]:
+        """
+        Returns the variable path for this field to be used in schema docs.
+        Like "rootkey.subkey.[].mykey".
```
gmuloc, line 184: is this notation a standard ?

I am more used to RESTCONF like path
```suggestion
        Like "rootkey.subkey[].mykey".
```

(which granted is less easy to parse :)

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+    """
+    A regular expression which will be matched on the variable value.
+    The regular expression should be valid according to the ECMA 262 dialect.
+    Remember to use double escapes.
+    """
+    valid_values: List[str] | None = None
+    """List of valid values"""
+    dynamic_valid_values: str | None = None
+    """
+    Path to variable under the parent dictionary containing valid values.
```
gmuloc, line 337: Can we remove the limitation of parent dict with this - by referencing full path in schema from the store? (maybe a later feature ;) )

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+        if not self.items:
+            return set()
+
+        descendant_tables = set()
+        descendant_tables.add(self.items._table)
+        descendant_tables.update(self.items._descendant_tables)
+
+        return descendant_tables
+
+    def model_post_init(self, __context: Any) -> None:
```
gmuloc, line 407: Neat! Just read about this in pydantic doc! Very good usage

python-avd/pyavd/schema/metaschema/meta_schema_model.py
```diff
@@ -0,0 +1,559 @@
...
+            - _is_first_list_key
+        """
+        if self.items:
+            self.items._parent_schema = self
+            if isinstance(self.items, AvdSchemaDict) and self.items.keys:
+                first_key = True
+                for key, grandchildschema in self.items.keys.items():
+                    if first_key:
+                        grandchildschema._is_first_list_key = True
+                        first_key = False
```
gmuloc, line 427: Maybe need some comment in the definition of the class variable to indicate it is used in documentation to select the indentation? (I had not remembered this variable from other files :) )

gmuloc (commented):
That;s a very great PR!

python-avd/pyavd/schema/generate_docs/utils.py
ClausHolbechArista: For doc generation I don't think we need to optimize this. I prefer to keep it readable.

python-avd/pyavd/schema/generate_docs/__init__.py
ClausHolbechArista: For now we will keep this internally and not expose as part of the pyavd API.

python-avd/pyavd/schema/generate_docs/mdtabsgen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: schema is a pydantic class but I use getattr here since we don't have valid_values on dict and list types. I don't think it makes sense to introduce get() here.

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: Did something like this (which was also the original plan. Issue was that the current md files did not have a space between the description and the deprecation, so I changed it later.)

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: done

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/utils.py
ClausHolbechArista: Removed

python-avd/pyavd/schema/generate_docs/yamlgen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: Fixed

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: done

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: done

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: Correct we only use this for large default values.

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: There is a difference between child fields and subclasses.

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: done

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: done

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: It would not enforce the actual types. I prefer to keep the alias.

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: I prefer to keep the alias so we get type checking of that too.

python-avd/pyavd/schema/metaschema/meta_schema_model.py
ClausHolbechArista: fixed

python-avd/pyavd/schema/generate_docs/tablerowgen.py
ClausHolbechArista: The nested function would not work well with subclasses so I will split this into two - a method and a @property to do this in a cleaner way.

python-avd/pyavd/schema/generate_docs/yamllinegen.py
ClausHolbechArista: There are slight differences, so I don't it would be beneficial (for readability) to write a helper.

carlbuchmann (approved):
LGTM

PR #3162, opened 2023-09-23
Feat(eos_designs): Automatic assignment of Node IDs using pool manager

ClausHolbechArista:
## Change Summary

Auto assignment of Node IDs using pool manager

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

```yaml
# PREVIEW: This feature is in marked as "preview", which means it is subject to change at any time.
#
# Assignment policies for numbers like Node ID.
fabric_numbering:

  # Assignment policy for Node ID.
  # Node ID is mainly used for IP address assignment but can also affect BGP AS and/or
  # interface assignments depending on other settings.
  node_id:

    # IDs will be automatically assigned according to the configured algorithm.
    # - `static` will use the statically set IDs under node setting.
    # - `pool_manager` will activate the pool manager for ID pools.
    #   Any statically set ID under node settings will be reserved in the pool if possible.
    #   Otherwise an error will be raised.
    algorithm: <str; "static" | "pool_manager"; default="static">

    # Path to file to use for storing ID pool data when using "pool_manager" as algorithm.
    # This can be an absolute path or a path relative to current working directory.
    #
    # By default the path is "<root_dir>/intended/data/<fabric_name>-ids.yml".
    #
    # Note: Since the pool manager will remove stale entries after every run, each fabric should be using it's own file.
    pools_file: <str>
```

## How to test

Added molecule coverage. Also needs manual testing by the reviewer.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs-twodc-5stage-clos/intended/data/test-ids.yml
```diff
@@ -0,0 +1,111 @@
+id_pools:
```
carlbuchmann: Add the following comment:
```
# This file is auto-generated by AVD eos_designs.
# When there is a merge conflict, rebase the branch, accept all current changes, and re-run AVD.
```

ansible_collections/arista/avd/docs/pyavd.md
```diff
@@ -113,3 +113,9 @@ md-toc>=8.1.8
     options:
       show_root_toc_entry: false
       paths: ../../../../python-avd
+
+::: pyavd.validation_result
+    options:
+      show_root_toc_entry: false
+      paths: ../../../../python-avd
```
ClausHolbechArista:
```suggestion
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+                id_pool = id_data_id_pool
+                break
+
+        if not id_pool:
+            # Create a new dict, add to the list of pools.
+            id_pool = {
+                "fabric_name": fabric_name,
+                "dc_name": shared_utils.dc_name,
+                "pod_name": shared_utils.pod_name,
+                "type": shared_utils.type,
```
jrecchia1029: I get why we need to match on the `type` but from the perspective of the campus studio leveraging pyavd, this may result in issues if a user were to re-architect their network from an L2 campus to an L3 campus because of the way Node IDs work in the campus studio.  Basically, for an L2 and L3 campus design in the campus studio we treat leafs and member leafs as different types.  In AVD, in an L2 campus the design is `l2ls` and leafs and member leafs are both `leafs` while in an L3 campus the design is `l3ls-evpn` and leafs are `l3leafs` and member leafs are `l2leafs`.  The solution to this may not require anything with the pool manager and may just involve adding another node type to the `l2ls` avd design.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_numbering.schema.yml
```diff
@@ -0,0 +1,44 @@
...
+                - `fabric_name`
+                - `dc_name`
+                - `pod_name`
+                - `type`
+
+                Note: This means changing any of these fields may renumber the devices!
+
+                Each pool will assign the first available ID starting from 1.
+
+                Stale entries will be reclaimed from each pool automatically after every run.
```
jrecchia1029: It may be overkill but also may be useful to offer an example for how to re-use stale entries. Something to the effect of `i.e. If you are replacing a particular node, node A, that is using id:1 with another host, node B that you wish to use id:1, first you will need to remove node A from the inventory file, run the eos_designs playbook successfully, then add node B to the inventory file (in the same location node A was), and then re-run the playbook a second time successfully.`

jrecchia1029 (commented):
Will want to do some more testing but the logic looks good to me.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: It would require some rework of the campus studio logic, to maybe introduce another type like member-leaf or whatever we decide. We have other things to discuss around this, like simpler data models for such member leafs etc. I am also okay with adding another "layer" in here if we feel for it, but let's talk first.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_numbering.schema.yml
ClausHolbechArista: I will try to move some of this doc to the markdown instead of the table. That will make it easier to be more elaborate.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: Another option is to also add the `rack` property we have already. It could be exposed as "IDF" in the campus scenario. But today different racks are still in the same pool of IDs, so maybe that is not a good idea.

carlbuchmann: I will address this after testing.

ansible_collections/arista/avd/molecule/eos_designs-twodc-5stage-clos/intended/data/test-ids.yml
```diff
@@ -0,0 +1,113 @@
+# This file is auto-generated by AVD eos_designs.
+# When there is a merge conflict, rebase the branch, accept all current changes, and re-run AVD.
+id_pools:
```
carlbuchmann: Would it be possible to indent the list with two spaces so that it allows to collapse each list easily (in IDE)?

ClausHolbechArista: Nope. We use PyYAML with integrated libyaml for performance. The only way to render the extra spaces is to use a custom dumper in pyyaml (loosing libyaml performance), or change to ruamel.yaml which would be a new requirement just for this.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_numbering.schema.yml
carlbuchmann: I will update the docs

carlbuchmann:
Converting back to draft - after discussions with @ClausHolbechArista we agreed to put a restriction allowing either ID pool manager or statically defined IDs but not both. AVD should raise an error if the ID pool manager is enabled and `id` is assigned to a node.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/fabric_numbering.schema.yml
ClausHolbechArista: updated.

ansible_collections/arista/avd/docs/pyavd.md
```diff
@@ -112,4 +110,7 @@ md-toc>=8.1.8
 ::: pyavd.validation_result
     options:
       show_root_toc_entry: false
-      paths: ../../../../python-avd
+
+::: pool_manager.avdpoolmanager
```
gmuloc: probably not a "for-now" question, but as this is brand new code and feature, can it not be directly in pyavd and we use it from there?

carlbuchmann (dismissed):
LGTM

- [x] Reviewed documentation.
- [x] Tested setting id on node and enabling pool manager, and id set on the node is considered and updates pool manager.
- [x] Update fabric-ids.yml manually to get desired id.
- [x] Manually set duplicate ids in fabric-ids.yml - resulted in structure config error.
- [x] Update dc_name after initial generation to update ids.
- [x] Tested on large multi-planar fabric with 700+ devices and no noticeable impact on performance.
- [x] Tested custom file path by setting `pools_file`.

Let me know if you would like more tests to be performed.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/fabric_numbering_id.yml
```diff
@@ -0,0 +1,11 @@
...
+
+spine:
+  defaults:
+    bgp_as: 65001-65100
+    loopback_ipv4_pool: 192.168.0.0/24
+    # id: 1 will be set by pool_manager
+
+fabric_numbering:
+  id:
+    algorithm: pool_manager
```
gmuloc: so here, and it is probably only for the test,  you set this in the host_vars but potentially another device could have the id set to 1 in group_vars given that the fabric_name is shared. 

I am not saying this is something we can catch (or maybe we can?) but could lead to quite confusing configuration

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/misc.py
```diff
@@ -39,6 +41,40 @@ def is_deployed(self: SharedUtils) -> bool:
...
+            Use pool manager
+
+        If running under eos_designs_facts and pool manager is _not_ activated:
+            Use 'self.switch_data_combined.id' which is the ID defined in the node type config or None.
+
+        """
+        # Check if we are running from eos_designs_structured_config ("switch" is a dict)
+        if isinstance(get(self.hostvars, f"avd_switch_facts..{self.hostname}..switch", separator=".."), dict):
+            # Return value of 'self.hostvars.switch.id' or None
+            return get(self.hostvars, f"avd_switch_facts..{self.hostname}..switch..id", separator="..")
```
gmuloc:
```
>>> test = {"id": 42}
>>> if isinstance((blah:=test), dict):
...   print(blah["id"])
...
42
```

```suggestion
        if isinstance(switch:=get(self.hostvars, f"avd_switch_facts..{self.hostname}..switch", separator=".."), dict):
            # Return value of 'self.hostvars.switch.id' or None
            return switch.get("id")
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/misc.py
```diff
@@ -39,6 +41,40 @@ def is_deployed(self: SharedUtils) -> bool:
 
     @cached_property
     def id(self: SharedUtils) -> int | None:
+        """
+        Node ID.
```
gmuloc: are we testing this in the tests we make in eos_designs?

for sure options 2 and 3 but where is the test for option 1?

ansible_collections/arista/avd/plugins/action/eos_designs_facts.py
```diff
@@ -55,7 +58,9 @@ def run(self, tmp=None, task_vars=None):
         # Get updated templar instance to be passed along to our simplified "templater"
         self.templar = get_templar(self, task_vars)
 
-        avd_switch_facts_instances = self.create_avd_switch_facts_instances(fabric_hosts, hostvars, result)
+        pool_manager = AvdPoolManager(output_dir)
```
gmuloc: is it always necessary to instantiate this?

gmuloc: hmm it looks simpler maybe to always have it

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/shared_utils.py
```diff
@@ -58,6 +65,7 @@ class SharedUtils(
     The class cannot be overriden.
     """
 
-    def __init__(self, hostvars: dict, templar) -> None:
+    def __init__(self, hostvars: dict, templar, pool_manager: AvdPoolManager | None = None) -> None:
```
gmuloc: do we need to make this optional if we always instantiate it?

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -110,6 +110,52 @@ roles/eos_designs/docs/tables/fabric-topology.md
...
+
+### Node ID Algorithm
+
+IDs will be automatically assigned according to the configured algorithm.
+
+- `static` will use the statically set IDs under [node setting](#node-type-common-configuration).
+- `pool_manager` will activate the pool manager for ID pools.
+
+!!! note
+    It is not allowed to override `pool_manager` with a static ID for the same device, since that could lead to allocation conflicts.
```
gmuloc: we can still end up with allocation conflict if the `fabric_numbering` algorithm is different per device based on ansible vars structure maybe let's warn for this (stating the potential issue and that it is the user responsibility to check?)

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -110,6 +110,52 @@ roles/eos_designs/docs/tables/fabric-topology.md
 roles/eos_designs/docs/tables/fabric-ip-addressing.md
 --8<--
 
+## Fabric Numbering
+
+Fabric Numbering controls how various numbers are derived across the fabric.
```
gmuloc, line 135: maybe mention the key name `fabric_numbering`

Maybe the name should be `fabric_allocation` and the sentence red `how various _variables_ are derived across the fabric` so that we can do more than numbers later without having to rename the top level key

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -110,6 +110,52 @@ roles/eos_designs/docs/tables/fabric-topology.md
...
+When using `pool_manager` for node IDs the pools are dynamically built and matched on the following device variables:
+
+- `fabric_name`
+- `dc_name`
+- `pod_name`
+- `type`
+
+Each pool will assign the first available ID starting from 1.
+
+It is important to make sure the combination of these variables is unique for each indended pool of devices.
```
gmuloc: would we not catch this?

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/utils.py
```diff
@@ -14,6 +14,11 @@ class UtilsMixin:
     """
     Mixin Class with internal functions.
     Class should only be used as Mixin to an AvdInterfaceDescriptions class
+
+    Functions here serve as an abstraction to shared_utils, so we can change shared_utils
+    without affecting third party AvdPoolManager subclasses.
```
gmuloc: Why does it say AvdPoolManager for descriptions?

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/utils.py
```diff
@@ -16,6 +16,11 @@ class UtilsMixin:
     """
     Mixin Class with internal functions.
     Class should only be used as Mixin to an AvdIpAddressing class
+
+    Functions here serve as an abstraction to shared_utils, so we can change shared_utils
+    without affecting third party AvdPoolManager subclasses.
```
gmuloc: Why does it say AvdPoolManager for ip addressing?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+from ansible_collections.arista.avd.plugins.filter.natural_sort import natural_sort
+from ansible_collections.arista.avd.plugins.plugin_utils.errors import AristaAvdError
+from ansible_collections.arista.avd.plugins.plugin_utils.utils import get, get_item
+
+if TYPE_CHECKING:
+    from ansible_collections.arista.avd.plugins.plugin_utils.eos_designs_shared_utils import SharedUtils
+
+FILE_HEADER = """\
+# This file is auto-generated by AVD eos_designs.
+# When there is a merge conflict, rebase the branch, accept all current changes, and re-run AVD.
```
gmuloc:
```suggestion
# When there is a merge conflict for this file, rebase the branch, accept all current changes for this file, and re-run AVD.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+    This class is imported and initialized once in eos_designs_facts
+    and given to shared_utils for each device.
+    """
+
+    def __init__(self, output_dir: str):
+        self.output_dir = Path(output_dir)
+        self.id_files: set[Path] = set()
+        self.changed_id_files: set[Path] = set()
+
+    def remove_stale_assignments(self):
```
gmuloc:
```suggestion
    def remove_stale_assignments(self) -> None:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Walk through all pools and in-place remove stale assignments and empty pools.
+
+        An assigment is deemed stale if "accessed" is not True.
+        The "accessed" field is also removed from the pool data.
+
+        Updates self.changed_id_files if any objects were removed from self.id_pool_data.
+
+        Note: This method is called from save_updated_pools and should _not_ be called manually.
```
gmuloc: maybe rename it `_remove_stale_assignments`

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        for id_file in self.id_files:
+            id_data = self.id_pool_data(id_file)
+            for id_pool in id_data["id_pools"]:
+                # Remove id_assignments that were not accesses using "last_accessed".
+                id_assignments_count = len(id_pool["id_assignments"])
```
gmuloc: maybe name it `on_file_id_assignments_count`, same for pools

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+                if id_assignments_count != len(id_pool["id_assignments"]):
+                    self.changed_id_files.add(id_file)
+
+            # Remove pools if there are no assignments left.
+            id_pools_count = len(id_data["id_pools"])
+            id_data["id_pools"] = [id_pool for id_pool in id_data["id_pools"] if id_pool["id_assignments"]]
+            if id_pools_count != len(id_data["id_pools"]):
+                self.changed_id_files.add(id_file)
+
+    def save_updated_pools(self, Dumper=SafeDumper):
```
gmuloc: not sure having capitalized keywords argument is legit
```suggestion
    def save_updated_pools(self, dumper: <type>=SafeDumper) -> None:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        for id_file in self.changed_id_files:
+            id_data = self.id_pool_data(id_file)
+            for id_pool in id_data["id_pools"]:
+                # Sort id_assignments on "id".
+                id_pool["id_assignments"] = natural_sort(id_pool["id_assignments"], "id")
+
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "type")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "pod_name")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "dc_name")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "fabric_name")
```
gmuloc: can we not make something better rather than sorting 4 times?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+
+            # Remove pools if there are no assignments left.
+            id_pools_count = len(id_data["id_pools"])
+            id_data["id_pools"] = [id_pool for id_pool in id_data["id_pools"] if id_pool["id_assignments"]]
+            if id_pools_count != len(id_data["id_pools"]):
+                self.changed_id_files.add(id_file)
+
+    def save_updated_pools(self, Dumper=SafeDumper):
+        """
+        Save cached data for all changed files.
```
gmuloc:
```suggestion
        Save cached data back to file for all changed files.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+            id_data = self.id_pool_data(id_file)
+            for id_pool in id_data["id_pools"]:
+                # Sort id_assignments on "id".
+                id_pool["id_assignments"] = natural_sort(id_pool["id_assignments"], "id")
+
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "type")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "pod_name")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "dc_name")
+            id_data["id_pools"] = natural_sort(id_data["id_pools"], "fabric_name")
+            id_file.write_text(FILE_HEADER + dump(id_data, Dumper=Dumper), encoding="UTF-8")
```
gmuloc: hmmm probably should try catch around this as we could get permission denied trying to write to a file when we are authorized to write it (HELLO CONTAINERS ;) 🐳 )

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        self.id_pool_data.cache_clear()
+        self.changed_id_files = set()
+        self.id_files = set()
+
+    @lru_cache
+    def id_pool_data(self, id_file: Path) -> dict:
```
gmuloc: docstring

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        if not id_file.exists():
+            # Try to create the dir and file.
+            id_file.parent.mkdir(mode=0o775, parents=True, exist_ok=True)
+            id_file.touch(mode=0o664)
```
gmuloc: maybe need some try /except here as well

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Read from file or assign an empty data model.
+        id_data = safe_load(id_file.read_text(encoding="UTF-8")) or {"id_pools": []}
```
gmuloc: catch `yaml.parser.ParserError` and maybe others - this file could have been ~~messed up~~ modified by the user (as our doc mentions it is an option)

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+
+        if not isinstance(id_data, dict) or not isinstance(id_data.get("id_pools"), list):
+            raise AristaAvdError(f"Invalid ID pool manager data when reading {id_file}. Expecting {'id_pools': []}")
```
gmuloc: are we expecting `{'id_pools': []}` always?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+
+        # Since we assigned a new ID, we have to mark the id pool as changed to have it saved once the build is done.
+        self.id_files.add(id_file)
```
gmuloc: reading the comment, I would expect the `id_file` to be added to the `self.changed_id_files` rather than just the `self.id_files` maybe rename the `self.id_files` to `self.dirty_id_files`?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+
+        id_pool = None
+        for id_data_id_pool in id_data["id_pools"]:
+            if (
+                id_data_id_pool.get("fabric_name") == fabric_name
+                and id_data_id_pool.get("dc_name") == shared_utils.dc_name
+                and id_data_id_pool.get("pod_name") == shared_utils.pod_name
+                and id_data_id_pool.get("type") == shared_utils.type
+            ):
+                id_pool = id_data_id_pool
```
gmuloc: almost thinking a pydantic class for a Pool would be nice

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        Returns the ID of one node if found in the pool.
+        Otherwise a new entry is inserted into the pool and the ID is returned.
+
+        The pools are dynamically built and matched on the following device data:
+        - Fabric Name (from 'shared_utils.hostvars.fabric_name')
+        - DC Name (from `shared_utils.dc_name`)
+        - POD Name (from `shared_utils.pod_name`)
+        - Node type (from `shared_utils.type`)
+
```
gmuloc: is this future-compatible with our known use cases (campus / wan / ... ) ?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
gmuloc: plus could write methods like `pool.first_available_id()` blah blah

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+            }
+            id_data["id_pools"].append(id_pool)
+
+        if id_assignment := get_item(id_pool["id_assignments"], "hostname", shared_utils.hostname):
+            # Found an ID, so return quickly.
+            id_assignment["accessed"] = True
+            return id_assignment["id"]
+
+        pool_sorted_on_id = sorted(id_pool["id_assignments"], key=lambda x: x.get("id"))
+        first_available_id = next((index for index, id_assignment in enumerate(pool_sorted_on_id + [{"id": 0}], 1) if id_assignment["id"] != index))
```
gmuloc: so if I understand the logic, we add the dummy assignments `[{"id": 0}]` in the list because we want ids starting at 1 BUT we enumerate also starting at 1 - so what's the purpose of adding the dummy id ?

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
```diff
@@ -978,6 +978,36 @@ keys:
       in the Fabric, **must** be an inventory group name.
     type: str
     required: true
+  fabric_numbering:
```
gmuloc: as mentioned above probably best to call it `allocations` rather than `numbering`

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
```diff
@@ -0,0 +1,168 @@
...
+if TYPE_CHECKING:
+    from ansible_collections.arista.avd.plugins.plugin_utils.eos_designs_shared_utils import SharedUtils
+
+FILE_HEADER = """\
+# This file is auto-generated by AVD eos_designs.
+# When there is a merge conflict, rebase the branch, accept all current changes, and re-run AVD.
+"""
+
+
+class AvdPoolManager:
```
gmuloc: This module would deserve proper pytest given its complexity

ansible_collections/arista/avd/docs/pyavd.md
ClausHolbechArista: I am not sure if you are talking about the docs and md file your are commenting on, or if this is a more general question.
1. For docs we have to keep it under the collection, since that is the mkdocs root.
2. For code we have to keep it together withing the ansible collection for it to be shipped as part of the collection. Remember that pyavd is not a dependency of ansible-avd (yet)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/fabric_numbering_id.yml
ClausHolbechArista: If ppl want to split and use fabric_numbering for a subset of devices, I think we can accept that. Ex. only use it for your l2leafs or something. We catch the issue of manual id being set together with pool-manager on one device.
1. It is very hard to catch this.
2. I am not sure we really want to prevent this. Being open for a step-wise introduction of pool-manager is good IMO.
3. We also accept overlapping IDs today, since ppl might have custom algorithms.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/misc.py
ClausHolbechArista: All code using `self.shared_utils.id` in `eos_designs_structured_config` use this.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/shared_utils.py
ClausHolbechArista: When shared_utils is instatiated from `avd_structured_config` it is _not_ getting a pool manager.
Later we might add a "read-only" mode to the pool manager, and then we can give it in that case too.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: We already have `fabric_ip_addressing`. to cover that case. I am not sure `fabric_allocation` is clear to the regular user, but I am open to find another name.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: No if ppl don't set pod_name or dc_name (which are optional), we would put all l2leaf into the same pool across multiple DCs. Should we require setting dc_name, pod_name?

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: Yes, we can add guidance to set this in a <FABRIC> group_vars file.

ansible_collections/arista/avd/roles/eos_designs/python_modules/interface_descriptions/utils.py
ClausHolbechArista: too much copy paste. I think we should change this to use an object instead (in another PR)

ansible_collections/arista/avd/roles/eos_designs/python_modules/ip_addressing/utils.py
ClausHolbechArista: too much copy paste. I think we should change this to use an object instead (in another PR)

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: ?

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: Not today. Unless those scenarios reuse the same variables.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: To catch the empty list. This is the most effective method, but we need to change this algorithm since it does not handle duplicate entries. So if a user edits, and create duplicates, we cannot find the correct next-available.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: yes we are getting to that point. Not sure if it will be pydantic or just some other class.

ansible_collections/arista/avd/roles/eos_designs/schemas/eos_designs.schema.yml
ClausHolbechArista: disputed in the other comment. I will close this so we only have the discussion once.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: Done. Please review. I focused on the behavior, and not the individual attributes, so we can refactor/rename anything internal to the class.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: done

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: I don't think that is the correct name. I have used `len_before_removing_stale_entries` for this one.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: I have changed the description, since this is _not_ the dirty files. It is _all_ files.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: as discussed offline, we will need some indirection to be added (outside of this PR).

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: I have changed the algorithm to something that is simpler to read (I hope) and also handles duplicate entries.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/fabric_numbering_id.yml
carlbuchmann: I agree with Claus' comments.

carlbuchmann (dismissed):
LGTM

- [x] Reviewed documentation
- [x] Tested negative test setting id on node and enabling pool manager
- [x] Update fabric-ids.yml manually to get desired id
- [x] Manually set duplicate ids in fabric-ids.yml - resulted in structure config error
- [x] Update dc_name after initial generation to update ids
- [x] Tested on small multi-planar fabric with 147 devices and generated same ids as I had set manually, and no noticeable impact on performance.

ansible_collections/arista/avd/docs/pyavd.md
ClausHolbechArista: no longer relevant

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: implemented classes

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: refactored everything so closing these old comments.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: refactored everything so closing these old comments.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: refactored everything so closing these old comments.

ansible_collections/arista/avd/roles/eos_designs/python_modules/pool_manager/avdpoolmanager.py
ClausHolbechArista: We should not continue if we cannot create the file, so I think raising is fine with the original error.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: added some warnings but also handling static vs pool id conflicts in code now (for one device, so still a problem between devices).

ClausHolbechArista:
change keys to dicts

carlbuchmann:
Results from large multi-planar fabric test (1107 nodes)

```bash
---

1. Base - No pool manager enabled

TASKS RECAP **************************************************************************************************************************************
Monday 20 January 2025  16:54:50 -0500 (0:02:18.457)       0:10:45.970 ********
===============================================================================
arista.avd.eos_designs : Generate device configuration in structured format ------------------------------------------------------------- 484.88s
arista.avd.eos_cli_config_gen : Generate eos intended configuration and device documentation -------------------------------------------- 138.46s
arista.avd.eos_designs : Set eos_designs facts ------------------------------------------------------------------------------------------- 12.02s
arista.avd.eos_designs : Generate fabric documentation ------------------------------------------------------------------------------------ 5.61s
arista.avd.eos_designs : Remove avd_switch_facts ------------------------------------------------------------------------------------------ 3.63s
arista.avd.eos_designs : Create required output directories if not present ---------------------------------------------------------------- 0.78s
arista.avd.eos_designs : Verify Requirements ---------------------------------------------------------------------------------------------- 0.42s
arista.avd.eos_cli_config_gen : Verify Requirements --------------------------------------------------------------------------------------- 0.09s

PLAYBOOK RECAP ***********************************************************************************************************************************
Playbook run took 0 days, 0 hours, 10 minutes, 45 seconds

2. Pool manager enabled 1st run (data/Fabric-ids.yml not present)

TASKS RECAP *****************************************************************************************************************************************************************************************************************************************************************************************************************************
Monday 20 January 2025  20:50:37 -0500 (0:01:09.331)       0:10:34.089 ********
===============================================================================
arista.avd.eos_designs : Generate device configuration in structured format ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 540.48s
arista.avd.eos_cli_config_gen : Generate eos intended configuration and device documentation ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 69.33s
arista.avd.eos_designs : Set eos_designs facts ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 12.75s
arista.avd.eos_designs : Generate fabric documentation --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 5.92s
arista.avd.eos_designs : Remove avd_switch_facts --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 4.19s
arista.avd.eos_designs : Create required output directories if not present ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 0.80s
arista.avd.eos_designs : Verify Requirements ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 0.43s
arista.avd.eos_cli_config_gen : Verify Requirements ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 0.11s

PLAYBOOK RECAP **************************************************************************************************************************************************************************************************************************************************************************************************************************
Playbook run took 0 days, 0 hours, 10 minutes, 34 seconds


3. Pool manager enabled 2nd run (data/Fabric-ids.yml present)

TASKS RECAP *****************************************************************************************************************************************************************************************************************************************************************************************************************************
Monday 20 January 2025  21:02:50 -0500 (0:01:10.851)       0:10:04.389 ********
===============================================================================
arista.avd.eos_designs : Generate device configuration in structured format ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 508.85s
arista.avd.eos_cli_config_gen : Generate eos intended configuration and device documentation ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 70.85s
arista.avd.eos_designs : Set eos_designs facts ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 12.44s
arista.avd.eos_designs : Generate fabric documentation --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 6.54s
arista.avd.eos_designs : Remove avd_switch_facts --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 4.24s
arista.avd.eos_designs : Create required output directories if not present ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 0.84s
arista.avd.eos_designs : Verify Requirements ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 0.43s
arista.avd.eos_cli_config_gen : Verify Requirements ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ 0.11s

PLAYBOOK RECAP **************************************************************************************************************************************************************************************************************************************************************************************************************************
Playbook run took 0 days, 0 hours, 10 minutes, 4 seconds
```

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -130,6 +130,61 @@ roles/eos_designs/docs/tables/fabric-topology.md
 roles/eos_designs/docs/tables/fabric-ip-addressing.md
 --8<--
 
+## PREVIEW - Fabric Numbering
```
carlbuchmann, line 133: I agree to include "PREVIEW" in the title, but we are inconsistent, i.e. "Zscaler Internet Exit". I think we should update Zscaler title, which can be done in another PR.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -130,6 +130,61 @@ roles/eos_designs/docs/tables/fabric-topology.md
...
+- `static` will use the statically set IDs under node setting.
+- `pool_manager` will activate the pool manager for ID pools.
+  Any statically set ID under node settings will be reserved in the pool if possible.
+  Otherwise an error will be raised.
+
+!!! note
+    It is strongly encouraged to use the same Node ID algorithm for all devices in the fabric.
+    Using different algorithms for groups of devices may lead to duplicates or inconsistent allocations.
+
+    The pool manager will not change IDs if already set under the node settings,
```
carlbuchmann:
```suggestion
    The pool manager will not change IDs if they are already set under the node settings,
```

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -130,6 +130,61 @@ roles/eos_designs/docs/tables/fabric-topology.md
...
+- `dc_name`
+- `pod_name`
+- `type`
+
+Each pool will assign the first available ID starting from 1. Any statically set ID under node settings will be reserved in the pool if possible, otherwise an error will be raised.
+
+It is important to make sure the *combination* of the variables above is unique for each intended pool of devices.
+
+!!! warning
+    This means changing any of these fields may renumber the node IDs, and in turn lead to renumbering of IP addresses etc!
```
carlbuchmann:
```suggestion
    This means changing any of these fields may renumber the node IDs and, in turn, lead to the renumbering of IP addresses, etc.
```

carlbuchmann (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -130,6 +130,61 @@ ansible_collections/arista/avd/roles/eos_designs/docs/tables/fabric-topology.md
...
+    Since stale entries are only reclaimed *after* every run, it is not possible to reuse an ID when removing and adding a new device
+    as part of the same execution of AVD.
+
+    To reuse a freed ID, first remove the old device and run AVD. Then add the new device and rerun AVD.
+
+The pool manager stores data in a YAML file per fabric. The default path is `<root_dir>/intended/data/<fabric_name>-ids.yml`
+
+!!! tip
+    It is possible to override the automatic assignments by editing the data files manually.
+    Just make sure to have a backup or use source control like Git and rerun AVD after changing the file.
```
gmuloc, line 186: so we still want to encourage this?

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
gmuloc: once we start growing this functionality, how do we anticipate documenting the IP addressing allocation with pool manager? will we end up with a centralized place with pool manager behavior or repeat in every section?

python-avd/pyavd/_eos_designs/schema/schema_fragments/fabric_numbering.schema.yml
```diff
@@ -0,0 +1,43 @@
...
+                Otherwise an error will be raised.
+          pools_file:
+            type: str
+            description: |-
+              Path to file to use for storing ID pool data when using "pool_manager" as algorithm.
+              This can be an absolute path or a path relative to current working directory.
+
+              By default the path is "<root_dir>/intended/data/<fabric_name>-ids.yml".
+
+              Note: Since the pool manager will remove stale entries after every run, each fabric should be using it's own file.
```
gmuloc:
```suggestion
              Note: Since the pool manager will remove stale entries after every run, each fabric should be using its own file.
```

python-avd/pyavd/api/pool_manager/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import TYPE_CHECKING, ClassVar, Literal
+
+from yaml import CSafeDumper
+
+from .node_id_pools import NodeIdPoolCollection
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+    from pyavd._eos_designs.shared_utils import SharedUtils
```
gmuloc: if we merge this before mixins then we will need to tackle it there

python-avd/pyavd/api/pool_manager/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+    This class is imported and initialized once in eos_designs_facts
+    and given to shared_utils for each device.
+    """
+
+    _pool_collection_types: ClassVar = {"node_id_pools": NodeIdPoolCollection}
+
+    _output_dir: Path
+    _pool_collections: dict[PoolType, dict[Path, PoolCollection]]
+
```
gmuloc:
```suggestion
```

python-avd/pyavd/api/pool_manager/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+class PoolManager:
+    """
+    Class used to handle pooled resources.
+
+    This class is imported and initialized once in eos_designs_facts
+    and given to shared_utils for each device.
+    """
+
+    _pool_collection_types: ClassVar = {"node_id_pools": NodeIdPoolCollection}
```
gmuloc, line 30: https://docs.python.org/3/library/typing.html

```suggestion
    _pool_collection_types: ClassVar[dict[str, type]] = {"node_id_pools": NodeIdPoolCollection}
```

python-avd/pyavd/api/pool_manager/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def get_pool(self, pool_type: PoolType, shared_utils: SharedUtils) -> Pool:
+        """Returns the pool of the given type for this device. Pool will be autocreated if missing."""
+        if pool_type not in self._pool_collection_types:
+            msg = f"Invalid pool type '{pool_type}'. Expected one of {', '.join(self._pool_collection_types.keys())}"
+            raise ValueError(msg)
+
+        pools_files_map = self._pool_collections.setdefault(pool_type, {})
+        pools_cls = self._pool_collection_types[pool_type]
+        pools_file = pools_cls._pool_file_from_shared_utils(self._output_dir, shared_utils)
```
gmuloc: This looks more correct?

```suggestion
        pool_file = pools_cls._pool_file_from_shared_utils(self._output_dir, shared_utils)
```

python-avd/pyavd/api/pool_manager/__init__.py
```diff
@@ -0,0 +1,84 @@
...
+
+        pool_key = pools_cls._pool_key_from_shared_utils(shared_utils)
+        return pool_collection.get_pool(pool_key)
+
+    def get_assignment(self, pool_type: PoolType, shared_utils: SharedUtils, requested_value: int | str | None = None) -> int:
+        """
+        Returns the assignment value for this device for the given pool type. Assignment and pool will be autocreated if missing.
+
+        Args:
+            pool_type: Currently only supports "node_id_poools".
```
gmuloc:
```suggestion
            pool_type: Currently only supports "node_id_pools".
```

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,282 @@
...
+T_AssignmentKeyType = TypeVar("T_AssignmentKeyType", bound="AssignmentKey")
+
+FILE_HEADER = """\
+# This file is auto-generated by AVD eos_designs.
+# When there is a merge conflict for this file, rebase the branch, accept all current changes for this file, and re-run AVD.
+"""
+
+
+@dataclass(frozen=True)
+class PoolKey:
```
gmuloc, line 33: some docstrings would be nice (othewise it is not obvious why we don't need just a Key class vs PoolKey and Assignment Key (yes we do, yes they are not the same))

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,282 @@
...
+    """One Pool of resources. Currently only supporting a pool of Integers."""
+
+    collection: PoolCollection[T_PoolKeyType, T_AssignmentKeyType, T_ValueType]
+    pool_key: T_PoolKeyType
+    assignments: dict[T_AssignmentKeyType, PoolAssignment[T_AssignmentKeyType, T_ValueType]]
+
+    def _is_value_available(self, value: T_ValueType) -> bool:
+        if self.collection.value_type is int and isinstance(value, int):
+            collection = cast(PoolCollection[T_PoolKeyType, T_AssignmentKeyType, int], self.collection)
+            assignments = cast(dict[T_AssignmentKeyType, PoolAssignment[T_AssignmentKeyType, int]], self.assignments)
```
gmuloc: why do we need to cast?

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,282 @@
...
+        self._remove_stale_assignments()
+        if not self.changed:
+            return False
+
+        if not self.pools_file.exists():
+            # Try to create the dir and file.
+            self.pools_file.parent.mkdir(mode=0o775, parents=True, exist_ok=True)
+            self.pools_file.touch(mode=0o664)
+
+        self.pools_file.write_text(FILE_HEADER + dump({self.pools_key: self.as_list()}, Dumper=dumper_cls))
```
gmuloc: need to handle permission error / OSError

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # currently we only support int pools
+        return False
+
+    def _next_available(self) -> T_ValueType:
+        if self.collection.value_type is int:
+            collection = cast(PoolCollection[T_PoolKeyType, T_AssignmentKeyType, int], self.collection)
+            assignments = cast(dict[T_AssignmentKeyType, PoolAssignment[T_AssignmentKeyType, int]], self.assignments)
+            existing_ids = {assignment.value for assignment in assignments.values()}
+            available_ids = set(range(collection.min_value, len(existing_ids) + 2)) - existing_ids
```
gmuloc: I undersrand the logic here though it is not obvious so probably we should add a comment explaining why this is enough to do this.

Also what do you mean by `range` ? here if my `collection.min_value=10` and I have 2 ids say 11 and 12 this yields:

```
>>> existing_ids = {11, 12}
>>> min_value = 10
>>> set(range(min_value, len(existing_ids) + 2)) - existing_ids
set()
>>> range(min_value, len(existing_ids) + 2)
range(10, 4)
```

I think you meant to say a range of length eixting_ids + 2 starting from colleciton.min_value

This does not show up in the people testing the PR because we always start from 1 in the ID pool

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,282 @@
...
+            # Existing item
+            self.assignments[key].accessed = True
+
+        return self.assignments[key]
+
+    def _remove_stale_assignments(self) -> None:
+        """
+        Remove stale assignments. An assignment is deemed stale if "accessed" is not True. Sets parent .changed if it removed anything.
+
+        Note: This method is called from the parent collection and should _not_ be called manually.
```
gmuloc, line 126: should it be moved to the parent collection class then? as it is really the object which has this responsability

gmuloc (changes requested):
7 files to go but sharing the first comments

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: sure. We made the file more accessible to still allow this.

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
ClausHolbechArista: The documentation would be different for IP addresses, since we would need to figure out how to make it work together with the legacy features. I don't think this is something we need to plan for now.

python-avd/pyavd/api/pool_manager/__init__.py
ClausHolbechArista: one file can hold multiple pools.

ClausHolbechArista: I will rename the method instead.

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: I don't think this worked in py3.10 type hinting. After upgrading to 3.12 I don't need this.

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: I think I have improved it a bit now by not creating an extra set from range. It will save on iterations in case of available IDs "in between" other IDs.

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: Yeah I guess it could now when we track everything from the collection layer.

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: reraising

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: these are nested through separate layers, and it is up to the resource (could be subclassed) to clean up their assignments.

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+        if key not in self.assignments:
+            # No existing assignment. Use the requested value if given and available. Otherwise use next available.
+            new_assignment_value = requested_value if requested_value is not None and self._is_value_available(requested_value) else self._next_available()
+            self.assignments[key] = self.collection.assignment_cls(key=key, value=new_assignment_value, accessed=True)
+            self.collection.changed = True
+        elif requested_value is not None and requested_value != self.assignments[key].value and self._is_value_available(requested_value):
+            # Existing assignment but not with the requested value. Change the existing assignment to the requested_value.
+            self.assignments[key] = self.collection.assignment_cls(key=key, value=requested_value, accessed=True)
+            self.collection.changed = True
```
gmuloc, line 115: would be cool to be able to have the previous call setting the bool `self.collection.changed`

Maybe a help funciton 
```
self.collection.assign(key, value):

class Collection
  def assign(self, key, value):
    self.assignment_cls(key=key, value=new_assignment_value, accessed=True)
    self.changed = True
```

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,295 @@
...
+            self._pools[pool_key] = self.pool_cls(collection=self, pool_key=pool_key, assignments={})
+            self.changed = True
+
+        return self._pools[pool_key]
+
+    def _remove_stale_assignments(self) -> None:
+        """
+        Remove stale assignments from all pools in the collection and remove empty pools afterwards.
+
+        Note: This method is called from the parent pool manager and should _not_ be called manually.
```
gmuloc: this is called from local `saved_updates` maybe we can prefix it with two `__` so that it cannot be called from outside the class?

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -0,0 +1,295 @@
...
+        """
+        len_before = len(self._pools)
+        [pool._remove_stale_assignments() for pool in self._pools.values()]
+        self._pools = {pool_key: pool for pool_key, pool in self._pools.items() if pool.assignments}
```
gmuloc: one loop
```suggestion
        non_empty_pools = {}
        for pool_key, pool in self._pools.items():
            pool._remove_state_assignments()
            if pool.assignments:
                non_empty_pools[pool_key] = pool
        self._pools = non_empty_pools
```

python-avd/pyavd/api/pool_manager/base_classes.py
gmuloc: ack

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: use __ and update docstring

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: awe will look more into it later - at least when we add some other classes.

gmuloc (approved):

PR #3206, opened 2023-10-05
Fix: Remove "anta" from requirements while the integration is in preview

ClausHolbechArista:
Remove "anta" from requirements while the integration is in preview

### DIFF
--- a/.github/workflows/pull-request-management.yml
+++ b/.github/workflows/pull-request-management.yml
@@ -343,10 +343,12 @@ jobs:
         avd_scenario:
           - 'eos_validate_state'
         ansible_version:
-          - 'ansible-core<2.16.0 --upgrade'
+          # Temporarily installing anta manually while the feature is in preview.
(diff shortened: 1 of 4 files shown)


.github/workflows/pull-request-management.yml
```diff
@@ -343,10 +343,12 @@ jobs:
         avd_scenario:
           - 'eos_validate_state'
         ansible_version:
-          - 'ansible-core<2.16.0 --upgrade'
+          # Temporarily installing anta manually while the feature is in preview.
+          - 'ansible-core<2.16.0 anta==0.9.0 --upgrade'
```
gmuloc: can we make a file for this so we can change the value in only one place instead?

gmuloc (commented):
Missing something in eos_validate_state task for verify requirements to check for anta

carl-baillargeon (commented):
Please update to 0.10.0. Thanks!

gmuloc (approved):

carl-baillargeon (approved):
LGTM

PR #3207, opened 2023-10-05
Refactor(eos_designs): Remove EVPN limitation for underlay_filter_peer_as

emilarista:
## Change Summary

Removes the check for device evpn_role: client/server when deciding whether to configure the route-map for underlay_filter_peer_as. Also raises an exception when configuring this knob together with default VRF EVPN services since this will create conflicting config.

Created negative unit test and updated documentation.

## Component(s) name

`arista.avd.eos_designs`

## How to test
Tested with molecule and customer repo.

### Repository Checklist


ClausHolbechArista (changes requested):
I have updated the negative tests. You only need to add a new play when the failure happens in the eos_designs_facts phase.
You still need to remove the EVPN restriction from the underlay filter feature and add a positive molecule test for that combination.

emilarista:
> I have updated the negative tests. You only need to add a new play when the failure happens in the eos_designs_facts phase. You still need to remove the EVPN restriction from the underlay filter feature and add a positive molecule test for that combination.

I already removed the restriction, now added positive molecule test!

ClausHolbechArista (dismissed):

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM tested the correct configuration with both RFC5549 and ipv4 underlay.
Tested negative test  with EVPN services in the default VRF

PR #3228, opened 2023-10-11
Feat(eos_designs): Allow 'default' in default_interfaces node types

jonxstill:
## Change Summary

Allow the use of 'default' in `default_interfaces.*.types` key.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Specifying 'default' in the list of types will apply these default interfaces to all node types (the node's platform must still match a platform type regular expression).

No change to the existing data model, other than documenting the use of the 'all' keyword in the schema.

```
default_interfaces:
  - types: [ default ]
    platforms: [ 7300X3 ]
    uplink_interfaces: [ Ethernet1/27-30/1 ]
    mlag_interfaces: [ Ethernet1/31-32/1 ]
    downlink_interfaces: [ Ethernet1/1-26/1 ]
```

## How to test
Existing molecule tests modified to include `types: [ default ]` similar to above.

Create a default_interface entry as per example above - should apply to all 7300X devices, regardless of node type.

### Repository Checklist


ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -168,13 +168,15 @@ def default_interfaces(self: SharedUtils) -> dict:
...
+                if search(f"^{platform}$", device_platform) and (
+                    self.type in default_interface.get("types", []) or "default" in default_interface.get("types", [])
+                ):
                     return default_interface
 
         # If not found, then look for a default default_interface that matches our type
         for default_interface in default_interfaces:
             for platform in default_interface.get("platforms", []):
-                if search(f"^{platform}$", "default") and self.type in default_interface.get("types", []):
+                if search(f"^{platform}$", "default") and (self.type in default_interface.get("types", []) or "default" in default_interface.get("types", [])):
```
ClausHolbechArista, line 186: This logic will not work for all cases. It is suspectible to reordering changes. Since first match will win, so if default is before the more specific match it will take the default. I think you need another for loop.

Another problem:
```
# , platform , type
1 , default  , default
2 , abc      , default
3 , default  , xyz
4 , abc      , xyz
```
The two middle ones will lead to ambiguity. Which one will it take, since they are equally imprecise. I am not sure which behavior we want here.

jonxstill: Yeah that makes sense.  I had considered if we just wanted to document that the first matching rule wins, or that any entries with 'default' in either list should be at the end of the list?  Let me try and make this more deterministic.

jonxstill: Latest commit reorders the default_interfaces like so:
```
        # #  platform  type
        # 1  abc       xyz
        # 2  abc       default
        # 3  default   xyz
        # 4  default   default
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AVD_LAB.yml
```diff
@@ -51,6 +51,14 @@ ntp:
 
 # Support for per-platform automatic interface allocation
 default_interfaces:
+  - types: [ default ]
+    platforms: [ default ]
+    # Higher-numbered interfaces used to differentiate from
+    # other definitions below
+    uplink_interfaces: [ Ethernet79-82/1 ]
```
gmuloc, line 58: Looks like we are not testing this

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -162,23 +162,46 @@ def default_interfaces(self: SharedUtils) -> dict:
...
+        # Re-order default_interfaces:
+        # #  platform  type
+        # 1  abc       xyz
+        # 2  abc       default
+        # 3  default   xyz
+        # 4  default   default
+        default_interfaces.sort(key=self.sort_default_interfaces)
+
         # First look for a matching default interface set that matches our platform and type
         for default_interface in default_interfaces:
```
gmuloc, line 176: I feel like now that it is sorted you actually don't need both loops but only one, given that you will first get the non default platforms first and then the default ones

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -162,23 +162,46 @@ def default_interfaces(self: SharedUtils) -> dict:
...
+        Set sort order for each default_interface, based on platforms and types keys
+        """
+        if "default" not in default_interface.get("platforms") and "default" not in default_interface.get("types"):
+            return 1
+        elif "default" not in default_interface.get("platforms") and "default" in default_interface.get("types"):
+            return 2
+        elif "default" in default_interface.get("platforms") and "default" not in default_interface.get("types"):
+            return 3
+        else:
+            return 4
```
gmuloc, line 203: so how do we end up sorting things like

```
  - types: [ spine ]
    platforms: [ "7050[SC]X3" ]
    uplink_interfaces: [ Ethernet49-52/1 ]
    mlag_interfaces: [ Ethernet53-54/1 ]
    downlink_interfaces: [ Ethernet1-30/1 ]

  - types: [ l3leaf, spine ]
    platforms: [ "7050[SC]X3" ]
    uplink_interfaces: [ Ethernet27-34 ]

  - types: [ spine, l3leaf ]
    platforms: [ "7050[SC]X3" ]
    uplink_interfaces: [ Ethernet42-48 ]
```

is it deterministic if I reverse the order of declaration ? or do we configure this config as wrong ? maybe we can sort the platform /type lists?

For this logic I am feeling like it could be worth writing some python tests with some inputs and expected outputs.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/default_interfaces.schema.yml
```diff
@@ -18,7 +18,10 @@ keys:
         types:
           type: list
           required: true
-          description: List of node type keys.
+          description: |
```
gmuloc, line 21:
```suggestion
          description: |-
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/default_interfaces.schema.yml
```diff
@@ -18,7 +18,10 @@ keys:
...
+            List of node type keys.
+            'default' will apply these default interfaces to node types where there is no more-specific match.
```
gmuloc, line 23: this also means that you cannot use the node type "default" if you define your own node_type_keys without breaking this - probably best to mention it somewhere (like `default` is reserved or whatever in node_type_keys options) ? @ClausHolbechArista

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/default_interfaces.schema.yml
```diff
@@ -18,7 +18,10 @@ keys:
...
+            Note that the node's platform must still match a platform type regular expression.
```
gmuloc, line 24: which can be `default` right (if not overwritten)?

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -162,23 +162,46 @@ def default_interfaces(self: SharedUtils) -> dict:
...
+                if search(f"^{platform}$", device_platform) and (
+                    self.type in default_interface.get("types", []) or "default" in default_interface.get("types", [])
+                ):
                     return default_interface
 
         # If not found, then look for a default default_interface that matches our type
         for default_interface in default_interfaces:
             for platform in default_interface.get("platforms", []):
-                if search(f"^{platform}$", "default") and self.type in default_interface.get("types", []):
+                if search(f"^{platform}$", "default") and (self.type in default_interface.get("types", []) or "default" in default_interface.get("types", [])):
```
gmuloc, line 186:
```
search(f"^{platform}$", "default")
```
could be
```
platform == "default"
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/default_interfaces.schema.yml
jonxstill: We should probably add a check for that.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
jonxstill: Yup, I see what you're saying there.  That's definitely worth adding to make it a little more deterministic.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/default_interfaces.schema.yml
jonxstill: Yup, will amend docs to make that clear.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
jonxstill: Unnecessary since we don't need that 2nd loop.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
jonxstill: Agreed.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
jonxstill: Actually, scratch my earlier comment.  If we're just using "in" to verify if the node's type is 'in' the types list, why do we care about the order of those lists?  What matters is the order of the sorted default_interfaces list.  The sort routine currently used preserves the order in the original list if two identical sort keys are returned.  Is this deterministic? No.  If you change the order of the two conflicting items, you will get different results.  However I don't think we can save users from every single mistake.  The sorting applied was designed to ensure that anything with a 'default' in it, was matched last.  This satisfies the original use case of being able to define one default/default option, and then override it as needed.  I think beyond that we need to make it clear that once we find a platform that matches and then a type that matches, we stop looking.

gmuloc: ok but so
```
  - types: [ default, spine ]
    platforms: [ "7050[SC]X3" ]
    uplink_interfaces: [ Ethernet49-52/1 ]
    mlag_interfaces: [ Ethernet53-54/1 ]
    downlink_interfaces: [ Ethernet1-30/1 ]

  - types: [ l3leaf, spine ]
    platforms: [ "7050[SC]X3" ]
    uplink_interfaces: [ Ethernet27-34 ]
```

you sort -> my spine get the second stuff only because it was with default in the first one

gmuloc: and yes I understand this is how it works and I even find it quite logical.
I would ask if we are guaranteed that the order of keys is respected when sorting keys with the same value for all python versions we support also (though I surely hope so)

jonxstill: OK.  So `platforms` was an existing key so I can't change the behaviour of that.  At present "default" as a platform has no intrinsic meeting - it's just a `platform_settings` entry named that.  If we remove consideration of "default" in the `platforms` key and just focus on sorting by `types` then we could:
* Sort by any `types` that are not-default, then default. 
* Enforce (becuse "default" as a type is a new concept) that "default" has to be the only item in the list
** i.e. `if "default" in types and len(types) > 1: raise AristaAvdError()`

This would eliminate the ambiguity in your example above and would also preserve the order of the list, with the exception of the new "default" entry.

@ClausHolbechArista - would appreciate your input in this too please.

ClausHolbechArista:
Putting this on ice since the complexity for the user is higher than the benefit.

ClausHolbechArista:
We will close this since it becomes too complex. The user just have to repeat default interfaces for each node type. Should not be too bad.

PR #3231, opened 2023-10-12
Feat(eos_cli_config_gen): Add ethernet interface dhcp server config

gmuloc:
## Change Summary

Adding
```
interface Ethernet2
  dhcp server ipv4
  dhcp server ipv6
```

the cli is introduced between `bfd` and `ipv6 enable` as per EOS cli:

```
ceos2(config-if-Et2)#show active
interface Ethernet2
   ip address 1.1.1.1/24
   bfd echo
   dhcp server ipv4
   dhcp server ipv6
   ipv6 enable
```

## Component(s) name

## How to test

molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/ethernet-interfaces.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/ethernet-interfaces.md
@@ -4,6 +4,8 @@
 
 - [Management](#management)
   - [Management Interfaces](#management-interfaces)
+- [DHCP Server](#dhcp-server)
+  - [DHCP Server interfaces](#dhcp-server-interfaces)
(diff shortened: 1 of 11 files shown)


ClausHolbechArista (approved):
It would be nice to get this implemented too for port-channels and SVIs if applicable, but we can take that another day :)

carlbuchmann (approved):
LGTM

pvinci-arista:
There are some caveats.  Do we want to add something to the jsonschema description or somewhere else to call them out?

The DHCP Server and DHCP Relay cannot be configured on the same router. The DHCP Relay has a higher precedence so the DHCP Server will be disabled when the DHCP Relay is configured.

If DHCPD is enabled on the router, configuring the EOS DHCP Server may cause DHCPD to crash or have unexpected behaviors.

PR #3239, opened 2023-10-13
Fix(eos_cli_config_gen): Remove requirement for MACSec license and FIPS

### LINKED ISSUE #3168: eos_cli_config_gen: Schema for macsec is too restrictive.
### Issue Summary

The schema for macsec has a lot of required keys, which are not required in the CLI.

eos_cli_config_gen

### How do you run AVD ?

None

### Steps to reproduce

```shell
N/A
```

### Relevant log output

```shell
N/A
```

### Contributing Guide

xaviramon:
## Change Summary

Today, AVD requests the user to define MACSec license file and key, a method we are changing for the EOS command `license import`. These variables are no longer a must to be able to use MACSec in the device.
Same way, FIPS is not a mandatory field for MACSec.

## Related Issue(s)

Fixes #3168 

## Component(s) name

## Proposed changes
Remove requirement for these variables in the schema.

## How to test

Molecule tests pass

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/mac_security.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/mac_security.schema.yml
@@ -15,8 +15,10 @@ keys:
         keys:
           license_name:
             type: str
+            required: true
           license_key:
             type: str
+            required: true
       fips_restrictions:
         type: bool
       profiles:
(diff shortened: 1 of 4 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/mac_security.schema.yml
```diff
@@ -12,17 +12,13 @@ keys:
     keys:
       license:
         type: dict
-        required: true
         keys:
           license_name:
             type: str
```
ClausHolbechArista, line 17: I guess these two keys inside the `license` dict should still be required. Since it does not make sense to have the dangling `license` dict without both of these. The user can still leave out the entire `license` since you remove that requirement in line 15.

carlbuchmann (approved):
LGTM

ClausHolbechArista (approved):

PR #3255, opened 2023-10-17
Feat(pyavd): Adding "path" attribute to validation_errors and deprecation_warnings.

Shivani-gslab:
## Change Summary

Adding "path" attribute to validation_errors and deprecation_warnings.

## Proposed changes

- Added `path` attribute to AvdValidationError, AvdConversionWarning and AvdDeprecationWarning class.
- Fixed the path for list elements in deprecation warnings.

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/errors/errors.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/errors/errors.py
@@ -50,7 +50,7 @@ class AvdValidationError(AristaAvdError):
 
         if isinstance(error, (jsonschema.ValidationError)):
             self.path = self._json_path_to_string(error.absolute_path)
-            self.message = f"'Validation Error: {self._json_path_to_string(error.absolute_path)}': {error.message}"
+            self.message = f"'Validation Error: {self.path}': {error.message}"
         else:
             self.message = message
         super().__init__(self.message)
@@ -59,7 +59,8 @@ class AvdValidationError(AristaAvdError):
 class AvdConversionWarning(AristaAvdError):
     def __init__(self, message: str = "Data was converted to conform to schema", key=None, oldtype="unknown", newtype="unknown"):
         if key is not None:
-            self.message = f"'Data Type Converted: {key} from '{oldtype}' to '{newtype}'"
+            self.path = self._json_path_to_string(key)
+            self.message = f"'Data Type Converted: {self.path} from '{oldtype}' to '{newtype}'"
         else:
             self.message = message
         super().__init__(self.message)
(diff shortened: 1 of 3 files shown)


ansible_collections/arista/avd/plugins/plugin_utils/errors/errors.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, message: str = "Schema Error", error=None):
             raise AristaAvdError('Python library "jsonschema" must be installed to use this plugin') from JSONSCHEMA_IMPORT_ERROR
 
         if isinstance(error, (jsonschema.ValidationError)):
+            self.path = self._json_path_to_string(error.absolute_path)
```
ClausHolbechArista, line 52: I think we should just take the `error.absolute_path` instead of having it as a string.

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
```diff
@@ -112,16 +112,18 @@ def convert_items(self, items: dict, data: list, schema: dict, path: list[str]):
             return
 
         for index, item in enumerate(data):
+            path[-1]+=f"[{index}]"
```
ClausHolbechArista: This will in-place change change `path`, so you will have an invalid path on the second and following loops. It was added below, to avoid modifying the original list.

ClausHolbechArista: You probably just need to edit the existing added items to be `+ [index]` instead of a string.

ansible_collections/arista/avd/plugins/plugin_utils/errors/errors.py
Shivani-gslab: it will be a list then, eg. deque(['router_bgp', 'peer_groups', 0, 'link_bandwidth'])

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
Shivani-gslab: if I do this `yield from self.convert_data(item, items, path + [str(index)])`, it will be like - router_bgp.peer_groups.0.peer_filter

ansible_collections/arista/avd/plugins/plugin_utils/errors/errors.py
Shivani-gslab: Fixed!

ansible_collections/arista/avd/plugins/plugin_utils/schema/avddataconverter.py
Shivani-gslab: Fixed!

ClausHolbechArista (approved):
LGTM! Thank you for fixing this.

carlbuchmann (approved):
LGTM, negative test case provides correct results.

PR #3264, opened 2023-10-18
Feat(eos_cli_config_gen): Add other valid_values for event-handler trigger 'on-boot'

gmuloc:
## Change Summary

Add additional valid_values for event-handler trigger 'on-boot'

## Component(s) name

## Proposed changes

Uses the value from EOS:

```
ceos2(config-handler-TOTO)#trigger ?
  on-boot            Trigger condition occurs on system boot
  on-config          Trigger the event handler when it is configured
  on-counters        Trigger condition occurs on evaluating statistical counters
  on-intf            trigger condition occurs on specified interface changes
  on-logging         Trigger condition occurs when regex match any log message
  on-maintenance     Trigger condition occurs on maintenance operation
  on-startup-config  trigger condition occurs on startup config changes
  vm-tracer          Trigger condition occurs on VmTracer events
```

## How to test

molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/event-handlers.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/event-handlers.md
@@ -46,11 +46,6 @@ interface Management1
 | CONFIG_VERSIONING | bash | <code>FN=/mnt/flash/startup-config; LFN="`ls -1 $FN.*-* \| tail -n 1`"; if [ -z "$LFN" -o -n "`diff -I 'last modified' $FN $LFN`" ]; then cp $FN $FN.`date +%Y%m%d-%H%M%S`; ls -1r $FN.*-* \| tail -n +11 \| xargs -I % rm %; fi</code> | on-startup-config |
 | evpn-blacklist-recovery | bash | <code>FastCli -p 15 -c "clear bgp evpn host-flap"</code> | on-logging |
 | trigger-on-boot | bash | <code>echo "on-boot"</code> | on-boot |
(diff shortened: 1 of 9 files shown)


ClausHolbechArista (commented):
I think each option supports/requires different options after the keyword on eos. So I think we may have to rewrite this model to a dict.

gmuloc:
> I think each option supports/requires different options after the keyword on eos. So I think we may have to rewrite this model to a dict.

indeed - only `on-boot` seems to work like this. I only need on-boot for now so will remove the other and log an issue to revisit the model later.

Though it will be hard to model - you can configure an event-handler without a trigger in CLI (yeah that is useless probably but valid config) and then it would have to be a list for each option.

`on-counters` is actually a mode to get in..

```
event-handler trigger-on-boot
   trigger on-boot
   action bash echo "on-boot"
event-handler trigger-on-config
   trigger on-config disabled
   action bash echo "on-config"
event-handler trigger-on-counters
   action bash echo "on-counters"
   !
   trigger on-counters
event-handler trigger-on-intf
   trigger on-intf Ethernet1 ip
   action bash echo "on-intf"
event-handler trigger-on-maintenance
   trigger on-maintenance enter bgp BLAH after stage linkdown
   trigger on-config disabled
   action bash echo "on-maintenance"
event-handler trigger-vm-tracer
   trigger vm-tracer vm
   action bash echo "vm-tracer"
```

ClausHolbechArista (approved):

ankudinov:
LGTM

ankudinov (approved):

carlbuchmann (approved):
Reviewed by Petr and Claus

PR #3265, opened 2023-10-18
Feat(eos_cli_config_gen): Implement management-ssh client-alive

gmuloc:
## Change Summary

Add `client-alive` options for `management-ssh`

## Component(s) name

## Proposed changes

CLI output

```
ceos2(config-mgmt-ssh)#client-alive ?
  count-max  Number of keep-alive packets that can be sent without a response before the connection is assumed dead
  interval   Time period ( in seconds ) to send SSH keep-alive packets

ceos2(config-mgmt-ssh)#client-alive count-max ?
  <1-1000>  Set Value

ceos2(config-mgmt-ssh)#client-alive interval ?
  <1-1000>  Set Value
```

Adding the following schema:

```
      client_alive:
        type: dict
        keys:
          count_max:
            type: int
            convert_types:
              - str
            min: 1
            max: 1000
            description: Number of keep-alive packets that can be sent without a response before the connection is assumed dead.
          interval:
            type: int
            convert_types:
              - str
            min: 1
            max: 1000
            description: Time period (in seconds) to send SSH keep-alive packets.
```

## How to test

molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-ssh.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/management-ssh.md
@@ -77,6 +77,8 @@ management ssh
    idle-timeout 15
    connection limit 50
    connection per-host 10
+   client-alive interval 666
+   client-alive count-max 42
    no shutdown
    log-level debug
    !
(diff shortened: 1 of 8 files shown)


carlbuchmann (approved):
LGTM and tested settings on DUT

ClausHolbechArista (approved):

PR #3273, opened 2023-10-20
Feat(eos_designs): Set spanning-tree priority per VLAN

ClausHolbechArista:
## Change Summary

Set spanning-tree priority per VLAN

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Allow setting `spanning_tree_priority` under l2vlans and svis for `rapid-pvst` only.
- Update schema doc for node setting to hint to this setting.

## How to test

- Added molecule case for rapid-pvst in eos_designs including this new feature.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
@@ -417,7 +417,8 @@ class AvdStructuredConfigBase(AvdFacts, NtpMixin, SnmpServerMixin):
             if spanning_tree_mode == "mstp":
                 spanning_tree["mst_instances"] = [{"id": "0", "priority": priority}]
             elif spanning_tree_mode == "rapid-pvst":
-                spanning_tree["rapid_pvst_instances"] = [{"id": "1-4094", "priority": priority}]
+                pass
+                # Per vlan spanning-tree priorities are set under network-services.
             elif spanning_tree_mode == "rstp":
                 spanning_tree["rstp_priority"] = priority
 
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/spanning_tree.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/spanning_tree.py
@@ -1,0 +1,62 @@
+# Copyright (c) 2023 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from functools import cached_property
+
+from ansible_collections.arista.avd.plugins.filter.list_compress import list_compress
+from ansible_collections.arista.avd.plugins.plugin_utils.utils import get
+from ansible_collections.arista.avd.roles.eos_designs.python_modules.network_services.utils import UtilsMixin
+
+
+class SpanningTreeMixin(UtilsMixin):
+    """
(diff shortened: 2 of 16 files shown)


ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
```diff
@@ -405,11 +405,12 @@ def spanning_tree(self) -> dict | None:
...
         if spanning_tree_mode is not None:
             spanning_tree["mode"] = spanning_tree_mode
-            priority = get(self.shared_utils.switch_data_combined, "spanning_tree_priority", default="32768")
+            priority = get(self.shared_utils.switch_data_combined, "spanning_tree_priority", default=32768)
             if spanning_tree_mode == "mstp":
                 spanning_tree["mst_instances"] = [{"id": "0", "priority": priority}]
             elif spanning_tree_mode == "rapid-pvst":
-                spanning_tree["rapid_pvst_instances"] = [{"id": "1-4094", "priority": priority}]
+                pass
+                # Per vlan spanning-tree priorities are set under network-services.
```
ankudinov: In my opinion it would be better to set the default priority for all STP modes in a single place. Either here or in the mixin class.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/spanning_tree.py
```diff
@@ -0,0 +1,62 @@
...
+                for svi in vrf["svis"]:
+                    if (priority := get(svi, "spanning_tree_priority")) is None:
+                        continue
+                    vlan_stp_priorities.setdefault(priority, set()).add(svi["id"])
+                    non_default_vlans.add(svi["id"])
+            for l2vlan in tenant["l2vlans"]:
+                if (priority := get(l2vlan, "spanning_tree_priority")) is None:
+                    continue
+                vlan_stp_priorities.setdefault(priority, set()).add(l2vlan["id"])
+                non_default_vlans.add(l2vlan["id"])
```
ankudinov, line 46: Shouldn't we set the default here for all STP modes unconditionally, walk over tenants irrespective of STP mode and raise an exception if STP mode is not `rapid-pvst` and some priority different to the global setting is defined? The downside - the code always have to walk over tenants at some performance cost. But current logic will silently ignore all settings if the mode is MSTP.

ankudinov (commented):
Apart from the comments above, happy to LGTM if changes are not feasible at this stage.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/spanning_tree.py
ClausHolbechArista: The performance hit is significant for large deployments. The same deployments that would most likely use MST.
In general we ignore a lot of invalid combinations in AVD. This would be addressed best in the schema, so we could define "only_allowed_if" or something similar.
So for now, I don't think we should catch this.

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/avdstructuredconfig.py
ClausHolbechArista: Moving things that are not relying on network services to network services doesn't seem right to me. I understand that contributing things from multiple places is confusing, but the alternative would also be confusing.

jrecchia1029 (approved):
LGTM

ankudinov (approved):
LGTM

PR #3278, opened 2023-10-20
Feat(eos_cli_config_gen): Add additional-path for evpn address-family peer-groups

gmuloc:
## Change Summary

Add capability to configure additional-paths at the EVPN address family peer-group level
Introduces in Jinja2 template the rendering of `no neighbor <PG> additional-path send any` that _IS_ not supported in the other places where we have this (not so great) model.

> Follow up to this PR would be to refactor the model for 5.0 and to add the `no ..` syntax in other places.

## Component(s) name

## Proposed changes

Reusing the same model as in other places in BGP schema, validated on cEOS

## How to test

molecule

### Repository Checklist


ClausHolbechArista (approved):

gmuloc:
As discussed - this model for additional-paths MUST
1/ be revisited for 5.0
2/ be made a defs in schema to reuse it everywhere it is needed

This will ensure consistency in every places we use it

emilarista (approved):
Tested and reviewed, LGTM!

PR #3283, opened 2023-10-20
Feat(eos_designs): Add dns_settings to replace name_servers

ClausHolbechArista:
## Change Summary

Add dns_settings to replace name_servers

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Add new data model:
  ```yaml
    # DNS settings
    dns_settings:

      # DNS domain name like 'fabric.local'
      domain: <str>
      servers: # >=1 items; required

          # If not set, the VRF is automatically picked up from the global setting `default_mgmt_method`.
          # The value of `vrf` will be interpreted according to these rules:
          # - `use_mgmt_interface_vrf` will configure the DNS server under the VRF set with `mgmt_interface_vrf` and set the `mgmt_interface` as DNS lookup source-interface.
          #   An error will be raised if `mgmt_ip` or `ipv6_mgmt_ip` are not configured for the device.
          # - `use_inband_mgmt_vrf` will configure the DNS server under the VRF set with `inband_mgmt_vrf` and set the `inband_mgmt_interface` as DNS lookup source-interface.
          #   An error will be raised if inband management is not configured for the device.
          # - Any other string will be used directly as the VRF name. Remember to set the `dns_settings.vrfs[].source_interface` if needed.
        - vrf: <str>

          # IPv4 or IPv6 address for DNS server.
          ip_address: <str; required>

          # Priority value (lower is first).
          priority: <int; 0-4>
      vrfs:

          # VRF name.

          # Source interface to use for DNS lookups in this VRF.
          # If set for the VRFs defined by `mgmt_interface_vrf` or `inband_mgmt_vrf`, this setting will take precedence.
          source_interface: <str>

      # Automatically set source interface when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
      # Can be set to `false` to avoid changes when migrating from old `name_servers` model.
      set_source_interfaces: <bool; default=True>
  ```

- Deprecate `name_servers`.
- Deprecate `source_interfaces.domain_lookup`
- Refactor code to support both.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/dns_settings.schema.yml
```diff
@@ -0,0 +1,41 @@
...
+        items:
+          type: dict
+          keys:
+            vrf:
+              type: str
+              description: |-
+                VRF Name.
+                Can be used in combination with "use_mgmt_interface_vrf" and "use_inband_mgmt_vrf" to configure the DNS server under multiple VRFs.
+              convert_types:
+                - int
```
gmuloc: Why convert_types ?
Should it be a list ?

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class IpNameServersMixin(UtilsMixin):
+    """
+    Mixin Class used to generate structured config for one key.
+    Class should only be used as Mixin to a AvdStructuredConfig class
+    """
+
+    @cached_property
+    def ip_name_servers(self) -> dict | None:
```
gmuloc: It used to return a list and now it returns a dict and the docstring mentions a set ?

looks like it actually returns `list | None`

gmuloc: the function could produce duplicate if both `name_servers` and `dns_settings` are set

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
```diff
@@ -0,0 +1,77 @@
...
+            if not any([vrfs, use_mgmt_interface_vrf, use_inband_mgmt_vrf]):
+                # If no VRFs are defined (and we are not just ignoring missing mgmt config)
+                vrfs.add("default")
+
+            # Ensure default VRF is added first
+            if "default" in vrfs:
+                vrfs.remove("default")
+                # Add server without VRF field
+                ip_name_servers.append(server)
+
```
gmuloc: looks like `default` should be rendered:

```
ceos2(config)#no ip name-server vrf default 42.42.42.66
ceos2(config)#show run | grep 66 | grep name
ceos2(config)#ip name-server 42.42.42.66
ceos2(config)#show run | grep 66 | grep name
ip name-server vrf default 42.42.42.66
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AVD_LAB.yml
```diff
@@ -30,9 +30,12 @@ snmp_settings:
 mgmt_gateway: 192.168.200.5
 
 # dns servers.
-name_servers:
-  - 192.168.200.5
-  - 8.8.8.8
+dns_settings:
```
gmuloc, line 69: Can we test also the extra VRF and in particular default?
also probably need to test the `use_inband_interface_vrf` somewhere

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/host1/name_servers.yml
```diff
@@ -0,0 +1,4 @@
+# name_servers is deprecated in 4.5.0. To be removed in 5.0.0
+name_servers:
+  - 192.168.200.5
```
gmuloc, line 3: config is not generated should we not expect "some config change" for the negative tests as it was not tested before?

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/dns_settings.schema.yml
ClausHolbechArista: VRF names can be numeric. So we always convert from int.

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
ClausHolbechArista: Fixed the type hint.

I am ok with it producing duplicates in case of both settings. I prefer that over overcomplicating the logic here for the deprecated variable.

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
ClausHolbechArista: I will make that the behavior for the new data model, but keep the old as-is.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AVD_LAB.yml
ClausHolbechArista: Yes I forgot to add a unit test for all variations. I will do that.

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/inventory/host_vars/host1/name_servers.yml
ClausHolbechArista: Good catch. The scenario was not included in MOLECULE_SCENARIOS or CI. I will add it in this PR.

ClausHolbechArista: https://github.com/aristanetworks/ansible-avd/pull/3283/commits/af9822f9d0a9652eceacc29a80e52a74f823ad51

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AVD_LAB.yml
ClausHolbechArista: https://github.com/aristanetworks/ansible-avd/pull/3283/commits/af9822f9d0a9652eceacc29a80e52a74f823ad51

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
ClausHolbechArista: https://github.com/aristanetworks/ansible-avd/pull/3283/commits/af9822f9d0a9652eceacc29a80e52a74f823ad51

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
ClausHolbechArista: https://github.com/aristanetworks/ansible-avd/pull/3283/commits/af9822f9d0a9652eceacc29a80e52a74f823ad51

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-1.yml
```diff
@@ -0,0 +1,51 @@
...
+      # priority: 1
+      # vrf: default
+      use_mgmt_interface_vrf: false
+      use_inband_mgmt_vrf: false
+
+      # This one should not be in the output, since this host has no inband mgmt
+    - ip_address: 10.10.10.4
+      # priority: 1
+      # vrf: default
+      # use_mgmt_interface_vrf: false
```
gmuloc: should we remove the comments?

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/dns_settings.schema.yml
gmuloc: what about making it a list of extra vrfs ?

ClausHolbechArista: I doubt it will ever be used. And if someone really needs it, they can just add the server twice with different VRFs. This follows the structure of snmp.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-1.yml
ClausHolbechArista: I can. I kept them to make it clear that these are *not* set, but sure I can clean up.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/dns_settings.schema.yml
gmuloc: fair enough. Though if I add the server twice, I could end up with duplicate VRF then given that the set is applied only at the `server` level (not saying we guard against it in SNMP!):

```
    dns_settings:
      domain: <str>
      servers:
        - vrf: MGMT
           ip_address: 1.1.1.1
        - use_mgmt_interface_vrf: true
           ip_address: 1.1.1.1
```

but I guess we can live with it.

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/ip_name_server.py
```diff
@@ -0,0 +1,70 @@
...
+        ip_name_servers set based on "dns_settings" data-model or the deprecated "name_servers" input data-model.
+
+        Servers in the new data model may have management VRFs dynamically set.
+        """
+        dns_settings_servers = get(self._hostvars, "dns_settings.servers", default=[])
+        ip_name_servers = self.get_deprecated_name_servers()
+
+        if not dns_settings_servers:
+            return ip_name_servers or None
+
```
gmuloc: if we reach here and `ip_name_servers` is not None, we could print a warning to the user

ClausHolbechArista: No this is to catch the legacy data model alone.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/dns_settings.schema.yml
ClausHolbechArista: Ah I think the duplicate detection is much more important than having multiple vrfs on a single server. So maybe we should add primary_key on the list of servers using the ip_address. It would rule out adding two extra VRFs, but we always have structured config.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -76,3 +83,66 @@ def _build_source_interfaces(
             )
 
         return source_interfaces
+
+    def _get_vrf_and_source_interface(
```
ClausHolbechArista, line 87: Handle `ipv6_mgmt_ip` according to the server ip version.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/source-interfaces.yml
```diff
@@ -12,9 +12,9 @@ l2leaf:
 mgmt_interface_vrf: default
 
 source_interfaces:
-  domain_lookup:
-    mgmt_interface: true
-    inband_mgmt_interface: true
+  # domain_lookup:  # Deprecated so moved to deprecated vars scenario
```
MaheshGSLAB: Either we can remove the comments or if we are keeping it for reason of removing then may be add a TODO 6.0 so we can remove the comments once deprecated

ClausHolbechArista: removed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
ClausHolbechArista: since this function is generic for all protocols, it will very likely use names instead of IPs, so we cannot rely on IP version expect for dns. So skipping this for now. @gmuloc ?

ansible_collections/arista/avd/roles/eos_designs/docs/tables/management-settings.md
```diff
@@ -83,6 +93,41 @@
 === "YAML"
 
     ```yaml
+    # DNS settings
+    # For DNS source-interfaces see "source_interfaces.domain_lookup"
```
MaheshGSLAB: Not getting the reason of pointing out a deprecated key.

MaheshGSLAB (commented):
LGTM. just a minor comment

ansible_collections/arista/avd/roles/eos_designs/docs/tables/management-settings.md
ClausHolbechArista: Mistake from previous revisions of this PR. I will fix it. Thanks.

python-avd/pyavd/_eos_designs/schema/schema_fragments/dns_settings.schema.yml
```diff
@@ -0,0 +1,61 @@
...
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  dns_settings:
+    documentation_options:
+      table: management-settings
+    type: dict
+    description: |-
+      DNS settings
+      For DNS source-interfaces see "source_interfaces.domain_lookup"
```
ClausHolbechArista:
```suggestion
```

MaheshGSLAB (approved):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -300,10 +300,29 @@ def queue_monitor_length(self) -> None:
...
+
+        if self.inputs.dns_settings.domain:
+            self.structured_config.dns_domain = self.inputs.dns_settings.domain
+
+        vrfs = self.inputs.dns_settings.vrfs
+        for server in self.inputs.dns_settings.servers:
+            server_vrf, source_interface = self._get_vrf_and_source_interface(
+                vrf_input=server.vrf,
+                vrfs=vrfs,
+                set_source_interfaces=self.inputs.dns_settings.set_source_interfaces,
```
laxmikantchintakindi, line 318: I think `self.inputs.dns_settings.set_source_interfaces` can directly be used in _get_vrf_and_source_interface, so it would not required to pass an argument.

MaheshGSLAB: Function `_get_vrf_and_source_interface` is going to be use for other settings as well, like logging settings. So it may require or not depend on type of settings. For logging setting its requires to set as static True

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -300,10 +300,29 @@ def queue_monitor_length(self) -> None:
 
     @structured_config_contributor
     def ip_name_servers(self) -> None:
-        """ip_name_servers set based on name_servers data-model and mgmt_interface_vrf."""
+        """Set ip name servers using old name_servers model and new dns_settings model. Results will be combined."""
```
laxmikantchintakindi, line 303:
```
ERROR! [dns-settings-1]: The input data model 'name_servers' is deprecated and cannot be used in conjunction with the new data model '.dns_settings.servers'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.
ERROR! [dns-settings-1]: 'Validation Error: name_servers[0]': Invalid type 'dict'. Expected a 'str'.
fatal: [dns-settings-1 -> 127.0.0.1]: FAILED! => {"changed": false, "msg": "2 errors found during schema validation of input vars."}
```

laxmikantchintakindi: It is throwing error when using both models together.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -76,3 +83,66 @@ def _build_source_interfaces(
...
+        """
+        Helper function to interpret the VRF field for a management protocol.
+
+        If not set, the VRF is automatically picked up from the global setting `default_mgmt_method`.
+        The value of `vrf` will be interpreted according to these rules:
+        - `use_mgmt_interface_vrf` will return `(<mgmt_interface_vrf>, <vrfs[].source_interface or mgmt_interface>)`.
+          An error will be raised if `mgmt_ip` or `ipv6_mgmt_ip` are not configured for the device.
+        - `use_inband_mgmt_vrf` will return `(<inband_mgmt_vrf>, <vrfs[].source_interface or inband_mgmt_interface>)`.
+          An error will be raised if inband management is not configured for the device.
+        - Any other string will return `(<vrf_input>, None)`
```
gmuloc: any reason to always return None if there is an interface in vrfs?

gmuloc: actually this won't return None in that case so just need to update teh docstring

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: the first error is expected as we don't want to sue both model at the same time (deprecated and replacement). The second error is it not wrong input from you?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-3.yml
```diff
@@ -0,0 +1,45 @@
...
+  nodes:
+    - name: dns-settings-3
+      mgmt_ip: 192.168.0.2/24
+      inband_mgmt_ip: 192.168.1.2/24
+      inband_mgmt_vrf: INBAND_MGMT
+
+default_mgmt_method: oob
+
+dns_settings:
+  # domain: testfabric.local
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-3.yml
```diff
@@ -0,0 +1,45 @@
+# Testing with both inband and oob management, overriding source interfaces on both.
```
gmuloc, line 2:
```suggestion
---
# Testing with both inband and oob management, overriding source interfaces on both.
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-2.yml
```diff
@@ -0,0 +1,32 @@
+# Testing with inband mgmt.
```
gmuloc, line 2:
```suggestion
---
# Testing with inband mgmt.
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-2.yml
```diff
@@ -0,0 +1,32 @@
...
+
+type: l2leaf
+l2leaf:
+  defaults:
+  nodes:
+    - name: dns-settings-2
+      # mgmt_ip: 192.168.0.2/24
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/dns-settings-1.yml
```diff
@@ -0,0 +1,33 @@
+# Testing with oob management
```
gmuloc, line 2:
```suggestion
---
# Testing with oob management
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/INBAND_MGMT_TESTS.yml
```diff
@@ -43,6 +45,8 @@ spine:
   defaults:
     loopback_ipv4_pool: 10.0.254.0/24
     bgp_as: 64999
+    # Dummy IP to avoid errors being raised because of dns_settings for device with oob management.
+    mgmt_ip: 192.168.100.10/24
```
gmuloc: Should this be a TODO? this feels clunky - how would a user know?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/inband-mgmt-spine2-ztp.cfg
```diff
@@ -7,7 +7,6 @@ daemon TerminAttr
    no shutdown
 !
 dhcp server
-   dns server ipv4 1.1.1.1 8.8.8.8
```
gmuloc: TODO check this is still covered somewhere

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/dns-settings-2.cfg
```diff
@@ -0,0 +1,42 @@
...
+!
+service routing protocols model multi-agent
+!
+hostname dns-settings-2
+ip domain lookup vrf INBAND_MGMT source-interface Vlan4092
+ip name-server vrf default 10.10.10.1 priority 1
+ip name-server vrf INBAND_MGMT 10.10.10.1 priority 1
+ip name-server vrf INBAND_MGMT 10.10.10.2
+ip name-server vrf other_vrf 10.10.10.2
+ip name-server vrf INBAND_MGMT 10.10.10.3
```
gmuloc, line 17: is the sorting order correct?

gmuloc: hmmm not sure how it is supposed to work..
```
site1-wan2#show run se name-ser
ip name-server vrf INBAND_MGMT 10.10.10.2
ip name-server vrf INBAND_MGMT 10.10.10.3
ip name-server vrf other_vrf 10.10.10.2
ip name-server vrf INBAND_MGMT 10.10.10.1 priority 1
ip name-server vrf default 10.10.10.1 priority 1
```

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
```diff
@@ -25,6 +25,18 @@
 {%         endif %}
 {%     endfor %}
 {% endfor %}
+{% set mgmt_interface_vrf = hostvars[groups[fabric_group][0]].mgmt_interface_vrf | arista.avd.default("MGMT") %}
+{% set nameservers = hostvars[groups[fabric_group][0]].name_servers | arista.avd.default([]) + (
```
gmuloc, line 30: hehe maybe this one deserves a comment to explain the 3 blocks. but I agree

gmuloc (changes requested):

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
ClausHolbechArista: I would love to get rid of all this....

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/dns-settings-2.cfg
ClausHolbechArista: I am not doing eos_cli_config_gen here.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/INBAND_MGMT_TESTS.yml
ClausHolbechArista: If they use dns_settings it will tell them.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
ClausHolbechArista:
```suggestion
        - Any other string will return `(<vrf_input>, <vrfs[].source_interface or None)`
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: > ERROR! [dns-settings-1]: The input data model 'name_servers' is deprecated and cannot be used in conjunction with the new data model '.dns_settings.servers'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.

This is expected. We do this for all deprecations since 5.0.

> ERROR! [dns-settings-1]: 'Validation Error: name_servers[0]': Invalid type 'dict'. Expected a 'str'.

Must be an error in your inputs.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/dns-settings-2.cfg
ClausHolbechArista: I guess no priority means priority 0 and prioirity is the first sort order. VRF is the second and ip third maybe. or maybe it is just random

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
gmuloc: I know :) so would I

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/dns-settings-2.cfg
gmuloc: yeha I was just curious - anyways - thats not for this PR :)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/INBAND_MGMT_TESTS.yml
gmuloc: well yeah but the fact that it is dummy is confusing. As a user I would be quite surprised to have to give a fake input to make things work.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
laxmikantchintakindi: Yes. My query is about the first error. Doc string says `the result is combined` but we cannot use both keys at the same time.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/INBAND_MGMT_TESTS.yml
ClausHolbechArista: moving to individual devices.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: the code here combines yes, which will not really matter, as the schema will prevent use of both.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/inband-mgmt-spine2-ztp.cfg
ClausHolbechArista: I missed this. I will fix

ansible_collections/arista/avd/roles/dhcp_provisioner/templates/ztp_configuration.j2
ClausHolbechArista: added a comment

gmuloc (approved):

PR #3300, opened 2023-10-26
Feat(eos_designs): Custom platform_settings and node_type_keys

### LINKED ISSUE #3152: Feat(eos_designs): Add `custom_foo` keys for large default
### Enhancement summary

The `platform_settings` and `node_typekeys` keys have large default (as in many lines of yaml) and today the only way to overwrite part of them while keeping the rest of the default is to fully copy the default and modify accordingly

This issue is for adding two keys `custom_platform_settings` and `custom_node_type_keys` to be prepended to the AVD defaults when defined to avoid copying over everything.

prepended because this way if you overwrite some existing default, because we match in order, it would take precedence.

NOTE: if both `custom_platform_settings` and `platform_settings` are defined by the user in the playbook, then we would prepend `custom_` to the user defined `platform_settings` though this is a quite convoluted way of working and only overriding `platform_settings` should be sufficient

NOTE: could be implemented also for connected_endpoints_keys and network_services_keys but it is less necessary because the defaults are very small

eos_designs

### Use case example

cf summary

cf summary

copying over large defaults

### Additional context

_No response_

### Contributing Guide

jonxstill:
## Change Summary

Allow use of `custom_platform_settings` and `custom_node_type_keys` to allow replacement of already-defined keys/platforms, or addition of new keys/platforms while retaining the default settings.

## Related Issue(s)

Fixes #3152 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Allows you to:
* Specify new platform settings and node_type_keys in addition to the AVD-defined defaults without having to copy/paste these into your inventory.
* Redefine AVD-defined default node_type_keys and platform_settings without having to redefine every single one.

No changes to the data model - just use:
* `custom_platform_settings` instead of `platform_settings`.
* `custom_node_type_keys` instead of `node_type_keys`.

## How to test
Molecule tests updated to cover these new options.  Problems encountered included AVD complaining about duplicate `node_type_keys` - this was fixed through use of `replace_or_append_item()`.

This can be tested in your own repo by defining a `custom_platform_settings` and/or `custom_node_type_keys` at a high level in your repo's hierarchy.  Defaults can be copied from the AVD docs and amended as required.

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
@@ -175,7 +175,7 @@ AVD provides the capability to customize your node types, supporting a variety o
 !!! note
     The default values will be overridden if this key is defined.
     If you need to change all the existing `node_type_keys`, it is recommended to copy the defaults and modify them.
-    If you need to add custom `node_type_keys`, create them under `custom_node_type_keys` - if named identically to default `node_type_keys` entries, custom entries will replace the equivalent default entry.
+    If you need to add custom `node_type_keys`, create them under `custom_node_type_keys`; if named identically to default `node_type_keys` entries, custom entries will replace the equivalent default entry.
 
 ??? example "Default value for design `l3ls-evpn`"
 
--- a/ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
+++ b/ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
@@ -173,7 +173,9 @@ To customize or create new node types, please refer to [node type customization]
 AVD provides the capability to customize your node types, supporting a variety of designs.
 
 !!! note
-    The default values will be overridden if defining this key, so it is recommended to copy the defaults and modify them.
+    The default values will be overridden if this key is defined.
+    If you need to change all the existing `node_type_keys`, it is recommended to copy the defaults and modify them.
+    If you need to add custom `node_type_keys`, create them under `custom_node_type_keys` - if named identically to default `node_type_keys` entries, custom entries will replace the equivalent default entry.
 
 ??? example "Default value for design `l3ls-evpn`"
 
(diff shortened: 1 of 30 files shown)


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_node_type_keys.schema.yml
```diff
@@ -0,0 +1,227 @@
+# Copyright (c) 2023 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../../plugins/plugin_utils/schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+$defs:
```
ClausHolbechArista: It would be ok to just keep the old schema as it was, and ref to that instead of the def. I think that would be simpler to read and maintain.

jonxstill: Can you point me to an example of where that's been done before?  I can then rework this.

ClausHolbechArista: If you revert the changes to the existing schema, you can just set `$ref: "eos_designs#/keys/node_type_keys"` under the custom one.

jonxstill: Thanks - have fixed in latest commit.

jonxstill:
Leaving as draft until #3078 is merged (first-class sflow support) as this PR is causing pyavd tests to fail currently.

ClausHolbechArista (approved):

ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.md
```diff
@@ -173,7 +173,9 @@ To customize or create new node types, please refer to [node type customization]
 AVD provides the capability to customize your node types, supporting a variety of designs.
 
 !!! note
-    The default values will be overridden if defining this key, so it is recommended to copy the defaults and modify them.
+    The default values will be overridden if this key is defined.
+    If you need to change all the existing `node_type_keys`, it is recommended to copy the defaults and modify them.
+    If you need to add custom `node_type_keys`, create them under `custom_node_type_keys` - if named identically to default `node_type_keys` entries, custom entries will replace the equivalent default entry.
```
carlbuchmann:
```suggestion
    If you need to add custom `node_type_keys`, create them under `custom_node_type_keys`; if named identically to default `node_type_keys` entries, custom entries will replace the equivalent default entry.
```

carlbuchmann (approved):
LGTM

PR #3301, opened 2023-10-27
Feat(eos_designs): Add cv_settings to replace cvp_* and terminattr_*

ClausHolbechArista:
## Change Summary

Add cv_settings to replace cvp_* and terminattr_*

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Settings for configuring CloudVision telemetry streaming and provisioning.

`cv_settings` replaces the following deprecated keys, which are ignored when `cv_settings` is set:
- `cvp_ingestauth_key`
- `cvp_instance_ip`
- `cvp_instance_ips`
- `cvp_token_file`
- `terminattr_disable_aaa`
- `terminattr_ingestexclude`
- `terminattr_ingestgrpcurl_port`
- `terminattr_smashexcludes`

New data model:
```yaml
    # Settings for CloudVision telemetry streaming and provisioning.
    cv_settings:

      # State streaming to CloudVision-as-a-Service.
      cvaas:

        # Enable streaming to CVaaS.
        # When enabled it will stream to 'apiserver.arista.io:443' using the VRF obtained from `default_mgmt_method` unless overridden under `clusters`.
        enabled: <bool; required>
        clusters: # default=[{'name': 'cvaas'}]

            # Short name for the CVaaS cluster. Required here, but only used when configuring multiple clusters.

            # Optionally set the region to stream to.
            # The "auto" region will use 'apiserver.arista.io:443' which will redirect to the correct region based on the device's serial number.
            region: <str; "auto" | "us-central1-a" | "us-central1-b" | "us-central1-c" | "apnortheast-1" | "euwest-2" | "ausoutheast-1" | "na-northeast1-b" | "uk-1" | "india-1" | "staging" | "dev" | "play"; default="auto">

            # The VRF used to connect to CloudVision.
            # The value will be interpreted according to these rules:
            # - `use_mgmt_interface_vrf` will configure the VRF set with `mgmt_interface_vrf` and configure the `mgmt_interface` as the source interface.
            #   An error will be raised if `mgmt_ip` or `ipv6_mgmt_ip` are not configured for the device.
            # - `use_inband_mgmt_vrf` will configure the VRF set with `inband_mgmt_vrf` and configure the `inband_mgmt_interface` as the source interface.
            #   An error will be raised if inband management is not configured for the device.
            # - `use_default_mgmt_method_vrf` will configure the VRF and source-interface for one of the two options above depending on the value of `default_mgmt_method`.
            # - Any other string will be used directly as the VRF name.
            vrf: <str; default="use_default_mgmt_method_vrf">

            # Path to the onboarding token used for certificate based authentication.
            # The path is on the EOS device and the token file must be copied to the device first.
            token_file: <str; default="/tmp/cv-onboarding-token">

            # Source-interface used to connect to CloudVision.
            # If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
            source_interface: <str>

      # On-premise CloudVision clusters.
      onprem_clusters:

          # Short name for the cluster. Required here, but only used when configuring multiple clusters.

          # CloudVision servers that makes up one cluster.
          servers: # 1-3 items; required

              # Server FQDN or IP address.
              port: <int; default=9910>

          # The VRF used to connect to CloudVision.
          # The value will be interpreted according to these rules:
          # - `use_mgmt_interface_vrf` will configure the VRF set with `mgmt_interface_vrf` and configure the `mgmt_interface` as the source interface.
          #   An error will be raised if `mgmt_ip` or `ipv6_mgmt_ip` are not configured for the device.
          # - `use_inband_mgmt_vrf` will configure the VRF set with `inband_mgmt_vrf` and configure the `inband_mgmt_interface` as the source interface.
          #   An error will be raised if inband management is not configured for the device.
          # - `use_default_mgmt_method_vrf` will configure the VRF and source-interface for one of the two options above depending on the value of `default_mgmt_method`.
          # - Any other string will be used directly as the VRF name.
          vrf: <str; default="use_default_mgmt_method_vrf">

          # Path to the onboarding token used for certificate based authentication.
          # The path is on the EOS device and the token file must be copied to the device first.
          token_file: <str; default="/tmp/token">

          # Source-interface used to connect to CloudVision.
          # If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
          source_interface: <str>

      terminattr:

        # Exclude paths from the shared memory table.
        smashexcludes: <str; default="ale,flexCounter,hardware,kni,pulse,strata">

        # Disable AAA authorization and accounting.
        # When setting this flag, all commands pushed from CloudVision are applied directly to the CLI without authorization
        disable_aaa: <bool; default=False>

        # Exclude paths from Sysdb on the ingest side.
        ingestexclude: <str>

      # Automatically set source interface when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
      # Can be set to `false` to avoid changes when migrating from old `cv_instances` model.
      set_source_interfaces: <bool; default=True>

```

## How to test

- Updated all existing molecule scenarios and examples to use the new data model
- Added the deprecated vars to `eos_designs_deprecated_vars`
- Added new unit test for `cv_settings` utilizing all options.

### User Checklist

- N/A

### Repository Checklist


jrecchia1029:
I think we want to add `source_interface` to the data model to be even more deterministic than `vrf` in inband scenarios.

So data model would be:
```
cv_settings:
  cvaas_clusters:
    - name: <str>
      server: <str>
      vrf: <str>
      source_interface: <str>
      token_file: <str>
  onprem_clusters:
    - name: <str>
      servers:
        - name: <str>
          port: <int>
      vrf: <str>
      source_interface: <str>
      key: <str>
      token_file: <str>
  terminattr_ingestexclude: <str>
  terminattr_smashexcludes: <str>
  terminattr_disable_aaa: <bool>
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/base/daemon_terminattr.py
```diff
@@ -0,0 +1,153 @@
...
+        if vrf == "use_mgmt_interface_vrf":
+            has_mgmt_ip = (self.shared_utils.mgmt_ip is not None) or (self.shared_utils.ipv6_mgmt_ip is not None)
+            if not has_mgmt_ip:
+                raise AristaAvdError(f"'{org_var}' is set to 'use_mgmt_interface_vrf' but this node is missing an 'mgmt_ip'")
+
+            return self.shared_utils.mgmt_interface_vrf
+
+        if vrf == "use_inband_mgmt_vrf":
+            if self.shared_utils.inband_mgmt_interface is None:
+                raise AristaAvdError("'ntp_settings.server_vrf' is set to 'use_inband_mgmt_vrf' but this node is missing configuration for inband management")
```
jrecchia1029: I believe `ntp_settings.server_vrf` should be `{org_var}`

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
```diff
@@ -0,0 +1,135 @@
+# Copyright (c) 2023-2025 Arista Networks, Inc.
```
ankudinov, line 1: I'd also add somewhere in the docs, that `daemon_terminattr` is not supposed to be used directly any more. This is the case for all my inventories. =) I was never using any of the keys we are deprecating.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,134 @@
...
+        type: list
+        description: CloudVision-as-a-Service.
+        primary_key: name
+        items:
+          type: dict
+          keys:
+            name:
+              type: str
+              description: |-
+                Short name for the CVaaS cluster. Used when configuring multiple clusters.
```
ankudinov: The description "Used when configuring multiple clusters" is a bit misleading, as it creates impression that this key is optional for single CVaaS/on-prem instance. However it's required and used as primary key. I'd also stress the fact that it must be unique. (If it's not unique - the error is raised, but adding that to the description is nice to have)
On a side note, I'd also consider an alternative way of implementing it:
1) I'd expect `fqdn` to be always unique and in that case it's feasible to use it as a primary key. Right now we can specify same fqdn in the same VRF multiple times by using unique names. If we can imagine a real case with the same fqdn in different VRFs (I can't) - I'd keep name as primary key, but it would be great to check that vrf/fqdn combination is unique.
2) Can we make name optional for single cluster cases? It's not too hard to add it, but single cluster will be the majority and less input is good.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,134 @@
...
+                If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
+      terminattr_ingestexclude:
+        type: str
+        description: Exclude paths from Sysdb on the ingest side.
+        default: "/Sysdb/cell/1/agent,/Sysdb/cell/2/agent"
+      terminattr_smashexcludes:
+        type: str
+        description: Exclude paths from the shared memory table.
+        default: "ale,flexCounter,hardware,kni,pulse,strata"
+      terminattr_disable_aaa:
```
ankudinov: Honestly, I'd make it a dict and replicate terminattr data model inside. But current way also works for me.

```yaml
cv_settings:
  daemon_terminattr:
    smashexcludes:
    ingestexclude:
    disable_aaa:
```

The only reason is to keep the original data model as it's the same thing in essence.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,134 @@
...
+          keys:
+            name:
+              type: str
+              description: |-
+                Short name for the CVaaS cluster. Used when configuring multiple clusters.
+              pattern: "^[a-zA-Z1-9-_]+$"
+            fqdn:
+              type: str
+              required: true
+              pattern: "^[a-z1-9\\.-]+\\.arista\\.io$"
```
ankudinov: daemon_terminattr > cvaddrs was allowing patterns like: `apiserver.cv-staging.corp.arista.io:443`. TBH, I was never using anything except 443, but should be keep an option to specify the port just in case? I'm OK to skip this though as worst case we can always use daemon_terminattr for corner cases.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,134 @@
...
+        primary_key: name
+        items:
+          type: dict
+          keys:
+            name:
+              type: str
+              description: |-
+                Short name for the CVaaS cluster. Used when configuring multiple clusters.
+              pattern: "^[a-zA-Z1-9-_]+$"
+            fqdn:
```
ankudinov: daemon_terminattr also had an option to stream to multiple nodes using the same token, which we are loosing in this case. Not sure it's critical, but can be helpful in case of DNS setup challenges.

ankudinov (changes requested):
Tested on Codespaces and CVaaS with the following settings:

```yaml
cv_settings:
  terminattr_disable_aaa: true
  cvaas:
    - name: test
      vrf: MGMT
      fqdn: "apiserver.cv-staging.corp.arista.io"
      token_file: "/mnt/flash/cv-onboarding-token"
  terminattr_ingestexclude: "/Sysdb/cell/1/agent,/Sysdb/cell/2/agent"
  terminattr_smashexcludes: "ale,flexCounter,hardware,kni,pulse,strata"
```

Also tested different excludes. Not tested the on-prem case.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ankudinov: As discussed internally - disregard the comment about unique vrf/fqdn.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: cvaas only supports 443. We can add a port knob if it ever becomes relevant.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: Right, you cannot enter more than one url per cluster, but is that really something people do? I would like to avoid adding a list level if possible.

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
ClausHolbechArista: We are not changing anything about that. Using cli_gen models will work as long as you don't try to _also_ set eos_designs inputs. If you wish to combine you need the `custom_structured_configuration_daemon_terminattr` variant.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: People can also just override with structured config

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ankudinov: Agreed. We can always use daemon_terminattr for that.

ankudinov (approved):
LGTM

dgonzalez-arista:
Summary of the testing:

Config on-prem:
```
cv_settings:
  onprem_clusters:
    - name: act
      servers:
        - name: 192.168.0.5
```
Renders:
```
-cvauth=token,/tmp/token
-cvaddr=192.168.0.5:9910

L2LS with OOB: -cvvrf=MGMT -cvsourceintf=Management1
L2LS Inband: -cvvrf=default  -cvsourceintf=Vlan4092
L2LS Inband non-default VRF: -cvvrf=InbandMGMT -cvsourceintf=Vlan4092
L3LS with OOB: -cvvrf=MGMT -cvsourceintf=Management1
L3LS with Inband: -cvvrf=default -cvsourceintf=Loopback0
```

Config cvaas:
```
cv_settings:
  cvaas:
    enabled: true
    clusters:
      - name: cvaas
        region: auto
  terminattr_disable_aaa: true
```
Renders:
```
-cvaddr=apiserver.arista.io:443
-cvauth=token-secure,/tmp/cv-onboarding-token
-disableaaa

L2LS with OOB: -cvvrf=MGMT -cvsourceintf=Management1
L2LS Inband: -cvvrf=default -cvsourceintf=Vlan4092
L2LS Inband non-default VRF: -cvvrf=InbandMGMT -cvsourceintf=Vlan4092
L3LS with OOB: -cvvrf=MGMT -cvsourceintf=Management1
L3LS with Inband: -cvvrf=default -cvsourceintf=Loopback0
```

LGTM!

ansible_collections/arista/avd/examples/cv-pathfinder/group_vars/WAN/management.yml
```diff
@@ -8,10 +8,12 @@ ntp_settings:
     - name: 0.pool.ntp.org
 
 # CloudVision/TerminAttr
-# **edit the cvp_instance_ips to match your CVaaS region**
```
gmuloc, line 11: you should probably keep this in the example as it is then rendered in the doc: https://avd.arista.com/5.4/ansible_collections/arista/avd/examples/cv-pathfinder/index.html#management

python-avd/pyavd/_eos_designs/schema/schema_fragments/cvp_token_file.schema.yml
```diff
@@ -14,3 +14,7 @@ keys:
       cvp_token_file is the path to the token file on the switch.
       If not set the default locations for on-premise or CVaaS will be used.
       See cvp_ingestauth_key for details.
+    deprecation:
+      warning: true
+      new_key: cv_settings.onprem_clusters[].token_file or cv_settings.cvaas.token_file
```
alexeygorbunov:
```suggestion
      new_key: cv_settings.onprem_clusters[].token_file or cv_settings.cvaas.clusters[].token_file
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+              default: "/tmp/token"
+            source_interface:
+              type: str
+              description: |-
+                Source-interface used to connect to CloudVision.
+                If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
+      terminattr_ingestexclude:
+        type: str
+        description: Exclude paths from Sysdb on the ingest side.
+        default: "/Sysdb/cell/1/agent,/Sysdb/cell/2/agent"
```
alexeygorbunov: CV/EoS documentation does not mention this key in default/suggested config. CV's ZTP and TA Studio logic do not render this key with it's default values as well (since late 2022). AVD should as well not render these default values.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
```diff
@@ -0,0 +1,45 @@
...
...vopt cvaas_inband_1.vrf=INBAND_MGMT_VRF -cvopt cvaas_inband_1.sourceintf=Vlan4092 -cvopt cvaas_inband_2.addr=apiserver.cv-prod-apnortheast-1.arista.io:443 -cvopt cvaas_inband_2.auth=token-secure,/tmp/anothercvaastoken -cvopt cvaas_inband_2.vrf=INBAND_MGMT_VRF -cvopt cvaas_inband_2.sourceintf=Loopback123 -cvopt cvaas_mgmt_intf.addr=apiserver.arista.io:443 -cvopt cvaas_mgmt_intf.auth=token-secure,/tmp/cv-onboarding-token -cvopt cvaas_mgmt_intf.vrf=INBAND_MGMT_VRF -cvopt cvaas_mgmt_intf.sourceintf=Loopback0 -cvopt onprem-oob.addr=onprem_oob_server_1:9911,onprem_oob_server_2:9910,onprem_oob_server_3:9910 -cvopt onprem-oob.auth=token,/tmp/token -cvopt onprem-oob.vrf=MGMT -cvopt onprem-oob.sourceintf=Management1 -cvopt onprem_inband.addr=onprem_inband_server_1:9910 -cvopt onprem_inband.auth=token,/tmp/anothertoken -cvopt onprem_inband.vrf=INBAND_MGMT_VRF -cvopt onprem_inband.sourceintf=Management1/1 -cvopt onprem_inband_two.addr=onprem_inband_server_2:9910 -cvopt onprem_inband_two.auth=token,/tmp/token -cvopt onprem_inband_two.vrf=INBAND_MGMT_VRF -cvopt onprem_inband_two.sourceintf=Vlan4092 -disableaaa -smashexcludes=this_is_smash_excluded -ingestexclude=this_is_ingest_excluded -taillogs
```
alexeygorbunov, line 6: Order of TA keys/settings rendered by AVD/cli_config_gen seem to not match the order of the default config in CV documentation and order of rendering keys/settings by CV's `Streaming Telemetry Agent` and ZTP logic.
This may create additional diff (sometimes just false-positive as keys are simply reordered without being changed) between DC and RC when user switches between different config approaches (ZTP/AVD/Studio). It would be helpful to use the same rendering order across all configuration approaches to make sure user can get to the same final result by using any config option.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+                    - us-central1-c
+                    - apnortheast-1
+                    - euwest-2
+                    - ausoutheast-1
+                    - na-northeast1-b
+                    - uk-1
+                    - india-1
+                    - staging
+                    - dev
+                    - play
```
alexeygorbunov, line 60: Should we put a note here to the customers that these clusters are for internal Arista purposes only and can not be used by the customers?

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_servers.py
```diff
@@ -69,19 +70,34 @@ def _update_subnets(self: AvdStructuredConfigUnderlayProtocol, dhcp_server: EosC
...
                     dhcp_server.subnets.append(subnet_item)
 
+    def _get_cvp_server_for_dhcp(self: AvdStructuredConfigUnderlayProtocol) -> str | None:
+        """Return the first CVP server using either new or old data models."""
+        if self.inputs.cv_settings.cvaas.enabled:
+            region = next(iter(self.inputs.cv_settings.cvaas.clusters)).region
+            return CV_REGION_TO_SERVER_MAP[region]
+
+        if self.inputs.cv_settings.onprem_clusters:
+            return next(iter(self.inputs.cv_settings.onprem_clusters)).name
```
alexeygorbunov: Name of the on-prem cluster here may not be DNS-resolvable, routable or responding to HTTP(s). Option 67 should be instead pointing to the FQDN/IP of the first node of the cluster
```suggestion
            return next(iter(next(iter(self.inputs.cv_settings.onprem_clusters)).servers)).name
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+                  An error will be raised if inband management is not configured for the device.
+                - `use_default_mgmt_method_vrf` will configure the VRF and source-interface for one of the two options above depending on the value of `default_mgmt_method`.
+                - Any other string will be used directly as the VRF name.
+            token_file:
+              type: str
+              description: |-
+                Path to the onboarding token used for certificate based authentication.
+                The path is on the EOS device and the token file must be copied to the device first.
+              default: "/tmp/token"
+            source_interface:
```
alexeygorbunov, line 138: TA binary has quite many configuration parameters. If we eventually need to add them here - we'll be duplicating them for both cvaas and on-prem. Is there a way to define them in one place and reuse?

There is also another approach (used by TA studio) which treats CVaaS just as a single-node cluster with port 443 instead of 9910. This would simplify schema a bit and would resolve any need for key duplication

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+            default:
+              - name: cvaas
+            items:
+              type: dict
+              keys:
+                name:
+                  type: str
+                  description: |-
+                    Short name for the CVaaS cluster. Required here, but only used when configuring multiple clusters.
+                  pattern: "^[a-zA-Z1-9-_]+$"
```
alexeygorbunov, line 39: Is this patterns enforced anywhere on TA side? It seems like TA is accepting any non-empty string and i would assume some customers would want to name their clusters using FQDNs (containing dots) as this would be matching on-prem clusters HTTPS LB VIPs

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
```diff
@@ -0,0 +1,142 @@
...
+            self._legacy_daemon_terminattr()
+            return
+
+        cv_settings = self.inputs.cv_settings
+
+        self.structured_config.daemon_terminattr._update(
+            ingestexclude=cv_settings.terminattr_ingestexclude,
+            smashexcludes=cv_settings.terminattr_smashexcludes,
+            disable_aaa=cv_settings.terminattr_disable_aaa,
+        )
```
alexeygorbunov:
```yaml
cv_settings:
  cvaas:
    enabled: false
```

produces TA config without any targeted clusters. Maybe we should not set `self.structured_config.daemon_terminattr` unless there is at least one cvaas or on-prem cluster present in the input?

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+                  type: str
+                  description: |-
+                    Path to the onboarding token used for certificate based authentication.
+                    The path is on the EOS device and the token file must be copied to the device first.
+                  default: "/tmp/cv-onboarding-token"
+                source_interface:
+                  type: str
+                  description: |-
+                    Source-interface used to connect to CloudVision.
+                    If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
```
alexeygorbunov:
```suggestion
                    If not set, the source interface may be set automatically when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+      onprem_clusters:
+        type: list
+        description: On-premise CloudVision clusters.
+        primary_key: name
+        unique_keys:
+          - servers.name
+        items:
+          type: dict
+          description: |-
+            Unless `key` is set, TerminAttr will be configured with certificate based authentication.
```
alexeygorbunov: Which `key`?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
alexeygorbunov, line 6: `cvass.clusters.token_file` and `onprem_clusters.servers.token_file` need to all be unique keys. Config line above (generated with multiple clusters all falling back to the default token_file) produce TA config where multiple "clusters/tenants" are pointing to the same token file (which is supposed to contain onboarding token).
This config will be OK if such TA config is applied to the switch where onboarding tokens were already (somehow) exchanged to the client certs from each tenant (those client certs will be saved into unique paths on the /persist, based on the cluster name).
But if such TA config is applied to the devices which is supposed to go through the onboarding (exchange of multiple onboarding tokens for the client certs) - this will not work as onboarding tokens must be unique per CVaaS tenent / onprem_cluster (as tenant/cluster will reject onboarding JWT unless it's issues/signed by the same tenant/cluster)

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
gmuloc: nitpicky but:
```suggestion
                    If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf`, `use_inband_mgmt_vrf` or `use_default_mgmt_method_vrf`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+              type: str
+              description: |-
+                Path to the onboarding token used for certificate based authentication.
+                The path is on the EOS device and the token file must be copied to the device first.
+              default: "/tmp/token"
+            source_interface:
+              type: str
+              description: |-
+                Source-interface used to connect to CloudVision.
+                If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
```
gmuloc: same as above:
```suggestion
                If not set, the source interface may be set automatically set when VRF is set to `use_mgmt_interface_vrf`, `use_inband_mgmt_vrf` or `use_default_mgmt_method_vrf`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+      terminattr_disable_aaa:
+        type: bool
+        description: |-
+          Disable AAA authorization and accounting.
+          When setting this flag, all commands pushed from CloudVision are applied directly to the CLI without authorization
+        default: false
+      set_source_interfaces:
+        type: bool
+        description: |-
+           Automatically set source interface when VRF is set to `use_mgmt_interface_vrf` and `use_inband_mgmt_vrf`.
```
gmuloc:
```suggestion
           Automatically set source interface when VRF is set to `use_mgmt_interface_vrf`, `use_inband_mgmt_vrf` or `use_default_mgmt_method_vrf`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
```diff
@@ -0,0 +1,164 @@
...
+        default: "ale,flexCounter,hardware,kni,pulse,strata"
+      terminattr_disable_aaa:
+        type: bool
+        description: |-
+          Disable AAA authorization and accounting.
+          When setting this flag, all commands pushed from CloudVision are applied directly to the CLI without authorization
+        default: false
+      set_source_interfaces:
+        type: bool
+        description: |-
```
gmuloc, line 162:
```suggestion
        # TODO: AVD7.0 remove mention of `cv_instances` in description
        description: |-
```

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -121,20 +121,24 @@ def _get_vrf_and_source_interface(
...
+        if source_interface_override:
+            return source_interface_override
+
+        match vrf_input:
+            case None | "" | "use_default_mgmt_method_vrf":
+                return self.shared_utils.default_mgmt_protocol_interface
+            case "use_mgmt_interface_vrf":
+                return self.shared_utils.mgmt_interface
+            case "use_inband_mgmt_vrf":
+                return self.shared_utils.inband_mgmt_interface
```
gmuloc, line 139: is no tool complaining about missing an explicit return?

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -121,20 +121,24 @@ def _get_vrf_and_source_interface(
...
-
-        if vrf in vrfs and vrfs[vrf].source_interface:
-            source_interface = vrfs[vrf].source_interface
+            source_interface = self.get_source_interface(vrf_input, source_interface_override=vrfs[vrf].source_interface if vrf in vrfs else None)
 
         return (vrf, source_interface)
 
+    def get_source_interface(self: AvdStructuredConfigBaseProtocol, vrf_input: str | None, source_interface_override: str | None) -> str | None:
+        """Returns source interface for the given vrf, letting the given override take precedence."""
+        # source_interface may be overridden below if given in the 'vrfs'
```
gmuloc: I am not sure this comment makes sense anymore
```suggestion
        # if an override is provided return it
```

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Class should only be used as Mixin to a AvdStructuredConfig class.
+    """
+
+    @structured_config_contributor
+    def daemon_terminattr(self: AvdStructuredConfigBaseProtocol) -> None:
+        """
+        Configures daemon_terminattr settings based on cv_settings and calls _legacy_daemon_terminattr for the legacy cv_* and terminattr_* models.
+
+        The schema will enforce that we only use either new or old models.
```
gmuloc, line 29: The schema would block a conjunction of say 
```
cv_settings:
  cvaas:
    enabled: true
```
 and `terminattr_smashexcludes` 

given that `terminattr_smashexcludes` would complain only if the following was set

```
cv_settings:
  cvaas:
    enabled: true
  terminattr_smashexcludes: ...
```

so if I have

```
cv_settings:
  cvaas:
    enabled: true
terminattr_smashexcludes: BLAH
```

I end up with

```
+  smashexcludes: ale,flexCounter,hardware,kni,pulse,strata
```

in my structured_config which can be counter intuitive and hard to catch

gmuloc: also the error message looks like:

```
The input data model 'terminattr_smashexcludes' is deprecated and cannot be used in conjunction with the new data model '.cv_settings.terminattr_smashexcludes'.
```

the `.` in front of `cv_settigns` could be confusing..

maybe we could add something that if the replacement key starts with `/` (for root) then we don't inject the statring `.`?

gmuloc (changes requested):

ansible_collections/arista/avd/examples/cv-pathfinder/group_vars/WAN/management.yml
ClausHolbechArista: Probably better to not have the region set here at all then. Then we need to set it for our testing.

ansible_collections/arista/avd/examples/cv-pathfinder/group_vars/WAN/management.yml
```diff
@@ -8,10 +8,12 @@ ntp_settings:
...
-cvp_instance_ips: [www.cv-staging.corp.arista.io]
-terminattr_smashexcludes: "ale,flexCounter,hardware,kni,pulse,strata"
-terminattr_ingestexclude: "/Sysdb/cell/1/agent,/Sysdb/cell/2/agent"
+cv_settings:
+  cvaas:
+    enabled: true
+    clusters:
+      - name: primary
+        region: staging
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: agreed. I will remove it. Thanks!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
ClausHolbechArista: I am not touching eos_cli_config_gen in this PR, so cannot influence that.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: They will find out :) I'd rather not attract extra attention on these.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_servers.py
ClausHolbechArista: absolutely. Thank you for catching.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: We can use $ref if it becomes important. But note these settings are only added per cluster, where we have fewer available settings. The rest are added under the common settings with exclude, disable_aaa etc.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
ClausHolbechArista: They may be onboarded one-by-one where it _will_ work with the same token file. But that is probably too confusing. But since we have a default value, I cannot also do unique_keys. And I prefer the default value :)

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: I used the same as we have on the others, but sure.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
ClausHolbechArista: Once CV team has standardized on one config in all cases we can rework ours to match.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: legacy stuff (this PR started two years ago). Removed :)

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
ClausHolbechArista: yes a fix for another PR :)

ClausHolbechArista: I can change the new_key to cv_settings on all of them if that is better?

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
ClausHolbechArista: fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
ClausHolbechArista: no because the type checker knows we cover all cases. Aparently ruff does as well :)

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: i added a comment after all

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
ClausHolbechArista: I did check various docs, but nothing is clearly stated. Since dot is used as separator on the terminattr commands, I doubt it is accepted as name. Ex. `-cvaddr=10.1.0.1:1234 -cvauth=none -cvopt foo.addr=1.2.3.4:5678 -cvopt foo.auth=none`

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
ClausHolbechArista: I updated the new_key on all of them to just say `cv_settings`. We still need to fix the leading .

ClausHolbechArista: Fixed:
```
ERROR! [pf1]: The input data model 'terminattr_smashexcludes' is deprecated and cannot be used in conjunction with the new data model 'cv_settings'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.
ERROR! [pf2]: The input data model 'terminattr_smashexcludes' is deprecated and cannot be used in conjunction with the new data model 'cv_settings'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.
ERROR! [site1-wan1]: The input data model 'terminattr_smashexcludes' is deprecated and cannot be used in conjunction with the new data model 'cv_settings'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.
ERROR! [site1-wan2]: The input data model 'terminattr_smashexcludes' is deprecated and cannot be used in conjunction with the new data model 'cv_settings'. This usually happens when a data model has been updated and custom structured configuration still uses the old model. See the porting guide on https://avd.arista.com for details.
```

ansible_collections/arista/avd/examples/cv-pathfinder/group_vars/WAN/management.yml
gmuloc: @joelbreton2 for information as this would break your pipeline

python-avd/pyavd/_eos_designs/structured_config/base/daemon_terminattr.py
gmuloc: this: ` See the porting guide on https://avd.arista.com for details.` may be problematic I suppose because there is nothing in the porting guide for it but ... not sure we can do anything except removing the suggestion

ClausHolbechArista: yeah once we add it to the porting guide we also remove the keys, so this error would not show up.

ClausHolbechArista: removed it

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
ClausHolbechArista: Accepting current behavior to keep the default value.

python-avd/pyavd/_eos_designs/schema/schema_fragments/cv_settings.schema.yml
alexeygorbunov: true! tested with 4.33.0 (TA 1.35.0) and although CLI accepts dotted cluster names - TA later fails to parse arguments

```bash
invalid value "staging.cluster.addr=apiserver.cv-staging.corp.arista.io:443" for flag -cvopt: expected form is cluster.option=value, invalid option specified: cluster.addr
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-settings.cfg
alexeygorbunov: I guess this also may become a problem later when switch with such TA config needs to be re-onboarded to one of the CV/CVaaS cluster/tenant. Pasting new onboarding token to the file shared by multiple clusters in TA's config and restarting TA may lead to unexpected results

ClausHolbechArista: I think it will just lead to it not working for one of the clusters :)

ClausHolbechArista: this is a corner case of a corner case. I don't think we need to keep digging into this. Few people stream to multiple cluster. Fewer even stream to multiple of each type.

gmuloc (approved):

PR #3315, opened 2023-10-31
Fix(eos_designs): Tags not inherited from profile

carlbuchmann:
## Change Summary

Tags not inherited from svi_profiles in network_services

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

TBD

## How to test

Updated molecule test


carlbuchmann:
This is a new feature and will take time to implement

PR #3340, opened 2023-11-14
Doc: Improve search

ClausHolbechArista:
Doc: Improve search by excluding raw tables.

Test by comparing a search for some key on https://avd.sh and https://ansible-avd--3340.org.readthedocs.build/en/3340/


gmuloc (approved):

ansible_collections/arista/avd/docs/requirements.txt
```diff
@@ -1,5 +1,5 @@
 mkdocs
-mkdocs-material==9.0.15
+mkdocs-material<10
```
gmuloc, line 2: why the cap? (which is much better than the hardcoded version ;) )
Can we lift it in a next PR?

mkdocs.yml
```diff
@@ -95,9 +99,6 @@ markdown_extensions:
   - pymdownx.caret
   - pymdownx.critic
   - pymdownx.details
-  - pymdownx.emoji:
-      emoji_index: !!python/name:materialx.emoji.twemoji
-      emoji_generator: !!python/name:materialx.emoji.to_svg
```
gmuloc, line 100: not used anymore?

ClausHolbechArista: They are builtin now.

ankudinov:
Definitely search is improved significantly.
I still see some corner cases where results are slightly suboptimal. For example, typing `prefix` gives better results than typing `prefix_list`. And typing `prefix_list:` breaks the search. But it's a huge improvement over current state.
LGTM

ansible_collections/arista/avd/docs/requirements.txt
ClausHolbechArista: Who knows that they will break in the next version...

ClausHolbechArista: If you prefer I can remove the cap.

gmuloc: let's keep it - I thought you were capping because it was already broken in next version and was curious of the why but actually 10.0 is not out yet

gmuloc (approved):

Issue #3304, opened 2023-10-29
Refactor(eos_validate_state): ANTA Slow catalog creation when high amount of devices


carl-baillargeon:
`_get_eos_validate_state_vars` is being called twice per device when it could be run once for all devices.

`get_anta_results.py` time logs:
```bash
[WARNING]: <p03l07> Logging setup time: 0.002478 seconds
[WARNING]: <p03l07> AvdTestHardware render time: 0.046718 seconds
[WARNING]: <p03l07> AvdTestHardware merge catalog time: 0.000320 seconds
[WARNING]: <p03l07> AvdTestNTP render time: 0.000074 seconds
[WARNING]: <p03l07> AvdTestNTP merge catalog time: 0.000029 seconds
[WARNING]: <p03l07> AvdTestInterfacesState render time: 0.074167 seconds
[WARNING]: <p03l07> AvdTestInterfacesState merge catalog time: 0.000024 seconds
[WARNING]: <p03l07> AvdTestLLDPTopology render time: 0.104166 seconds
[WARNING]: <p03l07> AvdTestLLDPTopology merge catalog time: 0.000024 seconds
[WARNING]: <p03l07> AvdTestMLAG render time: 0.006816 seconds
[WARNING]: <p03l07> AvdTestMLAG merge catalog time: 0.000020 seconds
[WARNING]: <p03l07> AvdTestP2PIPReachability render time: 0.059278 seconds
[WARNING]: <p03l07> AvdTestP2PIPReachability merge catalog time: 0.000058 seconds
[WARNING]: <p03l07> AvdTestBGP render time: 0.032573 seconds
[WARNING]: <p03l07> AvdTestBGP merge catalog time: 0.000024 seconds
[WARNING]: <p03l07> AvdTestRoutingTable render time: 24.519279 seconds
[WARNING]: <p03l07> AvdTestRoutingTable merge catalog time: 0.000238 seconds
[WARNING]: <p03l07> AvdTestInbandReachability render time: 0.022567 seconds
[WARNING]: <p03l07> AvdTestInbandReachability merge catalog time: 0.000194 seconds
[WARNING]: <p03l07> AvdTestLoopback0Reachability render time: 22.413227 seconds
[WARNING]: <p03l07> AvdTestLoopback0Reachability merge catalog time: 0.000258 seconds
[WARNING]: <p03l07> Catalog setup time: 47.769864 seconds
[WARNING]: <p03l07> Save catalog time: 1.276969 seconds
[...]

PR #3352, opened 2023-11-15
Feat(eos_validate_state): ANTA New action plugin to generate the reports

carl-baillargeon:
## Change Summary

New action plugin to generate the reports from the ANTA test results.

## Related Issue(s)

#3304

## Component(s) name

`arista.avd.eos_validate_state` **ANTA Mode**

## Proposed changes
`eos_validate_state_reports.py` is the action plugin.

The current `eos_validate_state_runner` will now save the results into a JSON file per host in the `test_results_dir` directory. The report plugin will load all results from all play hosts, process them and gather statistics.

`ResultsManager` class is responsible of storing the results and statistics and will be consumed by `CSVReport` and `MDReport` classes to generate the CSV and MD reports.

`only_failed_tests` can still be used to only write the failed tests in the reports.

`validation_report_csv` and `validation_report_md` can be set to false to avoid report generation and only run the tests.

## How to test
```yaml
---
- name: Validate AVD
  hosts: GLOBAL
  tasks:
    - name: Validate states on EOS devices
      import_role:
        name: arista.avd.eos_validate_state
      vars:
        # Some of these are defaults but can be changed
        validation_role:
          only_failed_tests: False
          validation_report_csv: True
          validation_report_md: True
        # To enable ANTA
        use_anta: True
        save_catalog: True
        logging_level: WARNING
```

### User Checklist

- [X] Module documentation and general documentation
- [X] Add a debug option to save the raw ANTA results
- [X] Add an option to have a report per device? No

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_validate_state/intended/test_catalogs/dc1-leaf1a-catalog.yml
```diff
@@ -194,17 +194,29 @@ anta.tests.connectivity:
       description: Loopback0 Reachability
 anta.tests.hardware:
 - VerifyEnvironmentPower:
+    result_overwrite:
```
ClausHolbechArista: Not the most intuitive key. Is this already in anta?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,139 @@
...
+    """Generate a report using the specified report class.
+
+    Args:
+    ----
+        report_path: The path to the report file.
+        report_class: The class used to generate the report.
+        test_results: The ResultsManager object containing the test results.
+    """
+    with Path(report_path).open("w", encoding="UTF-8") as report_file:
+        report = report_class(report_file, test_results)
```
ClausHolbechArista: IMO to keep sane, classes should be using title case even when passed as arguments. @gmuloc

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -91,20 +103,60 @@ def run(self, tmp=None, task_vars=None):
...
+        if cprofile_file:
+            profiler.disable()
+            stats = pstats.Stats(profiler).sort_stats("cumtime")
+            stats.dump_stats(cprofile_file)
 
         return result
 
 
-def setup_module_logging(hostname: str, result: dict) -> None:
+def create_temp_file(hostname: str, anta_results: list[dict]) -> str:
```
ClausHolbechArista: Please refactor to use https://docs.python.org/3.9/library/tempfile.html

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
ClausHolbechArista: After reading more of the code, I think you could simplify this a great deal by just passing the stream object to the generate_report, and then instantiate the relevant class in the calling function.

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,139 @@
...
+        ansible_play_hosts_all = task_vars.get("ansible_play_hosts_all", [])
+        # List to track all results temp files for each host
+        temp_files = []
+
+        try:
+            # Initialize an empty ResultsManager that will be used to store results and statistics
+            test_results = ResultsManager(only_failed_tests=only_failed_tests)
+            for host in sorted(ansible_play_hosts_all):
+                # Hosts marked as not deployed do not have any results
+                if not get(hostvars[host], "is_deployed"):
```
ClausHolbechArista: Since you do not have a default to true, this will print out for any device where is_deployed is not in the structured config. That would be very likely for structured configs generated by a user.

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,139 @@
...
+                    display.warning(f"No test results for host {host} since it's marked as not deployed.")
+                    continue
+                # Getting the host results temp file saved by eos_validate_state_runner action plugin
+                temp_file = get(hostvars[host], "anta_results.results_temp_file")
```
ClausHolbechArista: Getting variables one-by-one from hostvars like this is expensive, since hostvars is an object that performs templating etc. It is cheaper to assign `host_hostvars = hostvars[host]` and `get` from that.

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Good catch

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Yes this function uses `NamedTemporaryFile` from tempfile library and returns the name of the file.

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Good to know thanks

ansible_collections/arista/avd/molecule/eos_validate_state/intended/test_catalogs/dc1-leaf1a-catalog.yml
carl-baillargeon: Yes

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+        yield from load(file)
+
+
+class ActionModule(ActionBase):
+    def _get_and_validate_bool_arg(self, argument: str, *, default_value: bool) -> bool:
+        """Validate if a task argument is a boolean, returns the value if the validation succeeds."""
+        result = self._task.args.get(argument, default_value)
+        if not isinstance(result, bool):
+            raise AnsibleActionFail(f"'{argument}' must be a boolean, got {result}.")
+        return result
```
gmuloc: Could this be moved to a more central location?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+            task_vars = {}
+
+        result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        # Profiling
+        cprofile_file = self._task.args.get("cprofile_file")
+        if cprofile_file:
+            profiler = cProfile.Profile()
+            profiler.enable()
```
gmuloc: would be cool if in the future we had some decorator on functions to do this for all our plugins :) - future idea!

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+        temp_files = []
+
+        try:
+            # Initialize an empty ResultsManager that will be used to store results and statistics
+            test_results = ResultsManager(only_failed_tests=only_failed_tests)
+            for host in sorted(ansible_play_hosts_all):
+                host_hostvars = hostvars[host]
+                # Hosts marked as not deployed do not have any results
+                if not get(host_hostvars, "is_deployed", default=True):
+                    display.warning(f"No test results for host {host} since it's marked as not deployed.")
```
gmuloc:
```suggestion
                    display.warning(f"No test results for host {host} since 'is_deployed' is False.")
```

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+                    continue
+                # Getting the host results temp file saved by eos_validate_state_runner action plugin
+                temp_file = get(host_hostvars, "anta_results.results_temp_file")
+                if not isinstance(temp_file, str) or not Path(temp_file).exists():
+                    display.warning(f"Results temporary file path is invalid or does not exist for host {host}.")
```
gmuloc:
```suggestion
                    display.warning(f"Results temporary file path {temp_file} is invalid or does not exist for host {host}.")
```

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+            csv_report_path = self._get_and_validate_path_arg("csv_report_path")
+        if validation_report_md:
+            md_report_path = self._get_and_validate_path_arg("md_report_path")
+
+        # This is not all the hostvars, but just the Ansible Hostvars Manager object where we can retrieve hostvars for each host on-demand
+        hostvars = task_vars["hostvars"]
+        # For now we support the existing behavior of eos_validate_state, all hosts from the play
+        ansible_play_hosts_all = task_vars.get("ansible_play_hosts_all", [])
+        # List to track all results temp files for each host
+        temp_files = []
```
gmuloc: do we have multuple temp files for a single host? I am not sure I read the comment correctly

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+                    display.warning(f"Results temporary file path is invalid or does not exist for host {host}.")
+                    continue
+                # Track the temp file
+                temp_files.append(temp_file)
+
+                try:
+                    # Process the host test results
+                    for test_result in _test_results_gen(temp_file):
+                        test_results.update_results(test_result)
+                except Exception as error:
```
gmuloc: can we be more specifi regarding the exception?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+                    display.warning(f"Error while processing the test results of host {host}: {error}.")
```
gmuloc: Should we mark the task as failed when this happens?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+                    csv_report.generate_report()
+            if validation_report_md:
+                with Path(md_report_path).open("w", encoding="UTF-8") as report_file:
+                    md_report = ValidateStateReport(report_file, test_results)
+                    md_report.generate_report()
+
+        except Exception as error:
+            raise AnsibleActionFail(f"Error during plugin execution: {error}") from error
+        finally:
+            # Cleanup for each host's temporary file
```
gmuloc:
```suggestion
            # Cleanup temporary files
```

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -25,6 +31,23 @@ def ignore_aliases(self, data):
 
 
 class ActionModule(ActionBase):
+    def _get_and_validate_arg(
```
gmuloc: could this not ne used in the other plugin and moved to some more generic place?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+
+class ActionModule(ActionBase):
+    def _get_and_validate_bool_arg(self, argument: str, *, default_value: bool) -> bool:
+        """Validate if a task argument is a boolean, returns the value if the validation succeeds."""
+        result = self._task.args.get(argument, default_value)
+        if not isinstance(result, bool):
+            raise AnsibleActionFail(f"'{argument}' must be a boolean, got {result}.")
+        return result
+
+    def _get_and_validate_path_arg(self, argument: str) -> str:
```
gmuloc: it is called dir_arg in the oyther plugin

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
```diff
@@ -0,0 +1,130 @@
...
+            test_results = ResultsManager(only_failed_tests=only_failed_tests)
+            for host in sorted(ansible_play_hosts_all):
+                host_hostvars = hostvars[host]
+                # Hosts marked as not deployed do not have any results
+                if not get(host_hostvars, "is_deployed", default=True):
+                    display.warning(f"No test results for host {host} since it's marked as not deployed.")
+                    continue
+                # Getting the host results temp file saved by eos_validate_state_runner action plugin
+                temp_file = get(host_hostvars, "anta_results.results_temp_file")
+                if not isinstance(temp_file, str) or not Path(temp_file).exists():
```
gmuloc: we see this pattern a lot maybe create some function

```
def _valide_path(arg: Any) -> bool
  return isinstance(arg, str) and Path(arg).exists()
```

and use it in multiple places (even in your arg validator)

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -34,49 +57,40 @@ def run(self, tmp=None, task_vars=None):
         result = super().run(tmp, task_vars)
         del tmp  # tmp no longer has any effect
 
+        # Setup profiling
+        cprofile_file = self._task.args.get("cprofile_file")
+        if cprofile_file:
+            profiler = cProfile.Profile()
+            profiler.enable()
```
gmuloc: yeah ... profiling decorator? :)

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -91,20 +105,65 @@ def run(self, tmp=None, task_vars=None):
...
-    Create a Filter to add the hostname to generate logs,
+    temp_file_prefix = f"{hostname}_"
+    temp_file_directory = Path("/tmp")
+
+    # Search for leftover temp files with the hostname prefix and delete them.
+    for file in temp_file_directory.glob(f"{temp_file_prefix}*"):
+        file.unlink(missing_ok=True)
+
+    # Create the temp JSON file to save the ANTA results
+    with NamedTemporaryFile(mode="w", encoding="UTF-8", suffix=".json", prefix=f"{hostname}_", dir=str(temp_file_directory), delete=False) as temp_file:
```
gmuloc: Handle potential permission issue to write to `/tmp` ;)

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -91,20 +105,65 @@ def run(self, tmp=None, task_vars=None):
...
+    Returns:
+    -------
+      str: The absolute path of the temp file.
     """
-    Create a Filter to add the hostname to generate logs,
+    temp_file_prefix = f"{hostname}_"
+    temp_file_directory = Path("/tmp")
+
+    # Search for leftover temp files with the hostname prefix and delete them.
+    for file in temp_file_directory.glob(f"{temp_file_prefix}*"):
```
gmuloc: taht could be dramatic if I have a hostname called

`my_awesome_spine1`
and one called
`my_awesome_spine10`

;)

I may renove the temp_file of spine1 when running spine10

gmuloc: Just overwrite it and let the user cleanup stuff I guess

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,73 @@
...
+ANSIBLE_METADATA = {"metadata_version": "1.0", "status": ["preview"]}
+
+DOCUMENTATION = r"""
+---
+module: eos_validate_state_reports
+version_added: "4.5.0"
+author: Arista Ansible Team (@aristanetworks)
+short_description: Generates validation reports for the eos_validate_state role
+description:
+  - The C(arista.avd.eos_validate_state_reports) module, an Ansible Action Plugin, is designed to generate
```
gmuloc:
```suggestion
  - The C(arista.avd.eos_validate_state_reports) module is an Ansible Action Plugin designed to generate
```

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,73 @@
...
+    validation reports from the test results produced by the ANTA test framework.
+  - This plugin requires a temporary JSON file for each host in the Ansible play, which contains all test results.
+    The path to this temporary JSON file, used by the plugin, is supplied in the hostvars by the `eos_validate_state_runner` plugin.
```
gmuloc:
```suggestion
    The path to this temporary JSON file is supplied in the hostvars by the `eos_validate_state_runner` plugin.
```

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,73 @@
...
+  csv_report_path:
+    description:
+      - The absolute path where the CSV report will be saved.
+      - Required if C(validation_report_csv) is set to C(True).
+    type: str
+  md_report_path:
+    description:
+      - The absolute path where the Markdown report will be saved.
+      - Required if C(validation_report_md) is set to C(True).
+    type: str
```
gmuloc, line 49: do we need a boolean PLUS a path? Cannot the absence or presence of path be enough?

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,73 @@
...
+  cprofile_file:
+    description:
+      - The filename for storing cProfile data, useful for debugging performance issues.
+      - Be aware that enabling cProfile can affect performance, so use it only for troubleshooting.
```
gmuloc, line 56: So bold move but do we even need to document this? Do we want people to use it? Can we indicate it is for debug only?

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,73 @@
...
+seealso:
+  - name: ANTA website
+    description: Documentation for the ANTA test framework
+    link: https://anta.ninja
+notes:
+  - Enabling the cProfile feature for performance profiling may impact the plugin's performance, especially in production environments.
+  - The plugin manages temporary files created for processing test results, ensuring clean-up post-execution.
+  - Hosts marked as not deployed are automatically skipped, and no test results are processed for these hosts.
```
gmuloc: still would rather use `is_deployed` False to be clear

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
 module: eos_validate_state_runner
 version_added: "4.4.0"
 author: Arista Ansible Team (@aristanetworks)
-short_description: Leverage ANTA for eos_validate_state role
+short_description: Utilizes ANTA for eos_validate_state role validation
```
gmuloc: eos_validate_state role validation is a mouthful .. not sure how to reword this

```suggestion
short_description: Use ANTA for eos_validate_state role validation
```

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
 description:
   - The C(arista.avd.eos_validate_state_runner) module is an Ansible Action Plugin leveraging the ANTA test
     framework to validate that the generated structured configurations by AVD are applied to the devices and
     that the deployed network is working correctly.
-  - This plugin expects that the structued_configs of each device is present in hostvars, otherwise no test will be generated.
+  - This plugin requires that structured_configs for each device be present in hostvars; otherwise, no tests will be executed.
```
gmuloc: technically generated is more correct - you can generate the tests without executing them if you run in check mode.
```suggestion
  - This plugin requires that structured_configs for each device be present in hostvars; otherwise, no tests will be generated.
```

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
-        - Generate a per-device test catalog based on the structured_configs
-        - Run the generated tests against each device and generate a report in Markdown and CSV format.
-        - When using check_mode, only generate the test catalog and generate a report to preview what would tests be run against each device
-        - Dumping the per-device catalog to a file.
-        - Backward compatibility with existing ansible tags behavior for eos_validate_state to filter categories of tests.
+    The plugin offers the following capabilities:
+        - Generating a per-device test catalog based on the AVD structured_config.
+        - Running generated tests against each device, saving the results in a temporary JSON file,
+          and recording the file path in hostvars for use by the report plugin.
+        - In check_mode, only the test catalog is generated, and a report is created to preview the tests that would be run against each device.
```
gmuloc, line 25: how is the report created? is it using the report plugin ? if so this needs to be changed

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
+    description:
+      - The directory where device test results will be saved.
+      - Required if C(save_results) is set to C(True).
     type: str
   skipped_tests:
     description:
-      - A list of dictionaries containing the categories and tests to skip
-      - The keys for the dictionnary are C(categories) and C(tests).
+      - A list of dictionaries specifying categories and tests to skip.
+      - Each dictionary should have keys C(category) and C(tests).
```
gmuloc: should or must?

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
     type: list
     elements: dict
     suboptions:
       category:
         type: str
-        description: The name of one of the AvdTest categories. e.g., C(AvdTestHardware)
-        required: true
+        description: The name of an AvdTest category (e.g., C(AvdTestHardware)).
```
gmuloc, line 56: maybe a link to a list?

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
+      - Be aware that enabling cProfile can affect performance, so use it only for troubleshooting.
 seealso:
   - name: ANTA website
-    description: ANTA documentation
+    description: Documentation for the ANTA test framework
     link: https://anta.ninja
 notes:
-  - C(check_mode) is supported for this module and allows to generate a Test Report without running the tests.
+  - Enabling the cProfile feature for performance profiling may impact the plugin's performance, especially in production environments.
+  - The plugin manages the creation and cleanup of temporary JSON files used for storing test results.
```
gmuloc: let's see for cleanup

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
```diff
@@ -12,64 +12,80 @@
...
+  - This module supports C(check_mode), allowing the generation of test reports without executing the tests.
```
gmuloc, line 79: is it corect to state that you need to call the other plugin to generate the report?

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
```diff
@@ -91,20 +105,65 @@ def run(self, tmp=None, task_vars=None):
                 dry_run=ansible_check_mode,
                 yaml_dumper=AnsibleNoAliasDumper,
             )
+            # Call the function to handle temp file creation
+            temp_file_name = _create_temp_file(hostname, anta_results)
+
+            # Save test results
+            if save_results_name:
+                with Path(save_results_name).open("w", encoding="UTF-8") as results_file:
+                    dump(anta_results, results_file, indent=2, sort_keys=False)
```
gmuloc: Why not simply copying the temp_file content?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/catalog.py
```diff
@@ -34,7 +34,7 @@ class Catalog:
     Catalog class that handles management of ANTA catalogs from AVD.
     """
 
-    def __init__(self, device_name: str, hostvars: dict, skipped_tests: dict, custom_catalog: dict | None = None, custom_catalog_path: str | None = None):
+    def __init__(self, device_name: str, hostvars: dict, skipped_tests: list[dict], custom_catalog: dict | None = None, custom_catalog_path: str | None = None):
```
gmuloc: good catch!

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/csv_report.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def generate_report(self) -> None:
+        """Generate and writes the CSV report using the collected test results."""
+        fieldnames = ["test_id", "node", "test_categories", "test_description", "test", "result", "failure_reasons"]
+        writer = csv.DictWriter(self.csvfile, fieldnames=fieldnames)
+        writer.writeheader()
+
+        # writerows seems to perfom better when writing in batches
+        batch_size = 5000
+        batch = []
```
gmuloc: how did you estimate the "right" batch size?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/csv_report.py
```diff
@@ -0,0 +1,64 @@
...
+
+        for row in self.generate_rows():
```
gmuloc, line 53: probably can use `isslice` here  https://docs.python.org/3/library/itertools.html#itertools.islice
https://stackoverflow.com/questions/6335839/python-how-to-read-n-number-of-lines-at-a-time

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
```diff
@@ -0,0 +1,251 @@
...
+        self.mdfile.write(f"{heading}\n\n")
+
+    def write_table_header(self, headers: list[str]) -> None:
+        """Write a table header to the markdown file.
+
+        Args:
+        ----
+            headers (list[str]): List of header strings for the table.
+        """
+        self.mdfile.write("\n".join(headers) + "\n")
```
gmuloc: is it required to be a function? It is used in one place only

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
```diff
@@ -0,0 +1,251 @@
...
+    def generate_table(self, table_header: list[str], row_generator: Callable[[], Generator[str, None, None]]) -> None:
+        """Generate a table with a header and multiple rows.
+
+        Args:
+        ----
+            table_header (list[str]): The table header.
+            row_generator (Callable): The rows generator function.
+        """
+        self.write_table_header(table_header)
+        for row in row_generator():
```
gmuloc: Use a method called `self.generate_rows()` instead with the following in the base class
```
def generate_rows(self) -> Generator:
  raise NotImplementedError
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
```diff
@@ -0,0 +1,251 @@
...
+        ----
+            headers (list[str]): List of header strings for the table.
+        """
+        self.mdfile.write("\n".join(headers) + "\n")
+
+
+class ValidateStateReport(MDReportBase):
+    """Generate the `# Validate State Report` section of the markdown report."""
+
+    def generate_report(self, cls: type | None = None, *, is_root: bool = True) -> None:
```
gmuloc: I feel like the recursivity here is overkill - given the number of sections plus this makes ythe file dependant of order of the subclasses!

can we not just write some class that says

```
class ValidateStateReport(MDReportBase):

   def generate_report():
     self.generate_test_result_summary()
     self.generate_failed_test_summary()
     self.generate_all_test_results()

  def generate_test_result_summary():
    self.generate_summary_total()
    self.generate_summary_total_per_dut()
    self.generate_summary_total_per_category()

...
```

gmuloc: we can keep the base classe idea

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+class ResultsManager:
+    """Manages and stores test results from eos_validate_state running ANTA.
+
+    This class processes individual test result and maintains statistics like total tests passed, failed, and skipped.
+
+    An instance of this class holding all results and statistics can be used to generate the validation reports.
+    """
+
+    def __init__(self, *, only_failed_tests: bool = True) -> None:
+        """Initialize the ResultsManager with default values and stats counters set to 0.
```
gmuloc, line 22: Maybe add some explanation about the defuatdict stuff here

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+        -------
+            dict: The parsed and standardized test result.
+        """
+        test_categories = [
+            " ".join(word.upper() if word.lower() in ACRONYM_CATEGORIES else word.title() for word in category.split())
+            for category in result.get("categories", [])
+        ]
+
+        # Mapping the ANTA results
+        result_mapping = {"success": "PASS", "failure": "FAIL", "error": "FAIL", "skipped": "SKIPPED", "unset": "NOT RUN"}
```
gmuloc: should be a constant in this file

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+        anta_result = result.get("result", "")
```
gmuloc, line 77: should we not log something here if we don't get a `result` ?
Same if the `anta_result` does not exist in the mapping

we are silently ignoring stuff

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+        new_result = result_mapping.get(anta_result, "")
+
+        # Create the parsed result dictionary
+        return {
+            "test_id": self.test_id,
+            "node": result.get("name", ""),
+            "test_categories": test_categories,
+            "test_description": result.get("description", ""),
+            # Since AVD tests can have the same description and category, ANTA's custom_field is used to differentiate tests
+            "test": result.get("custom_field", "") if result.get("custom_field") != "None" else result.get("test", ""),
```
gmuloc, line 87: are we returning `"None"` as a string from ANTA? this looks wrong

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+            "result": new_result,
+            "failure_reasons": result.get("messages", []),
```
gmuloc: is failure_reasons the right term if we have skipped reasons / other?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+        }
+
+    def _increment_stats(self, test_result: str, dut: str, categories: list[str]) -> None:
+        """Increment test statistics based on the test result.
+
+        Args:
+        ----
+            test_result (str): The test result.
```
gmuloc: maybe status? (I KNOW is is called result in ANTA..)

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+            dut (str): The name of the device under test.
+            categories (list[str]): The categories of the test.
+        """
+        stats_mapping = {
```
gmuloc: CONSTANT?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+            "PASS": ("total_tests_passed", "tests_passed"),
+            "FAIL": ("total_tests_failed", "tests_failed", "categories_failed"),
+            "SKIPPED": ("total_tests_skipped", "tests_skipped", "categories_skipped"),
+            "NOT RUN": ("total_tests_not_run", "tests_not_run"),
+        }
+        stats_to_increment = stats_mapping.get(test_result, ())
```
gmuloc: This function is a bit hard to get as the logic based on prfix in the for loop makes it quite tricky to follow
espcially the for category loop at the end that depends on the prefix yet again and is implicit about the categpries_ one?

gmuloc: I would probably just write a function that does 
```
if result == PASS:
  total_passed +=1
  passed += 1
if result == SKIPED 
   total_skipped +=1
   skipped +=1
   _update_categories( # using an inneer function)
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+        ----
+            result (dict): The test result data to be added and processed.
+        """
+        self.test_id += 1
+        parsed_result = self._parse_result(result)
+        test_result = parsed_result["result"]
+        categories = parsed_result["test_categories"]
+        dut = parsed_result["node"]
+
+        self._increment_stats(test_result, dut, categories)
```
gmuloc: Maybe call this one inside the _parse_result function ?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
```diff
@@ -0,0 +1,144 @@
...
+
+        if test_result == "FAIL":
+            self.failed_tests.append(parsed_result)
+        if not self.only_failed_tests:
+            self.all_tests.append(parsed_result)
```
gmuloc: do we need to maintain two lists here?

Can we not give a function to retrieve only failed_result? (maybe less efficient?) but if you have 1M tests and 1M failed tests ..

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
```diff
@@ -30,13 +30,27 @@ title: Ansible Collection Role eos_valudate_state - Preview Integration with ANT
...
-- All tests will be skipped for a device flagged as undeployed using the host level variable [`is_deployed: false`](https://avd.sh/en/stable/roles/eos_designs/docs/input-variables.html#flagging-a-device-as-not-deployed). Additionally, all tests take into account the `is_deployed` variable value and skip tests accordingly.
+- Tests skipped by ANTA will be marked as `SKIPPED` in the final reports.
+- All tests will be removed from the catalog for a device flagged as undeployed using the host level variable [`is_deployed: false`](https://avd.sh/en/stable/roles/eos_designs/docs/input-variables.html#flagging-a-device-as-not-deployed). Additionally, all tests take into account the `is_deployed` variable value and remove tests accordingly.
+
+!!! warning
+    Tests can also be automatically removed from the catalogs depending on the structured configuration of the devices. ANTA is therefore not aware of these tests and they will not appear in the final report. For example, the `AvdTestMLAG` tests will not be present in a test catalog of a device that do not have an MLAG configuration in his structured configuration.
```
gmuloc: wondering if we should not word it as "the tests are not generated in the catalog, so not forwarded to the test framework (ANTA) and so not in the report".

```suggestion
    Tests can also be automatically removed from the catalogs depending on the structured configuration of the devices. ANTA is therefore not aware of these tests and they will not appear in the final report. For example, the `AvdTestMLAG` tests will not be present in the test catalog of a device that does not have an MLAG configuration in its structured configuration.
```

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
```diff
@@ -65,13 +79,20 @@ title: Ansible Collection Role eos_valudate_state - Preview Integration with ANT
...
 
-- You can now run the eos_validate_state role in check_mode. This will produce a report of tests that will be performed without running the tests on your network.
+- You can now run the eos_validate_state role in check_mode. This will produce a report of tests that will be performed without running the tests on your network. Tests will be marked as `NOT RUN` in the final reports.
 
   ```shell
   ansible-playbook playbooks/fabric-validate.yaml --check
   ```
 
-- You have the option to save the test catalog generate by the role for each device in the `intended/test_catalogs` folder by setting the variable `save_catalog` to `true`.
+- You can increase the Ansible verbosity by adding multiple `-v` when running the playbook. This will give you visibility on which [test categories](#test-categories) are being removed from the catalogs by AVD according to the structured configurations.
```
gmuloc: for each device?

ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
```diff
@@ -40,6 +44,12 @@ accepted_pwr_supply_states: "{{ validation_role.pwr_supply_states | arista.avd.d
 # Allow different states for fans
 accepted_fan_states: "{{ validation_role.fan_states | arista.avd.default(['ok']) }}"
 
+# Generate CSV results file
+validation_report_csv: "{{ validation_role.validation_report_csv | arista.avd.default(true) }}"
+
+# Generate MD results file
+validation_report_md: "{{ validation_role.validation_report_md | arista.avd.default(true) }}"
```
gmuloc, line 51: are we testing setting these to false?

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
```diff
@@ -136,30 +157,17 @@ test_catalogs_dir: '{{ output_dir }}/{{ test_catalogs_dir_name }}'
...
-      - VerifyRoutingProtocolModel
-  - category: AvdTestHardware
-    tests:
-      - VerifyEnvironmentCooling
-
-# Fabric Name, required to match Ansible Group name covering all devices in the Fabric | Required and **must** be an inventory group name.
-fabric_name: "all"
-
 # Allow different manufacturers
 accepted_xcvr_manufacturers: "{{ validation_role.xcvr_own_manufacturers | arista.avd.default(['Arastra, Inc.', 'Arista Networks']) }}"
```
gmuloc, line 178: is this actually correct?

gmuloc: is it really an input variable or should it be
```
validation_role:
  xcvr_own_manufcatures: [dsjalkdajs]
```

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -42,28 +43,51 @@
 ##  Run eos_validate_state using ANTA  ##
 #########################################
 
-- name: Run eos_validate_state_runner leveraging ANTA
+- name: Include ANTA tests
```
gmuloc: hmmm what does it mean "include tests" ? we are not including tests but including tasks

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
```diff
@@ -42,28 +43,51 @@
...
   ansible.builtin.include_tasks: "anta_tests.yml"
-  when: use_anta
+  when: use_anta | bool   # Ensure that `use_anta` is evaluated as a boolean when using `-e` or `--extra-vars`
```
gmuloc: is the comment like this passing ansible lint?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: One temp file per host. I fixed the comment.

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
```diff
@@ -0,0 +1,65 @@
+<!--
+  ~ Copyright (c) 2023 Arista Networks, Inc.
+  ~ Use of this source code is governed by the Apache License 2.0
+  ~ that can be found in the LICENSE file.
+  -->
+
+# eos_validate_state_reports
```
JulioPDX: Consider changing this to `# EOS validate state reports`,in the navigation it looks a bit funky

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
```diff
@@ -0,0 +1,65 @@
...
+    Always use the FQCN (Fully Qualified Collection Name) `arista.avd.eos_validate_state_reports` when using this plugin.
+
+!!! warning "This module is in **preview** mode"
+    This module is not guaranteed to have a backwards compatible interface.
+
+Generates validation reports for the eos\_validate\_state role
+
+## Synopsis
+
+The <code>arista.avd.eos\_validate\_state\_reports</code> module, an Ansible Action Plugin, is designed to generate validation reports from the test results produced by the ANTA test framework.
```
JulioPDX: Any particular reason to use html code versus backticks?

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
```diff
@@ -0,0 +1,65 @@
...
+
+This plugin requires a temporary JSON file for each host in the Ansible play, which contains all test results. The path to this temporary JSON file, used by the plugin, is supplied in the hostvars by the \`eos\_validate\_state\_runner\` plugin.
```
JulioPDX: Likewise here, the syntax highlighting is not working on the preview site with these escape characters

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
```diff
@@ -0,0 +1,65 @@
...
+
+The plugin offers the following functionalities\:
```
JulioPDX: This is also rendering wrong, consider adding blank lines around a list

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
ClausHolbechArista: This file is generated by our tooling. Some of the over-eager escaping is an artifact of that.

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: No I think we should keep going for other play hosts.

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
carl-baillargeon: It is documented in our other plugins. I think the message is clear here that it is for debug only no?

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
carl-baillargeon: We are using these variables today in eos_validate_state:
https://avd.sh/en/stable/roles/eos_validate_state/index.html#default-variables

Since we provide a default path for the CSV and MD reports, I think it makes sense to just have a boolean to turn on/off the report generation.

ansible_collections/arista/avd/plugins/action/eos_validate_state_runner.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/csv_report.py
carl-baillargeon: With more testing on different systems, writing in batches is not conclusive. Removing it.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/csv_report.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/md_report.py
carl-baillargeon: I decided to remove the recursion but kept the class concept. It is easier to work with, especially when each section has multiple functions.

Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: This is coming from ANTA `TestResult` which is a Pydantic model. We technically shouldn't be missing these fields.

carl-baillargeon: If it's missing, it will appear in the final report. Do you think we should still log it?

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Yes we are. Probably somewhere in ANTA when we serialize `TestResult`. It should be `null` in JSON.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Fixed in 14811244bf088911103283bdb6dcacde39f65536

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: When `only_failed_tests` is true there is only 1 list and we don't have to store all tests.

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
carl-baillargeon: Yes this is what's being used today:
https://avd.sh/en/stable/roles/eos_validate_state/index.html#default-variables

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
carl-baillargeon: Fixed in 2f7c91b74a5914fd1e7fabd3bb86711dcbba55b7

ansible_collections/arista/avd/roles/eos_validate_state/tasks/main.yml
carl-baillargeon: Fixed in 2f7c91b74a5914fd1e7fabd3bb86711dcbba55b7

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
carl-baillargeon: Fixed in 2f7c91b74a5914fd1e7fabd3bb86711dcbba55b7

ansible_collections/arista/avd/roles/eos_validate_state/defaults/main.yml
carl-baillargeon: Should we create another molecule scenario?

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/roles/eos_validate_state/ANTA-Preview.md
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Outdated

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_runner.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
carl-baillargeon: Fixed in 68479d7a6549495d8a3a4e5ce970b358e51ea5bc

ansible_collections/arista/avd/plugins/action/eos_validate_state_reports.py
carl-baillargeon: Fixed in 418a598cba52e1dcd878bc5773675179d53b82dc

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Fixed in 418a598cba52e1dcd878bc5773675179d53b82dc

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Fixed in 418a598cba52e1dcd878bc5773675179d53b82dc

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: Lots of repetition if we do `if result == PASS` etc. I refactored a little bit in 418a598cba52e1dcd878bc5773675179d53b82dc Easier to understand now.

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/results_manager.py
carl-baillargeon: I prefer to have 2 distinct functions since they serve different purposes.

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
```diff
@@ -0,0 +1,76 @@
...
+    description:
+      - The filename for storing cProfile data, useful for debugging performance issues.
+      - Be aware that enabling cProfile can affect performance, so use it only for troubleshooting.
+seealso:
+  - name: ANTA website
+    description: Documentation for the ANTA test framework
+    link: https://anta.ninja
+notes:
+  - Enabling the cProfile feature for performance profiling may impact the plugin's performance, especially in production environments.
+  - Hosts with `is_deployed` is False are automatically skipped, and no test results are processed for these hosts.
```
gmuloc:
```suggestion
  - Hosts with C(is_deployed) is False are automatically skipped, and no test results are processed for these hosts.
```

ansible_collections/arista/avd/plugins/modules/eos_validate_state_reports.py
gmuloc: understood - this makes for a lots of knobs but I guess people can deal with it

ansible_collections/arista/avd/plugins/plugin_utils/eos_validate_state_utils/contants.py
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2023 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
```
gmuloc: Ah this file is called `contants.py` can we rename it to `constants.py`?

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
gmuloc: this is auto generated

ansible_collections/arista/avd/docs/plugins/Modules and action plugins/eos_validate_state_reports.md
gmuloc: auto generated

gmuloc:
To summarize most of @JulioPDX's comments in the PR - the auto generated doc for module where we are using `ansible-doc-extractor` would need some love for lists and some escaping being done (e.g. `;`). which renders weirdly in mkdocs

gmuloc (approved):
LGTM :shipit:

carlbuchmann (approved):
LGTM - validated in lab

PR #3369, opened 2023-11-21
Fix(eos_cli_config_gen): Correct syntax for "redistribute bgp" in router bgp

### LINKED ISSUE #3356: BGP leaked Route redistribution not working
### Issue Summary

The following config:
```
redistribute_routes:
  - source_protocol: bgp
     route_map: RM-DEFAULT-ROUTE-ONLY
     include_leaked: true
```
will result in the following eos cli command:
```
redistribute bgp include leaked route-map RM-DEFAULT-ROUTE-ONLY
```

However the correct eos cli command would be:
```
redistribute bgp leaked route-map RM-DEFAULT-ROUTE-ONLY
```

That is not the case for static or connected. There the 'include' is needed. This is a bgp specific issue.

eos_cli_config_gen

### How do you run AVD ?

None

### Steps to reproduce

```shell
Try to redistribute bgp routes including leaked ones with structured config. (config snipped above)
```

### Relevant log output

```shell
fatal: [Borderleaf1]: FAILED! => {"changed": false, "module_stderr": "Invalid input (at token 2: 'include')", "module_stdout": "", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error"}
```
CLI help from my 4.30.3M vEOS (note the missing 'include'):
```
redistribute bgp ?
  leaked  Include leaked routes of this protocol while redistributing
```

Static looks like this for example:
```
redistribute static ?
  include    Include following routes while redistributing
  rcf        Routing control functions configuration to filter and modify paths
  route-map  Name a route map
  <cr>
```
```


### Contributing Guide

### PR DESCRIPTION
**Pull request by carlbuchmann**
## Change Summary

Correct syntax in "restribute bgp leaked" in router bgp.
- Remove incorrect `include` statement being generated when source routing protocol is bgp
- Since `leaked` is required, include_leaked: true is no longer required when `redistributing bgp` in router bgp. Instead it will automatically be generated when the source protocol is BGP.

## Related Issue(s)

Fixes #3356

## Component(s) name

## Proposed changes

In `router_bgp` jinja template add logic to look if source_protocol == bgp, then automatically generate `leaked`.

## How to test

See new molecule test cases.

### Repository Checklist

### CODE REVIEW

File: ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/router-bgp-base.yml [hunk 1]
```diff
@@ -68,6 +68,13 @@ router_bgp:
       attribute_map: RM-ATTRIBUTE
       match_map: RM-MATCH
       advertise_only: true
+  redistribute_routes:
+    - source_protocol: ospf
+      include_leaked: true
+    - source_protocol: bgp
+      route_map: RM-REDISTRIBUTE-BGP
+      # this should not do anything as leaked is generated when source protocol is bgp
+      include_leaked: true
```
ClausHolbechArista, line 77: we should remove it here and deprecate the key.

carlbuchmann: The current model needs to support all routing protocols i.e ospf, bgp, static. etc., so we can't deprecate the inlcule_leaked feature.
My intent was to test to ensure that when configuring BGP, `include leaked` should not be generated.

EOS CLI 4.31.0F

```eos
router bgp 65000
   redistribute ospf include leaked route-map ospf-bgp
   redistribute bgp leaked route-map bgp-bgp
   !
   vrf blah
      redistribute ospf include leaked route-map ospf-bgp
      redistribute bgp leaked route-map bgp-bgp
      !
      address-family ipv6
         redistribute bgp leaked route-map bgp-bgp
         redistribute ospfv3 include leaked route-map ospf-bgp
```

We should create a per-protocol data model in the future and deprecate this knob entirely, but as discussed, that is a lot of work and will be added to the backlog.

ClausHolbechArista (approved):

gmuloc (approved):

tgodaA (approved):
LGTM

PR #3385, opened 2023-11-27
Feat(eos_designs): Add uplink_type at nodes level

gmuloc:
## Change Summary

Add the possibility to overwrite the uplink_type at the node level

the usecase is to be able to have a default uplink_type for a given node_type but be able to overwrite it at the nodes level

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Added `uplink_type` via a ref in the node_type schema and make it having better preference in the shared_utils plugin

## How to test
molecule

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/override_uplink_type-d.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/override_uplink_type-d.cfg
@@ -1,0 +1,45 @@
+!RANCID-CONTENT-TYPE: arista
+!
+vlan internal order ascending range 1006 1199
+!
+transceiver qsfp default-mode 4x10G
+!
+service routing protocols model multi-agent
+!
+hostname override_uplink_type-d
+!
+spanning-tree mode none
+!
+no enable password
+no aaa root
+!
+vrf instance MGMT
+!
+interface Port-Channel1
+   description OVERRIDE_UPLINK_TYPE-U_Po1
+   no shutdown
+   switchport
+   switchport trunk allowed vlan none
+   switchport mode trunk
+!
+interface Ethernet1
+   description OVERRIDE_UPLINK_TYPE-U_Ethernet1
+   no shutdown
+   channel-group 1 mode active
+!
+interface Loopback0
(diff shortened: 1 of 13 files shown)


ClausHolbechArista (approved):
LGTM

philippebureau:
The override works locally on the nodes defined but will not change the remote node(s) port 
in cases of L3LS design and fail with error on the spine node due to the interface being switched to layer2:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.arista.avd.plugins.plugin_utils.errors.errors.AristaAvdError: Configuring eBGP neighor without a remote_as
fatal: [dc1-spine1 -> localhost]: FAILED! => changed=false 
  msg: Configuring eBGP neighor without a remote_as
```

Is this an expected / not covered use case?

ClausHolbechArista:
> The override works locally on the nodes defined but will not change the remote node(s) port in cases of L3LS design and fail with error on the spine node due to the interface being switched to layer2:
> 
> ```
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible_collections.arista.avd.plugins.plugin_utils.errors.errors.AristaAvdError: Configuring eBGP neighor without a remote_as
> fatal: [dc1-spine1 -> localhost]: FAILED! => changed=false 
>   msg: Configuring eBGP neighor without a remote_as
> ```
> 
> Is this an expected / not covered use case?

Yeah the issue here is that we don't support for combination of `underlay_router: true` and `uplink_type: port-channel`. Maybe we need to fix that as well. Either catch the bad combo and raise with a good error message, or make it work somehow.

To be clear, port-channel uplink type means L2 port-channel. We do not intend to add support for l3 port-channel uplink with this change.

philippebureau:
> 
> Yeah the issue here is that we don't support for combination of `underlay_router: true` and `uplink_type: port-channel`. Maybe we need to fix that as well. Either catch the bad combo and raise with a good error message, or make it work somehow.
> 
> To be clear, port-channel uplink type means L2 port-channel. We do not intend to add support for l3 port-channel uplink with this change.

Thank you for confirming.  

LGTM

philippebureau (approved):
LGTM

PR #3439, opened 2023-12-15
Feat(eos_designs): Add l3_edge and core_interfaces VRF support

### LINKED ISSUE #3005: Add support for `vrfs` under `l3_edge` / `core_interfaces`
Related to https://github.com/aristanetworks/avd/issues/2399

philippebureau:
## Change Summary

Add capability to set a VRF on both side of p2p_link nodes 

## Related Issue(s)

Fixes #3005 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
add a key called 'vrf' under p2p_links

## How to test
l3_edges:  ##this could be core_interfaces
  p2p_links:
    - id: 
      vrf: TEST40

### Repository Checklist


gmuloc (changes requested):
so this is quite much more complicated than it looks like.
The PR as it is does not render the VRF so would generate invalid configuration if the VRF is not created by other means.

The solution forward we would like to go with with @ClausHolbechArista  is to make:
1. Validate that the VRF exists in network_services tenants
2. if there is a p2p_link with the VRF defined under it, make sure the VRF is included in the filtered VRF, in the filtered tenants in network services to render it.

2. implies moving the filtered_tenants to shared_utils which is quite some work

Also it would be nice to be able to support having the VRF being named differently at both ends for a VRF-Lite scenario.

philippebureau:
This is way more complicated than I expected.

Would implementing "Validate that the VRF exists in network_services tenants" be enough as a phase 1?

I can make sure this PR would not close issue 3005

philippebureau:
TODO: figure out the logic to move the BGP neighbors in the right VRF

PR #3474, opened 2024-01-08
Fix: Minor container fixes

ankudinov:
## Change Summary

This PR adds xz-utils to the base container, that is required for Docker import to work when D-in-D is installed.
It also adds `/home/${USER}/.ansible/collections/ansible_collections/arista/avd` explicitly to `dev` image, as otherwise mount to non-existing path can fail in some environments.

## Component(s) name

containers

## How to test

Merge to rebuild containers. Pull and check that xz-utils are installed.
The best way to test for mount issues to non-existing path is to start Codespace for AVD PR and check if AVD collection was installed successfully.
Once `base` is re-build, `dev` and `universal` builds must be triggered manually to get xz-utils.

### DIFF
--- a/containers/base/.devcontainer/Dockerfile
+++ b/containers/base/.devcontainer/Dockerfile
@@ -7,6 +7,8 @@ RUN apt-get update \
     && apt-get install -y --no-install-recommends \
     make \
     git \
+    # xz-utils required for docker import to work
+    xz-utils \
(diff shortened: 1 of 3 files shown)


ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #3483, opened 2024-01-11
CI: build avd universal container on pre-release

ankudinov:
## Change Summary

Start building universal container on pre-release.

## Component(s) name

containers

## Proposed changes
trigger CI workflow on `prereleased` event

## How to test
- Create a test repository.
- Add a Github Actions workflow similar to: https://github.com/ankudinov/gh-actions-playground/blob/master/.github/workflows/test-pre-release.yml
- Create pre-release and verify that CI is triggered as expected.

Also check if universal container will be created on the next pre-release.

### DIFF
--- a/.github/workflows/container_build_universal.yml
+++ b/.github/workflows/container_build_universal.yml
@@ -13,7 +13,7 @@ on:
     branches:
       - devel
   release:
-    types: [published]
+    types: [published, prereleased]
 
 jobs:
   build_universal_container:


ClausHolbechArista:
We will get some issues with stability if we build on every merge/prerelease. Please make sure that pre-releases won't get `latest` and also make sure the pre-release version works for containers.

ankudinov:
@ClausHolbechArista We currently build universal on every PR merge. So, usually :latest will point to a container with AVD collection installed from a dev branch. And according to the Docker best practices, it's never supposed to be used unless any random and latest version is fine. For stable setup, a stable tag must be specified explicitly.  
I can imagine people using :latest in the field occasionally, but it their fault and do educate not to do that.  
Also the way how it currently works looks like:
- In the matrix file for specific container file we specify what Python version is latest
- When container build is triggered by any event (PR merge, manual trigger, release and pre-release in future), the main workflow triggers the child workflow that is the same for all container kinds
- In there we check if any custom tags were defined and if latest is in the list, we add it on top of auto-created tag. We could theoretically add some logic to check if the container build was triggered by release event and avoid adding :latest in any other case. But 2 concerns here:
  - It's quite some shell complexity, must be covered in a separate PR
  - This defeats the idea of :latest. Which is really just latest and must not be stable.

TLDR: this PR is not changing current concept and causing any additional instability.

ClausHolbechArista:
@ankudinov thanks, now the penny dropped. One day it would be nice to get a stable and a prerelease or edge tag, but that is not for this PR :)

ClausHolbechArista (approved):

PR #3486, opened 2024-01-11
Feat(eos_designs): Loopbacks data model under tenant vrfs

emilarista:
## Change Summary

Adds data model for loopback_interfaces under tenant vrfs similar to l3_interfaces.

```yaml
...
    vrfs:
      - name: <name>
        loopbacks:
          - loopback: <int> # Ex. 123
            ip_address: <ip/mask>
            node: <str>
            description: <str>
            ospf:
              enabled: <bool>
              area: <str> # Ex. 0.0.0.0 or 0
            raw_eos_cli: |
              <multiline cli>
```

## Component(s) name

`arista.avd.eos_designs`

## How to test

Tested with molecule scenario

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/virtual_source_nat_vrfs.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/virtual_source_nat_vrfs.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 
 from functools import cached_property
 
+from ansible_collections.arista.avd.plugins.plugin_utils.utils import append_if_not_duplicate
+
 from .utils import UtilsMixin
 
 
@@ -23,20 +25,34 @@ class VirtualSourceNatVrfsMixin(UtilsMixin):
         Return structured config for virtual_source_nat_vrfs
 
         Only used by VTEPs with L2 and L3 services
+        Using data from loopback_interfaces to avoid duplicating logic
         """
         if not (self.shared_utils.overlay_vtep and self.shared_utils.network_services_l2 and self.shared_utils.network_services_l3):
             return None
 
+        if (loopback_interfaces := self.loopback_interfaces) is None:
+            return None
+
         virtual_source_nat_vrfs = []
-        for tenant in self._filtered_tenants:
-            for vrf in tenant["vrfs"]:
-                if (loopback_interface := self._get_vtep_diagnostic_loopback_for_vrf(vrf)) is not None:
-                    virtual_source_nat_vrfs.append(
-                        {
-                            "name": loopback_interface["vrf"],
-                            "ip_address": loopback_interface["ip_address"].split("/", maxsplit=1)[0],
-                        }
-                    )
+        for loopback_interface in loopback_interfaces:
+            if (vrf := loopback_interface.get("vrf", "default")) is None:
+                continue
+
+            # Using append_if_not_duplicate but with ignore_same_dict and ignore_keys.
+            # It will append the dict unless the same "name" is already in the dict.
+            # It will never raise since we only have these two keys.
+            append_if_not_duplicate(
+                virtual_source_nat_vrfs,
+                primary_key="name",
+                new_dict={
+                    "name": vrf,
+                    "ip_address": loopback_interface["ip_address"].split("/", maxsplit=1)[0],
+                },
+                context="virtual_source_nat_vrfs",
+                context_keys=["name"],
+                ignore_same_dict=True,
+                ignore_keys={"ip_address"},
+            )
 
         if virtual_source_nat_vrfs:
             return virtual_source_nat_vrfs
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/virtual_source_nat_vrfs.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/virtual_source_nat_vrfs.py
@@ -23,27 +23,20 @@ class VirtualSourceNatVrfsMixin(UtilsMixin):
         Return structured config for virtual_source_nat_vrfs
 
         Only used by VTEPs with L2 and L3 services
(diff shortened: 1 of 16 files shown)


ClausHolbechArista:
I would prefer just to have one ip address, interface, node per entry.

emilarista:
> I would prefer just to have one ip address, interface, node per entry.

Changed the data model according to this.

philippebureau (dismissed):
LGTM

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/virtual_source_nat_vrfs.py
```diff
@@ -33,12 +33,17 @@ def virtual_source_nat_vrfs(self) -> list | None:
...
+            for vrf in virtual_source_nat_vrfs:
+                if loopback_interface["vrf"] == vrf["name"]:
+                    vrf_exists = True
+            if not vrf_exists:
+                virtual_source_nat_vrfs.append(
+                    {
+                        "name": loopback_interface["vrf"],
+                        "ip_address": loopback_interface["ip_address"].split("/", maxsplit=1)[0],
+                    }
+                )
```
ClausHolbechArista: As discussed we need to find a better way for this logic. I will look into it.

ClausHolbechArista (changes requested):

ClausHolbechArista:
@philippebureau I modified this PR a bit since your review. Can you take another round? Notice the schema changed slightly. Thanks.

ClausHolbechArista (approved):

ankudinov (commented):
Tested the PR in Codespace (adjust devcontainer.json to the new image if you want to reproduce), AVD examples single-dc inventory.

Following `/workspaces/ansible-avd/ansible_collections/arista/avd/examples/single-dc-l3ls/group_vars/NETWORK_SERVICES.yml` produces confusing results:

```yaml
---
tenants:
  # Definition of tenants. Additional level of abstraction to VRFs
  - name: TENANT1
    # Number used to generate the VNI of each VLAN by adding the VLAN number in this tenant.
    mac_vrf_vni_base: 10000
    vrfs:
      # VRF definitions inside the tenant.
      - name: VRF10
        # testing https://github.com/aristanetworks/avd/pull/3486
        loopbacks:
          - loopback: 1234
            ip_address: 1780.10.10.30/32
            node: dc1-leaf1a
            description: just-a-test-loopback
            raw_eos_cli:
              description description-from-raw-eos-cli
          - loopback: 2345
            ip_address: 192.10.10.20/234
            node: dc1-leaf1b
            description: test2345
          - loopback: 987
            ip_address: 192.30.30.20/32
            node: dc1-leaff1b
            description: test2345
        # VRF VNI definition.
        vrf_vni: 10
        # Enable VTEP Network diagnostics
        # This will create a loopback with virtual source-nat enable to perform diagnostics from the switch.
        vtep_diagnostic:
          # Loopback interface number
          loopback: 10
          # Loopback ip range, a unique ip is derived from this ranged and assigned
          # to each l3 leaf based on it's unique id.
          loopback_ip_range: 10.255.10.0/27
        svis:
          # SVI definitions.
          - id: 11
            # SVI Description
            name: VRF10_VLAN11
            enabled: true
            # IP anycast gateway to be used in the SVI in every leaf.
            ip_address_virtual: 10.10.11.1/24
          - id: 12
            name: VRF10_VLAN12
            enabled: true
            ip_address_virtual: 10.10.12.1/24
      - name: VRF11
        loopbacks:
          - loopback: 4321
            ip_address: 1780.10.10.30/32
            node: dc1-leaf2a
            description: just-a-test-loopback
        vrf_vni: 11
        vtep_diagnostic:
          loopback: 11
          loopback_ip_range: 10.255.11.0/27
        svis:
          - id: 21
            name: VRF11_VLAN21
            enabled: true
            ip_address_virtual: 10.10.21.1/24
          - id: 22
            name: VRF11_VLAN22
            enabled: true
            ip_address_virtual: 10.10.22.1/24

    l2vlans:
      # These are pure L2 vlans. They do not have a SVI defined in the l3leafs and they will be bridged inside the VXLAN fabric
      - id: 3401
        name: L2_VLAN3401
      - id: 3402
        name: L2_VLAN3402
```

1. There is no verification for IP address format. `1780.10.10.30/32` is accepted and inserted in the config as well as `192.10.10.20/234`. Not sure about complexity and if we should plan IP address validation at this stage, but it's feasible to list that somewhere in limitations for the data model.
2. Creating a loopback on non-existing leaf fails silently. Typing hostnames incorrectly can be quite common in the field and this should be detected to avoid hunting that in future inventories.
3. I'd consider switching `node` and `ip_address` to list. As the case of provisioning same loopback for a similar purpose on multiple switches can be quite common.

If required - happy to LGTM and park in a different issue as the PR works for me in general.

ClausHolbechArista:
> There is no verification for IP address format. 1780.10.10.30/32 is accepted and inserted in the config as well as 192.10.10.20/234. Not sure about complexity and if we should plan IP address validation at this stage, but it's feasible to list that somewhere in limitations for the data model.

We don't have that schema support yet. But schema descriptions should mention for expected format until then.

> Creating a loopback on non-existing leaf fails silently. Typing hostnames incorrectly can be quite common in the field and this should be detected to avoid hunting that in future inventories.

That is the same everywhere in AVD today. It would be quite "expensive" to perform validation of these fields in a scaled setup.

> I'd consider switching node and ip_address to list. As the case of provisioning same loopback for a similar purpose on multiple switches can be quite common.

If you check the history of this PR, it started as lists. Since we already have a model for common loopbacks, this is more thought as a form of override for special cases. Also we would like to avoid more models with multiple lists that the user has to align.


Next review, please comment in the code/schema as separate points so we can track/reply/resolve individually. These large comments makes it harder to follow.

ankudinov (approved):
LGTM

PR #3541, opened 2024-01-24
Fix: Remove devcontainer mounts to address error 16 with molecule

ankudinov:
## Change Summary

This PR addresses a bug with molecule tests failing in AVD dev container due to `error 16: device or resource busy` when bind mount from workspace subdirectory to Ansible collection path is present.
As this bind mount is causing significant confusion and increases dev container complexity - it's feasible to remove this option until we have a real use case.
Container can be always rebuild after the change if required to reinstall the collection.

## Component(s) name

AVD dev container

## Proposed changes

- remove `mounts` from `devcontainer.json`
- update comments in the `devcontainer.json`
- consider replacing `onCreateCommand` with `postCreateCommand` (additional testing required before moving this PR out of draft stage)

## How to test
Start dev container from this branch and test if molecule tests are successful.


ankudinov:
Switched from `onCreateCommand` and `postCreateCommand ` to `postStartCommand`. The first 2 are initiating install script 2 early in the process, before VSCode finishes all the tasks, like copying git credentials and marking git repo as trusted with `git config --global --add safe.directory /workspaces/ansible-avd`. If the collection install script starts too early it fails with `fatal: detected dubious ownership in repository at '/workspaces/ansible-avd'`.  
The downside of `postStartCommand` is that it is triggered every time when container starts. To prevent triggering installation every time the script now verifies Ansible binary presence first with `command -v`.

ankudinov:
I'm keeping the logic required to install collections from a bind mount part of the script to cover some special cases that may come in the future and to avoid re-testing the logic. But open for suggestions and alternative opinions.  
The mount option will be marked as unsupported later (once images are release) to avoid IO problems like molecule case in the field.

ClausHolbechArista (approved):

.devcontainer/devcontainer.json
```diff
@@ -1,18 +1,16 @@
...
     //
     // "containerEnv": {
     //     "AVD_GITHUB_REPO": "aristanetworks/avd",
     //     "AVD_BRANCH_NAME": "devel"
     // },
     // Run entrypoint script manually as it's ignored by dev container CLI otherwise.
     // The dev entrypoint is used to install Ansible collections and requirements, as they are not included with the dev version.
-    // "true" is required to exit "onCreateCommand" without entering ZSH.
-    "onCreateCommand": "/bin/entrypoint.sh true"
+    // "true" is required to exit "postCreateCommand" without entering ZSH.
```
gmuloc: Should it look like this?
```suggestion
    // "true" is required to exit "postStartCommand" without entering ZSH.
```

gmuloc (approved):

PR #3548, opened 2024-01-25
Feat(eos_designs): Add the possibility to set CPU max allocation

gmuloc:
## Change Summary

For AutoVPN RRs and Pathfinders, it is recommended to limit the number of CPUs allocated to Data Plane. This PR allows to set it for every device and enforce it for WAN Routes Reflectors (this may need to be relaxed).

Setting this as default at the platform level would requires to maintain a list of Cloud platforms which would be quite cumbersome.

## Component(s) name

`arista.avd.eos_designs`

## How to test

molecule tests added

### Repository Checklist


ClausHolbechArista (approved):
I would have preferred to have sfe_ in front of the key, since that would be the natural thing to search for when configuring it. But I am ok with this too.

carlbuchmann (approved):
LGTM

PR #3560, opened 2024-01-29
Feat(eos_designs): Inject default VRF policy with a match-all statement when missing

gmuloc:
## Change Summary

In the current implementation, when there is no policy defined under `wan_virtual_topology.vrfs` for VRF default. AVD will raise an error.

This PR replaces this behavior with auto configuring a policy with one default statement (`default_virtual_topology`) matching all local path_groups in the load balance policy.

Of course, because it is VRF default, the Control plane special virtual topology is still injected.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

When the VRF default is missing from `wan_virtual_topology.vrfs`, AVD creates a policy called `DEFAULT-AVT-POLICY` that is configured with one default statement to match all traffic using all locally available path-groups

## How to test
molecule updated to test it in AutoVPN use case.

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml
```diff
@@ -100,8 +100,7 @@ tenants:
 
 wan_virtual_topologies:
   vrfs:
-    - name: default
-      policy: DEFAULT-AVT-POLICY
+    # Not configuring default VRF policy on purpose to test auto-generation
```
ClausHolbechArista: Why are these changes causing the IT stuff to get removed?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
```diff
@@ -42,10 +41,10 @@ router path-selection
       path-group INET
    !
    policy DEFAULT-AVT-POLICY-WITH-CP
+      default-match
```
ClausHolbechArista: The order on EOS is opposite I believe? default-match being last.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml
ClausHolbechArista: Ah because you are adding CUSTOM-DEFAULT-AVD-POLICY but not referring to it anywhere. We are loosing a lot of test coverage with this change.

gmuloc: this is covered in CV PATHFINDER

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
gmuloc: not for router path-selction policies

```
router path-selection
   tcp mss ceiling ipv4 ingress
   !
   policy BLAH
      default-match
         load-balance BLAH
      !
      10 application-profile TITI
         load-balance TITI

File: ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml (hunk 1 above)
Comment (gmuloc) [reply ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml]: otherwise I need two new examples in eos_designs just for this

File: ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py [hunk 3]
```diff
@@ -316,14 +316,24 @@ def _wan_load_balance_policies(self) -> list:
                     context_keys=["name"],
                 )
 
-            default_virtual_topology = get(policy, "default_virtual_topology", required=True)
+            default_virtual_topology = get(
```
ClausHolbechArista, line 322: Something seems wrong about the order here. You error out suggesting the drop_unmatched _above_ the line where you check for drop_unmatched.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -316,14 +316,24 @@ def _wan_load_balance_policies(self) -> list:
...
+                ),
+            )
             if not get(default_virtual_topology, "drop_unmatched", default=False):
                 name = get(default_virtual_topology, "name", default=f"{policy['name']}-DEFAULT")
                 context_path = f"wan_virtual_topologies.policies[{policy['name']}].default_virtual_topology"
                 append_if_not_duplicate(
                     list_of_dicts=wan_load_balance_policies,
                     primary_key="name",
-                    new_dict=self._generate_wan_load_balance_policy(f"LB-{name}", default_virtual_topology, context_path),
+                    new_dict=self._generate_wan_load_balance_policy(
```
ClausHolbechArista: Why not just do the _default_all check in the function you call instead of here where it seems a bit "clunky"?

ClausHolbechArista (commented):
The logic becomes increasingly hard to follow. More comments in the code might help.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
gmuloc: so not really - line 319 is checking for the presence of the key that must be there
the second line is checking that either `drop_unmatched` is set (and then we render nothing) or then we render path-groups.

This is to force explicit configuration of `defaut_virtual_topology` every time as we discussed - if you want to drop unmatched then you need to do it explicitly rather than having it done silently for you (I need to check if the dict is required in schema)

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
gmuloc: hmm the mechanism using input was already used some other place for control_plane LB policy and here I am reusing this but I can just use the key instead and change the behavior. Let me refactor to make it less clunky

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
ClausHolbechArista: The error message on line 319 is that the default_virtual_topology must be specified. This will error out no matter what since you have required=True. So setting `drop_unmatched` to True will not stop that.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml
ClausHolbechArista: You could just add another VRF with the old policy, so we keep all the checks also for autovpn case.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
gmuloc: yes thats the goal
1. you must define `default_virtual_topology` with either `drop_unmatched` or something else - line 319 checks that it is defined if not it errors out
2. either you have `drop_unmatched` and nothing is rendered or we render the rest

ClausHolbechArista: For someone getting this error, this last line will only confuse them. I think this org_key should just be the full variable path to the missing key. But it would be simpler to just require it in the schema and just use `['default_virtual_topology']` here.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -316,14 +316,24 @@ def _wan_load_balance_policies(self) -> list:
...
-            default_virtual_topology = get(policy, "default_virtual_topology", required=True)
+            default_virtual_topology = get(
+                policy,
+                "default_virtual_topology",
+                required=True,
+                org_key=(
+                    f"A 'default_virtual_topology must be defined for policy {policy['name']}. "
+                    "It is possible to disable default-match by setting 'drop_unmatched' to 'true'."
+                ),
+            )
```
gmuloc, line 324: check for policy keyword line 329 and raise if not present

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -391,25 +401,30 @@ def _filtered_wan_policies(self) -> list:
...
+                vrf_default_policy,
+                required=True,
+                custom_error_msg=(
+                    f"The policy {vrf_default_policy} defined for vrf default under 'wan_virtual_topologies.vrfs' "
+                    "is not defined under 'wan_virtual_topologies.policies'."
+                ),
+            ).copy()
+        else:
+            # Injecting a hidden key _default_all, used when generating the relevant Load Balance Policy
+            default_policy = {"name": "DEFAULT-AVT-POLICY", "default_virtual_topology": {"_default_all": True}}
```
gmuloc: use all path-groups here and in control-plane to avoid extra flags

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/AUTOVPN_TESTS.yml
gmuloc: adding new hosts instead for thsi specifically

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -265,6 +254,17 @@ def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context
 
         return wan_load_balance_policy
 
+    @cached_property
+    def _local_path_groups_connected_to_pathfinder(self) -> list:
+        """
+        Return list of local path_groups connected to pathfinder
+        """
```
ClausHolbechArista:
```suggestion
    def _local_path_groups_connected_to_pathfinder(self) -> list:
        """
        Return list of names of local path_groups connected to pathfinder
        """
```

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM, works as expected in test topology

PR #3582, opened 2024-02-02
CI: Update gitattributes to avoid conflicts for generated files

ClausHolbechArista:
Update gitattributes to avoid conflicts for generated files.

Using "merge=union" should merge the files in some random way, but since we run pre-commit automatically, the files will be regenerated and fixed in a commit done by the bot.


gmuloc (approved):
:shipit: 🎲

PR #3584, opened 2024-02-02
Refactor: Exclude schema pickle files from repo and build at runtime

ClausHolbechArista:
Refactor: Exclude schema pickle files from repo and build at runtime


python-avd/scripts/build-schemas.py
```diff
@@ -117,7 +80,8 @@ def build_schema_tables(schema_store):
 
 def main():
     combine_schemas()
-    schema_store = compile_schemas()
+    print("Rebuilding pickled schemas")
```
carl-baillargeon, line 83: print? Why not a logger?

carl-baillargeon (dismissed):
LGTM if CI passes

python-avd/scripts/build-schemas.py
ClausHolbechArista: This is just a cli script printing to pre-commit.

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
 from yaml import safe_load
 
-from ansible_collections.arista.avd.plugins.plugin_utils.schema.default_schemas import DEFAULT_PICKLED_SCHEMAS, DEFAULT_SCHEMAS
+from .avdschemaresolver import AvdSchemaResolver
+from .default_schemas import DEFAULT_PICKLED_SCHEMAS, DEFAULT_SCHEMAS
 
 
 @lru_cache
-def create_store(load_from_yaml=False):
+def create_store(load_from_yaml=False, force_rebuild=False):
```
gmuloc: maybe a docstring here?

gmuloc: what should this function return?

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
+
+def _should_recompile_schemas() -> bool:
+    """
+    Returns true if pickled schemas should be recompiled
+    """
+    # Check if any pickled schema is missing
+    for id, pickle_file in DEFAULT_PICKLED_SCHEMAS.items():
+        if not pickle_file.exists():
+            return True
+    # Check if any hash file is missing
```
gmuloc: Using values instead of items as not using id + using any
```suggestion
if any(not pickle_file.exists() for pickle_file in DEFAULT_PICKLED_SCHEMAS.values()):
  return True
```

gmuloc: reading the rest, actually rather than doing three loops, why not just do the three checks in a single loop?

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
+    for schema_name, pickle_file in DEFAULT_PICKLED_SCHEMAS.items():
+        if schema_name == "avd_meta_schema":
+            # Do not resolve $ref in the meta schema.
+            resolved_schema = temp_schema_store[schema_name]
+        else:
+            # Copying so we can pop below.
+            resolved_schema = _resolve_schema(temp_schema_store[schema_name].copy(), temp_schema_store)
+
+            # Since the schema is now fully resolved we can drop the $defs.
+            resolved_schema.pop("$defs", None)
```
gmuloc: This assumes that `defs` are all top level - which is the case today but maybe worth adding a comment

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
+
+            # Inplace update the schema store with the resolved variant without $def.
+            # This is needed so eos_designs will not resolve to a schema with another $ref.
+            temp_schema_store[schema_name] = resolved_schema
```
gmuloc: could also use the `resolved_schema_store` when resolving the schema for `eos_designs` ?

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
+                pickle_dump(resolved_schema, stream, HIGHEST_PROTOCOL)
+
+            # Update the .sha1 file with the new hash of the yaml schema file.
+            schema_file = DEFAULT_SCHEMAS[schema_name]
+            new_hash = sha1(schema_file.read_bytes(), usedforsecurity=False).hexdigest()
+            schema_file.with_suffix(".sha1").write_text(new_hash, encoding="UTF-8")
+        except PermissionError:
+            # Ignoring PermissionError so we can operate in read-only environments
+            # (like "ansible-test units" containers).
+            pass
```
gmuloc, line 119: Maybe a log

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
```diff
@@ -1,29 +1,129 @@
...
+
+    # We rely on eos_cli_config_gen being before eos_designs,
+    # so anything in eos_cli_config_gen can be resolved and $def popped before resolving from eos_designs.
+    for schema_name, pickle_file in DEFAULT_PICKLED_SCHEMAS.items():
+        if schema_name == "avd_meta_schema":
+            # Do not resolve $ref in the meta schema.
+            resolved_schema = temp_schema_store[schema_name]
+        else:
+            # Copying so we can pop below.
+            resolved_schema = _resolve_schema(temp_schema_store[schema_name].copy(), temp_schema_store)
```
gmuloc: do we need this when we are also deeocopying in the `_resolve_schema` function? Also should we not pop `$defs` in the `_resolve_schema` funciton instead?

.github/workflows/pull-request-management.yml
```diff
@@ -471,14 +464,16 @@ jobs:
...
       - uses: actions/checkout@v3
       - name: Install requirements
         # Install the specific version of galaxy-importer used on galaxy.ansible.com
-        # The old version conflicts with our versions of other tooling,
+        # The version conflicts with our versions of other tooling,
         # so we let the galaxy-importer version resolve remaining requirements.
+        # Also install deepmerge and jsonschema for use in rebuilding schemas.
         run: |
-          pip install "galaxy-importer==0.4.0.post1"
+          pip install "galaxy-importer==0.4.19" "jsonschema>=4.5.1,<4.18" "deepmerge>=1.1.0" "pydantic"
```
gmuloc: Can we make this a file instead? we are tracking versions dependencies all over the place.
Also it says pydantic line 476 but not in the comment line 474 - is it on purpose?

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: We don't do logging in AVD today.

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: We only support defs at the top.

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: I am looping over different files. And to avoid doing 3 checks on each I wanted to exit quickly for the most common case.

.github/workflows/pull-request-management.yml
ClausHolbechArista: We can only install a subset of our usual requirements, and I think adding another requirements file will confuse. Here it is clear why it is needed.
I will add pydantic to the comment.

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
gmuloc: the outer loop is for the same files line 46 and 51

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
gmuloc: we do today yes

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
gmuloc: this is just failing silently - which can be hard to troubleshoot

.github/workflows/pull-request-management.yml
gmuloc: but it means we need to maintain this file as well with dependencies versions

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: yes but since I replace in one, the check is not for the same file.

.github/workflows/pull-request-management.yml
ClausHolbechArista: yes for this special case where we have to match whatever Ansible Galaxy decides is the current version.

ClausHolbechArista: refactored

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: I wanted to avoid running the hash operation unless strictly needed, but I have combined the last two to make you happy :)

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: It does not fail as such. We will just not save a pickled file, so next time we run, the schema will be resolved again.

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: done

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: fixed

carlbuchmann (approved):
LGTM

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
gmuloc: Understood - do you think it could be puzzling in some troubleshooting situations maybe? Like seeing the pickle file not being updated silently because force_rebuild is set to True but permission is failing and this is just eaten by the function?

Not willing to block the PR for this but wondering if it could lead to head banging troubleshooting sessions :)

gmuloc (approved):
LGTM

ansible_collections/arista/avd/plugins/plugin_utils/schema/store.py
ClausHolbechArista: force rebuild is only used by the pre-commit / schema tooling scripts, so not something part of AVD. We have to move on and there is no alternative for this available today.

PR #3594, opened 2024-02-05
Fix(eos_designs): Remove WAN RR BGP peering when no common path-group

gmuloc:
## Change Summary

Right now the code appropriately does not configure STUn profile and so on but still tries to establish a peering between an Edge and a RR when they don't have coexisting path-groups making the BGP peering never establishing.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Removing this peering for now - this need to be revisited later when implementing the import path-group statement under `router path-selection`

## How to test

molecule test correctly changed

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-common-path-group.cfg
```diff
@@ -214,7 +214,6 @@ application traffic recognition
       application CUSTOM-VOICE-APPLICATION
    !
    field-set ipv4 prefix CONTROL-PLANE-APP-DEST-PREFIXES
```
ClausHolbechArista, line 216: So this setup will not work since we have no pathfinders available. Should we error out in this case instead of producing a non-working config?

ClausHolbechArista (approved):
Approved since the current behavior is to produce a partial config, but I am questioning whether we should emit an error instead when we have no pathfinders.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-common-path-group.cfg
gmuloc: so the purpose of the PR is to make sure we don't configure the neighbor which was done erroneously until now. In a scenario with multiple pathfinders where at least one would have a common path-group we want to configure only the ones with common path-groups.

However I agree that if there is no common path-group with ALL configured pathfinders, we should probably error out, though not sure if it could be valid for HA site for instance. Will track this and ask engineering

carlbuchmann (approved):
LGTM

PR #3602, opened 2024-02-08
Feat(eos_designs): Implement WAN/LAN redistribution for eBGP LAN

gmuloc:
## Change Summary

* aligned node types underlay protocols in node_type_keys and underlay_routing_protocol
* changed default underlay_routing_protocol of `wan_xx` router
* updated 
* Added FLOW tracking on uplinks for cv-pathfinder
* Removed IPSec IKE profile to be generated for dataplane profile (will need to add it back in HA case, or rather, whenever there is a static peer configured in WAN case to be more generic in the mechanism)
 
## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

**NOTE: The behavior below is for WAN router**

* Mark all locally learned route on the router with the SoO extended community
* Configure BGP route-map out towards underlay peers (LAN side) to advertise SoO marked routes (learned locally) and WAN routes (using AS-PATH)
* Configure BGP route-map in from underlay peers (LAN side) to deny WAN routes (potentially learned from another router connected on the WAN) and to mark the prefixes received from the LAN with the SoO
* configured the route-map conn-2-bgp to set the SoO on redistributed routes
* Changed the EVPN-EXPORT-VRF-DEFAULT route-map to match on the SoO to decide what to send over EVPN in VRF default

## How to test

molecule

### User Checklist

- [ ] Also support AutoVPN edges
- [ ] Test in a small lab setup to verify redistribution
- [ ]  Update documentation

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-common-path-group.cfg
```diff
@@ -232,36 +254,64 @@ ip routing vrf IT
 no ip routing vrf MGMT
 ip routing vrf PROD
 !
+ip extcommunity-list ECL-WAN-HA-SOO permit soo 192.168.42.2:511
```
ClausHolbechArista: Maybe just ECL-WAN-SOO

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-common-path-group.cfg
```diff
@@ -232,36 +254,64 @@ ip routing vrf IT
...
+!
 ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
    seq 10 permit 192.168.42.0/24 eq 32
 !
+route-map RM-BGP-UNDERLAY-PEERS-IN deny 20
+   description Deny prefixes from WAN
+   match as-path AP-WAN
```
ClausHolbechArista: AP- IMO is not a good prefix. ASACL?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-default-policy.cfg
```diff
@@ -190,9 +186,10 @@ ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
 !
 route-map RM-CONN-2-BGP permit 10
    match ip address prefix-list PL-LOOPBACKS-EVPN-OVERLAY
+   set extcommunity soo 192.168.42.1:511 additive
 !
-route-map RM-EVPN-EXPORT-VRF-DEFAULT permit 30
-   match ip address prefix-list PL-LOOPBACKS-EVPN-OVERLAY
+route-map RM-EVPN-EXPORT-VRF-DEFAULT permit 10
+   match extcommunity soo 192.168.42.1:511 additive
```
ClausHolbechArista: match additive? Also I believe this command takes the extended community list as argument.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge-no-default-policy.cfg
```diff
@@ -190,9 +186,10 @@ ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
 !
 route-map RM-CONN-2-BGP permit 10
    match ip address prefix-list PL-LOOPBACKS-EVPN-OVERLAY
+   set extcommunity soo 192.168.42.1:511 additive
```
ClausHolbechArista, line 192: Is it possible to set with the community list? So we only have to have the unique value in one place?

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -345,3 +345,40 @@ def should_connect_to_wan_rs(self: SharedUtils, path_groups: list) -> bool:
...
+        """
+        Return True is the current wan_mode is cv-pathfinder and the device is either an edge or a transit device
+        """
+        if self.wan_mode != "cv-pathfinder":
+            return False
+        if self.cv_pathfinder_role == "cv-pathfinder":
+            return False
+        if self.cv_pathfinder_role in ["edge", "transit region"]:
+            return True
+        return False
```
MaheshGSLAB: If we dont add  condition `if self.wan_mode != "cv-pathfinder"` and `if self.cv_pathfinder_role == "cv-pathfinder"` in that case also we are returning False as default so we can remove those conditions.

gmuloc: no because the mode is not the same as the role order is important.

we could remove the 380 `if self.cv_pathfinder_role == "cv-pathfinder":` though

ClausHolbechArista: Simplified the logic a bit.

gmuloc (commented):
Tested with current SoO usage in Lab and it works

gmuloc:
@ClausHolbechArista for the last commit 7807e3d it feels like we are missing some filtering on static routes (as we would not want the ones coming from WAN to be advertised, but we would want the one coming from tenant)

ClausHolbechArista (approved):

PR #3606, opened 2024-02-09
Refactor(eos_cli_config_gen): Insert EOF at the end, if not present in banner string

### LINKED ISSUE #2782: Banner without EOF can break eAPI management
### Issue Summary

According to the AVD [documentation](https://avd.sh/en/stable/roles/eos_cli_config_gen/index.html?h=banner#banners), EOF is required when configuring banner string.
That looks like a "lazy" implementation, as EOF is only required to break banner in the terminal and is redundant in the YAML file. But the problem is that EOS is accepting the configuration when EOF is not specified, but treating everything after the banner string as part of the banner. eAPI configuration comes after and a simple playbook run to add a banner results in broken management connectivity to the switch. It also breaks any `eos_cli` config as it's appended to the end as well.

I'd suggest one of the following fixes:
- remove EOF requirement in AVD docs and always enforce it in the template (Disadvantage: that can affect existing implementations)
- make EOF optional, check if it's present and add if not

_No response_

### How do you run AVD ?

None

### Steps to reproduce

```shell
Add the banner without EOF, run the playbook.
```

### Relevant log output

```shell
None
```

### Contributing Guide

Shivani-gslab:
## Change Summary

Checking if the **EOF** is present in the last line of banner string, if not then inserting it.

## Related Issue(s)

Fixes #2782  

## Component(s) name

## Proposed changes
Update jinja2 template to check if EOF is present, if not insert it.

## How to test

Try giving banner.login and banner.motd value without "EOF" at the end, it should automatically be inserted in the config.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/banners_without_eof.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/banners_without_eof.md
@@ -1,0 +1,165 @@
+# banners_without_eof
+
+## Table of Contents
+
+- [Management](#management)
+  - [Management Interfaces](#management-interfaces)
+  - [Management SSH](#management-ssh)
+  - [Management Console](#management-console)
+  - [Management API HTTP](#management-api-http)
+- [Management Security](#management-security)
+  - [Management Security Summary](#management-security-summary)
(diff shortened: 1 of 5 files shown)


emilarista (approved):
Tested, LGTM!

ClausHolbechArista (approved):

PR #3618, opened 2024-02-12
Doc(eos_cli_config_gen, eos_designs): Consistent descriptions for BGP AS schema fields re asdot notation

ClausHolbechArista:
## Change Summary

- Improve descriptions for BGP AS schema fields re asdot notation

We should not implement convert_types from `float` since it would introduce a risk when the user inputs `65000.100`, we would configure `65000.1`. **We already have that a few places, so to avoid breaking changes, this is kept in this PR.**

## Component(s) name

`arista.avd.eos_designs`

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/tables/router-bgp.md
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/tables/router-bgp.md
@@ -8,7 +8,7 @@
     | Variable | Type | Required | Default | Value Restrictions | Description |
     | -------- | ---- | -------- | ------- | ------------------ | ----------- |
     | [<samp>router_bgp</samp>](## "router_bgp") | Dictionary |  |  |  |  |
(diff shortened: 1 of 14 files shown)


carl-baillargeon (commented):
The added description is missing under network service: `<network_services_keys.name>.[].vrfs.[].bgp_peers.[].remote_as`.

ClausHolbechArista:
> The added description is missing under network service: `<network_services_keys.name>.[].vrfs.[].bgp_peers.[].remote_as`.

The scope was only `eos_cli_config_gen`, but I have fixed it for `eos_designs` as well now, and I actually found a few places where we converted from `float`. So since I change that I have changed this to a fix instead of doc.

carl-baillargeon (commented):
The `*must*` is not rendered properly. I _think_ it was fine when I first reviewed.

image

carlbuchmann:
> The `*must*` is not rendered properly. I _think_ it was fine when I first reviewed.
> 
> image

That is to be expected in the YAML output. It is italic in the MD table as expected.But perhaps we want bold?

carlbuchmann (approved):
LGTM, small comment on format

chetryan (approved):
LGTM - much clearer descriptions than before.

PR #3623, opened 2024-02-14
Feat(eos_designs): Option to define evpn_vlan_bundle for SVIs

### LINKED ISSUE #3297: It is better to have a possibility to configure one vlan-aware-bundle for VLANs which belonged into different VRFs
### Enhancement summary

For now, the vlan-aware-bundle is only possible to include VLANs which belong to the same VRF. But in real life, SVIs which belonging to different VRFs, their VLANs could be in same evpn instance (mac-vrf) with same RD/RT (which could also be defined by network administrator). To intercommunication with this situation and also flexible usage. I suggest you could also add this feature. From the capability of the switch and principle, it should work.

### Use case example

The configs would be like this:

In BGP sub-view:
vlan-aware-bundle service-vlans
      rd xxxx:yyyy
      route-target both xxxx:yyyy
      redistribute learned
      vlan 1007-1008,1301,1401,1496-1499,1780-1789,1799,2000-2001,2005-2008

For SVIs configs:
interface Vlan1799
   mtu 9100
   vrf OS
   arp gratuitous accept
   ip address virtual xx.xx.xx.xx/zz
!
interface Vlan1007
   mtu 9100
   vrf INET
   arp gratuitous accept
   ip address virtual aa.bb.cc.dd/ee

Maybe "evpn_vlan_bundles" dictionary (in 4.4) could be referred by Network services VRF SVIs configuration (like Network services L2 VLANs configuration)

_No response_

### Additional context

_No response_

### Contributing Guide

bjmeuer:
## Change Summary

Add the possibilty to define a vlan-aware-bundle name which should be used for one or several SVIs, even if the SVIs are defined in different VRFs.

## Related Issue(s)

Fixes #3297 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Adding the option 'evpn_vlan_bundle' to the 'svis' section. It behaves the same way as it is for 'l2vlans' today.

## How to test
```
tenants:
  - name: SIT
    vlan_aware_bundle_number_base: 300
    mac_vrf_vni_base: 20000
    vrfs:
      - name: SIT_VRF
        vrf_vni: 789654
        svis:
          - id: 1008
            name: SVI_1
            enabled: true
            evpn_vlan_bundle: bundle3
            tags: ['sit']
      - name: SIT2
        vrf_vni: 654789
        svis:
          - id: 1011
            name: SVI_4
            enabled: true
            evpn_vlan_bundle: bundle3
            tags: ['sit']
```

### User Checklist
- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
```diff
@@ -49,6 +49,44 @@ tenants:
...
+            name: SVI_4
+            enabled: true
+            evpn_vlan_bundle: bundle3
+            tags: ['sit']
+          - id: 1012
+            name: SVI_5
+            enabled: true
+            evpn_vlan_bundle: bundle3
+            tags: ['sit']
+          - id: 1013
```
ClausHolbechArista, line 95: Let's try having one VRF test without any SVIs falling back to the default created bundle. We should not create that bundle at all if there are no vlans for it.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
```diff
@@ -49,6 +49,44 @@ tenants:
...
+        svis:
+          - id: 1011
+            name: SVI_4
+            enabled: true
+            evpn_vlan_bundle: bundle3
+            tags: ['sit']
+          - id: 1012
+            name: SVI_5
+            enabled: true
+            evpn_vlan_bundle: bundle3
```
ClausHolbechArista, line 93: please try to combine one of them with l2vlans and svis in the same bundle.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
bjmeuer: will add a test for this

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
bjmeuer: this will not work, as of now the l2vlan bundles and svi bundles can't overlap. The duplicate check should kick in here.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
bjmeuer: test added

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml
ClausHolbechArista: We should figure out how to solve this. Maybe move the addition of the bundles to the end.

bjmeuer: will figure that out later this week

bjmeuer: now l2vlans and svis can share a defined evpn_vlan_bundle

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -428,7 +465,29 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                        evpn_vlan_bundle = self._get_evpn_vlan_bundle(svis[0], bundle_name)
+
+                        if bundle_name in l2vlan_svi_vlan_aware_bundles:
+                            l2vlan_svi_vlan_aware_bundles[bundle_name]["l2vlan_svis"].extend(svis)
+                        else:
+                            l2vlan_svi_vlan_aware_bundles[bundle_name] = {"evpn_vlan_bundle": evpn_vlan_bundle, "l2vlan_svis": svis}
+
+                        # Remove the SVIs which have an evpn_vlan_bundle from the VRF list
+                        for bundle_svi in svis:
+                            vrf["svis"].remove(bundle_svi)
```
ClausHolbechArista: Why are you removing SVIs here? This will modify the original data, and should be avoided. It would be better to give a default value to the 
`_get_vlan_aware_bundle_name_tuple()` function, so you can give it the vrf-specific bundle. This way we don't have to duplicate the logic and parse through the lists multiple times.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -490,20 +526,40 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                        "BGP VLAN-Aware Bundles defined under network services. A common reason is that an 'l2vlan' or 'svi' name overlaps with an"
+                        " 'evpn_vlan_bundle' name"
                     ),
                     context_keys=["name"],
                 )
 
         return bundles or None
 
-    def _router_bgp_vlan_aware_bundles_vrf(self, vrf, tenant) -> dict | None:
+    def _router_bgp_vlan_aware_bundles_vrf(self, vrf, tenant) -> list | None:
```
ClausHolbechArista:
```suggestion
    def _router_bgp_vlan_aware_bundles_vrf(self, vrf, tenant) -> dict | None:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -490,20 +526,40 @@ def _router_bgp_vlan_aware_bundles(self) -> list | None:
...
+                            "BGP VLAN-Aware Bundles defined under network services. A common reason is that an 'l2vlan' name overlaps with an"
+                            " 'evpn_vlan_bundle' name"
+                        ),
+                        context_keys=["name"],
+                    )
+
+            # L2VLANs and SVIs which have an evpn_vlan_bundle defined
+            for bundle_name, bundle_dict in l2vlan_svi_vlan_aware_bundles.items():
+                evpn_vlan_bundle = bundle_dict["evpn_vlan_bundle"]
+                l2vlans_svis = bundle_dict["l2vlan_svis"]
```
ClausHolbechArista, line 552: I think it would be simpler to maintain and understand if we had one dict of bundles to create like `{<bundle_name>: [<vlan number>, <vlan number> ...]}` which is populated during parsing of all vlans and svis. Then at the end here you can just parse that dict a create the bundles.

ClausHolbechArista (approved):

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -406,14 +406,61 @@ def _router_bgp_vlans_vlan(self, vlan, tenant, vrf) -> dict | None:
...
+        """
+        if vlan.get("evpn_vlan_bundle") is not None:
+            return (str(vlan.get("evpn_vlan_bundle")), True)
+        return ("", False)
+
+    def _get_evpn_vlan_bundle(self, vlan: dict, bundle_name: str) -> dict:
+        """
+        Return an evpn_vlan_bundle dict if it exists, else raise an exception.
+        """
+        if (evpn_vlan_bundle := get_item(self._hostvars["evpn_vlan_bundles"], "name", bundle_name)) is None:
```
carl-baillargeon: `self._hostvars["evpn_vlan_bundles"]` will throw a `KeyError` if `evpn_vlan_bundles` does not exist. Should we handle this as well and include it in the `raise AristaAvdMissingVariableError` message below?

ClausHolbechArista: Good catch. Thanks! I will fix this.

ClausHolbechArista: Fixed in https://github.com/aristanetworks/avd/pull/3623/commits/ed5bdd66df71aed06d8963386f34f191d9021abd

carl-baillargeon (approved):
LGTM

PR #3648, opened 2024-02-21
Feat(eos_designs): Add default_preference and excluded for wan_path_groups

ayushmittal-arista:
## Change Summary

Whenever AVD automatically generates a policy including a path group (e.g. for Control Plane), currently we use preference 1. Give an option to user in wan_path_gorups to specify default_preference which will be used instead of 1.

Also provide an exclude_from_auto_lb_policies parameter so user can exclude this path group from auto generated load balance policies if need be

## Related Issue(s)

Fixes #
## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add a knob to define default preference used when no preference is defined under `wan_virtual_topologies.polices` or when AVD is auto-generating a policy.

`excluded` means that the path-group is not inserted.

```
wan_path_groups
  - name: <str>
    default_preference : str < preferred | alternate | excluded | priority as integer>
```

## How to test
molecule

### User Checklist

- N/A

### Repository Checklist

- [ ] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
@@ -23,6 +23,8 @@ router path-selection
    !
    load-balance policy LB-DEFAULT-AUTOVPN-POLICY-CONTROL-PLANE
       path-group INET
+      path-group LTE
+      path-group MPLS
    !
    load-balance policy LB-DEFAULT-AUTOVPN-POLICY-IT
       path-group MPLS
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
@@ -22,9 +22,8 @@ router path-selection
    path-group MPLS id 100
    !
    load-balance policy LB-DEFAULT-AUTOVPN-POLICY-CONTROL-PLANE
-      path-group INET
-      path-group LTE
       path-group MPLS
+      path-group INET priority 42
    !
    load-balance policy LB-DEFAULT-AUTOVPN-POLICY-IT
       path-group MPLS
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
@@ -117,6 +117,8 @@ router path-selection
    peer dynamic source stun
    tcp mss ceiling ipv4 ingress
    !
+   path-group AWS id 105
+   !
    path-group Equinix id 103
    !
    path-group INET id 101
@@ -143,25 +145,32 @@ router path-selection
       path-group INET priority 2
    !
    load-balance policy LB-DEFAULT-AVT-POLICY-CONTROL-PLANE
+      path-group AWS
       path-group INET
       path-group LAN_HA
       path-group MPLS
+      path-group Equinix priority 2
+      path-group Satellite priority 2
    !
    load-balance policy LB-DEFAULT-AVT-POLICY-DEFAULT
-      path-group Equinix
       path-group INET
       path-group LAN_HA
-      path-group MPLS priority 42
+      path-group Equinix priority 2
+      path-group MPLS priority 4223
    !
    load-balance policy LB-DEFAULT-AVT-POLICY-VIDEO
+      path-group Equinix
       path-group INET
       path-group LAN_HA
       path-group MPLS
    !
    load-balance policy LB-DEFAULT-POLICY-DEFAULT
+      path-group AWS
       path-group INET
       path-group LAN_HA
       path-group MPLS
+      path-group Equinix priority 2
+      path-group Satellite priority 2
    !
    load-balance policy LB-PROD-AVT-POLICY-DEFAULT
       path-group INET
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
@@ -93,7 +93,7 @@ wan_router:
               wan_carrier: Comcast-5G
               wan_circuit_id: AF830
               ip_address: 172.20.20.20/31
-              connected_to_pathfinder: False
+              connected_to_pathfinder: false
         - name: cv-pathfinder-edge-no-common-path-group
           id: 2
           uplink_switch_interfaces: [Ethernet2]
@@ -108,7 +108,7 @@ wan_router:
               wan_circuit_id: 212
               dhcp_accept_default_route: true
               ip_address: dhcp
-              connected_to_pathfinder: False
+              connected_to_pathfinder: false
     # SITE_HA_ENABLED
     - group: Site423
       cv_pathfinder_region: AVD_Land_West
@@ -255,7 +255,7 @@ wan_path_groups:
   - name: LTE
     id: 102
     # The expectation here is that LTE shouldn't be present in the default lb policy.
-    default_preference: excluded
+    excluded_from_default_policy: true
   - name: Equinix
     id: 103
     default_preference: alternate
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
@@ -101,6 +101,12 @@ wan_router:
               wan_circuit_id: S511
               dhcp_accept_default_route: true
               ip_address: dhcp
+            - name: Ethernet2
+              wan_carrier: AWS-1
+              wan_circuit_id: 212
+              dhcp_accept_default_route: true
+              ip_address: dhcp
+              connected_to_pathfinder: False
     # SITE_HA_ENABLED
     - group: Site423
       cv_pathfinder_region: AVD_Land_West
@@ -241,16 +247,21 @@ wan_rr:
 wan_path_groups:
   - name: MPLS
     ipsec: false
-    # TODO remove one once auto-id is implemented - for now required in schema
     id: 100
   - name: INET
     id: 101
   - name: LTE
     id: 102
+    # The expectation here is that LTE shouldn't be present in the default lb policy.
+    default_preference: excluded
   - name: Equinix
     id: 103
+    default_preference: alternate
   - name: Satellite
     id: 104
+    default_preference: alternate
+  - name: AWS
+    id: 105
 
 wan_carriers:
   - name: Comcast
@@ -273,6 +284,8 @@ wan_carriers:
     path_group: LTE
   - name: Inmrasat
     path_group: Satellite
+  - name: AWS-1
+    path_group: AWS
 
 tenants:
   - name: TenantA
@@ -373,11 +386,13 @@ wan_virtual_topologies:
           # No prefence to check the default.
           - names: [INET, Equinix]
           - names: [MPLS]
-            preference: 42
+            preference: 4223
       application_virtual_topologies:
         - application_profile: VIDEO
           path_groups:
-            - names: [MPLS, INET]
+            # The expectation here is that Equinix will be configured with priority 1/preferred,
+            # since we are overriding the default_preference value (2/alternate).
+            - names: [MPLS, INET, Equinix]
               preference: preferred
           id: 3
     - name: TRANSIT-AVT-POLICY
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/UPLINK_LAN_TESTS.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/UPLINK_LAN_TESTS.yml
@@ -26,9 +26,19 @@ wan_router: # dynamic_key: node_type
       id: 1
       uplink_switch_interfaces: [Ethernet1]
       uplink_native_vlan: 10
+      l3_interfaces:
+        - name: Ethernet1
+          wan_carrier: Comcast
+          wan_circuit_id: 999
+          ip_address: 10.9.9.9/31
     - name: uplink_lan_wan_router2
       uplink_switch_interfaces: [Ethernet2]
       id: 2
+      l3_interfaces:
+        - name: Ethernet1
+          wan_carrier: Comcast
+          wan_circuit_id: 999
+          ip_address: 10.9.9.1/31
 
 tenants: # dynamic_key: network_services
   - name: TEST
@@ -74,8 +84,14 @@ wan_ipsec_profiles:
 wan_mode: autovpn
 
 wan_route_servers: []
-wan_path_groups: []
-wan_carriers: []
+wan_path_groups:
+  - name: INET
+    ipsec: false
+    # TODO remove one once auto-id is implemented - for now required in schema
+    id: 100
+wan_carriers:
+  - name: Comcast
+    path_group: INET
 bgp_peer_groups:
   wan_overlay_peers:
     listen_range_prefixes: [0.0.0.0/0]
--- a/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
@@ -88,6 +88,10 @@ class WanMixin:
             if get(interface, "wan_carrier") is not None:
                 wan_interfaces.append(interface)
 
+        if not wan_interfaces:
+            raise AristaAvdError(
+                "At least one WAN interface must be configured on a WAN router. Add WAN interfaces under `l3_interfaces` node setting with `wan_carrier` set."
+            )
         return wan_interfaces
 
     @cached_property
@@ -131,9 +135,44 @@ class WanMixin:
 
         return list(local_carriers_dict.values())
 
+    def path_group_preference_to_eos_priority(self, path_group_preference: int | str, context_path: str) -> int:
+        """
+        Convert "preferred" to 1 and "alternate" to 2. Everything else is returned as is.
+
+        Arguments:
+        ----------
+        path_group_preference (str|int): The value of the preference key to be converted. It must be either "preferred", "alternate" or an integer.
+        context_path (str): Input path context for the error message.
+        """
+        if path_group_preference == "preferred":
+            return 1
+        if path_group_preference == "alternate":
+            return 2
+
+        failedConversion = False
+        try:
+            priority = int(path_group_preference)
+        except ValueError:
+            failedConversion = True
+
+        if failedConversion or not (1 <= priority <= 65535):
+            raise AristaAvdError(
+                f"Invalid value '{path_group_preference}' for Path-Group preference - should be either 'preferred', "
+                f"'alternate' or an integer[1-65535] for {context_path}."
+            )
+
+        return priority
+
     @cached_property
     def wan_path_groups(self: SharedUtils) -> list:
-        return get(self.hostvars, "wan_path_groups", required=True)
+        path_groups = get(self.hostvars, "wan_path_groups", required=True)
+        for path_group in path_groups:
+            default_preference = get(path_group, "default_preference", default=1)
+            priority = 0
+            if default_preference != "excluded":
+                priority = self.path_group_preference_to_eos_priority(default_preference, f"wan_path_groups[{path_group['name']}]")
+            path_group["default_priority"] = priority
+        return path_groups
 
     @cached_property
     def wan_local_path_groups(self: SharedUtils) -> list:
--- a/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
@@ -135,42 +135,15 @@ class WanMixin:
 
         return list(local_carriers_dict.values())
 
-    def path_group_preference_to_eos_priority(self, path_group_preference: int | str, context_path: str) -> int | None:
-        """
-        Convert "preferred" to 1 and "alternate" to 2 and "excluded" to None. Everything else is returned as is.
-
-        Arguments:
-        ----------
-        path_group_preference (str|int): The value of the preference key to be converted. It must be either "preferred", "alternate" or an integer.
-        context_path (str): Input path context for the error message.
-        """
-        if path_group_preference == "preferred":
-            return 1
-        if path_group_preference == "alternate":
-            return 2
-        if path_group_preference == "excluded":
-            return None
-
-        failedConversion = False
-        try:
-            priority = int(path_group_preference)
-        except ValueError:
-            failedConversion = True
-
-        if failedConversion or not (1 <= priority <= 65535):
-            raise AristaAvdError(
-                f"Invalid value '{path_group_preference}' for Path-Group preference - should be either 'preferred', "
-                f"'alternate' or an integer[1-65535] for {context_path}."
-            )
-
-        return priority
-
     @cached_property
     def wan_path_groups(self: SharedUtils) -> list:
+        """
+        List of path-groups defined in the top level key `wan_path_groups`
+        Updating default preference for each path-group to 'preferred' if not set.
+        """
         path_groups = get(self.hostvars, "wan_path_groups", required=True)
         for path_group in path_groups:
-            default_preference = get(path_group, "default_preference", default=1)
-            path_group["default_priority"] = self.path_group_preference_to_eos_priority(default_preference, f"wan_path_groups[{path_group['name']}]")
+            path_group["default_preference"] = get(path_group, "default_preference", default="preferred")
         return path_groups
 
     @cached_property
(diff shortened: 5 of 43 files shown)


ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,20 @@ keys:
               local:
                 type: str
                 description: Optional, if not set, the path-group `name` is used as local.
+        exclude_from_auto_lb_policies:
+          type: bool
+          description: |-
+            When set to `true`, the path-group is not used for auto-generated load-balace policies.
+        default_preference:
```
gmuloc, line 52: please add an option here callled `excluded` and remove `exclude_from_auto_lb_policies`

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,20 @@ keys:
...
+          type: str
+          convert_types:
+            - int
+          description: |-
+            Required when `exclude_from_auto_lb_policies` is not set.
```
gmuloc: Remove the `required` and document at the end what `excluded` means

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,20 @@ keys:
...
+            Preference to be used in auto-generated load balance policies.
+
+            Valid values are 1-255 | preferred | alternate.
```
gmuloc:
```suggestion
            Valid values are 1-255 | preferred | alternate | excluded.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -268,6 +268,29 @@ def _local_path_groups_connected_to_pathfinder(self) -> list:
...
+
+            default_preference = get(path_group, "default_preference", 1)
+            context_path = f"wan_path_groups[{path_group_name}]"
+            priority = self._path_group_preference_to_eos_priority(default_preference, context_path)
+            if not path_group.get("exclude_from_auto_lb_policies"):
+                if priority not in filtered_path_groups:
+                    filtered_path_groups[priority] = []
+                filtered_path_groups[priority].append(path_group_name)
+
+        return [{"names": names, "preference": priority} for priority, names in filtered_path_groups.items()]
```
gmuloc: refactor this and see if this is really needed - maybe can just include default preference in `self.shared_utils.wan_local_path_groups` instead so that you can use it when required in the existing functions

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
```diff
@@ -209,10 +213,12 @@ wan_path_groups:
     ipsec: false
     # TODO remove one once auto-id is implemented - for now required in schema
     id: 100
+    default_preference: alternate
   - name: INET
     id: 101
   - name: LTE
     id: 102
+    exclude_from_auto_lb_policies: True
```
gmuloc: replace with `default_preference: excluded`

gmuloc (changes requested):

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
ayushmittal-arista: This looks good to me, `wan_local_path_groups` has `default_preference` field if you want to fetch it. 

The question is wether you want to use this default_preference for manual policies as well ( but without preference specified ) or just for auto generated policies.

If just for auto generated policies this seems like a good place. 

I thought that for an auto policy we will have a list of path groups, we call this function, which gives us policy and then `_generate_wan_load_balance_policy` takes that policy.

The other option is to have `_generate_wan_load_balance_policy` take some parameter, if the policy is auto or not, and then pick up the preference from wan_local_path_groups, and then do magic, but because the input parameter is policy for it, it won't look clean, as now even within a policy row, you will need to check preference. 

Let me know if you have something else in mind.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,16 @@ keys:
...
+          convert_types:
+            - int
+          description: |-
+            Preference to be used in auto-generated load balance policies.
+
+            Valid values are 1-255 | preferred | alternate | excluded.
+
+            preferred is converted to priority 1.
+            alternate is converted to priority 2.
+            If excluded, path group is not considered for automatically generated policies
```
gmuloc:
```suggestion
            excluded means that the path-group is not considered for automatically generated policies.
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
gmuloc: > The question is wether you want to use this default_preference for manual policies as well ( but without preference specified ) or just for auto generated policies.

That was not the idea, the idea is that by default we would be using `1` (`preferred`) but thinking further about this it could actually make sense to use the default policy set under wan_path_groups. @ClausHolbechArista any opinion?

Even if this is for autogenerated policies, is it not possible to avoid looping though all the path_groups  here but rather using watever is is in `wan_local_path_groups` when the automatic policy is being used?

ayushmittal-arista: > That was not the idea, the idea is that by default we would be using 1 (preferred) but thinking further about this it could actually make sense to use the default policy set under wan_path_groups. @ClausHolbechArista any opinion?

This doesn't properly suit the inputs structure, 
as someone could in policy specify 
- names : [P1, P2]

where P1 has default pref of 1 and P2 has 2, we can obviously now throw error, but seems confusing behaviour.

---

> Even if this is for autogenerated policies, is it not possible to avoid looping though all the path_groups here but rather using watever is is in wan_local_path_groups when the automatic policy is being used?

That is what we are doing, no ? it's a helper function which iterates over path groups which are to be used in load balance policy and then reads the priority from wan_local_path_groups, and returns.

Doing it in `_generate_wan_load_balance_policy ` will increase code complexity further.

gmuloc: > This doesn't properly suit the inputs structure,
> as someone could in policy specify
> 
> names : [P1, P2]
> where P1 has default pref of 1 and P2 has 2, we can obviously now throw error, but seems confusing behaviour.

good point so probably a bad idea for non automatically generated policies

ayushmittal-arista: Refactored as discussed.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
ayushmittal-arista: Done.

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -211,7 +211,7 @@ def _wan_policy_key(self) -> str:
         """
         return "policy" if self.shared_utils.wan_mode == "cv-pathfinder" else "path_selection_policy"
 
-    def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context_path: str) -> dict:
+    def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context_path: str, ignore_priority_1: bool = False) -> dict:
```
gmuloc: @ayushmittal-arista I had to add this to rebase for now but we need to figure this out

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
gmuloc: @ayushmittal-arista - we discussed with @ClausHolbechArista and actually we are coming back on this

>The question is wether you want to use this default_preference for manual policies as well ( but without preference specified ) or just for auto generated policies.

Let's implement this also in this PR and detail that if you don;t specify a preference for an entry, each path-group in the list will pick the default preference you have implemented (and yes, they may not be the same)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
```diff
@@ -140,7 +140,7 @@ router path-selection
    load-balance policy LB-CONTROL-PLANE-PROFILE
       path-group INET
       path-group LAN_HA
-      path-group MPLS
+      path-group MPLS priority 2
```
gmuloc: for Control PLane for pathfinder we should put all the path-groups

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
ayushmittal-arista, line 1: Not sure if this was intentional, but a bug here. 
Previously even when there were no path groups, it would take as preference 1, and pass the validation. 
But now it fails.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/autovpn-rr1.cfg
```diff
@@ -23,6 +23,8 @@ router path-selection
    !
    load-balance policy LB-DEFAULT-AUTOVPN-POLICY-CONTROL-PLANE
       path-group INET
+      path-group LTE
+      path-group MPLS
```
gmuloc: This is actually fixing a bug where we want the full view of the policy on pathfinders! thanks

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
```diff
@@ -143,25 +145,32 @@ router path-selection
       path-group INET priority 2
    !
    load-balance policy LB-DEFAULT-AVT-POLICY-CONTROL-PLANE
+      path-group AWS
       path-group INET
       path-group LAN_HA
       path-group MPLS
+      path-group Equinix priority 2
+      path-group Satellite priority 2
```
gmuloc, line 156: same here!

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
```diff
@@ -247,10 +253,15 @@ wan_path_groups:
     id: 101
   - name: LTE
     id: 102
+    default_preference: excluded
```
gmuloc: lets add a comment in the test to state we don't want this in default policy

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
```diff
@@ -377,7 +390,7 @@ wan_virtual_topologies:
       application_virtual_topologies:
         - application_profile: VIDEO
           path_groups:
-            - names: [MPLS, INET]
+            - names: [MPLS, INET, Equinix]
```
gmuloc, line 404: let's add here that the expectation is that Equinix is configured with pref 1 and not the defualt policy of 2 in the output (good test!)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/UPLINK_LAN_TESTS.yml
```diff
@@ -74,8 +84,14 @@ wan_ipsec_profiles:
 wan_mode: autovpn
 
 wan_route_servers: []
-wan_path_groups: []
-wan_carriers: []
+wan_path_groups:
+  - name: INET
+    ipsec: false
+    # TODO remove one once auto-id is implemented - for now required in schema
```
gmuloc, line 95:
```suggestion
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -88,6 +88,8 @@ def wan_interfaces(self: SharedUtils) -> list:
             if get(interface, "wan_carrier") is not None:
                 wan_interfaces.append(interface)
 
+        if not wan_interfaces:
+            raise AristaAvdError("No wan interfaces found for the node. - should have altleast one l3_interface with wan_carrier configured.")
```
gmuloc:
```suggestion
            raise AristaAvdError("At least one WAN interface must be configured on a WAN router. Add WAN interfaces under `l3_interfaces` node setting with `wan_carrier` set .")
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -131,9 +133,41 @@ def wan_local_carriers(self: SharedUtils) -> list:
 
         return list(local_carriers_dict.values())
 
+    def _path_group_preference_to_eos_priority(self, path_group_preference: int | str, context_path: str) -> int:
```
gmuloc:
```suggestion
    def path_group_preference_to_eos_priority(self, path_group_preference: int | str, context_path: str) -> int:
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -131,9 +133,41 @@ def wan_local_carriers(self: SharedUtils) -> list:
...
+            default_preference = get(path_group, "default_preference", default=1)
+            priority = 0
+            if default_preference != "excluded":
+                priority = self._path_group_preference_to_eos_priority(default_preference, f"wan_path_groups[{path_group['name']}]")
+            path_group["default_priority"] = priority
+        return path_groups
+
+    @cached_property
+    def wan_path_groups_dict(self: SharedUtils) -> dict:
+        return {path_group["name"]: path_group for path_group in self.wan_path_groups}
```
gmuloc: lets not use this

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -206,7 +206,7 @@ def _wan_control_plane_application_profile(self) -> str:
         """
         return "APP-PROFILE-CONTROL-PLANE"
 
-    def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context_path: str) -> dict:
+    def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context_path: str, filter_path_groups: list | None = None) -> dict:
```
gmuloc: doc string update

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -216,7 +216,9 @@ def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context
         input_dict (dict): The dictionary containing the list of path-groups and their preference.
         context_path (str): Key used for context for error messages.
         """
-        wan_local_path_group_names = [path_group["name"] for path_group in self.shared_utils.wan_local_path_groups]
+        if not filter_path_groups:
```
gmuloc:
```suggestion
        if not filtered_path_groups:
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -236,15 +238,19 @@ def _generate_wan_load_balance_policy(self, name: str, input_dict: dict, context
         for policy_entry in policy_entries:
             # TODO check if it cannot be optimized further in shared_utils or validated in a global fashion - maybe
             # schema? check that the LB policy has at least one prio 1 / preferred EVEN if the path group is not configured.
-            if (
-                priority := self._path_group_preference_to_eos_priority(
-                    get(policy_entry, "preference", default=1), f"{context_path}[{policy_entry.get('names')}]"
-                )
-            ) == 1:
-                at_least_one_priority_1_found = True
+            policy_priority = self.shared_utils._path_group_preference_to_eos_priority(
```
gmuloc:
```suggestion
            policy_priority = self.shared_utils.path_group_preference_to_eos_priority(
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/utils.py
```diff
@@ -303,7 +288,7 @@ def _wan_load_balance_policies(self) -> list:
         control_plane_virtual_topology = get(
             self._hostvars,
             "wan_virtual_topologies.control_plane_virtual_topology",
-            default={"path_groups": [{"names": self._local_path_groups_connected_to_pathfinder}]},
+            default={"path_groups": [{"names": self.shared_utils.wan_path_groups_dict.keys()}]},
```
gmuloc: refactor to just get the names from the shared_utils list

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,16 @@ keys:
...
+          convert_types:
+            - int
+          description: |-
+            Preference to be used in auto-generated load balance policies.
+
+            Valid values are 1-255 | preferred | alternate | excluded.
+
+            preferred is converted to priority 1.
+            alternate is converted to priority 2.
+            excluded means that the path-group for automatically generated policies.
```
gmuloc: missing something

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
ClausHolbechArista:
```suggestion
    # Excluding this pathgroup from auto-generated default policy.
    default_preference: excluded
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_PATHFINDER_TESTS.yml
ClausHolbechArista, line 404:
```suggestion
              # The expectation here is that Equinix will be configured with priority 1/preferred,
              # since we are overriding the default_preference value (2/alternate).
            - names: [MPLS, INET, Equinix]
```

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -131,9 +133,41 @@ def wan_local_carriers(self: SharedUtils) -> list:
...
+        ----------
+        path_group_preference (str|int): The value of the preference key to be converted. It must be either "preferred", "alternate" or an integer.
+        context_path (str): Input path context for the error message.
+        """
+        if path_group_preference == "preferred":
+            return 1
+        if path_group_preference == "alternate":
+            return 2
+        try:
+            return int(path_group_preference)
```
ClausHolbechArista: While we are at it we should check that the value is between 1 and 255. The schema is not checking this, since it is a string.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,16 @@ keys:
...
                 type: str
                 description: Optional, if not set, the path-group `name` is used as local.
+        default_preference:
+          type: str
+          convert_types:
+            - int
+          description: |-
+            Preference to be used in auto-generated load balance policies.
+
+            Valid values are 1-255 | preferred | alternate | excluded.
```
ClausHolbechArista:
```suggestion
            Valid values are 1-255 | "preferred" | "alternate" | "excluded".
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/wan_path_groups.schema.yml
```diff
@@ -49,3 +49,16 @@ keys:
               local:
                 type: str
                 description: Optional, if not set, the path-group `name` is used as local.
+        default_preference:
+          type: str
+          convert_types:
+            - int
+          description: |-
+            Preference to be used in auto-generated load balance policies.
```
ClausHolbechArista: This would also be the default preference if a preference is not given in a policy.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
ayushmittal-arista: Why is it 1-255 here, when in eos cli gen it's 1 - 65535. 

Also do we need to validate it here, since cli gen will validate it ?

gmuloc: the limitiation is checked in `eos_cil_config_gen` 

and the range is

```
pf2(config-load-balance-policy-LB-CONTROL-PLANE-PROFILE)#path-group PPP priority ?
  <1-65535>  Priority of this group
```

ayushmittal-arista: I copied over the range 1-255 from the policy schema. Added the check in this function as well, now.

ClausHolbechArista: If we support 65535 we should indeed change that all over.

Re validating inputs, we should make sure eos_designs produce good output when possible. And part of that is also validating the inputs. From a user perspective it does not make much sense to run through minutes of build time, just to get an input error at the end.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/wan.py
```diff
@@ -131,9 +135,43 @@ def wan_local_carriers(self: SharedUtils) -> list:
...
+        Arguments:
+        ----------
+        path_group_preference (str|int): The value of the preference key to be converted. It must be either "preferred", "alternate" or an integer.
+        context_path (str): Input path context for the error message.
+        """
+        if path_group_preference == "preferred":
+            return 1
+        if path_group_preference == "alternate":
+            return 2
+        if path_group_preference == "excluded":
```
gmuloc: So here we are going a bit back and forth on this with @ayushmittal-arista 

basically adding "excluded" here feels more natural but it also brings the issue that the funciton is used to verify both
`wan_path_groups.default_preference` AND `wan_virtual_topologies.policies[].path_groups[].preference`.
The first one allows for `excluded` but not the second one. And with this implementation it allows `excluded` for the second one.

@ClausHolbechArista for discussion

gmuloc: from @ClausHolbechArista 

> Pathgroup LTE - it is specifically set in the VIDEO policy but because the default preference is excluded it is now showing up.
> so for excluded I think we need to only respect it for auto generated policies, and raise on other policies

error message: Path group blah has a default pref of excluded so a preference must be specified under the policy for this pathgroup.

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_virtual_topology.schema.yml
```diff
@@ -55,7 +55,7 @@ $defs:
               convert_types:
                 - int
               description: |-
-                Valid values are 1-255 | preferred | alternate.
+                Valid values are 1-65535 | preferred | alternate.
```
ClausHolbechArista:
```suggestion
                Valid values are 1-65535 | "preferred" | "alternate".
```

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_path_selection.py
```diff
@@ -37,14 +38,180 @@ def router_path_selection(self) -> dict | None:
...
+
+        for policy_entry in policy_entries:
+            policy_entry_priority = None
+            if preference := get(policy_entry, "preference"):
+                policy_entry_priority = self._path_group_preference_to_eos_priority(preference, f"{context_path}[{policy_entry.get('names')}]")
+
+            for path_group_name in policy_entry.get("names"):
+                if (priority := policy_entry_priority) is None:
+                    # No preference defined at the policy level, need to retrieve the default preference
+                    # TODO error message
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/roles/eos_designs/schemas/schema_fragments/defs_virtual_topology.schema.yml
```diff
@@ -55,8 +55,10 @@ $defs:
...
+                Valid values are 1-65535 | "preferred" | "alternate".
 
-                preferred is converted to priority 1.
-                alternate is converted to priority 2.
-              default: "preferred"
+                "preferred" is converted to priority 1.
+                "alternate" is converted to priority 2.
+
+                If not set, each path-group in `names` will be attributed its `default_preference`.
+                In that casem the `default_preference` cannot be "excluded".
```
ClausHolbechArista:
```suggestion
```

gmuloc: [3e92169](https://github.com/aristanetworks/avd/pull/3648/commits/3e921697663c79594fc4ae23c792a3a5d0548c20)

ClausHolbechArista (approved):

gmuloc (approved):
LGTM - thanks @ayushmittal-arista for tackling this. It was a bumpy road

PR #3667, opened 2024-02-27
Feat(eos_designs): Update platform.py with 7358X4

colinmacgiolla:
## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Set the default management interface for the 7358X4 series to `Management1/1`

## How to test
Set the platform type to `7358X4` - the management interface should render as `Management1/1`

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -95,6 +95,14 @@
             "non_mlag": 1020,
         },
     },
+    {
```
ClausHolbechArista, line 98: Please check the various options under `feature_support` for platform settings and anything not following the defaults. At least I would expect `queue_monitor_length_notify: false` is needed (which should also be the case for all other X-series but is apparently not done)

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
```diff
@@ -95,6 +95,14 @@
...
+        "platforms": ["7358X4"],
+        "management_interface": "Management1/1",
```
ClausHolbechArista, line 100: Seems weird that 7368X4 has Ma0 but this one has Ma1/1. I trust you are right, but then we may need to fix the others. Can you please confirm?

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
ClausHolbechArista: Please also update the schema which will reflect in docs.

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
colinmacgiolla: Yeah, I'm wondering about that - trying to hunt down a 7368 in the lab today to check. I have a 7358, hence the PR :)

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
colinmacgiolla: Will verify later today. Interestingly `bgp_update_wait_install` is not currently supported, so good point 👍

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
ClausHolbechArista: Turns out that Ma0 is _not_ optimal for these platforms. It is _not_ mapped to Ma1/2 and it is somehow dependent on the Strata agent to load, so it has a 5 min delay on startup.
We will stick with ma1/1. Thank you for listening :)

carl-baillargeon (commented):
I don't see the new platform when I click on `+` beside `platform_settings` in the YAML documentation.

carl-baillargeon:
> I don't see the new platform when I click on `+` beside `platform_settings` in the YAML documentation.

Right, you need to update `platform_settings.schema.yml` and re-run pre-commit

EDIT: Looks like Claus already commented about updating the schema... Sorry for the noise :)

ansible_collections/arista/avd/plugins/plugin_utils/eos_designs_shared_utils/platform.py
ClausHolbechArista: @colinmacgiolla you still need to update the default value in the schema so it will be reflected in the docs.

ClausHolbechArista (approved):

carl-baillargeon (approved):
LGTM.

PR #3704, opened 2024-03-06
Doc(eos_cli_config_gen, eos_designs): Add missing schema files

ClausHolbechArista:
Several keys where not shown in docs because they were never added to the schema. The schema fragments were not having `.schema.yml` suffix, so they got ignored.


python-avd/scripts/build-schemas.py
```diff
@@ -24,7 +24,7 @@
 from ansible_collections.arista.avd.plugins.plugin_utils.schema.avdtojsonschemaconverter import AvdToJsonSchemaConverter
 from ansible_collections.arista.avd.plugins.plugin_utils.schema.store import create_store
 
-FRAGMENTS_PATTERN = "*.schema.yml"
+FRAGMENTS_PATTERN = "*.yml"
```
gmuloc, line 27: not sure if this impact some documentation as well

ClausHolbechArista: I don't think we specify the file names anywhere, and clearly it leads to missing docs.

ClausHolbechArista:
```suggestion
# Schemas should be named '*.schema.yml', but we include all yml files since we are humans.
FRAGMENTS_PATTERN = "*.yml"
```

carlbuchmann (approved):
LGTM, and I agree that all  `.yml` files in fragment folder are evaluated. It will prevent this from happening in the future.

gmuloc (approved):

PR #3706, opened 2024-03-06
Fix(plugins): YAML Dumper/Loader not working without libyaml

ClausHolbechArista:
## Change Summary

YAML Dumper not working without libyaml

## Proposed changes

Minimal impact since this is not used widely (yet).
Affecting:
- PyAVD when running from an unsupported install without libyaml
- deploy_to_cv when running without libyaml


gmuloc (approved):

carl-baillargeon (approved):
LGTM

PR #3725, opened 2024-03-08
Feat(plugins): Add "unique_keys" in avdschema.

ClausHolbechArista:
## Change Summary

Add "unique_keys" in avdschema.

## Component(s) name

AVD Schema validation

## Proposed changes

Support `unique_keys` in the AVD Meta schema so we can validate uniqueness of one or more keys which is not the `primary_key`.
Also supports checking uniqueness across nested data models by using dot-notated variable path.

Limitation: If the test for nested keys overlap with the primary_key of an underlying dict, it will emit errors from both checks. This should be solved by removing `primary_key` on the nested dict, and just set `required: true` on the key instead.

## How to test

Added unit tests for both positive and negative cases.
Also tested manually with `cv_pathfinder_regions` where we want `unique_keys: [name, sites.name"` to ensure unique site names across multiple regions.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/schema/test_avdschema.py
+++ b/ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/schema/test_avdschema.py
@@ -73,6 +73,44 @@ INVALID_ACL_DATA = [
 ]
 
 
+UNIQUE_KEYS_SCHEMAS = [
+    {
+        "type": "list",
+        "unique_keys": ["key", "nested_list.nested_list_key"],
+        "items": {
+            "type": "dict",
+            "keys": {"key": {"type": "str"}, "nested_list": {"type": "list", "items": {"type": "dict", "keys": {"nested_list_key": {"type": "int"}}}}},
+        },
+    }
+]
+
+UNIQUE_KEYS_VALID_DATA = [
+    [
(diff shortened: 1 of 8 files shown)


ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/schema/test_avdschema.py
```diff
@@ -73,6 +73,44 @@
...
+        {"key": "a", "nested_list": [{"nested_list_key": 1}, {"nested_list_key": 2}]},
+        {"key": "b", "nested_list": [{"nested_list_key": 3}, {"nested_list_key": 4}]},
+    ],
+]
+
+UNIQUE_KEYS_INVALID_DATA = [
+    [
+        {"key": "a", "nested_list": [{"nested_list_key": 1}, {"nested_list_key": 2}]},
+        {"key": "b", "nested_list": [{"nested_list_key": 3}, {"nested_list_key": 4}]},
+    ],
```
gmuloc: I may be blind but I dont see the difference between this one and the valid data?

ClausHolbechArista: correct. Which is worrying since the unit tests pass on my system.

gmuloc (approved):
LGTM

carlbuchmann (approved):
LGTM

PR #3195, opened 2023-10-03
Fix(eos_designs): Improve evpn_multicast error handling

jonxstill:
## Change Summary

Raise errors when `evpn_l2_multicast` or `evpn_l3_multicast` is defined but `evpn_multicast` is not.

## Related Issue(s)

Fixes #3153 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Raises AristaAvdError when EVPN  L2/L3 multicast is enabled at the VLAN, VRF or Tenant level, but not globally.

## How to test
Negative tests added to check error is triggered.

### Repository Checklist

- [X] My code has been rebased from devel before I start
- [X] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [X] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #3731, opened 2024-03-11
Fix(eos_designs): evpn_multicast error handling

### LINKED ISSUE #3153: Automatically set software-forwarding SFE when configuring EVPN multicast in tenants
### Enhancement summary

As discussed, with the following configuration at the fabric level:
_underlay_multicast: true_

And tenants files set with EVPN multicast:
_evpn_l2_multicast: true_

SFE was not configured:
```
router multicast
   ipv4
      routing
```
I required configuring at the fabric level:
_evpn_multicast: true_

This can be enhanced to set "software-forwarding sfe" when EVPN multicast is set already in the tenant files.

eos_designs

### Use case example

L2 EVPN multicast: https://www.arista.com/en/support/toi/eos-4-26-0f/14669-layer-2-multicast-evpn

Automatically set software-forwarding SFE when EVPN multicast is enabled in the tenant files.

_No response_

### Additional context

_No response_

### Contributing Guide

carlbuchmann:
## Change Summary

Remove the EVPN multicast checks and error handling were too restrictive and prevented some valid designs and use cases.
Enforce evpn_multicast: true when enable l2_evpn_multicast.

## Fixes

Fixes #3153 correctly by not configuring L2 EVPN Multicast.
Resolves an issue where you had L3 leafs and didn't want EVPN Multicast configured.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Revert checks implemented in PR #3195
- Removed associated eos_designs negative unit tests
- related to issue https://github.com/aristanetworks/avd/issues/3153
- Add unit test to explicitly disable evpn_multicast on node even when it is enabled in tenant network services
   - See diff to evaluate in commit (node was first configured with EVPN multicast): 6f13f00
- Enforce evpn_multicast: true when enable l2_evpn_multicast
- Update schema description to be clear on this requirement.
   

## How to test

Added a test case in the EVPN Multicast scenario with a node `EVPN-MULTICAST-DISABLED` with EVPN Multicast disabled.
The behavior is that no EVPN multicast configuration gets generated, which is what we would expect.


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/EVPN-MULTICAST-DISABLED.yml
```diff
@@ -0,0 +1,5 @@
+# Disable EVPN Multicast while still having EVPN Multicast to be configured on node.
```
gmuloc: This comment looks wrong is it rather:
```suggestion
# Disable EVPN Multicast while still having underlay Multicast to be configured on node.
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/EVPN-MULTICAST-DISABLED.yml
```diff
@@ -0,0 +1,5 @@
...
+# We should see that no EVPN services be configured with EVPN Multicast
```
gmuloc:
```suggestion
# Expecting that no EVPN service is configured with EVPN Multicast.
```

gmuloc (changes requested):

gmuloc (approved):
LGTM

PR #3813, opened 2024-04-04
Fix(dhcp_provisioner): Support custom `node_type_keys` as a list of dicts.

philippebureau:
## Change Summary

Add filter to the creation of node type list to match new AVD 4.x syntax of node type keys (dict of dict -> list of dict)

## Related Issue(s)

Fixes aristanetworks/avd-internal#154

This was preventing the use of custom node type keys with the new AVD 4.0 syntax

## Component(s) name

`arista.avd.dhcp_provisioner`

## Proposed changes
add | map(attribute='key')

## How to test
create custom node types and run the role
see molecule test for example

### Repository Checklist

- [ x ] My code has been rebased from devel before I start
- [ x ] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [ x ] My change requires a change to the documentation and documentation have been updated accordingly.
- [ x ] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)


ansible_collections/arista/avd/molecule/dhcp_configuration/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -177,3 +184,67 @@ bfd_multihop:
...
+    type: l2leaf
+    connected_endpoints: true
+    mlag_support: true
+    network_services:
+      l2: true
+    underlay_router: false
+    uplink_type: port-channel
+
+  - key: custom_l2leaf
+    type: customl_2leaf
```
danieltudares: fix typo

philippebureau: fixed

philippebureau:
after discussing internally, to prevent braking old syntax "| arista.avd.convert_dicts('key')" was added to convert the old syntax so it works with both.
Thank you to @gmuloc for the recommendation

ansible_collections/arista/avd/molecule/dhcp_provisioning/inventory/group_vars/DC1_L2LEAFS.yml
```diff
@@ -1 +1,65 @@
 type: l2leaf
+
+# Custom node types with old syntax
```
gmuloc:
```suggestion
# Custom node types with AVD 3.x syntax using a dict of dicts
```

ansible_collections/arista/avd/molecule/dhcp_provisioning/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -175,3 +185,67 @@ bfd_multihop:
   interval: 1200
   min_rx: 1200
   multiplier: 3
+
+# Custom node types
```
gmuloc: Hmm given that the role takes only the first device in the template to derive the node_type_keys (yeah thats not great we agree) I don't think everything is being tested as we would expect here.
We probably need to create separate fabrics for the default behavior (that was tested before and is not now), the new syntax node_type_keys as you have here and the AVD 3.8 syntax node_type_keys.

I guess lets write 3 tests fabric / modify the playbook to run at three levels?

ansible_collections/arista/avd/molecule/dhcp_configuration/inventory/group_vars/DC1_CUSTOM_L2LEAFS.yml
```diff
@@ -0,0 +1,65 @@
+type: custom_l2leaf
+
+# Custom node types with old syntax
```
gmuloc:
```suggestion
# Custom node types with AVD 3.8 syntax
```

ansible_collections/arista/avd/molecule/dhcp_provisioning/inventory/group_vars/DC1_FABRIC.yml
philippebureau: I will work on the 3 fabrics test

danieltudares (approved):
LGTM!

gmuloc (dismissed):
LGTM - thanks @philippebureau for taking the time to fix this!

ansible_collections/arista/avd/molecule/dhcp_provisioning/molecule.yml
```diff
@@ -1,29 +1,24 @@
...
-    - prepare
     - converge
   cleanup_sequence:
     - destroy
 dependency:
   name: galaxy
   enabled: false
 driver:
-  name: docker
+  name: default
```
ClausHolbechArista, line 19: Why do we need these changes to the molecule scenario?

gmuloc: it is more that we don't need docker for offline mode and so we are triggering extra hooks in the scenario that are not needed. The other scenario `dhcp_configruation` (and yes we can argue that the names are confusing) for online mode still requires docker containers.

ClausHolbechArista (approved):

PR #3867, opened 2024-04-16
Feat(eos_cli_config_gen): Add capability to disable address-only for connectivity monitors

gmuloc:
## Change Summary

> [!IMPORTANT]  
> You need a very recent engineering version of EOS to test - this is not yet released.

This PR is required for WAN Internet Exit configuration towards SWG

## Component(s) name

## Proposed changes

* Add a boolean `address_only` in the four places where it can now be disabled
* Set the boolean to `true` by default to be backward compatible
    * In AVD 5.0 we may decide to make it default False -> **this needs to be tracked as a 5.0 issue if we decide so** 
* Exercise the new knob in 3 of the four places - can make it work as well in the last one but requires an extra host in `eos_cli_config_gen` molecule.

## How to test

molecule

### Reviewer Checklist

- [x] For documentation, rather than adding a template we could decide to add `(address-only)` in the local interfaces cell when this is configured and nothing otherwise, if the reviewer feels strongly about this, we can change that. -> NO
 
### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/monitor-connectivity.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/monitor-connectivity.md
@@ -79,7 +79,7 @@ interface Management1
 
 ##### Host Parameters
 
-| Host Name | Description | IPv4 Address | Probing Interface Set | Address only | URL |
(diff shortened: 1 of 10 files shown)


ClausHolbechArista (approved):

MaheshGSLAB (approved):
LGTM, tested by generating the .cfg files.

PR #3901, opened 2024-04-24
Fix(eos_designs)!: Add 'vxlan-routing' TCAM profile to R3 devices

fhibler:
Related to SR546455. This change will be hitful, as the TCAM profile on R3 devices is changed and hence the forwarding agent is restarted.

## Change Summary

The 'vxlan-routing' TCAM profile is required on Jericho2-based devices for proper handling of VXLAN IPv6 NS packets.

## Related Issue(s)

Related to SR546455. No GitHub/AVD Issue opened for this.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add TCAM profile 'vxlan-routing' to R3 devices (7280R3/7500R3/7800R3) to make it consistent with the configuration we see for R/R2 devices.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/snmp.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/snmp.md
@@ -167,4 +167,5 @@ snmp-server host 10.6.75.100 vrf MGMT version 3 priv USER-READ-AUTH-PRIV
 snmp-server enable traps
 no snmp-server vrf default
 snmp-server vrf MGMT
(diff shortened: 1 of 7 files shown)


ClausHolbechArista:
I am not sure I agree that we should add this in the defaults at this moment. It only solves a specific case for IPv6 and this would affect every customer already using AVD.

Maybe we could make this a knob `jericho_tcam_profile_for_ipv6: <str>` under `platform_settings` so we detect if IPv6 is in use, and then configure the profile?

ClausHolbechArista (changes requested):
Blocking this until we agree on how we want to address it.

PR #3918, opened 2024-04-29
CI: Remove pre-commit from Github Workflow

ClausHolbechArista:
## Change Summary

Remove pre-commit from Github Workflow

Since we have implemented the pre-commit.ci application, we no longer need to run pre-commit in the workflow.


gmuloc (approved):

carlbuchmann (approved):
LGTM

PR #3976, opened 2024-05-09
Fix(eos_cli_config_gen): Don't render 'vrf default' on 'router ospf'

pvinci-arista:
## Change Summary

This change changes the rendering of `router ospf <x> vrf default` to `router ospf <x>`

## Component(s) name

## Proposed changes
added test for default vrf

## How to test
tested with molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-ospf.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-ospf.md
@@ -153,6 +153,7 @@ interface Vlan24
 | 400 | - | disabled |- | disabled | default | disabled | disabled | - | - | - | - |
 | 500 | - | disabled |- | disabled | default | disabled | disabled | - | - | - | - |
 | 600 | - | disabled |- | disabled | default | disabled | disabled | - | - | - | - |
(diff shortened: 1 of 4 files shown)


ClausHolbechArista (approved):
LGTM

ClausHolbechArista (changes requested):
On second thought, since EOS accepts the current behavior, we should not change this in a minor release. If you don't want the default VRF, you should not set it in the input. So maybe the bug here is in eos_designs. Nevertheless we would have to wait for 5.0.

pvinci-arista:
This is an artifact of configuring `ospf` under `network services` in the `default` vrf.  Based on where it is configured, I don't see how to not set it.  To remove it would seem to require structured_config on each node.

```
default:
  - name: "default"
    vrfs:
       - name: "default"       
          ospf:
            enabled: true

            process_id: 123
            nodes:
              - SW1
              - SW2
              - SW3
              - SW4
```
  
I agree that this is cosmetic.  Eos does accept the existing rendered config.  Feel free to tag this to the appropriate SemVer release.

ClausHolbechArista:
@pvinci-arista This should be addressed in `eos_designs` so please adjust the fix accordingly. In parallel we will discuss if this can be merged in a minor or will have to wait for 5.0. @aristanetworks/avd-maintainers

ClausHolbechArista:
Replaced with #4334

PR #3986, opened 2024-05-13
Fix(cv_deploy): Detect pathfinder metadata studio version and skip unsupported fields

ClausHolbechArista:
## Change Summary

Detect pathfinder metadata studio version and skip unsupported fields

## Component(s) name

`arista.avd.cv_deploy`

## How to test

- Tested manually with wrong versions and older versions.
- More manual testing will be done when merged.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -286,38 +343,52 @@ async def deploy_cv_pathfinder_metadata_to_cv(cv_pathfinder_metadata: list[CVPat
...
+        return {}, []
 
     LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Found %s 'internet_exit_policies' for device: %s", len(internet_exit_policies), device.hostname)
 
     services_dict = {}
+    warnings = []
+
     for internet_exit_policy in internet_exit_policies:
         # We currently only support zscaler
         if internet_exit_policy.get("type") != "zscaler":
```
ClausHolbechArista:
```suggestion
        if internet_exit_policy.get("type") not in ["zscaler", "direct"]:
```

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
```diff
@@ -40,6 +50,79 @@ class StudioMixin:
...
+            if isinstance(e, CVResourceNotFound):
+                # Continue execution if we did not find any state in the workspace.
+                # This simply means the studio itself was not changed in this workspace.
+                pass
+            else:
+                raise e
+
+        # If we get here, it means no studio was returned by the workspace call.
+        # So now we fetch the studio config from the workspace to see if the studio was deleted in this workspace.
+        request = StudioConfigStreamRequest(
```
ayushmittal-arista, line 91: Is this working?

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -23,6 +25,47 @@
...
+    version = studio_version_field.string_props.default_value
+    LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Found metadata studio version: %s", version)
+    try:
+        return StudioVersion(float(version))
+    except ValueError:
+        accepted_versions = [str(v.value) for v in StudioVersion]
+        warning = (
+            f"deploy_cv_pathfinder_metadata_to_cv: Got invalid metadata studio version '{version}'. "
+            f"This plugin only accepts versions '{accepted_versions}'. Skipping upload of metadata."
+        )
```
gmuloc: So this means that we are locking each AVD version to a limited set of studio versions?

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -286,38 +343,52 @@ async def deploy_cv_pathfinder_metadata_to_cv(cv_pathfinder_metadata: list[CVPat
...
-                "deploy_cv_pathfinder_metadata_to_cv: Ignoring unsupported internet exit policy '%s' with type '%s' for device: %s.",
-                internet_exit_policies.get("name"),
-                internet_exit_policy.get("type"),
-                device.hostname,
+        if internet_exit_policy.get("type") not in ["zscaler", "direct"]:
+            warning = (
+                f"deploy_cv_pathfinder_metadata_to_cv: Ignoring unsupported internet exit policy '{internet_exit_policies.get('name')}' "
+                f"with type '{internet_exit_policy.get('type')}' for device: {device.hostname}."
+            )
+            LOGGER.info(warning)
```
gmuloc, line 371: Should the warning be added to `warnings` here as well just like in line 391?

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ayushmittal-arista: I am not familiar with the wrappers you have implemented, pretty neat. It's a hassle to deal with the proto gen python clients.

 But the first argument needs to be list.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
```diff
@@ -40,6 +50,79 @@ class StudioMixin:
...
+            StudioConfig(
+                key=StudioKey(studio_id=studio_id, workspace_id=workspace_id),
+                remove=True,
```
ayushmittal-arista: 1.) remove=True is disregarded in getAll. Only the key will be used. 
As such we need to check the responses to see if we get remove=true, set, instead of raising error for any response received
2.) time field needs to timebound, ignore if auto converted. For StudioConfigRequest it's timestamp, for StudioConfigStreamRequest it's timebounds

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
```diff
@@ -40,6 +50,79 @@ class StudioMixin:
...
+            ),
+            time=time,
+        )
+        client = StudioConfigServiceStub(self._channel)
+        try:
+            responses = client.get_all(request, metadata=self._metadata, timeout=timeout)
+            async for response in responses:
+                # If we get here it means we got an entry with "removed: True" so no need to look further.
+                raise CVResourceNotFound("The studio was deleted in the workspace.", f"Studio ID '{studio_id}, Workspace ID '{workspace_id}'")
```
gmuloc, line 105: probably a bit nitpicking.

this piece of code will trigger the `except` line 110 - do we want this ? Or would we rather want to run the async for loop afterwards?
Feels like we are going to call `get_cv_client_exception` on an already crafted by ourselves exception otherwise

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
```diff
@@ -40,6 +50,79 @@ class StudioMixin:
...
+
+        except Exception as e:
+            raise get_cv_client_exception(e, f"Studio ID '{studio_id}, Workspace ID '{workspace_id}'") or e
+
+        # If we get here, it means there are no inputs in the workspace and they are not deleted, so we can fetch from mainline.
+        request = StudioRequest(
+            # First attempt to fetch studio from workspace.
+            key=StudioKey(studio_id=studio_id, workspace_id=""),
+            time=time,
```
ayushmittal-arista, line 114: Ideally, the time needs to be Workspace.CreatedAt: i.e the time at which workspace was created. 

I say ideally, because we are almost near latest when using avd. But if you are looking to have this function be fail-proof, and generic, we need to do a WorkspaceConfig Request, which will give us the WorkspaceCreatedAt timestamp, which we then need to pass here, when fetching studio from mainline.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -84,6 +128,18 @@ def upsert_pathfinder(metadata: dict, device: CVDevice, studio_inputs: dict) ->
         "tags": {"query": f"device:{device.serial_number}"},
     }
 
+    pathfinder_location = {key: metadata.get(key) for key in ("region", "site", "address")}
+    if any(pathfinder_location):
+        if is_feature_supported("pathfinder_location", studio_version):
+            pathfinder_metadata.update(pathfinder_location)
```
ayushmittal-arista: Maybe I am missing something but don't we need to do,
we need to do pathfinder_metadata["inputs"]["value"].update(..)

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ClausHolbechArista: Well get_cv_client_exception will ignore it and return none, so the `or e` will raise the original.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: No this is only for the debug output.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -286,38 +343,52 @@ async def deploy_cv_pathfinder_metadata_to_cv(cv_pathfinder_metadata: list[CVPat
...
+            continue
+
+        if not is_feature_supported("internet_exit_zscaler", studio_version):
+            warning = (
+                "deploy_cv_pathfinder_metadata_to_cv: Ignoring Zscaler internet-exit information since "
+                f"it is not supported by metadata studio version {studio_version.value}."
             )
+            LOGGER.info(warning)
+            warnings.append(warning)
             continue
```
ayushmittal-arista, line 383: break instead of continue if feature not supported case.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: Yes. We have no way of filling out the correct fields otherwise.

ayushmittal-arista (changes requested):

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: We could have other internet exit policies which are supported.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: fixed

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -25,45 +25,37 @@
...
+    return bool(get_v2(studio_schema, "fields.values.zscaler"))
 
 
-async def get_metadata_studio_version(result: DeployToCvResult, cv_client: CVClient) -> StudioVersion | None:
+async def get_metadata_studio_schema(result: DeployToCvResult, cv_client: CVClient) -> InputSchema | None:
     """
-    Extract the version from the metadata studio.
-    This is stored as a default value on the studioVersion field.
+    Extract and return the input schema from the metadata studio.
+    Returns None if the metaschema studio is not found.
```
ayushmittal-arista: metaschema -> metadata

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ayushmittal-arista: Yeap, sounds good.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ClausHolbechArista: Are you sure about the disregard for remove=true? I tested it for some other APIs and it worked well.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ClausHolbechArista: OK. I will log this as a general issue, since I reuse this style everywhere. IMO it should be solved serverside instead of requiring me to do multiple API calls to get the state of the workspace.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ClausHolbechArista: fixed. Thanks

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: Changed to check supported fields in the schema instead. Thanks.

gmuloc: don't we prefer to think that we are non breaking going forward and just adding things in this metadata studio? This is the same discussion we had regarding having dependencies locked at some version - it will bites us in the future

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista:
```suggestion
    Returns None if the metadata studio is not found.
```

ClausHolbechArista: fixed

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
gmuloc: so in one case you have `warnings.append(warning)` for a non supported feature and in the other case you don't - it does not feel consistent

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
gmuloc: true but this could be rewritten to avoid this double processing by putting the for loop outside of the try block

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
gmuloc: ok my comment came too late :)

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/client/studio.py
ClausHolbechArista: Tested the API and it returns the config only if a studio is removed.

ClausHolbechArista: Added timebound to our internal tracker to fix later.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista: I did not want to make noise for "direct" or other future types, but now I added handling of "direct, so I can make it as a warning now.

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -23,6 +25,39 @@
...
+
+
+def is_internet_exit_zscaler_supported(studio_schema: InputSchema) -> bool:
+    """Detect if zscaler internet exit is supported by the metadata studio"""
+    return bool(get_v2(studio_schema, "fields.values.zscaler"))
+
+
+async def get_metadata_studio_schema(result: DeployToCvResult, cv_client: CVClient) -> InputSchema | None:
+    """
+    Extract and return the input schema from the metadata studio.
```
ayushmittal-arista: comment can be reworded maybe 

`Download and return the input schema for the cv pathfinder metadata studio`

ayushmittal-arista (approved):

ansible_collections/arista/avd/plugins/plugin_utils/cv_client/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ClausHolbechArista:
```suggestion
    Download and return the input schema for the cv pathfinder metadata studio.
```

ClausHolbechArista: fixed

gmuloc (approved):

PR #3244, opened 2023-10-14
Feat(plugins): Support braces in range_expand filter

ClausHolbechArista:
## Change Summary

Support braces in range_expand filter

## Component(s) name

`arista.avd.range_expand`

## Proposed changes

Add support for braces with ranges or comma-separated numbers:
```yaml
- "{{ 'eth{7,9,11-13}/1,21/1,26/1' | range_expand }}"
# -> ["eth7/1", "eth9/1", "eth11/1", "eth12/1", "eth13/1", "eth21/1", "eth26/1"]```
```

## How to test

Added complete unit tests for the rewritten python code and added to the existing unit tests for the Jinja filter.

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/ansible-avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #4013, opened 2024-05-17
Feat(plugins): Revert support braces in range_expand filter (#3244)"

carlbuchmann:
## Change Summary

Revert support braces in range_expand filter PR #3244 

The current implementation support undocumented syntax: `65000.1-65000.20`, and the new refactored filter create invalid configuration i.e `65000.1.20`. 

We will re-implement this in 5.0 since this is a potential breaking change.

## How to test

CI must pass!


python-avd/pyavd/templater.py
```diff
@@ -57,9 +57,9 @@ def import_filters_and_tests(self) -> None:
         from .vendor.j2.filter.hide_passwords import hide_passwords
         from .vendor.j2.filter.list_compress import list_compress
         from .vendor.j2.filter.natural_sort import natural_sort
+        from .vendor.j2.filter.range_expand import range_expand
```
carlbuchmann, line 60: @ClausHolbechArista - When I reverted the commit there was a merge conflict on this file which I resolved, but please double-check I did this correctly!

ClausHolbechArista: Looks good.

ClausHolbechArista (approved):

gmuloc (approved):

PR #4020, opened 2024-05-21
Fix(eos_designs): Connected endpoints ESI should only be configured on EVPN VTEPs.

### LINKED ISSUE #3064: eos_designs: Connected endpoints ESI should only matter for EVPN VTEPs.
Connected endpoints ESI should only matter for EVPN VTEPs, so please change the test to l3leaf instead.
But this also reveals that our code for l2leaf is configuring ESI even though we should not. (Opening seperate issue on that.)

_Originally posted by @ClausHolbechArista in https://github.com/aristanetworks/ansible-avd/pull/3027#discussion_r1276170347_

laxmikantchintakindi:
## Change Summary

 Connected endpoints ESI should only matter for EVPN VTEPs. It is not needed to configure short_esi on l2leafs. 

## Related Issue(s)

Fixes #3064

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Change python module to skip generating short_esi for l2leafs.

## How to test
Add short_esi under l2leaf and make sure it is not generated in structured config.

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
@@ -97,18 +97,15 @@ class UtilsMixin:
         """
         Return short_esi for one adapter
         """
-        if len(set(adapter["switches"])) < 2:
+        if len(set(adapter["switches"])) < 2 and not self.shared_utils.overlay_evpn and not self.shared_utils.overlay_vtep:
             # Only configure ESI for multi-homing.
             return None
 
         # short_esi is only set when called from sub-interface port-channels.
         if short_esi is None:
             # Setting short_esi under port_channel will be removed in AVD5.0
-            if self._hostvars["type"] != "l2leaf":
-                port_channel_short_esi = get(adapter, "port_channel.short_esi")
-                if (short_esi := get(adapter, "ethernet_segment.short_esi", default=port_channel_short_esi)) is None:
-                    return None
-            else:
+            port_channel_short_esi = get(adapter, "port_channel.short_esi")
+            if (short_esi := get(adapter, "ethernet_segment.short_esi", default=port_channel_short_esi)) is None:
                 return None
 
         endpoint_ports: list = adapter.get("endpoint_ports")
--- a/ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
+++ b/ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
@@ -97,7 +97,7 @@ class UtilsMixin:
         """
         Return short_esi for one adapter
         """
-        if len(set(adapter["switches"])) < 2 and not self.shared_utils.overlay_evpn and not self.shared_utils.overlay_vtep:
+        if len(set(adapter["switches"])) < 2 or not self.shared_utils.overlay_evpn or not self.shared_utils.overlay_vtep:
             # Only configure ESI for multi-homing.
             return None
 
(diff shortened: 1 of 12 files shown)


ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -104,8 +104,11 @@ def _get_short_esi(self, adapter: dict, channel_group_id: int, short_esi: str =
         # short_esi is only set when called from sub-interface port-channels.
         if short_esi is None:
             # Setting short_esi under port_channel will be removed in AVD5.0
-            port_channel_short_esi = get(adapter, "port_channel.short_esi")
-            if (short_esi := get(adapter, "ethernet_segment.short_esi", default=port_channel_short_esi)) is None:
+            if self._hostvars["type"] != "l2leaf":
```
ClausHolbechArista: This is not how we should check for it. Types are customizable, so we have to check for capabilities instead. I suggest checking this:
```suggestion
            if self.shared_utils.overlay_evpn and self.shared_utils.overlay_vtep:
```

ClausHolbechArista: Also this will not capture all cases, since you add it below the short_esi is none check. Please move these two conditions (negated) to the `if` in line 100. So we return None if these two are NOT true.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
ClausHolbechArista: My bad. It should also be for evpn-mpls so I will update my comment above

laxmikantchintakindi: if self.shared_utils.overlay_evpn and self.shared_utils.overlay_vtep:
 
 can we write 'or' instead of 'and'?

ClausHolbechArista: you can implement it any way you like. Just make sure you understand the logic and how it would work first.

laxmikantchintakindi: Fixed.

ansible_collections/arista/avd/roles/eos_designs/python_modules/connected_endpoints/utils.py
```diff
@@ -97,7 +97,7 @@ def _get_short_esi(self, adapter: dict, channel_group_id: int, short_esi: str =
         """
         Return short_esi for one adapter
         """
-        if len(set(adapter["switches"])) < 2:
+        if len(set(adapter["switches"])) < 2 and not self.shared_utils.overlay_evpn and not self.shared_utils.overlay_vtep:
```
ClausHolbechArista: This logic will not work correctly if only one of the condtions are false.
```suggestion
        if len(set(adapter["switches"])) < 2 or not self.shared_utils.overlay_evpn or not self.shared_utils.overlay_vtep:
```

Please add a test case in molecule to make sure we see the problem and then see that your solution solves it.

laxmikantchintakindi: Fixed.

ClausHolbechArista (approved):

Shivani-gslab (approved):

PR #4050, opened 2024-05-29
Feat(eos_designs): Allow reuse of cross-device BGP peer ip

### LINKED ISSUE #3914: Bug(eos_designs): Relax cross-device BGP peer IP reuse validation
### Issue Summary

Today, if a given bgp peer IP is used on device A and reused on device B, validation fails:

```
[WARNING]: [SJC13EXTLF01-A]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
are not unique as required.
[WARNING]: [SJC13EXTLF01-B]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
are not unique as required.
```

However, although not a recommended design, this is a valid config. 
Currently, validation has to be run set to 'warning' and the user just ignores those messages increasing margin for errors.

Please allow IP reuse for BGP peerings across different devices.

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
On an MLAG pair with -A and -B peers, configure:

On -A:
        bgp_peers:
          - ip_address: 10.95.250.201  # svi 1608

On -B:
        bgp_peers:
          - ip_address: 10.95.250.201  # svi 1609
```

### Relevant log output

```shell
TASK [arista.avd.eos_designs : Set eos_designs facts] *********************************************************************
[WARNING]: [SJC13MARLF01-A]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
are not unique as required.
[WARNING]: [SJC13MARLF01-B]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
are not unique as required.
```

### Contributing Guide

laxmikantchintakindi:
## Change Summary

Currently if a given BGP peer IP is used on device A and reused on device B, validation fails. However, although not a recommended design, this is a valid config. This PR will relax the BGP peer ip reuse.

## Related Issue(s)

Fixes #3914

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Today, if a given bgp peer IP is used on device A and reused on device B, validation fails.

```
[WARNING]: [SJC13EXTLF01-A]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
are not unique as required.
[WARNING]: [SJC13EXTLF01-B]: 'Validation Error: Marr-Managed-WAN[0].vrfs[0].bgp_peers': Values of Primary key 'ip_address'
```

However, although not a recommended design, this is a valid config.
Currently, validation has to be run set to 'warning' and the user just ignores those messages increasing margin for errors.
Allow IP reuse for BGP peerings across different devices.

### User Checklist

- N/A

### Repository Checklist


pmprado:
> Review docs on [Read the Docs](https://ansible-avd--4050.org.readthedocs.build/en/4050/)
> 
> To test this pull request:
> 
> ```shell
> # Create virtual environment for this testing below the current directory
> python -m venv test-avd-pr-4050
> # Activate the virtual environment
> source test-avd-pr-4050/bin/activate
> # Install all requirements including PyAVD
> pip install "pyavd[ansible] @ git+https://github.com/laxmikantchintakindi/avd.git@bgp_peer_ip_use_validation#subdirectory=python-avd" --force
> # Install Ansible collection
> ansible-galaxy collection install git+https://github.com/laxmikantchintakindi/avd.git#/ansible_collections/arista/avd/,bgp_peer_ip_use_validation --force
> # Optional: Install AVD examples
> cd test-avd-pr-4050
> ansible-playbook arista.avd.install_examples
> ```

Feedback - it should be "python3" instead of "python"

pmprado:
I've tested it and it does not warn anymore for when we have repeated IPs across devices. It builds correctly.
However, shouldn't it warn on this? In my testing it didn't warn at all:

```
        bgp_peers:
          - ip_address: 1.2.3.4 
            nodes: [EXTLF01-B]
          - ip_address: 1.2.3.4 
            nodes: [EXTLF01-B]

```
That's IP reuse on the same VRF and on the same device, not sure that's a valid config that we should warn or not.

pmprado (changes requested):
The code now allows duplicates in general and does not warn.
While this does address the problem of duplicates across devices, it is blindly allowing duplicates on the same device, and when that happens, it will create a single entry in the resulting config. 
If the duplicates have different data, I believe only the lastly defined will be considered.
IMHO duplicates on the same device and same VRF are an error, likely due to a typo or copy/paste error... @ClausHolbechArista what do you think?

ClausHolbechArista:
This code should not allow duplicates. Have you tested? Your example above had the same exact info twice. Such duplication is silently ignored. Only one peering will be added to the structured config. Please confirm.
@pmprado

pmprado:
> This code should not allow duplicates. Have you tested? Your example above had the same exact info twice. Such duplication is silently ignored. Only one peering will be added to the structured config. Please confirm. @pmprado

Yes I did test. Indeed it was an exact duplicate, to simulate a typo on a situation where we have 2 peers that are basically identical except for the IP. Do you mean that if they are not _exact_  duplicates they won't be silently ignored? Still, shouldn't any duplicate be warned about?

ClausHolbechArista:
> > This code should not allow duplicates. Have you tested? Your example above had the same exact info twice. Such duplication is silently ignored. Only one peering will be added to the structured config. Please confirm. @pmprado
> 
> Yes I did test. Indeed it was an exact duplicate, to simulate a typo on a situation where we have 2 peers that are basically identical except for the IP. Do you mean that if they are not _exact_ duplicates they won't be silently ignored?

Correct

> Still, shouldn't any duplicate be warned about?

No, since you may have situations where you define the VRF in multiple tenants, and they get merged, possibly leading to the same peers being repeated (it is a very complicated corner case). There is no reason to error out when there is no conflict.

ClausHolbechArista (approved):

MaheshGSLAB (approved):

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
```diff
@@ -196,7 +196,14 @@ def _router_bgp_vrfs(self) -> list | None:
...
                     )
-                    bgp_vrf.setdefault(address_family, {}).setdefault("neighbors", []).append(neighbor)
+
+                    append_if_not_duplicate(
+                        list_of_dicts=bgp_vrf.setdefault(address_family, {}).setdefault("neighbors", []),
+                        primary_key="ip_address",
+                        new_dict=neighbor,
+                        context="BGP peer defined under VRFs",
+                        context_keys=["ip_address"],
+                    )
```
gmuloc, line 206: Can you add a negative unit test for this? 
like a duplicate for the same host with different configs? 
Seems important that we don't lose track of this.

gmuloc (dismissed):

ansible_collections/arista/avd/roles/eos_designs/python_modules/network_services/router_bgp.py
laxmikantchintakindi: Added.

ClausHolbechArista (approved):

PR #4063, opened 2024-06-03
Feat(eos_cli_config_gen): Add Patch-Panel Connector commands

ccsnw:
## Change Summary

Extend existing Patch-Panel template by adding "connector" global commands.

```
patch_panel:
  connector:
    interface:
      patch:
        bgp_vpws_remote_failure_errdisable: <bool>
      recovery:
        review_delay: 
          min: <10-600>
          max:  <15-900>
```

## Related Issue(s)

Fixes #<N/A>

## Component(s) name

## Proposed changes
Add connector interface
 - patch
 - recovery

with currently featured sub-commands

## How to test
extended existing molecule test

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/patch_panel.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/patch_panel.schema.yml
@@ -9,6 +9,37 @@ keys:
   patch_panel:
     type: dict
     keys:
+      connector:
+        type: dict
+        keys:
+          interface:
+            type: dict
+            keys:
+              patch:
+                type: dict
+                keys:
+                  bgp_vpws_remote_failure_errdisable:
+                    type: bool
+              recovery:
+                type: dict
+                keys:
+                  review_delay:
+                    type: dict
+                    keys:
+                      min:
+                        type: int
+                        required: true
+                        min: 10
+                        convert_types:
+                          - str
+                        description: Minimum delay (10-600).
+                      max:
+                        type: int
+                        required: true
+                        min: 15
+                        convert_types:
+                          - str
+                        description: Maximum delay (15-900).
       patches:
         type: list
         primary_key: name
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/patch_panel.schema.yml
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/patch_panel.schema.yml
@@ -44,7 +44,7 @@ keys:
         type: list
         primary_key: name
         convert_types:
-        - dict
+          - dict
         items:
           type: dict
           keys:
@@ -56,7 +56,7 @@ keys:
               type: list
               primary_key: id
               convert_types:
-              - dict
+                - dict
               min_length: 2
               max_length: 2
               description: Must have exactly two connectors to a patch of which at least one must be of type "interface".
@@ -66,12 +66,12 @@ keys:
                   id:
                     type: str
                     convert_types:
-                    - int
+                      - int
                   type:
                     type: str
                     valid_values:
-                    - interface
-                    - pseudowire
+                      - interface
+                      - pseudowire
                     required: true
                   endpoint:
                     type: str
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
@@ -39,6 +39,7 @@ Patch Panel Connector Interface Recovery Review Delay Max: {{ patch_panel.connec
 {%     if patch_panel.connector.interface.patch.bgp_vpws_remote_failure_errdisable  is arista.avd.defined(true) %}
 
 Patch Panel Connector Interface Path BGP VPWS Remote Failure Errdisable is enabled.
+
 {%     endif %}
 ### Patch Panel Device Configuration
 
--- a/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
+++ b/ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
@@ -28,7 +28,18 @@
 | {{ patch.name }} | {{ patch_panel_patch.enabled }} | {{ patch_panel_patch.connector_a_type }} | {{ patch_panel_patch.connector_a_endpoint }} | {{ patch_panel_patch.connector_b_type }} | {{ patch_panel_patch.connector_b_endpoint }} |
 {%         endif %}
 {%     endfor %}
+{%     if patch_panel.connector.interface.recovery.review_delay.min is arista.avd.defined %}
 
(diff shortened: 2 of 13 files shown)


ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
```diff
@@ -28,7 +28,19 @@
 | {{ patch.name }} | {{ patch_panel_patch.enabled }} | {{ patch_panel_patch.connector_a_type }} | {{ patch_panel_patch.connector_a_endpoint }} | {{ patch_panel_patch.connector_b_type }} | {{ patch_panel_patch.connector_b_endpoint }} |
 {%         endif %}
 {%     endfor %}
+{%     if patch_panel.connector.interface.recovery.review_delay.min is arista.avd.defined %}
```
ClausHolbechArista: Let's move these above the table (right after the `### Patch Panel Summary` heading and then insert a heading after (above the table) with `#### Patch Panel Connections`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/patch-panel.j2
```diff
@@ -7,6 +7,27 @@
 {% if patch_panel is arista.avd.defined %}
 !
 patch panel
+{%     if patch_panel.connector is arista.avd.defined %}
+{%         set pp_connector_cli = "connector " %}
+{%         if patch_panel.connector.interface is arista.avd.defined %}
+{%             set pp_connector_cli = pp_connector_cli ~ "interface " %}
```
ClausHolbechArista: You can collapse tests, since testing for deeper keys always includes the check for parents.

Also I don't think we need a cli variable here, since you are not really adding multiple tokens "building" a long command.
```suggestion
{%     if patch_panel.connector.interface is arista.avd.defined %}
```

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/patch-panel.j2
```diff
@@ -7,6 +7,27 @@
...
+{%             if patch_panel.connector.interface.recovery is arista.avd.defined %}
+{%                 set pp_connector_cli_recovery = pp_connector_cli ~ "recovery " %}
+{%                 if patch_panel.connector.interface.recovery.review_delay.min is arista.avd.defined and patch_panel.connector.interface.recovery.review_delay.max is arista.avd.defined %}
+{%                     set pp_connector_cli_recovery = pp_connector_cli_recovery ~ "review delay " ~ patch_panel.connector.interface.recovery.review_delay.min ~ " " ~ patch_panel.connector.interface.recovery.review_delay.max %}
```
ClausHolbechArista: Same here you can just check for the required delay min/max and maybe just print it directly with `   connector interface recovery review delay {{ foo }} {{ bar }}`

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/patch-panel.j2
```diff
@@ -7,6 +7,27 @@
...
+   {{ pp_connector_cli_recovery }}
+{%                 endif %}
+{%             endif %}
+{%             if patch_panel.connector.interface.patch is arista.avd.defined %}
+{%                 set pp_connector_cli_patch = pp_connector_cli ~ "patch " %}
+{%                 if patch_panel.connector.interface.patch.bgp_vpws_remote_failure_errdisable is arista.avd.defined(true) %}
+{%                     set pp_connector_cli_patch = pp_connector_cli_patch ~ "bgp vpws remote-failure errdisable" %}
+   {{ pp_connector_cli_patch }}
+{%                 endif %}
+{%             endif %}
```
ClausHolbechArista:
```suggestion
{%             if patch_panel.connector.interface.patch.bgp_vpws_remote_failure_errdisable is arista.avd.defined(true) %}
   connector interface bgp vpws remote-failure errdisable
{%             endif %}
```

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/eos/patch-panel.j2
ccsnw: the idea was to prepare for possible other commands in the future, but i agree to slim it down for now

ClausHolbechArista (approved):
LGTM. Verified command validity and ordering on EOS.

ansible_collections/arista/avd/roles/eos_cli_config_gen/templates/documentation/patch-panel.j2
```diff
@@ -9,6 +9,20 @@
...
 
 ### Patch Panel Summary
+{%     if patch_panel.connector.interface.recovery.review_delay.min is arista.avd.defined %}
+
+Patch Panel Connector Interface Recovery Review Delay Min: {{ patch_panel.connector.interface.recovery.review_delay.min }}
+{%     endif %}
+{%     if patch_panel.connector.interface.recovery.review_delay.max is arista.avd.defined %}
+
+Patch Panel Connector Interface Recovery Review Delay Max: {{ patch_panel.connector.interface.recovery.review_delay.max }}
+{%     endif %}
```
gmuloc, line 19: here the logic is slightly different from the config template in the sense that if I only give min delay it wont' show up in the config but will show up in the doc (as the if statements are disjoint for doc)

gmuloc: hmm ok with the schema (assuming it is not disabled) it won't happen so let's ignore this.

ccsnw: Hi Guillaume,  i could also check if min AND max is set and only if this is the case render both outputs on the doc.. does that make more sense?

ClausHolbechArista: yes and then combine them to a single line in docs showing min and max.

gmuloc (approved):

ansible_collections/arista/avd/roles/eos_cli_config_gen/schemas/schema_fragments/patch_panel.schema.yml
```diff
@@ -9,11 +9,42 @@ keys:
...
+                  review_delay:
+                    type: dict
+                    keys:
+                      min:
+                        type: int
+                        required: true
+                        min: 10
+                        convert_types:
+                          - str
+                        description: Minimum delay (10-600).
```
Shivani-gslab, line 35: we can set `max: 600` here
same for max key below - `max: 900`

gmuloc:
I merged too fast ! wil take care of the comments

ccsnw:
> I merged too fast ! wil take care of the comments

thanks Guillaume, I will take a look once you have updated it. From what Shivani is syaing, I guess i could do sth like:

```
 min:
   type: int
   required: true
   min: 10
   **max: 600**
   convert_types:
     - str
   description: Minimum delay (10-600).
```
right?

PR #4083, opened 2024-06-06
Refactor: Improve handling of missing libs in Ansible plugins

ClausHolbechArista:
## Change Summary

Improve handling of missing libs in Ansible plugins

Currently only the default filter which is using pyavd.

This change will be needed for moving other filters since they are imported in various python code.


ClausHolbechArista:
will move this into one of the pending prs instead.

PR #4169, opened 2024-07-01
Feat(eos_designs): Underlay OSPF authentication

jonxstill:
## Change Summary

Automatic encryption of underlay OSPF message digest keys for use in underlay.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Because OSPF message-digest keys are encrypted in EOS configurations, using the interface name as the key, it requires inline Jinja2 (or offline encryption of keys) to generate the type 7 passwords required.  This PR automates this process.

Schema Changes:
```
underlay_ospf_authentication:
  enabled: true
  message_digest_keys:
    - id: 1
      hash_algorithm: sha512
      key: arista123
```

## How to test
Molecule tests added in `eos_designs_unit_tests` and `evpn_underlay_ospf_overlay_ebgp`.
Test manually by adding above YAML at fabric level with ospf configured as the underlay.

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,40 @@
...
+              - sha1
+              - sha256
+              - sha384
+              - sha512
+              default: sha512
+            key:
+              type: str
+              description: |
+                Key password.
+                Only plaintext passwords are supported here as the password will need to be encrypted for every underlay interface.
```
ClausHolbechArista: Are we sure we need this? The key could be set under the ospf process instead of per-interface. I think the process id is then used in the hash, so the user could leverage our encrypt filter to generate the correct hash.

jonxstill: As discussed just now, EOS only supports per-interface authentication, not per-process or per-area.

gmuloc: Can we reword this to state that AVD is doing the job of encrypting the password? here it feels like `will need to be encrypted` implies extra work required.

Something like:
```suggestion
                Only plaintext passwords are supported here, the per-interface hashed password is automatically created by `eos_designs` using the relevant keys.
```

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -99,6 +100,33 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlay) -> list | None:
...
+                            if not ("id" in ospf_key and "key" in ospf_key):
+                                continue
+
+                            ospf_keys.append(
+                                {
+                                    "id": ospf_key["id"],
+                                    "hash_algorithm": ospf_key.get("hash_algorithm", "sha512"),
+                                    "key": encrypt(
+                                        ospf_key["key"],
+                                        passwd_type="ospf_message_digest",
```
ClausHolbechArista:
```suggestion
                                        passwd_type="ospf_message_digest",  # NOSONAR
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,40 @@
...
+  underlay_ospf_authentication:
+    documentation_options:
+      table: ospf-settings
+    type: dict
+    keys:
+      enabled:
+        type: bool
+        default: false
+      message_digest_keys:
+        type: list
```
ClausHolbechArista, line 19: Should we add some min length and max length on this list? As far as I understand the doc, EOS supports one or two keys.
```suggestion
        type: list
        required: true
        primary_key: id
        min_length: 1
        max_length: 2
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,40 @@
...
+              type: str
+              valid_values:
+              - md5
+              - sha1
+              - sha256
+              - sha384
+              - sha512
+              default: sha512
+            key:
+              type: str
```
ClausHolbechArista, line 41:
```suggestion
              type: str
              required: true
              convert_types:
                - int
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,40 @@
...
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  underlay_ospf_authentication:
+    documentation_options:
+      table: ospf-settings
+    type: dict
+    keys:
+      enabled:
+        type: bool
```
ClausHolbechArista, line 15:
```suggestion
        type: bool
        required: true
```

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/MLAG_OSPF_TESTS.yml
```diff
@@ -1,5 +1,10 @@
 ---
 underlay_routing_protocol: ospf
+underlay_ospf_authentication:
+  enabled: true
+  message_digest_keys:
+    - id: 1
+      key: arista123arista123
```
ClausHolbechArista: please add two keys here

python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,48 @@
...
+              - sha256
+              - sha384
+              - sha512
+              default: sha512
+            key:
+              type: str
+              required: true
+              convert_types:
+              - int
+              description: |
```
ClausHolbechArista:
```suggestion
              description: |-
```

ClausHolbechArista (approved):
Looks good to me overall. Suggesting improving the test a bit for two keys but otherwise all good.

python-avd/pyavd/_eos_designs/schema/schema_fragments/underlay_ospf_authentication_schema.yml
```diff
@@ -0,0 +1,48 @@
...
+              valid_values:
+              - md5
+              - sha1
+              - sha256
+              - sha384
+              - sha512
+              default: sha512
+            key:
+              type: str
+              required: true
```
ClausHolbechArista, line 42:
```suggestion
              required: true
              max_length: 16
              min_length: 1
```

gmuloc (approved):
LGTM

PR #4199, opened 2024-07-10
Fix(plugins): Prevent deprecation warnings when deprecated filters are not used

gmuloc:
## Change Summary

in 4.9.0 3 filters have been deprecated to be removed in 5.0.0

When running AVD and not using the filters in the playbook, users sees the deprecation warnings and they cannot do anything about it. This is because of the way ansible loads filters. A workaround has been found in this PR where all the deprecated filters are grouped in the same python file and the documentation is kept on the side.

## Related Issue(s)

Fixes issue reported on slack and in the field

## Component(s) name

`plugins`

## Proposed changes

cf summary +

* updated pre-commit to render documentation from yaml files for these filters and ignore the deprecated_filter.py file

## How to test

when running in molecule no more warnings.

### Repository Checklist


pmprado (approved):
Tested on a Caravan repository and all the warnings were gone.

carlbuchmann (approved):
LGTM + added test in deprecated vars to ensure warning is raised:
image

PR #4205, opened 2024-07-11
Feat(eos_designs): Add keys to set mlag_peer_ipv4 and mlag_peer_l3_ipv4 directly

### LINKED ISSUE #4202: Feat(eos_designs): Create options to set mlag_peer_ipv4 and mlag_peer_l3_ipv4 direcly
### Enhancement summary

Right now eos_designs takes pools for mlag ip addressing. The idea is to set these statically (like loopback interfaces #2719).

eos_designs

### Use case example

This would be useful in brownfield deployments or when using IPAM tools such as Netbox. 

Have an option to statically set the IP done within ipAddressing model that takes precedence over the pool.

_No response_

### Additional context

_No response_

### Contributing Guide

jmussmann:
This PR adds two node level keys:

mlag_peer_ipv4 and mlag_peer_l3_ipv4 to set the addresses directly without using pools.

## Related Issue(s)

Fixes #4202 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- The two keys take precedence over the _pool version
- AVD does not check if they are unique (or valid - it only knows them as string)

## How to test

added a molecule test in DC1_FABRIC.yml

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/DC1-SVC3B.yml
+++ b/ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/DC1-SVC3B.yml
@@ -45,7 +45,7 @@ router_bgp:
     - name: EVPN-OVERLAY-PEERS
       activate: false
   neighbors:
-  - ip_address: 10.255.251.6
+  - ip_address: 192.168.44.41
     peer_group: MLAG-PEERS
     peer: DC1-SVC3A
     description: DC1-SVC3A
@@ -118,7 +118,7 @@ router_bgp:
     - source_protocol: connected
       route_map: RM-CONN-2-BGP-VRFS
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -138,7 +138,7 @@ router_bgp:
     - source_protocol: connected
       route_map: RM-CONN-2-BGP-VRFS
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -157,7 +157,7 @@ router_bgp:
     redistribute_routes:
     - source_protocol: connected
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -179,7 +179,7 @@ router_bgp:
     - source_protocol: connected
       route_map: RM-CONN-2-BGP-VRFS
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -218,7 +218,7 @@ router_bgp:
     redistribute_routes:
     - source_protocol: connected
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -237,7 +237,7 @@ router_bgp:
     redistribute_routes:
     - source_protocol: connected
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -256,7 +256,7 @@ router_bgp:
     redistribute_routes:
     - source_protocol: connected
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
@@ -275,7 +275,7 @@ router_bgp:
     redistribute_routes:
     - source_protocol: connected
     neighbors:
-    - ip_address: 10.255.251.6
+    - ip_address: 192.168.44.41
       peer_group: MLAG-PEERS
     updates:
       wait_install: true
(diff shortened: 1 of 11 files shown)


python-avd/pyavd/_eos_designs/ip_addressing/__init__.py
```diff
@@ -139,8 +149,13 @@ def mlag_l3_ip_primary(self) -> str:
         """
         Return IP for L3 Peerings for MLAG Primary
 
+        If "mlag_peer_l3_ipv4_address" is set, it is used.
         Default pool is "mlag_peer_l3_ipv4_pool"
         """
+
+        if self._mlag_peer_l3_ipv4_address:
+            return self._mlag_peer_l3_ipv4_address
```
gmuloc, line 157: so this is not 100% correct - the problem here is that you will end up with the same value for primary and secondary which should not be the case

python-avd/pyavd/_eos_designs/ip_addressing/__init__.py
```diff
@@ -85,8 +85,13 @@ def mlag_ip_primary(self) -> str:
         """
         Return IP for MLAG Primary
 
+        If "mlag_peer_ipv4_address" is set, it is used.
         Default pool is "mlag_peer_ipv4_pool"
         """
+
+        if self._mlag_peer_ipv4_address:
+            return self._mlag_peer_ipv4_address
```
gmuloc, line 93: same as below - this is not really correct to always pick this address for both primary and secondary

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -142,11 +142,15 @@ l3leaf:
           id: 4
           mgmt_ip: 192.168.200.108/24
           mac_address: '0c:1d:c0:1d:62:01'
+          mlag_peer_ipv4_address: 192.168.44.41
+          mlag_peer_l3_ipv4_address: 192.168.44.41
```
gmuloc, line 146: let's not use the same address for non l3 and l3 in the test to avoid confusion in the results (like 192.168.45.1 for the next one)

gmuloc: Also the proper `192.168.44.40/31` and `192.168.44.41/31` rather than 41 and 42 if we want it to be "networkingly" correct

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/DC1-SVC3B.yml
```diff
@@ -1494,9 +1494,9 @@ prefix_lists:
 - name: PL-MLAG-PEER-VRFS
   sequence_numbers:
   - sequence: 10
-    action: permit 10.255.251.6/31
-  - sequence: 20
     action: permit 172.31.11.6/31
+  - sequence: 20
+    action: permit 192.168.44.42/31
```
gmuloc, line 1499: so here really we should be careful regarding the generated config as it should be really the network IP of the subnet - this comes from the fact that the primary IP is the same as the secondary IP

python-avd/pyavd/_eos_designs/ip_addressing/__init__.py
jmussmann: Yes, thanks for the hint. Maybe it would be better to set the IPs on the group level like this:

```yml
# If two nodes (and only two) are in the same node_group, they will automatically form an MLAG pair
  node_groups:
    # Definition of a node group that will include two devices in MLAG.
    # Definitions under the group will be inherited by both nodes in the group
    - group: DC1_L3_LEAF1
      # ASN to be used by BGP for the group. Both devices in the MLAG pair will use the same BGP ASN
      bgp_as: 65101
      mlag_peer_primary_ipv4_address: 192.168.42.40
      mlag_peer_secondary_ipv4_address: 192.168.42.41
      mlag_peer_l3_primary_ipv4_address: 192.168.43.40
      mlag_peer_l3_secondary_ipv4_address: 192.168.43.41
      nodes:
        # Definition of hostnames under the node_group
        - name: dc1-leaf1a
          id: 1
          mgmt_ip: 172.16.1.101/24
          loopback_ipv4_address: 172.16.1.101/24
          vtep_loopback_ipv4_address: 172.16.1.101/24
          # Definition of the port to be used in the uplink device facing this device.
          # Note that the number of elements in this list must match the length of 'uplink_switches' as well as 'uplink_interfaces'
          uplink_switch_interfaces:
            - Ethernet1
            - Ethernet1
        - name: dc1-leaf1b
          id: 2
          mgmt_ip: 172.16.1.102/24
          uplink_switch_interfaces:
            - Ethernet2
            - Ethernet2
```
An alternative might be something like this, but I think the "group" level would be better. What do you think?

```python
def mlag_ip_primary(self) -> str:
  if self.shared_utils.mlag_role == "primary":
              if self._mlag_peer_ipv4_address:
                  return self._mlag_peer_ipv4_address
          else:
              if self.shared_utils.get_mlag_peer_fact("mlag_peer_ipv4_address"):
                  return self.shared_utils.get_mlag_peer_fact("mlag_peer_ipv4_address")
```

ClausHolbechArista: IMO we should just have the direct ip options for those who wants to control things. They have to set it per node or build inline jinja.
Note we already have the option of reusing the same mlag subnet under `fabric_ip_addressing`, and this could be used together with setting `mlag_peer_ipv4_pool` per node group to override things per group.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
ClausHolbechArista: Maybe we could just expose `mlag_peer_ipv4_subnet` and always assign the first and second host address of the given subnet. That would make it easier to use group inheritance etc.

PR #4206, opened 2024-07-11
Doc: Fix incorrect keys specified in default_mgmt_method

carlbuchmann:
## Change Summary

Documentation fix of incorrect keys specified in `default_mgmt_method`


gmuloc (approved):
Indeed these two keys are pending PRs from @ClausHolbechArista which have not merged yet: #3301 and #3283 respectively

ClausHolbechArista (changes requested):
The proposed changes are misleading, since the keys in question do *not* take this setting into account. Only their new replacements would.

python-avd/pyavd/_eos_designs/schema/schema_fragments/default_mgmt_method.schema.yml
```diff
@@ -12,8 +12,8 @@ keys:
     type: str
     description: |
       `default_mgmt_method` controls the default VRF and source interface used for the following management and monitoring protocols configured with `eos_designs`:
-        - `cv_settings`
-        - `dns_settings`
+        - `cvp_instance_ips`
+        - `name_servers`
```
gmuloc:
```suggestion
```

ClausHolbechArista (approved):

PR #4208, opened 2024-07-12
Fix(eos_designs)!: VARPv6 config is not generated even when "ipv6_enable: true" is specified

### LINKED ISSUE #4203: eos_designs: VARPv6 config is not generated even when "ipv6_enable: true" is specified
### Issue Summary

When I have "ipv6_enable: true" set I expect the ipv6_virtual_router_addresses config to work, but it does not. Reason is that in the code it is only checked if an "ipv6 address" is configured on the SVI:

Current code:

```
        # Only set VARPv6 if ipv6_address is set
        if vlan_interface_config["ipv6_address"] is not None:
            vlan_interface_config["ipv6_virtual_router_addresses"] = svi.get("ipv6_virtual_router_addresses")
            _check_virtual_router_mac_address(vlan_interface_config, ["ipv6_virtual_router_addresses"])
```

eos_designs

### How do you run AVD ?

Ansible CLI with AVD Runner

### Steps to reproduce

```shell
tenants:
  - name: All-Switches
    mac_vrf_vni_base: 10000

    vrfs:
      - name: HOSTING
        vrf_id: 1
        mlag_ibgp_peering_vlan: 4093
        ...
        svis:
          - id: 10
            name: hosting
            tags: ['fabric']
            description: hosting SVI
            enabled: true
            ip_address_virtual: 1.1.1.1/24
            ipv6_enable: true
            ipv6_virtual_router_addresses:
              - fe80::1
```

### Relevant log output

_No response_

### Contributing Guide

bjmeuer:
## Change Summary

VARPv6 is now configured not only when "ipv6_address" is given, it also is configured if "ipv6_enable:true" is given

## Related Issue(s)

Fixes #4203 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
before for VARPv6 this was checked:
`if vlan_interface_config["ipv6_address"] is not None:`

after the change it is:
`if vlan_interface_config["ipv6_address"] is not None or vlan_interface_config["ipv6_enable"]:`

## How to test
```
tenants:
  - name: All-Switches
    mac_vrf_vni_base: 10000

    vrfs:
      - name: HOSTING
        vrf_id: 1
        mlag_ibgp_peering_vlan: 4093
        ...
        svis:
          - id: 10
            name: hosting
            tags: ['fabric']
            description: hosting SVI
            enabled: true
            ip_address_virtual: 1.1.1.1/24
            ipv6_enable: true
            ipv6_virtual_router_addresses:
              - fe80::1
```

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py
```diff
@@ -126,7 +126,7 @@ def _check_virtual_router_mac_address(vlan_interface_config: dict, variables: li
                 vlan_interface_config["pim"] = {"ipv4": pim_config_ipv4}
 
         # Only set VARPv6 if ipv6_address is set
-        if vlan_interface_config["ipv6_address"] is not None:
+        if vlan_interface_config["ipv6_address"] is not None or vlan_interface_config["ipv6_enable"]:
```
gmuloc, line 132: We should change the comment above right?

gmuloc: so now this also implies that the next condition in line 134 is going to be true more often - not sure about the implications?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/varpv6.yml
```diff
@@ -0,0 +1,52 @@
+# Testing duplicate VRF definitions according to the following rules documented in the schema.
+# Also see eos_designs_negative_unit_tests for testing of the raised errors.
+
+# It is recommended to only define a VRF in one Tenant. If the same VRF name is used across multiple tenants and those tenants
+# are accepted by `filter.tenants` on the node, any object set under the duplicate VRFs must either be unique or be an exact match.
```
gmuloc: is this relevant?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/varpv6.yml
```diff
@@ -0,0 +1,52 @@
...
+          - TENANT1
+
+tenants:
+  - name: TENANT1
+    mac_vrf_vni_base: 10000
+    vrfs:
+      - name: VRF1
+        vrf_id: 1
+        svis:
+          - id: 100
```
gmuloc, line 23: can you add comment for each test about what you are testing? also you have twice the name `VARPv6_not_configured_2` for VLAN 200 and 400 - is it what you need here?

python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py
bjmeuer: adjusted the comment. Yes it implies that condition in 134 does match more often, but in my opinion that is the correct way as if in CLI you configure "ipv6 enable" and "ipv6 virtual router address..." it will activate VARPv6. Additionally I think in line 134 we have a mistake today as it checks the IPv4 VARP but should check the IPv6 VARP. I fixed that.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/varpv6.yml
bjmeuer: added some comments

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/varpv6.yml
```diff
@@ -0,0 +1,50 @@
...
+            enabled: true
+            name: VARPv6_configured_1
+            ipv6_virtual_router_addresses:
+              - fe80::1
+            nodes:
+              - node: varpv6
+                ipv6_address: 2000:1000::2/64
```
gmuloc: let's use `2001:db8::/32 a` subnet for our tests. (we may. not always be doing it yet but we aim to remove any potential real IP address from our documentation and tests)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/varpv6.yml
```diff
@@ -0,0 +1,50 @@
...
+            ipv6_virtual_router_addresses:
+              - fe80::1
+          # VARPv6 is configured if ipv6_enable and ipv6_virtual_router_addresses is set
+          - id: 300
+            enabled: true
+            name: VARPv6_configured_2
+            ipv6_enable: true
+            ipv6_virtual_router_addresses:
+              - fe80::1
+          # VARPv6 is not configured if ipv6_enable is set to false
```
gmuloc, line 44: Could we add a test where VARPv6 is not configured (because ipv6_enable is False and ipv6_address is not set) and it leads to `ipv6_address_virtuals` being configured?

python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py
gmuloc: thanks for fixing the bug.
This should be mention somewhere as sadly it is a breaking change.
Before that I guess it was possible to end up with both vipv6 irtual-router address and IPv6 address virtual - I will test it out.

gmuloc (dismissed):
So I think we would need to adjust the schema description for the key under SVI:

```
                # IPv6 VARP addresses.
                # Requires an IPv6 address to be configured on the SVI.
                # If ipv6_address_virtuals is also set, ipv6_virtual_router_addresses will take precedence
                # _if_ there is an ipv6_address configured for the node.
                ipv6_virtual_router_addresses:
```

The second line needs to be changed. Also the fourth line is not correct anymore. Can you please adjust the documentation with what you feel makes sense?


**Point of attention**

Right now I feel like this Fix is a breaking change to the current behavior for at least one case, the existing behavior is that:
* If `ipv6_enable` is true BUT no ipv6 address is set, then `ipv6_address_virtuals` (Anycast IPv6 addresses) are configured, as `ipv6_virtual_router_address` is ignored if present.
* With your change, `ipv6_virtual_router_address` is not ignored anymore in this situation (which is what your fix is about) and so we would not be configuring the Anycast IPv6 addresses anymore.

From a behavior point of view it can break existing deployment where people would have both `ipv6_address_virtuals` and `ipv6_virtual_router_address` set (probably a bad input choice as one is useless today), no IPv6 address defined, and `ipv6_enable: true`. I know this is probably a corner case but as we are following semver this is something we need to consider.

@ClausHolbechArista for review when he is back.

gmuloc:
The following inputs would give a different outcome before/ after the PR:

### Change of behavior for `ipv6_enable: true` 

```yaml
          - id: 666
            enabled: true
            name: VARPv6_vs_AnycastV6
            ipv6_enable: true
            ipv6_virtual_addresses:
              - 2001:db8::cafe
            ipv6_virtual_router_addresses:
              - fe80::1
 ```

### Bug fix where AVD is wrongly checking for `ip_router_virtual_address` instead of `ipv6_router_virtual_addresses`

With the config below, because AVD is wrongly checking for IPv4 VARP to decide or not to configure IPv6 Anycast vurtal addresses, this will generate IPv6 Anycast config before the PR but not after.

```yaml
          - id: 42
            enabled: true
            name: AnycastGW_checking_v4VARP_Bug
            # the next two keys will generate the IPv4 VARP
            ip_virtual_router_addresses:
              - 10.10.10.10
            nodes:
              - node: dummy
                 ip_address: 19.168.250.5/24
            # Then because of the bug the next key will generate the Anycast IPv6 virtual IPs
            ipv6_virtual_addresses:
              - 2001:db8::cafe
 ```

**Comment by ClausHolbechArista** [comment]
We will move this to 5.0 to avoid any concerns of breaking changes. We need to add some info in the release notes to tell about the potential breakage although it is a very small risk.
Moving the PR to draft until 4.10 has been released.

File: python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py (hunk 1 above)
Comment (ClausHolbechArista) [reply python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py:132]: I have updated the release notes with both behavioral changes.

File: ansible_collections/arista/avd/docs/release-notes/5.x.x.md [hunk 6]
```diff
@@ -66,6 +66,21 @@ for details.
 
 TODO: add level4 sections for breaking changes outlining the change and referring to porting guide for migration.
 
+#### ipv6_virtual_router_addresses are now configured if ipv6_enable is true on SVIs
+
+For SVIs with `ipv6_virtual_router_addresses` set, AVD versions below 5.0.0 only rendered in the configuration the addresses if `ipv6_address`
```
laxmikantchintakindi:
```suggestion
For SVIs with `ipv6_virtual_router_addresses` set, AVD versions below 5.0.0 only rendered the configuration if `ipv6_address`
```

laxmikantchintakindi (approved):

ClausHolbechArista (approved):
Great test cases with good descriptions. Thank you for fixing these issues.

PR #4209, opened 2024-07-12
CI: Add step to run make dep when installing pyavd in editable mode

gmuloc:
## Change Summary

When running editable install of pyavd, pickle files may not exist and so running AVD will fail

## Related Issue(s)

Reported in the field when running dev containers. When devcontainer is started without a specified branch, the following is done: https://github.com/aristanetworks/avd/blob/d8dcef092c789354ba16a1d0683a6c3100cb09fd/containers/dev/.devcontainer/entrypoint.sh#L19

running AVD fails because pickle files are missing. It is possible to fix this by running manually `make dep` or `pre-commit`

## Component(s) name

`pyavd`

## Proposed changes

Added a step in the custom build backend we use to run `make dep` on editable install as well

## How to test

1. Check issue does not appear in container anymore
2. run editable install and verifies pickle files and compiled templates are generated

### Repository Checklist

### DIFF
--- a/python-avd/scripts/custom_build_backend.py
+++ b/python-avd/scripts/custom_build_backend.py
@@ -66,9 +66,18 @@ def get_requires_for_build_wheel(config_settings=None):
     print("Fetch version from ansible.avd ansible collection and insert into __init__.py")
     _insert_version()
 
-    print("Running 'make dep' to vendor various scripts and templates from arista.avd ansible collection")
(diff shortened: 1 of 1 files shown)


ankudinov (approved):
LGTM. Tested in a codespace on `ci-make-dep-in-build-editable` branch - works great and fixes the issue with schema files not available in container.

ClausHolbechArista (approved):

PR #4230, opened 2024-07-19
Feat(eos_cli_config_gen): Add ptp profiles for ITU-T G8275 support

colinmacgiolla:
## Change Summary

Add support for settting the ITU-T G.8275.1 and .2 support, as well as allowing the handling of PTP frames on interfaces. See [here](https://www.arista.com/en/support/toi/eos-4-25-1f/14634-ptp-itu-t-g8275-1-support) for details on the feature.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.pyavd.eos_cli_config_gen`

## Proposed changes
On the PTP level, we are just adding the `profile` key, which can be set to `g8275_1` or `g8275_2` profiles.
On the interface level we need to indicate if we should recognise either the `forwardable` or `non-forwardable` PTP frames.

I've set the the following model on the port-channel and ethernet interface level:
```yaml
            profile:
              type: dict
              keys:
                g8275_1:
                  type: dict
                  keys:
                    destination_mac_address:
                      type: str
                      valid_values:
                        - forwardable
                        - non-forwardable
```

## How to test
Updated molecule tests to reflect device config

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/ethernet_interfaces.schema.yml
```diff
@@ -644,6 +644,20 @@ keys:
             delay_mechanism:
               type: str
               valid_values: ["e2e", "p2p"]
+            profile:
+              type: dict
+              keys:
+                g8275_1:
+                  type: dict
+                  keys:
+                    destination:
```
ClausHolbechArista: I think we should collapse into `destination_mac_address`.

ClausHolbechArista (changes requested):
Minor data model request but other than that looks good to me.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/ethernet_interfaces.schema.yml
ClausHolbechArista: We could even collapse further, since there are no other options under profile, but maybe it is good to keep those more open.

colinmacgiolla: yeah, I've flattened it to `destination_mac_address`, but I think its better to leave the option of additional profiles and options open

ClausHolbechArista (dismissed):

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ptp.j2
```diff
@@ -37,6 +37,9 @@ ptp mode {{ ptp.mode }} one-step
 ptp mode {{ ptp.mode }}
 {%         endif %}
 {%     endif %}
+{%     if ptp.profile is arista.avd.defined %}
```
MaheshGSLAB: for me the config order as below, so if this is same in your platform then move this logic just before to ptp ttl
```
ptp clock-identity 11:11:11:ff:ff:11:11:11
ptp domain 17
ptp message-type event dscp 46 default
ptp message-type general dscp 36 default
ptp mode boundary one-step
ptp priority1 101
ptp priority2 102
ptp profile g8275.1    <--------------
ptp ttl 12
ptp forward-unicast
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ethernet-interfaces.j2
```diff
@@ -685,6 +685,9 @@ interface {{ ethernet_interface.name }}
 {%     if ethernet_interface.ptp.enable is arista.avd.defined(true) %}
    ptp enable
 {%     endif %}
+{%     if ethernet_interface.ptp.profile.g8275_1.destination_mac_address is arista.avd.defined %}
```
MaheshGSLAB: Config order is different from running config. Please move the logic after ptp delay-req interval. I know other options are not even in order but moving profile after delay-req interval would be better 
```
interface Ethernet1
   description P2P_LINK_TO_BORDER1-DC1_Ethernet3
   mtu 9214
   no switchport
   ip address 10.100.0.129/31
   ptp enable
   ptp announce interval 3
   ptp announce timeout 9
   ptp delay-req interval -7
   ptp profile g8275.1 destination mac-address forwardable
   ptp sync-message interval 1
```
Same for port-channel

ClausHolbechArista (approved):

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ptp.j2
```diff
@@ -21,6 +21,9 @@ ptp priority2 {{ ptp.priority2 }}
 {%     if ptp.ttl is arista.avd.defined %}
 ptp ttl {{ ptp.ttl }}
 {%     endif %}
+{%     if ptp.profile is arista.avd.defined %}
+ptp profile {{ ptp.profile }}
+{%     endif %}
```
MaheshGSLAB:
```suggestion
{%     if ptp.profile is arista.avd.defined %}
ptp profile {{ ptp.profile }}
{%     endif %}
{%     if ptp.ttl is arista.avd.defined %}
ptp ttl {{ ptp.ttl }}
{%     endif %}
```

MaheshGSLAB (changes requested):

colinmacgiolla:
@MaheshGSLAB - I'm not understanding why the negative test is failing on something that has nothing to do with my changes?

ClausHolbechArista:
> @MaheshGSLAB - I'm not understanding why the negative test is failing on something that has nothing to do with my changes?

Fixed now. Something sneaked in from another PR.

ClausHolbechArista (approved):

MaheshGSLAB (approved):

PR #4258, opened 2024-07-24
Fix(eos_cli_config_gen): Add a pseudo task to handle tags on import role

### LINKED ISSUE #4232: Fix(eos_cli_config_gen): When using the role in a playbook and adding a tag on it, the default values for generate_device_config / doc are False
### Issue Summary

When refactored for 4.9 - there was an attempt to conserve ansible tag behavior, this works well as long as no other tags is applied on the role in the playbook
```
      generate_device_config: "{{ ansible_run_tags is arista.avd.contains(['all', 'build', 'provision']) and ansible_skip_tags is not arista.avd.contains(['build', 'provision']) }}"
      generate_device_doc: "{{ generate_device_documentation | arista.avd.default(true) and ansible_run_tags is arista.avd.contains(['all', 'build', 'provision', 'documentation']) and ansible_skip_tags is not arista.avd.contains(['build', 'provision', 'documentation']) }  }"
```

otherwise these default to false and this is sad

eos_cli_config_gen

### How do you run AVD ?

None

### Steps to reproduce

```shell
Run eos_cli_config_gen in a playbook where the role is imported with a tag. Run with -vvv to see that the action plugin never renders config
```

### Relevant log output

_No response_

### Contributing Guide

gmuloc:
## Change Summary

With 4.9 move to pyavd, there was a requirement to keep backward compatibility when using tags for generating documentation / skippind documentation. However the pushed version did not cater for tags being applied on an imported role in a playbook (which is not a good idea in general) and this broke some playbook. This fixes it with a hack until we can deprecate this in 5.0 (and  probably remove the tag behavior)

## Related Issue(s)

Fixes #4232

## Component(s) name

## Proposed changes

add an intermediate task with `never` that would be run if the role is applied a tag when imported

## How to test

Add a tag on a task importing eos_cli_config_gen in a playbook and see the config and doc generated

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_cli_config_gen_deprecated_vars/create.yml
```diff
@@ -9,3 +9,4 @@
       ansible.builtin.import_role:
         name: arista.avd.build_output_folders
       run_once: true
+      tags: [test]
```
gmuloc, line 13: Add TODOs for AVD 5.0 on all of the tags

gmuloc: done in 9a5ad9a

PR #4298, opened 2024-07-31
Feat(plugins): Add strict mode and ignore_case flags to natural_sort filter

### LINKED ISSUE #4242: Feat(plugins): Allow natural_sort to be case sensitive for strings
### Enhancement summary

Today, `natural_sort` filter is case insensitive (calling `lower` on the keys to sort)
This enhancement is suggesting adding a param to make it case sensitive (not by default) in order to be able to cater to some area of EOS CLI

plugins

### Use case example

EOS configuration
```
  application traffic recognition
     application ipv4 Data-App
        protocol tcp udp destination port field-set Data-port
     !
     application ipv4 Voice-App
        protocol tcp udp destination port field-set Voice-port
     !
     application ipv4 prefix-App
        source prefix field-set prefix
     !
```

AVD generated configuration

```
  application traffic recognition
     !
     application ipv4 Data-App
        protocol tcp destination port field-set Data-port
        protocol udp destination port field-set Data-port
     !
     application ipv4 prefix-App
        source prefix field-set prefix
     !
     application ipv4 Voice-App
        protocol tcp destination port field-set Voice-port
        protocol udp destination port field-set Voice-port
     !
```

The different ordering comes from the case insensitive sorting

case sensitive sorting flag

_No response_

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

This PR adds two new flags to the `natural_sort` filter
1. a `strict` flag to be able to raise when the `sort_key` is missing in one item in the iterable dict or Namespace (default is `False` for backward compatibility). 
2. a `ignore_case` flag to be able to not ignore the case. Today natural_sort calls `lower` by default on some strings and this leads to AVD generating config not in the same order as EOS which does not use `lower`.

## Related Issue(s)

Fixes #4242

## Component(s) name

`plugins`

## Proposed changes

Adding two flags to the function, with backward compatible default values to avoid impacting the rest of AVD
The goal is then to review the config ordering we have today against EOS and to use the flags where appropriate.

## How to test

pytests have been added

### Note to reviewers

- [x] ignore_case: False for a list of string effectively reimplements builtin `sort` from Jinja -> is it worth it?

### Repository Checklist


python-avd/pyavd/j2filters/natural_sort.py
```diff
@@ -9,35 +9,59 @@
...
-    Args:
+def natural_sort(iterable: list | dict | str | None, sort_key: str | None = None, *, strict: bool = False, ignore_case: bool = True) -> list:
+    """Sorts an iterable in a natural (alphanumeric) order.
+
+    Parameters
+    ----------
         iterable (list | dict | str | None): Input iterable.
         sort_key (str | None, optional): Key to sort by, defaults to None.
-    Returns:
+        strict (bool, optional): If strict is True, raise an error is the sort_key is missing.
```
ClausHolbechArista, line 39: I think strict should be default. Since we are introducing this in a major version, we can just document the change

ClausHolbechArista (approved):

ClausHolbechArista (changes requested):
approved by mistake

python-avd/pyavd/j2filters/natural_sort.py
gmuloc: may need to change a couple of things around - let's see what breaks

ClausHolbechArista:
The special catch in the template of `hardware_counters` should be removed in 5.0.0. The negative test of that is causing the tests here to fail.

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

gellis713:
Sorry but I feel like there is a better way to name this. "ignore_case" flag would naturally imply that we're ignoring case, which would further imply the default behavior is being case-sensitive. The fact that "ignore_case" must be set so that we would NOT ignore_case is odd behavior in my opinion.

PR #4304, opened 2024-07-31
Feat(eos_designs): Support for L3 Inband ZTP

jrecchia1029:
## Change Summary

Adding L3 In band ZTP functionality

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Enable L3 In Band ZTP by making the network device upstream of the ZTP a dhcp server that will allocate the appropriate IP address on the connected p2p interface.

## How to test

Updated eos_designs_unit_tests for the INBAND_MGMT_TESTS group 

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/base/dhcp_server.py
```diff
@@ -0,0 +1,105 @@
...
+                    uplink["peer"] == self.shared_utils.hostname
+                    and uplink["type"] == "underlay_p2p"
+                    and uplink.get("ip_address")
+                    and "unnumbered" not in uplink["ip_address"]
+                    and get(peer_facts, "inband_ztp")
+                ):
+                    subnet = {
+                        "subnet": str(ip_network(f"{uplink['peer_ip_address']}/{uplink['prefix_length']}", strict=False)),
+                        "ranges": [{"start": str(uplink["ip_address"]), "end": str(uplink["ip_address"])}],
+                        "name": f"p2p to {peer}-{uplink['interface']}",
```
jrecchia1029: add `inband ztp` to description and move all logic to underlay

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,116 @@
...
+        if ztp_bootfile := self._ipv4_ztp_boot_file:
+            dhcp_server["tftp_server"] = {"file_ipv4": ztp_bootfile}
+        # Set DNS servers
+        if dns_servers := self._dns_servers:
+            dhcp_server["dns_servers_ipv4"] = dns_servers
+        # Set NTP servers
+        if ntp_servers := self._ntp_servers:
+            dhcp_server["ipv4_vendor_options"] = [ntp_servers]
+        dhcp_servers.append(dhcp_server)
+        return dhcp_servers
```
jrecchia1029: Would be good to get these DHCP options based on the downstream switch's values instead of this switch's values.

ClausHolbechArista: This may or may not be possible, depending on the capabilities of our dhcp server. The downstream switches may give us different server info. So it would lead to different properties per scope.

jrecchia1029: Yes, you are correct.  Assuming we're ZTPing downstream devices within the same fabric, the DHCP option values gathered from the upstream device should be sufficient for the downstream devices.

emilarista (commented):
Data model, code and documentation looks good, have done some testing with ACT and it seems to work well. Could it be worth mentioning in the documentation that this knob is not compatible with unnumbered fabrics?

emilarista (approved):
LGTM!

ClausHolbechArista:
Please rebase and resolve conflicts

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+        """Returns the list of name servers."""
+        dns_servers = get(self._hostvars, "name_servers")
+        if not dns_servers:
+            return None
+
+        return dns_servers
+
+    @cached_property
+    def _ntp_servers(self) -> list | None:
+        """Returns the list of name servers."""
```
alexeygorbunov:
```suggestion
        """Returns the list of NTP servers."""
```

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+        custom_bootfile = get(self._hostvars, "inband_ztp_bootstrap_file")
+        if custom_bootfile:
+            return custom_bootfile
+        cvp_instance_ips = get(self._hostvars, "cvp_instance_ips")
+        if not cvp_instance_ips:
+            return None
+        if "arista.io" in cvp_instance_ips[0]:
+            return "https://www.arista.io/ztp/bootstrap"
+
+        return f"https://{cvp_instance_ips[0]}/ztp/bootstrap"
```
alexeygorbunov: Redirection of the EoS running <4.30.0F to US-1A CVaaS (www.arista.io) would fail if switch is mapped to another region (only EOS >=4.30.0F has embedded support for redirector service). It might be better just to send it to whatever first CV FQDN/IP is configured under `cvp_instance_ips`?

```suggestion

        return f"https://{cvp_instance_ips[0]}/ztp/bootstrap"
```

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+            return None
+
+        ntp_servers = []
+        for ntp_server in ntp_servers_settings:
+            # Validate NTP server IP address
+            try:
+                ntp_server_ip = IPv4Address(ntp_server["name"])
+            except AddressValueError:
+                continue
+            ntp_servers.append(str(ntp_server_ip))
```
alexeygorbunov: This would only append item if it is set as an IP and not an FQDN. In case of FQDN - what if we try to resolve it (on a runner) into an IP and append resolved IP (if we succeeded)? Lack of NTP sync during CVaaS onbording may not be super critical anymore as this gets synced over TA

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+    def _dns_servers(self) -> list | None:
+        """Returns the list of name servers."""
+        dns_servers = get(self._hostvars, "name_servers")
+        if not dns_servers:
+            return None
+
+        return dns_servers
+
+    @cached_property
+    def _ntp_servers(self) -> list | None:
```
alexeygorbunov: Function returns a dict | None

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+                continue
+            ntp_servers.append(str(ntp_server_ip))
+
+        if ntp_servers:
+            return {"vendor_id": "NTP", "sub_options": [{"code": 42, "array_ipv4_address": ntp_servers}]}
+
+        return None
+
+    @cached_property
+    def dhcp_servers(self) -> dict | None:
```
alexeygorbunov: Function is returning a list | None

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,114 @@
...
+                    and uplink["type"] == "underlay_p2p"
+                    and uplink.get("ip_address")
+                    and "unnumbered" not in uplink["ip_address"]
+                    and get(peer_facts, "inband_ztp")
+                ):
+                    subnet = {
+                        "subnet": str(ip_network(f"{uplink['peer_ip_address']}/{uplink['prefix_length']}", strict=False)),
+                        "ranges": [{"start": str(uplink["ip_address"]), "end": str(uplink["ip_address"])}],
+                        "name": f"inband ztp for {peer}-{uplink['interface']}",
+                        "default_gateway": f"{uplink['peer_ip_address'].split('/')[0]}",
```
alexeygorbunov: do we expect `/<prefix_length>` to be returned inside `uplink['peer_ip_address']`?

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
alexeygorbunov, line 1: @ClausHolbechArista should file name 1to1 match the name of the returned/set variable? aka dhcp_server***s*** (plural)?

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
jrecchia1029: I'm not sure but regardless of if it is or isn't, since we're picking out the first item from the split operation (which will always be the IP address without the subnet), we'll wind up with the value we need.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
ClausHolbechArista: yes that would be the best.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
ClausHolbechArista: We should not force someone to use IP as NTP server. DNS names are fine, so please remove this check. We should make sure to also give the dns options in dhcp in this feature.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
alexeygorbunov: right, it's just that `uplink['peer_ip_address']` is used on lines 42 and 45 and line 42 is not expecting it to have a prefix length (no splitting there) but line 45 does.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
jrecchia1029: EOS doesn't accept an FQDN for the NTP server DHCP option value.  That's why there is the check.  So we'd have to resolve the FQDN to an IP like @alexeygorbunov suggested.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
```diff
@@ -0,0 +1,115 @@
...
+        """Returns the file name to allow for ZTP to CV."""
+        custom_bootfile = get(self._hostvars, "inband_ztp_bootstrap_file")
+        if custom_bootfile:
+            return custom_bootfile
+        cvp_instance_ips = get(self._hostvars, "cvp_instance_ips")
+        if not cvp_instance_ips:
+            return None
+
+        if "arista.io" in cvp_instance_ips[0]:
+            cvp_instance_ips[0] = f"www.{cvp_instance_ips[0].replace('https://', '').replace('www.', '').replace('apiserver.', '')}"
```
jrecchia1029: @ClausHolbechArista Wondering what the best way to handle this is

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
jrecchia1029: Oh I see.  Good point. I'll commit the suggestion.

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
ClausHolbechArista: with regex

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
ClausHolbechArista: I would prefer not to resolve stuff during build, since we would no longer be idempotent. It sounds weird to me that we cannot use dns name here... I will unicast you.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/inband-mgmt-parent-vrf.yml
```diff
@@ -78,7 +104,27 @@ management_api_http:
   enable_vrfs:
   - name: MGMT
   enable_https: true
+ntp:
+  local_interface:
+    name: null
+    vrf: null
```
ClausHolbechArista, line 110: This looks wrong

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/inband-mgmt-spine1-ztp.cfg
```diff
@@ -0,0 +1,128 @@
...
+      name inband ztp for inband-mgmt-parent-Ethernet1
+      default-gateway 172.16.255.0
+   !
+   subnet 172.16.255.2/31
+      !
+      range 172.16.255.3 172.16.255.3
+      name inband ztp for inband-mgmt-parent-vrf-Ethernet1
+      default-gateway 172.16.255.2
+   !
+   vendor-option ipv4 NTP
```
ClausHolbechArista, line 25: please try to use regular option 42 instead of vendor option

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_server.py
ClausHolbechArista: So after discussion we settle on ignoring NTP servers with names, and then only error out if we have no NTP servers left.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/inband-mgmt-spine1-ztp.cfg
ClausHolbechArista: This will be kept for now, since regular options require EOS 4.31.0, which we would prefer not to require at this point.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/structured_configs/inband-mgmt-parent-vrf.yml
jrecchia1029: Maybe true but I think that is unrelated to this change.

alexeygorbunov: Looks like it's coming from [here](https://github.com/aristanetworks/avd/blob/71a101e7f722c855354d09e85578528ad5a368ad/python-avd/pyavd/_eos_designs/structured_config/base/ntp.py#L48-L54)

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_servers.py
```diff
@@ -0,0 +1,118 @@
...
+from functools import cached_property
+from ipaddress import AddressValueError, IPv4Address, ip_network
+
+from pyavd._errors import AristaAvdInvalidInputsError
+from pyavd._utils import get
+
+from .utils import UtilsMixin
+
+
+class DhcpServerMixin(UtilsMixin):
```
alexeygorbunov: Sorry, just to respect existing "naming convention"
```suggestion
class DhcpServersMixin(UtilsMixin):
```

python-avd/pyavd/_eos_designs/structured_config/underlay/__init__.py
```diff
@@ -5,6 +5,7 @@
 
 from .agents import AgentsMixin
 from .as_path import AsPathMixin
+from .dhcp_servers import DhcpServerMixin
```
alexeygorbunov:
```suggestion
from .dhcp_servers import DhcpServersMixin
```

python-avd/pyavd/_eos_designs/structured_config/underlay/__init__.py
```diff
@@ -41,6 +42,7 @@ class AvdStructuredConfigUnderlay(
     MplsMixin,
     AgentsMixin,
     IpAccesslistsMixin,
+    DhcpServerMixin,
```
alexeygorbunov:
```suggestion
    DhcpServersMixin,
```

alexeygorbunov (commented):
Last 3 cosmetic items. Rest LGTM

alexeygorbunov (approved):
LGTM

carlbuchmann (approved):
LGTM

PR #4390, opened 2024-08-26
Fix(eos_designs)!: Do not render EVPN address-family on MPLS devices unless `overlay_address_families` includes `evpn`

### LINKED ISSUE #4176: MPLS ISIS-LDP configuration enables EVPN address family
### Issue Summary

There is this MPLS ISIS-LDP guide example, which nicely goes over the required config to set this up: https://avd.arista.com/4.8/examples/isis-ldp-ipvpn/index.html

For some reason we create the RR-OVERLAY-PEERS peer group which is enabled for vpn-ipv4 address family as expected, but it is also activated for the EVPN address family in the RR and a cosmetic line in the PEs:

RR:
```
  address-family evpn
      neighbor RR-OVERLAY-PEERS activate
   !
   address-family vpn-ipv4
      neighbor MPLS-OVERLAY-PEERS activate
      neighbor RR-OVERLAY-PEERS activate
```
PE:
```
address-family evpn
```

We are seeing this when testing this for our solution guides and a project where we are delivering an MPLS core.

eos_designs

### How do you run AVD ?

### Steps to reproduce

```shell
Follow the guide to configure MPLS LDP-ISIS (also i have provided internally in the AVD field my full testing repository).
```

### Relevant log output

_No response_

### Contributing Guide

ClausHolbechArista:
## Change Summary

Do not render EVPN address-family on MPLS devices unless `overlay_address_families` includes `evpn`

## Related Issue(s)

Fixes #4176

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Check if overlay is EVPN + MPLS before configuring EVPN address family on MPLS nodes.

## How to test

Existing example exposed the issue on both RR and PE.

Tested manually by adding `overlay_address_families: [evpn, vpn-ipv4]` on the RR and it added back the EVPN address family.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -297,7 +298,7 @@ def _address_family_evpn(self: AvdStructuredConfigOverlay) -> dict:
             }
             address_family_evpn["neighbors"] = [{"ip_address": self._wan_ha_peer_vtep_ip(), "activate": True}]
 
-        return address_family_evpn
+        return address_family_evpn or None
```
gmuloc, line 301: need to adjust the return type in the function signature - but do we really want to return None rather than an empty dict as we tend to do everywhere else? Usually we would expect the caller to not render empty (strip_empty..)

gmuloc: toh

Ok we are doing this in main.. so not removing empty dicts. sad sad
```
       # Need to keep potentially empty dict for redistribute_routes
        return strip_empties_from_dict(router_bgp, strip_values_tuple=(None, ""))
```

let's change the signature then

ClausHolbechArista: fixed

ClausHolbechArista: fixed

gmuloc (approved):

dgonzalez-arista:
Tested and all good now we do not render in RRs and the PEs the EVPN address family

PR #4357, opened 2024-08-13
Fix(cv_deploy): Fix Ansible Logging and increase API timeouts

ClausHolbechArista:
## Change Summary

Fix Ansible Logging and increase API timeouts to 600 seconds.

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

- Logging did not work correctly after we moved code to pyavd
- API Timeouts of 10 seconds has proved to be too short, so increasing to 600 seconds as per dev recommendation.

## How to test

Tested manually with molecule scenario `cv_deploy`

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #4408, opened 2024-08-28
Fix(cv_deploy): Fix Ansible Logging and increase API timeouts (#4357)

gmuloc:
## Change Summary

Cherry-pick #4357

### DIFF
--- a/ansible_collections/arista/avd/plugins/action/cv_workflow.py
+++ b/ansible_collections/arista/avd/plugins/action/cv_workflow.py
@@ -44,46 +44,44 @@ LOGGER = logging.getLogger("ansible_collections.arista.avd")
 LOGGING_LEVELS = ["DEBUG", "INFO", "ERROR", "WARNING", "CRITICAL"]
 
 if HAS_PYAVD:
-    ARGUMENT_SPEC = (
-        {
-            "configuration_dir": {"type": "str", "required": True},
-            "structured_config_dir": {"type": "str", "required": True},
-            "structured_config_suffix": {"type": "str", "default": "yml"},
-            "device_list": {"type": "list", "elements": "str", "required": True},
-            "strict_tags": {"type": "bool", "required": False, "default": False},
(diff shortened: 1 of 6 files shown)


ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -44,44 +44,46 @@
...
+                "options": {
+                    "workspace_build_timeout": {"type": "float", "default": CVTimeOuts.workspace_build_timeout if HAS_PYAVD else 300.0},
+                    "change_control_creation_timeout": {"type": "float", "default": CVTimeOuts.change_control_creation_timeout if HAS_PYAVD else 300.0},
+                },
+                "return_details": {"type": "bool", "required": False, "default": False},
             },
         },
-        "return_details": {"type": "bool", "required": False, "default": False},
-    }
+    )
```
ClausHolbechArista: Why is this changing all over? Looks like an extra parenthesis by mistake.

python-avd/pyavd/_cv/client/tag.py
```diff
@@ -6,6 +6,8 @@
 from datetime import datetime
 from typing import TYPE_CHECKING, Literal
 
+from pyavd._cv.api.arista.time import TimeBounds
+
 from ..api.arista.tag.v2 import (
```
ClausHolbechArista:
```suggestion
from pyavd._cv.api.arista.tag.v2 import (
```

gmuloc: yeah this one is tough as this wont be consistent with other files in 4.10 as we moved to absolute path later in devel but lets be consistent at file level

ansible_collections/arista/avd/plugins/action/cv_workflow.py
gmuloc: let me recheck this - it was causing issue on merge - there is an extra HAS_PYAVD in one of the two files

ClausHolbechArista (approved):

PR #4419, opened 2024-09-02
Refactor(cv_deploy): Optimize push of configlets with version aware API calls

ClausHolbechArista:
## Change Summary

Optimize push of configlets with version aware API calls

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

- Detect version of CloudVision
- Use `SetSome` for newer versions for pushing configlets and containers
- Retain current batching of parallel `Set` for older versions.

## How to test

Tested manually:
- [x] CVaaS with 388 devices (very large configs). Use new `SetSome` method with overflow of max message sizes.
- [x] 2024.1.0 with 1030 devices. Use old `Set` in batches of 20 parallel operations. (fails after ~1000 because of CV issue)
- [x] 2024.2.0 with 1030 devices. Use new `SetSome` method packed to a single message.
- [x] 2024.3.0 EFT3 with 1030 devices. Use new `SetSome` method packed to a single message.
- [x] 2024.3.0 EFT4 with 1030 devices. Use new `SetSome` method packed to a single message.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -143,3 +146,37 @@ def _set_token(self) -> None:
...
             msg = "Unable to get token from CloudVision server. Please supply service account token instead of username/password."
             raise CVClientException(msg) from e
+
+    def _set_version(self) -> None:
+        """
+        Fetch the CloudVision version via REST and set self._cv_version.
+
+        This version is used to decide which APIs to use later.
+
+        TODO: Handle multinode clusters
```
gmuloc, line 160: should we document this as a limitation in the documentation?

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -143,3 +146,37 @@ def _set_token(self) -> None:
...
+            raise CVClientException(msg)
+
+        if not self._verify_certs:
+            # Accepting SonarLint issue: We are purposely implementing no verification of certs.
+            context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)  # NOSONAR
+            context.check_hostname = False
+            context.verify_mode = ssl.CERT_NONE  # NOSONAR
+        else:
+            # None means default context which will verify certs.
+            context = None
```
gmuloc: Looks like we are repeating this block of code - lets create a helper function for it

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -0,0 +1,144 @@
...
+from .exceptions import CVMessageSizeExceeded
+from .versioning import CvVersion
+
+LOGGER = getLogger(__name__)
+
+
+def grpc_msg_size_handler(list_field: str) -> Callable:
+    def decorator_grpc_msg_size_handler(func: Callable) -> Callable:
+        func_signature = signature(func)
+        if isinstance(func_signature.return_annotation, list):
```
gmuloc: this looks wrong in terms of logic if the return annotation is a list then we raise?

```python
>>> from pyavd._cv.client.async_decorators import  grpc_msg_size_handler
>>> from inspect import signature
>>> @grpc_msg_size_handler("toto")
... def test() -> bool:
...   return True
...
>>>
>>> test
<function test at 0x104d93400>
>>> test()
<coroutine object test at 0x102d431b0>
>>> signature(test)
<Signature () -> bool>
```

The issue is that the return_annotation is never a `list` or a `bool` but always a `type`. So the guard never works anyways

```
>>> signature(test).return_annotation
<class 'bool'>
>>> isinstance(signature(test).return_annotation, bool)
False
>>> isinstance(signature(test).return_annotation, type)
True

File: python-avd/pyavd/_cv/client/async_decorators.py [hunk 4]
```diff
@@ -0,0 +1,144 @@
...
+            bound_arguments = func_signature.bind(*args, **kwargs)
+            arguments = bound_arguments.arguments
+            if list_field not in arguments:
+                msg = f"grpc_msg_size_handler decorator is unable to find the list_field '{list_field}' in the given arguments to '{func.__name__}'."
+                raise KeyError(msg)
+
+            list_value: list = arguments[list_field]
+            if not isinstance(list_value, list):
+                msg = f"grpc_msg_size_handler decorator expected the value of the list_field '{list_field}' to be a list. Got '{type(list_value)}'"
+                raise TypeError(msg)
```
gmuloc, line 48: these would deserve unit tests.
Tried them both and it works

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -0,0 +1,144 @@
...
+            (<min_cv_version>, <max_cv_version>): <method>,
+            (<min_cv_version>, <max_cv_version>): <method>,
+        }
+    }
+    """
+
+    def __init__(self, min_ver: str = "2024.1.0", max_ver: str = CVAAS_VERSION_STRING) -> None:
+        """__init__ is called with the arguments of the decorator."""
+        # Storing these on the instance so it can be read in the __call__ method.
+        self.min_version = CvVersion(min_ver)
```
gmuloc, line 106: CvVersion could raise (yes it is used internally but maybe this should be handled in the CvVersion class?)

```
>>> LimitCvVersion("blah", "17")
  File "<stdin>", line 1, in <module>
  File "/Users/guillaumemulocher/Github/gmuloc/ansible-avd/python-avd/pyavd/_cv/client/async_decorators.py", line 100, in __init__
    self.min_version = CvVersion(min_ver)
  File "/Users/guillaumemulocher/Github/gmuloc/ansible-avd/python-avd/pyavd/_cv/client/versioning.py", line 13, in __init__
    self.version_value = self._version_as_value()
  File "/Users/guillaumemulocher/Github/gmuloc/ansible-avd/python-avd/pyavd/_cv/client/versioning.py", line 29, in _version_as_value
    return int(f"{elements[0]:04}{elements[1]:02}{elements[2]:02}")
IndexError: list index out of range
```

python-avd/pyavd/_cv/client/configlet.py
```diff
@@ -157,20 +165,68 @@ async def set_configlet_containers(
...
         try:
-            responses = client.set_some(request, metadata=self._metadata, timeout=timeout + len(request.values) * 0.5)
+            responses = client.set_some(request, metadata=self._metadata, timeout=timeout)
             assignment_keys = [response.key async for response in responses]
         except Exception as e:
             raise get_cv_client_exception(e, f"Workspace ID '{workspace_id}', Containers '{containers}'") or e
 
         return assignment_keys
 
+    @LimitCvVersion(max_ver="2024.1.99")
```
gmuloc, line 183: does this version exist?

python-avd/pyavd/_cv/client/versioning.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def __hash__(self) -> int:
+        return self.version_value
+
+    def _version_as_value(self) -> int:
+        if self.version == CVAAS_VERSION_STRING:
+            return 99999999
+
+        elements = self.version.split(".", maxsplit=2)
+        return int(f"{elements[0]:04}{elements[1]:02}{elements[2]:02}")
```
gmuloc, line 38: this can be quite fragile though I guess we are not expecting anyone but pyavd to use the method so thats probably ok

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: Spending extra runtime to check that the developer gave correct input seems inefficient to me.

python-avd/pyavd/_cv/client/configlet.py
ClausHolbechArista: I will change max_ver to below_ver so I can use an existing version number instead of this.

python-avd/pyavd/_cv/client/__init__.py
ClausHolbechArista: done

python-avd/pyavd/_cv/client/__init__.py
ClausHolbechArista: done

python-avd/pyavd/_cv/client/versioning.py
ClausHolbechArista: fixed.

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: fixed

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: fixed

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: added tests

python-avd/pyavd/_cv/client/configlet.py
ClausHolbechArista: I cannot change this. The idea of 2024.1.99 means "every version below 2024.2.0" but since I cannot predict which 2024.1.x" version will be the last, I cannot set this.
Should we change the logic to use regex matching instead? It would be a bit ugly with CVaaS in there.

gmuloc: lets aim at simplicity and add a comment on top explaining the why for future us

gmuloc (approved):

alexeygorbunov (approved):
Tested and worked on 2024.1.2 and CVaaS:

```bash
wrapper_cv_version: Called 'set_configlets_from_files' with version '2024.1.2'
set_configlets_from_files: Deploying 10 configlets in batches of 20.
set_configlets_from_files: Batch 1
Adding (b':path', b'/arista.configlet.v1.ConfigletConfigService/Set')

wrapper_cv_version: Called 'set_configlets_from_files' with version 'CVaaS'
wrapper_grpc_msg_size_handler: Called 'set_configlets_from_files' with '10' items
Adding (b':path', b'/arista.configlet.v1.ConfigletConfigService/SetSome')
```

PR #4463, opened 2024-09-13
Fix(eos_designs): Do not filter AVT on HA device if one path-group is present on peer

gmuloc:
## Change Summary

Until now, AVD is filtering completely a VT if none of the path-group listed is present locally,
This works as expected but there is one use case where this breaks potential connectivity:

Consider an HA site with two routers, `router1` connected to INET only and `router2` connected to MPLS only.
If some matched traffic is configured to be sent only on MPLS (let's say voice), then AVD would not generate the profile on `router1` for this MPLS only traffic.
When traffic for voice reaches `router1`, it will not be matched correctly and may be matched in one of the next application profile, default, or if no default is present, could be dropped.
The correct behavior is to configure the profile targeting a load-balance policy containing only the LAN_HA path-group so that the traffic can be sent to the peer and forwarded out of MPLS.

## Related Issue(s)

Reported during testing by AVD maintainers

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

This PR adds a check where if for a given VT, no path-group is present locally but any path-group is present on the HA peer, then AVD will now generate a load-balancing policy with only the `LAN_HA` path group and will generate the profile.

## How to test

Molecule - some molecule tests were already demonstrating this.

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -199,6 +199,10 @@ def _configure_bgp_mlag_peer_group(self: AvdStructuredConfigNetworkServices) ->
 
         return False
 
+    ################
```
ClausHolbechArista: Maybe a separate mixin?

ClausHolbechArista (approved):
Nice!

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
gmuloc: yeah probably a btter plan it starts to grow a bit big

carlbuchmann (approved):
LGTM

PR #4491, opened 2024-09-18
Fix(eos_designs)!: l3_edge with rfc5549 underlay and 'ebgp: true' should not have next-hop address-family ipv6

### LINKED ISSUE #4379: Feat(eos_designs): Support for IPv4 p2p_links in an RFC5549 Fabric
### Enhancement summary

Currently if `include_in_underlay_protocol` is `true` then the p2p_link is given the same peer group as the rest of the underlay. If the underlay fabric is an RFC5549 fabric using IPv6 but the p2p_link is IPv4, then due to the peer group it gets the attribute `neighbor IPv4-UNDERLAY-PEERS next-hop address-family ipv6 originate` set under the IPv4 address family. This will prevent routes from being shared. 

### Use case example

Two DC fabrics that use RFC5549 as their individual underlays but with a desire to make all connections out of the fabric IPv4, including the DCI. 

The quickest solution (from my naive perspective) is to add a key to allow overriding the peer-group used on the p2p_link. This would allow multiple benefits, not only could one control the `ipv6 originate` knob, but in the case of the DCI they could more easily tune other knobs independent of the rest of the underlay. 

Alternatively the existing `ipv6_enable` key controls the interface itself but has no impact on the neighborship AFAIK. This key could be used to automatically add the `no ... next-hop address-family ipv6 originate` for the neighbor if the underlay is set to RFC5549, or to add logic to create an RFC5549 underlay peer-group as well as a non-rfc5549 peer group (ideally only if needed). This solution adds more nested logic and would have to be careful to make sure it's backwards compatible. Even though I'm not sure how it would be working, existing installations might be using rfc5549 with ipv6 set to false and this would be a change in behavior. 

- Adding the manual config to the bgp ipv4 address family to override `next-hop address-family ipv6 originate` for the specific p2p neighbors. Feels confusing and not clear to have it defined in the peer group and then override it. 
- Manually build out all the neighborships for the DCI without using p2p_links. Undesirable, but likely the short term solution

### Additional context

I'm willing to work on a PR that implements this. Just looking for guidance on the best way to approach this issue before I start working. 

### Contributing Guide

nathanmusser:
## Change Summary

When `routing_protocol` is set to `ebgp` and the underlay is set to `rfc5549` we now override the `next-hop address-family ipv6` directive the neighbor inherits from the underlay peer-group.

## Related Issue(s)

Partially fixes #4379 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Override the `next-hop address-family ipv6` directive the neighbor inherits from the underlay peer-group.

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista (dismissed):
Please update the description for the `routing_protocol` key in the schema fragment to explain the new behavior.

ClausHolbechArista:
Also please add a test in molecule for this scenario. It should be enough to add another interface to one of the existing tests. I suggest `ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/l3_edge_bgp.yml`. To update the outputs (and verify that is is doing what you expect) run
```
molecule converge -s eos_designs_unit_tests -- --limit l3_edge_bgp
```

nathanmusser:
Started a refactor of this after a discussion with @ClausHolbechArista . New iteration is such that when the underlay is set to RFC5549 and `routing_protocol` is set to `ebgp` we explicitly override the next-hop ipv6 command for that specific neighbor (since it's part of the underlay peer-group that gets it due to enabling RFC5549).

eos_cli_config_gen had the next-hop ipv6 knob for peer groups in `router_bgp.address_family_ipv4` and for neighbors in the `router_bgp.vrfs[].address_family_ipv4` path but did not have it exposed for neighbors for the default vrf `router_bgp.address_family_ipv4`. Since that's what we needed that key(s) had to be added. 

I haven't done extensive testing of the latest commit yet, just verified it did what I wanted against a prod fabric. Need to finish testing the various edge cases and make sure it's implemented correctly. Need to update molecule, probably need to flesh out the documentation to correctly describe the eos_cli_config_gen keys as well as the behavior in eos_designs. 

Ultimately I still want to explore having the ability to override the peer group for l3_edge neighbors, but not sure I want to tackle that here.

ClausHolbechArista:
Thanks Nathan

> Ultimately I still want to explore having the ability to override the peer group for l3_edge neighbors, but not sure I want to tackle that here.

We would not want that capability in the `l3_edge` model. If you need custom peers and peer-groups, you should use the models under `tenants` to do it. There you can build interfaces, peers and peer-groups. With AVD 5.0 it also works well for VRF default.

nathanmusser:
> We would not want that capability in the `l3_edge` model. If you need custom peers and peer-groups, you should use the models under `tenants` to do it. There you can build interfaces, peers and peer-groups. With AVD 5.0 it also works well for VRF default.

Respectfully I disagree here. I started to type a novel justifying my thoughts, but I figure we can revisit after 5.0. I did successfully prototype custom peer_groups yesterday after my initial comment. I'm sure I haven't considered all the implications but it wasn't as difficult as I imagined. 

In the meantime I'll keep this PR focused on the above solution, I'll update the title and description today to better reflect the new focus.

nathanmusser:
Added a new molecule test `l3_edge_rfc5549` to validate the various interactions between rfc5549 and l3_edge. Though this edge case was present in some of the existing molecule tests so some of those artifacts have been updated to reflect this.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/l3_edge_rfc5549.yml
```diff
@@ -0,0 +1,48 @@
...
+      include_in_underlay_protocol: true
+
+      # P2P link without routing and ebgp.
+    - nodes: [l3_edge_rfc5549, peer3]
+      interfaces: [ethernet3, ethernet3]
+      as: [65000, 65003]
+      ip_pool: pool1
+      id: 3
+      include_in_underlay_protocol: false
+      routing_protocol: ebgp
```
ClausHolbechArista, line 36: Not sure why we are rendering BGP for this neighbor

ansible_collections/arista/avd/molecule/evpn_underlay_rfc5549_overlay_ebgp/intended/configs/DC1-BL1A.cfg
```diff
@@ -250,6 +250,7 @@ router bgp 65104
       no neighbor EVPN-OVERLAY-PEERS activate
       neighbor UNDERLAY_PEERS activate
       neighbor UNDERLAY_PEERS next-hop address-family ipv6 originate
+      no neighbor 10.23.23.2 next-hop address-family ipv6
```
ClausHolbechArista, line 253: @kornoa you implemented this 9 months ago. If we implement this change, we will no add this extra line, but as I see it, the config would actually be more correct with this change. Do you have a running network using the rfc5549 + the `ebgp` knob under l3_edge without ipv6_enable?

ansible_collections/arista/avd/molecule/evpn_underlay_rfc5549_overlay_ebgp/intended/configs/DC1-BL1B.cfg
```diff
@@ -251,6 +251,7 @@ router bgp 65105
       no neighbor EVPN-OVERLAY-PEERS activate
       neighbor UNDERLAY_PEERS activate
       neighbor UNDERLAY_PEERS next-hop address-family ipv6 originate
+      no neighbor 10.23.23.6 next-hop address-family ipv6
```
ClausHolbechArista, line 254: @nathanmusser I think we need to keep this neighbor with the next-hop, since we have ipv6_enable. Now I understand why you asked about this a long time ago :-/

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/router-bgp.j2
```diff
@@ -1167,6 +1167,17 @@ router bgp {{ router_bgp.as }}
...
+{%                 if neighbor.next_hop.address_family_ipv6.enabled is arista.avd.defined(true) %}
+{%                     set ipv6_originate_cli = "neighbor " ~ neighbor.ip_address ~ " next-hop address-family ipv6" %}
+{%                     if neighbor.next_hop.address_family_ipv6.originate is arista.avd.defined(true) %}
+{%                         set ipv6_originate_cli = ipv6_originate_cli ~ " originate" %}
+{%                     endif %}
+{%                 elif neighbor.next_hop.address_family_ipv6.enabled is arista.avd.defined(false) %}
+{%                     set ipv6_originate_cli = "no neighbor " ~ neighbor.ip_address ~ " next-hop address-family ipv6" %}
+{%                 endif %}
+      {{ ipv6_originate_cli }}
+{%             endif %}
```
laxmikantchintakindi:
```suggestion
{%             if neighbor.next_hop.address_family_ipv6.enabled is arista.avd.defined(true) %}
{%                 set ipv6_originate_cli = "neighbor " ~ neighbor.ip_address ~ " next-hop address-family ipv6" %}
{%                 if neighbor.next_hop.address_family_ipv6.originate is arista.avd.defined(true) %}
{%                     set ipv6_originate_cli = ipv6_originate_cli ~ " originate" %}
{%                 endif %}
      {{ ipv6_originate_cli }}
{%             elif neighbor.next_hop.address_family_ipv6.enabled is arista.avd.defined(false) %}
{%                 set ipv6_originate_cli = "no neighbor " ~ neighbor.ip_address ~ " next-hop address-family ipv6" %}
      {{ ipv6_originate_cli }}
{%             endif %}
```

ClausHolbechArista:
Moving to draft until comments have been addressed.

ansible_collections/arista/avd/molecule/evpn_underlay_rfc5549_overlay_ebgp/intended/configs/DC1-BL1A.cfg
kornoa: Sorry, the customer's project had finished. I don't have access to this implementation anymore.
I think this line is redundant. An IPv4-based neighbor session with an IPv4 address-family defaults to a IPv4 next-hop within the NLRI. But anyway, it doesn't harm.

nathanmusser: > An IPv4-based neighbor session with an IPv4 address-family defaults to a IPv4 next-hop within the NLRI.

This is true, but if the neighbor is is a member of the UNDERLAY_PEERS group (which l3_edge does) then line 252 forces the advertisements to advertise with ipv6 next-hops instead. Or at least this was my experience in my testing. Hence the need for the suggested 253.

nathanmusser:
I will likely not have the cycles for this until mid/late December or January. I plan to complete at that time but just adding that for transparency here.

ansible_collections/arista/avd/molecule/evpn_underlay_rfc5549_overlay_ebgp/intended/configs/DC1-BL1A.cfg
kornoa: Copy. :-)

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
```diff
@@ -54,7 +56,17 @@ def router_bgp(self: AvdStructuredConfigCoreInterfacesAndL3EdgeProtocol) -> None
...
                 peer=p2p_link_data["peer"],
                 description=p2p_link_data["peer"],
-                peer_group=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name,
+                peer_group=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name if p2p_link.include_in_underlay_protocol else Undefined,
                 bfd=p2p_link.bfd,
                 local_as=p2p_link_data["bgp_as"] if p2p_link_data["bgp_as"] != self.shared_utils.bgp_as else None,
             )
+
+            # For the combination of underlay-routing, rfc5549 and ebgp we will add the neighbor using the regular logic above,
+            # but since it is also included in the underlay peer group whichis configured for rfc5549,
```
ClausHolbechArista:
```suggestion
            # but since it is also included in the underlay peer group which is configured for rfc5549,
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/router-bgp.j2
ClausHolbechArista: implemented a slightly different optimization.

ankudinov (approved):
LGTM

gmuloc (approved):

PR #4492, opened 2024-09-18
Fix(eos_cli_config_gen): Do not render entries with only sequence number from ip_access_list

### LINKED ISSUE #4423: Fix(eos_cli_config_gen): Do not render entries with only sequence number and nothing else from ip_access_list
### Issue Summary

Reported from the field

This input

```yaml
ipv4_acls:
  - name: WAN_ACL_IN_Et1
    entries:
      - source: any
        destination: any
        sequence: 10
        action: permit
```

Will generate the following invalid config:

```
ip access-list WAN_ACL_IN_Et1
   10 
!
```

Because the `eos_designs` `ipv4_acls` key will pass along the entry untouched to the underlying `ip_access_list` model.

## Expected resolution

Please make sure no sequence number is generated when both `remark` and `protocol` are missing.
The problem is in this template where the acl entry string is ALWAYS rendered https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2#L133 even if neither the `if` https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2#L34 nor the `elif` https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2#L38 are matched.

In this scenario, no sequence number should be rendered

eos_cli_config_gen

### How do you run AVD ?

### Steps to reproduce

```shell
cf example above or use directly the eos_cli_config_gen model.
```

### Relevant log output

_No response_

### Contributing Guide

gmuloc:
## Change Summary

Skip issues if `remark` or any field in `[protocol, action, source, destination]` is missing 

## Related Issue(s)

Fixes #4423

## Component(s) name

## Proposed changes

Test if the entry is empty, if it is skip rendering it (and the sequence number)

## How to test

molecule - first commit demonstrate the issue, second solves it

### Repository Checklist


python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2
```diff
@@ -130,7 +126,14 @@ ip access-list {{ acl.name }}
 {%                     endif %}
 {%                 endif %}
 {%             endif %}
+{#             - only render if the string is not empty #}
```
ClausHolbechArista:
```suggestion
{#             Only render if the string is not empty #}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2
```diff
@@ -130,7 +126,14 @@ ip access-list {{ acl.name }}
...
+{%             if acl_entry.string != "" %}
+{#                 - prepend sequence number if defined and space #}
```
ClausHolbechArista:
```suggestion
{#                 Prepend sequence number if defined #}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2
gmuloc: I was matching the rest of the file with the `-` do you want to change line 28 and 26 as well?

ClausHolbechArista: It is not a style we usually have. Feel free to update elsewhere.

Vibhu-gslab (approved):
LGTM!

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ip-access-lists.j2
gmuloc: this was done in last commit already

ClausHolbechArista (approved):

PR #4493, opened 2024-09-19
Feat(eos_cli_config_gen): Global logging keys for congestion-drops, link-status, and repeat-messages

nathanmusser:
## Change Summary

Added new logging keys to disable `repeat-messages`, change the logging interval for the `congestion-drops` event, and set the global command for the `link-status` logging event. 

## Related Issue(s)

N/A

## Component(s) name

## Proposed changes
Added the following data model:
```
logging:
  repeat_messages: <bool>
  event:
    congestion_drops:
      interval: <int>
    link_status:
      global: <bool>
```
And added the associated j2 templates to render the associated commands:
```
no logging repeat-messages
logging event congestion-drops interval 600
logging event link-status global
```

## How to test
Used the example dual-dc-l3ls topology and added the various commands to the `FABRIC.yml` group_var. Tested removing them and adding them and setting invalid values to make sure behavior is as expected. Commands are only rendered when the new model is present, etc..  

### User Checklist

- N/A

### Repository Checklist

_**Note:** I'm still new to contributing here, but for the repo checklist item referring to documentation. My understanding is that for these simple knobs the documentation is generated from the schema. AFAIK there's nowhere else I need to add documentation? Should I be checking this box in those cases? Or leaving it unchecked since there is no documentation beyond what is auto-generated that needs to be modified?_

_Also I know I didn't open an issue for this first. It felt like a simple enough PR that there wouldn't be enough discussion to warrant an issue. Feel free to correct that assumption and I can create one as well as do so going forward for all PRs._


ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/logging.cfg
```diff
@@ -1,7 +1,10 @@
 !
+no logging repeat-messages
 logging buffered 1000000 warnings
 no logging trap
 logging console errors
+logging event congestion-drops interval 10
+logging event link-status global
```
MaheshGSLAB: this is not correct order of config compare to actual EOS cli. For `logging event congestion-drops interval 10` we need to create new j2 module and include it in eos-intended-config.j2

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
```diff
@@ -128,6 +132,21 @@ keys:
...
+          congestion_drops:
+            type: dict
+            keys:
+              interval:
+                type: int
+                convert_types:
+                  - str
+                min: 1
+                max: 65535
+                description: Logging interval in seconds
```
Shivani-gslab:
```suggestion
                description: Logging interval in seconds.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
```diff
@@ -33,6 +33,10 @@ keys:
             valid_values: ["alerts", "critical", "debugging", "emergencies", "errors", "informational", "notifications", "warnings", "disabled"]
             description: |
               Buffer logging severity level.
+      repeat_messages:
+        type: bool
+        description: |
+          Summarize concurrent repeat messages
```
Shivani-gslab:
```suggestion
        description: Summarize concurrent repeat messages.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
```diff
@@ -128,6 +132,21 @@ keys:
       event:
         type: dict
         keys:
+          congestion_drops:
```
Shivani-gslab: we can create a single key - `congestion_drops_interval` instead of a dict

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
```diff
@@ -128,6 +132,21 @@ keys:
...
+            type: dict
+            keys:
+              interval:
+                type: int
+                convert_types:
+                  - str
+                min: 1
+                max: 65535
+                description: Logging interval in seconds
+          link_status:
```
Shivani-gslab: this key could be `global_link_status` instead of a dict

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/logging.j2
```diff
@@ -33,6 +36,12 @@ no logging monitor
 {%     elif logging.monitor is arista.avd.defined %}
 logging monitor {{ logging.monitor }}
 {%     endif %}
+{%     if logging.event.congestion_drops.interval is arista.avd.defined %}
+logging event congestion-drops interval {{ logging.event.congestion_drops.interval }}
+{%     endif %}
+{%     if logging.event.link_status.global is arista.avd.defined(true) %}
+logging event link-status global
```
gmuloc, line 44: wondering if we would not rather need the `no` version of this command:
```
wan2-site1(config)#show run se logging
logging event congestion-drops interval 43
logging repeat-messages
no logging event link-status global  <<<<<<<<<<<<<<  show run shows the `no` version
wan2-site1(config)#
wan2-site1(config)#
wan2-site1(config)#logging event link-status global  <<<< enabling it and not rendered
wan2-site1(config)#show run se logging
logging event congestion-drops interval 43
logging repeat-messages
wan2-site1(config)#
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
nathanmusser: Is there a standard for when to use the multi-line pipe for the description and when to not? I noticed the logging.schema.yml had mixed usage even for single line comments.

I ask because there might be a pattern here I'm missing, but if there's not I can fix any other descriptions in this file if there's an effort to clean them up.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
nathanmusser: I don't have an opinion, just trying to better understand so I can make cleaner PRs in the future. Is there a standard that drives when to use nested keys vs a single key? As an example in the same file there's this structure:

https://github.com/aristanetworks/avd/blob/997453813a4fe391464636e632c48fdf81b609d7/python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml#L45-L53

which is needlessly a dictionary. It could be `synchronous_level`. I assumed this was done to match the structure of the EOS CLI command and allow for future expansion of that command. I don't mind merging these two suggestions just trying to understand the standards so I can make sure I comply going forward.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/logging.cfg
nathanmusser: I noticed there are other parts of the logging schema that are not rendered into the correct spot in the EOS configuration. As an example the existing `logging event storm-control` looks like it should be rendered after `eos/aliases.j2` and before `eos/daemon-terminattr.j2` to match the ordering of the EOS configuration. While fixing my own additions should I look to bring the rest of the logging data model into compliance here? If so under this PR or a new one?

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/logging.j2
nathanmusser: This is a good point. I'm looking over an AVD codebase that had eos_cli adding `logging event link-status global`, but as you pointed out this appears to be the default behavior. Which means this isn't really something I need anymore, since AVD is rendering the configuration as desired. But since I'm here I can refactor this to reflect the actual behavior. Is it good enough to change the default value to `true` and have it only render when `false`?

carlbuchmann: With `eos_cli_config_gen` templates, we want to print the command regardless of the default behaviour when set. So it should print when set to true or false, and no default value.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
Shivani-gslab: since there is only one command under congestion_drops, I think it should be combined to `congestion_drops_interval`

ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/logging.cfg
Shivani-gslab: The reordering of existing stuff is covered in this PR - https://github.com/aristanetworks/avd/pull/4449/

MaheshGSLAB: Yes so just fix the order of your changes, others will be handle in above PR.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
Shivani-gslab: We use multi-line pipe in multiline strings for readability and proper formatting. So here its not required.
Regarding the existing ones, if its a lot of changes you can ignore it in this PR and just fix the new changes.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
gmuloc: @nathanmusser so essentially we are trying to limit the nested-ness of things these days. There may be things from before (maybe not so long sometimes things slip) that have more levels than necessary but we don't go through the deprecation process and live with it. But as a rule of thumb if there is only one key - we can have ~dict~ single key.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/logging.cfg
nathanmusser: Done

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
nathanmusser: Done, as well as the existing ones in the same file.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/logging.schema.yml
nathanmusser: Done

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos-intended-config.j2
```diff
@@ -32,6 +32,7 @@
 {% include 'eos/terminal.j2' %}
 {# aliases #}
 {% include 'eos/aliases.j2' %}
+{# logging event storm-control #}
```
gmuloc:
```suggestion
```

gmuloc (approved):

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/logging.j2
```diff
@@ -6,6 +6,9 @@
 {# eos - logging #}
 {% if logging is arista.avd.defined %}
 !
+{%     if logging.repeat_messages is arista.avd.defined(false) %}
+no logging repeat-messages
```
Shivani-gslab, line 10: shall we add `logging repeat-messages` here when its `true`?
@gmuloc

gmuloc: yes let's do it

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/logging.j2
```diff
@@ -6,6 +6,9 @@
...
+{%     endif %}
```
Shivani-gslab:
```suggestion
{%     if logging.repeat_messages is arista.avd.defined(false) %}
no logging repeat-messages
{%     elif logging.repeat_messages is arista.avd.defined %}
logging repeat-messages
{%     endif %}
```

Shivani-gslab (approved):
LGTM!

PR #4497, opened 2024-09-19
Feat(eos_designs): Add possibility to change network mask for direct WAN HA link

gmuloc:
## Change Summary

cf title

## Related Issue(s)

Reported from field

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Add a setting to manipulate the default value for HA mask

## How to test

molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -331,3 +331,24 @@ def evpn_underlay_l2_multicast_group(
         """Return IP address to be used for EVPN underlay L2 multicast group."""
         offset = vlan_id - 1 + underlay_l2_multicast_group_ipv4_pool_offset
         return get_ip_from_pool(underlay_l2_multicast_group_ipv4_pool, 32, offset, 0)
+
+    def get_wan_ha_ip_address(self, local: bool) -> str | None:
```
ClausHolbechArista: We do not call these methods `get_xxx`. Also I would prefer to have two methods similar to uplinks, mlag etc. (I am aware that they will be very similar, but it will be slightly more efficient).

```suggestion
    def wan_ha_ip_address(self, local: bool) -> str | None:
```

gmuloc: this is a copy of an exitsting method. I can split it in two method though

gmuloc: also there is no primary or secondary so thats a bit tricky

gmuloc: fixed

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -331,3 +331,27 @@ def evpn_underlay_l2_multicast_group(
         """Return IP address to be used for EVPN underlay L2 multicast group."""
         offset = vlan_id - 1 + underlay_l2_multicast_group_ipv4_pool_offset
         return get_ip_from_pool(underlay_l2_multicast_group_ipv4_pool, 32, offset, 0)
+
+    def wan_ha_ip_local(self) -> str:
```
ClausHolbechArista: For consistency with the other methods

```suggestion
    def wan_ha_ip(self) -> str:
```

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -331,3 +331,27 @@ def evpn_underlay_l2_multicast_group(
...
+        """Render ipv4 address for wan_ha_ip_address using dynamically loaded python module."""
```
ClausHolbechArista: We are inside the dynamically loaded module, so we should say more specifically what we are doing.

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -331,3 +331,27 @@ def evpn_underlay_l2_multicast_group(
...
+        prefixlen = self._fabric_ip_addressing_wan_ha_ipv4_prefix_length
+
+        if self.shared_utils.is_first_ha_peer:
+            ip_address = get_ip_from_pool(wan_ha_ipv4_pool, prefixlen, 0, 0)
+        else:
+            ip_address = get_ip_from_pool(wan_ha_ipv4_pool, prefixlen, 0, 1)
+
+        return f"{ip_address}/{prefixlen}"
+
+    def wan_ha_ip_remote(self) -> str:
```
ClausHolbechArista:
```suggestion
    def wan_ha_peer_ip(self) -> str:
```

python-avd/pyavd/api/ip_addressing/utils.py
```diff
@@ -193,3 +193,7 @@ def _get_p2p_ipv4_pool_and_offset(self: AvdIpAddressing, uplink_switch_index: in
             return (uplink_pool, uplink_offset)
 
         return (downlink_pool, downlink_offset)
+
+    @cached_property
```
ClausHolbechArista: You are reading the pool directly from shared_utils. We should do it consistently one or the other way.

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM, reviewed and tested with systest inventory.

PR #3730, opened 2024-03-11
Feat(eos_cli_config_gen): Redo the model for additional-paths in BGP

laxmikantchintakindi:
## Change Summary
Redo the model for additional-paths in BGP to reduce number of key-value pairs and avoid repeated use of block of code using defs.

## Related Issue(s)

Fixes #3279

## Component(s) name

`arista.avd.eos_cli_config_gen`

## Proposed changes
Enhancing code by replacing multiple key-value pairs with boolean values with the help of valid_values key.
Using defs to avoid repeated use of a block of code.

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #4515, opened 2024-09-25
Feat(eos_cli_config_gen): Add support BGP Additional-Paths on Root of BGP

ccsnw:
## Change Summary

Add support for "bgp additional-paths" in root level of the BGP config.
Updated Doc to show the commands under "BGP Tuning" parameters.

## Related Issue(s)

Fixes #<N/A>

## Component(s) name

## Proposed changes
Using the same model as "additional-paths" in router_bgp for more specific sections

## How to test
Added molecule testing to "router-bgp-path-selection".
Let me know if dedicated test is preferred.

### User Checklist

- N/A

### Repository Checklist

- [X} My change requires a change to the documentation and documentation have been updated accordingly.

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-path-selection.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/router-bgp-path-selection.md
@@ -47,6 +47,11 @@ ASN Notation: asplain
 | ------ | --------- |
 | 65101 | 192.168.255.42 |
 
+| BGP Tuning |
+| ---------- |
+| bgp additional-paths receive |
+| bgp additional-paths send ecmp limit 30 |
+
 #### Router BGP Peer Groups
 
 ##### PATH-SELECTION-PG-1
@@ -122,6 +127,8 @@ ASN Notation: asplain
 !
 router bgp 65101
    router-id 192.168.255.42
+   bgp additional-paths receive
+   bgp additional-paths send ecmp limit 30
    neighbor PATH-SELECTION-PG-1 peer group
    neighbor PATH-SELECTION-PG-1 remote-as 65001
    neighbor PATH-SELECTION-PG-2 peer group
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-path-selection.cfg
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/intended/configs/router-bgp-path-selection.cfg
@@ -6,6 +6,8 @@ interface Management1
 !
 router bgp 65101
    router-id 192.168.255.42
+   bgp additional-paths receive
+   bgp additional-paths send ecmp limit 30
    neighbor PATH-SELECTION-PG-1 peer group
    neighbor PATH-SELECTION-PG-1 remote-as 65001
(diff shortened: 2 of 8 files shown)


ClausHolbechArista:
This is possibly a partial duplicate of #3730.

Please take a look at that one, and if it is indeed a duplicate we would appreciate your feedback on the existing PR. Thanks.

ccsnw:
Hi Claus,

I think the #3730 is breaking existing code, mine shouldn't (it would just give the user the option to add the additional path commands to the root of the bgp config using the currently existing model - so would be additive to what we have and not breaking).

Is this file "python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/defs_bgp_additional_paths.schema.yml" in the other PR used to make that part of the schema re-usable under different sections? This would make sense.

In my opinion you could merge my request to the current version, since not breaking. 

From what I understood, the other PR should allow the use of additional_paths under multiple sections of the bgp config, but since it's breaking, I would add it to a new "major version" and it should make all the legacy duplicates under the various sections obsolete?

ClausHolbechArista:
> Hi Claus,
> 
> I think the #3730 is breaking existing code, mine shouldn't (it would just give the user the option to add the additional path commands to the root of the bgp config using the currently existing model - so would be additive to what we have and not breaking).
> 
> Is this file "python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/defs_bgp_additional_paths.schema.yml" in the other PR used to make that part of the schema re-usable under different sections? This would make sense.
> 
> In my opinion you could merge my request to the current version, since not breaking.
> 
> From what I understood, the other PR should allow the use of additional_paths under multiple sections of the bgp config, but since it's breaking, I would add it to a new "major version" and it should make all the legacy duplicates under the various sections obsolete?

We are in the process of preparing 5.0, and you are absolutely correct that the other PR is breaking. We wish to bring it to the new model within the next few weeks, so I would prefer to just focus on the new stuff. Is the CLI you need covered by the other PR? Thanks.

ccsnw:
Hi Claus,

As far as I can see, the other PR does 2 things (please correct me if I'm wrong):

a) Change the model and Jinja Template accordingly
b) Outsources the Schema to a file that can be referenced (to reduce lines of code in the schema)

"Is the CLI you need covered by the other PR?"
>> From a command perspective: yes, but I need the commands in the top level of the BGP config (not under "address-family ipv4" etc)

What I can do: If you want I can try to extend #3730 by modifying the router-bgp.j2 so we can use the new data model to also render the config to the bgp root... and modify the schema to also refer to the outsourced file.

So once 5.0 is released, the commands will also be available there?

ClausHolbechArista (changes requested):
Apologies for being "slow" here :) I get it now. So if you could please update this PR to use the same model as the other PR, but in the root context. You can just copy the `def_*` schema fragment from the other one and reference it in your context. Then the other one will come later to update the other contexts to the same model (as a breaking change).

ccsnw:
Thanks Claus! Got it - we can do it like this, because the cli doesn't exist in the bgp root context yet. 

This way we can introduce the new model for the new context, but not break the existing implementation + later convert for the other contexts.

I will add this tomorrow.

Thanks for clarifying

ccsnw:
I'll close this PR and add a new one using the new model

PR #4522, opened 2024-09-27
Fix(eos_designs): Add redistribute connected under BGP for VRF default if no underlay

ClausHolbechArista:
## Change Summary

Add redistribute connected under BGP for VRF default if no underlay

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

In the corner case where we build BGP config for VRF default as part of network services (but not for underlay/overlay), we were missing "redistribute connected".

There will be a PR later to provide control for redistribute connected.

## How to test

Existing molecule exposed the issue and show the fix.

### User Checklist

- N/A

### Repository Checklist


carlbuchmann (approved):
LGTM

ClausHolbechArista:
@carlbuchmann I don't think this will apply to 4.x since we did not support the corner case we hit here. It is part of the new BGP behavior when not using overlays.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -191,6 +191,12 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServices) -> dict:
                     # Resetting bgp_vrf so we only add global keys if there are any neighbors for VRF default
                     bgp_vrf = {}
 
+                if vrf_name == "default" and self.shared_utils.underlay_routing_protocol == "none":
+                    # We need to add redistribute connected for the default VRF when underlay_routing_protocol is "none"
+                    bgp_vrf |= {
+                        "redistribute_routes": [{"source_protocol": "connected"}],
+                    }
```
```suggestion
                    bgp_vrf["redistribute_routes"] = [{"source_protocol": "connected"}]
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -191,6 +191,12 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServices) -> dict:
                     # Resetting bgp_vrf so we only add global keys if there are any neighbors for VRF default
                     bgp_vrf = {}
 
+                if vrf_name == "default" and self.shared_utils.underlay_routing_protocol == "none":
```
gmuloc: Could we move this up to line 180 with an `else` (as we know it is VRF default), then add a check for `/ if bgp_vrf is not empty and 2/ routing protocol==none

This saves one comparison in this if statement and groups everything happening for VRF default in one block (which we could move toa function

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista: done

gmuloc (approved):

PR #4540, opened 2024-10-02
Refactor(eos_designs): Move debug vars dump to action plugin instead of it's own task

ClausHolbechArista:
## Change Summary

Move debug vars dump to action plugin instead of it's own task

## Related Issue(s)

Increased run time after we removed tags since Ansible gathers all vars for every device before evaluating `when` on the task.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Move dump of hostvars to `eos_designs_structured_config` action plugin, since that already runs per device and has access to all vars.

## How to test

Tested manually and compared output between devel and branch:

```
molecule converge -s eos_designs_unit_tests -- -e avd_eos_designs_debug="{{ true }}"
```

Note some reordering is expected where the internal data structures are built using sets.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -44,14 +44,22 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
             profiler = cProfile.Profile()
             profiler.enable()
 
+        hostname = task_vars["inventory_hostname"]
+
+        if self._task.args.get("debug_vars") is True and (debug_vars_file := self._task.args.get("debug_vars_file")):
```
gmuloc, line 49: probably would be good to start adding some logs to track what we are doing in this action.
(not in this PR?)

gmuloc (approved):

carlbuchmann (approved):
LGTM - tested in local repo... it's fast!

PR #4551, opened 2024-10-04
Feat(eos_designs)!: Use tags for marking local routes to be redistributed to WAN

gmuloc:
## Change Summary

Right now, the routes in VRF default are marked twice with SOO: when being redistributed into BGP (from LAN, CONN or STATIC) and when exiting over the WAN EVPN session. The SOO is used to decide which VRF default routes (and only VRF default not the others) are injected in EVPN. This double SOO marking is not useful.

This PR replaces marking the routes with TAG 10 and matching on this to decide which route should make it in to VRF default EVP.

The PR also allows to revert back to the AVD 4.x config 

## Related Issue(s)

open point for 5.0.0 in aristanetworks/avd-internal#157

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

cf summary

## How to test

molecule show the changes

### User Checklist

- [ ] Add test reverting the changes to validate it works
- [ ] Update WAN how-to (and diagrams ...)

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/wan_use_sso_for_route_injection.schema.yml
```diff
@@ -0,0 +1,17 @@
...
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  wan_use_soo_for_route_injection:
+    documentation_options:
+      table: wan-settings
+    description: |-
+      In AVD 5.0.0, tags are used instead of SOO to manage which routes are
```
ClausHolbechArista:
```suggestion
      By default, route tags are used instead to mark which routes are exported to EVPN.
```

ClausHolbechArista: We should not talk about versions in schemas, as they will quickly become outdated. It would be better to talk about the two behaviors.
The version stuff needs to go in the porting guide.

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -252,7 +255,11 @@ def _redistribute_connected_to_bgp_route_map(self: AvdStructuredConfigNetworkSer
                 "match": ["ip address prefix-list PL-SVI-VRF-DEFAULT"],
             }
             if self.shared_utils.wan_role:
-                sequence_30["set"] = [f"extcommunity soo {self.shared_utils.evpn_soo} additive"]
+                if self.shared_utils.wan_use_soo_for_route_injection:
+                    set_statements = [f"extcommunity soo {self.shared_utils.evpn_soo} additive"]
+                else:
+                    set_statements = ["tag 10"]
```
ClausHolbechArista: Should we put this in the _utils file so we don't have to check it everywhere?

ClausHolbechArista (changes requested):

laxmikantchintakindi:
Need a test case with wan_use_soo_for_route_injection: true.

gmuloc:
> Need a test case with wan_use_soo_for_route_injection: true.

yes it is in the User checklist

ClausHolbechArista:
Setting to draft until planned changes have been implemented.

gmuloc:
Closing this PR as match tag is not working in the export to EVPN route-maps and so it breaks. SoO which is a community works so we shall keep it for now.

PR #4554, opened 2024-10-04
Feat(eos_designs,eos_cli_config_gen): Recirculation Channel Interface

### LINKED ISSUE #4495: Requesting the addition of support for a platform-specific recirculation channel to mirror egress traffic to the CPU.
### Enhancement summary

For the following platforms, Egress mirroring to Cpu requires configuring a recirculation channel. For which, we currently don't have support through AVD. 
CCS-710P-12***
CCS-710P-16***
DCS-7010TX-48***
CCS-720DF***
CCS-720DP***
CCS-720DT***
CCS-720XP-24ZY4***
CCS-720XP-24Y6***
CCS-720XP-48Y6***
CCS-720XP-48ZC2***
CCS-720XP-96ZC2***

eos_designs

### Use case example

We want to be able to mirror the traffic in their OT-DC campus enviornment which have CCS-720XP-48Y4 switches. We need this recirculation channel configured in order to be configure mirror sessions. Please refer to this TOI for more info. https://www.arista.com/en/support/toi/eos-4-32-0f/13722-mirror-cpu

We are thinking this solution should be tied to the platform type since it's only for specific platforms. Or maybe we don’t need to—regular support for this key might be sufficient, and we can control which group gets that config through AVD.

We used `raw_eos_cli` at the group level, but we already have `eos_cli` configured under `global_variables`. The `eos_cli` at the global level is being overridden by the new group-level `eos_cli` config. For now, we're using a custom Jinja template as a workaround.

### Additional context

_No response_

### Contributing Guide

nathanmusser:
## Change Summary

Add support for recirculation channel interfaces for cpu-mirroring and adding a front panel interface to the recirculation.

## Related Issue(s)

Fixes #4495 

## Component(s) name

`arista.avd.eos_designs`
`arista.avd.cli_config_gen`

## Proposed changes
Add `recirculation_interfaces` model to `eos_cli_config_gen` to generate the recirculation interface. Add the `traffic-loopback source system device mac` to `ethernet_interfaces` model. Add the ability to put an ethernet interface into a recirculation channel group. Potentially expose all this in the platform settings so it can be configured per platform with an easy toggle.

## How to test
TBD

### User Checklist

- [x] Add recirculation_interfaces data model
- [x] Add recirculation_interfaces documentation
- [x] Add recirculation_interfaces molecule testing
- [ ] Add traffic-loopback key to ethernet_interfaces
- [ ] Add traffic-loopback key documentation
- [ ] Add traffic-loopback key molecule testing
- [x] Add recirc channel group assignment
- [x] Add recirc channel group documentation
- [x] Add recirc channel group molecule testing 

### Repository Checklist


nathanmusser:
When adding an interface to the recirculation channel you use the syntax `channel-group recirculation x` but our `channel_group` data model under `ethernet_interfaces` is
```
    channel_group:
      id: <int>
      mode: <str; "on" | "active" | "passive">
```
I'm suggesting we add a new key `recirculation` to the channel group, and the associated logic to the template to render the channel-group as it currently is when false, and when true add the recirculation keyword before the ID. 
```
    channel_group:
      id: <int>
      mode: <str; "on" | "active" | "passive">
      recirculation: <bool>
```

I've already added the recirculation_channel interface. I generally referenced the format of other interface types (port-channel, dps, etc..) when structuring the data model. But one thing I don't love is that we're keying off a user supplied interface "name" but there's no validation that the user is supplying a valid name for a recirculation interface. But none of the other interface types seem to validate this either. As an example a user could create a tunnel interface using the name "Recirc1" and eos_cli_config_gen would allow it but it would have options under it that would not be accepted on the switch. I've seen the `pattern:` key in the schema used elsewhere but I don't see it on any of the interfaces and wasn't sure if there was a reason why. 

I still wouldn't really like the `pattern:` though, as we now define the interface as `Recirc-channel x` in one place but refer to it later as just the integer ID. This appears to be how all the other interface types are done in AVD but makes for more friction between the structured_config and other tooling that might want to interact with it. It seems to me that since we know the interface name is always going to start with "Recirc-channel" that doesn't need to be provided by the user and instead the `name` field could just be `id`.  I didn't want to deviate from the established pattern though with asking about it.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/recirculation_interfaces.yml
```diff
@@ -0,0 +1,28 @@
...
+keys:
+  recirculation_interfaces:
+    type: list
+    primary_key: name
+    items:
+      type: dict
+      keys:
+        name:
+          type: str
+          description: "Recirculation interface name."
```
MaheshGSLAB:
```suggestion
          description: Recirculation interface name.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/recirculation_interfaces.yml
```diff
@@ -0,0 +1,28 @@
...
+        description:
+          type: str
+        shutdown:
+          type: bool
+        recirculation_features:
+          type: str
+          valid_values: ["vxlan", "telemetry inband", "openflow", "cpu-mirror"]
```
MaheshGSLAB:
```suggestion
          valid_values: 
            - vxlan
            - telemetry inband
            - openflow
            - cpu-mirror
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/recirculation-interfaces.j2
```diff
@@ -0,0 +1,27 @@
...
+#}
+{# doc - recirculation interfaces #}
+{% if recirculation_interfaces is arista.avd.defined %}
+
+### Recirculation Interfaces
+
+#### Recirculation Interfaces Summary
+
+| Interface | Description | Shutdown | Recirculation Features |
+| --------- | ----------- | -------- | --------------- |
```
MaheshGSLAB:
```suggestion
| Interface | Description | Shutdown | Recirculation Features |
| --------- | ----------- | -------- | ---------------------- |
```

MaheshGSLAB (commented):
First round of review

nathanmusser:
When implementing the `traffic-loopback` key, for my purposes I only really need a `traffic-loopback source system device mac` flag. When creating the key should I create the structure to represent the options available at the CLI that I don't plan to implement?

```
sw0-arista(config-if-Et3)#show cli commands | inc traffic-loopback
[no|default] traffic-loopback source ( ( system device ( SYSTEMPHY | mac ) ) | ( network device NETWORKPHY ) )
```

My current plan was to implement the schema as:
```
# ...
  ethernet_interfaces:
    type: list
    primary_key: name
    items:
      type: dict
      keys:
        # ...
        traffic_loopback_source:
          type: dict
          keys:
            system_device_mac:
              type: bool
              description: "Implement loopback in MAC layer."
``` 

To allow for the potential of the rest of the CLI structure being fleshed out later if needed.

ClausHolbechArista:
Closing stale PR. It is probably easier to start over at this point. Please split in one for eos_cli_config_gen and another for eos_designs

PR #4562, opened 2024-10-08
Refactor(plugins): Emit errors when both deprecated and new keys are set

ClausHolbechArista:
## Change Summary

Emit errors when both deprecated and new keys are set

## Proposed changes

During schema validation/conversion we will now check if both the deprecated and replacement values are set. If so we will fail validation and stop execution (even when validation mode is warning).

This is implemented to protect users from mixing new and old data models which can lead to surprising results.

## How to test

- Added a negative test for eos_cli_config_gen.
- Tested manually.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_schema/avddataconverter.py
```diff
@@ -173,11 +175,31 @@ def deprecation(self, deprecation: dict, _data: Any, _schema: dict, path: list)
         if not deprecation.get("warning", True):
             return
 
-        yield AvdDeprecationWarning(
+        new_key = deprecation.get("new_key")
+        removed = deprecation.get("removed", False)
+
+        # If new_key set, we can check for collision where both new and old key are set.
```
gmuloc, line 181: maybe this should make it to some page where we explain what we are able to catch and what we are not

python-avd/pyavd/avd_schema_tools.py
```diff
@@ -49,18 +48,27 @@ def convert_data(self, data: dict) -> list[AvdDeprecationWarning]:
                 Input variables which should be converted according to the schema.
 
         Returns:
-            List of AvdDeprecationWarnings
+            Validation result object with any validation errors or deprecation warnings.
```
gmuloc:
```suggestion
            ValidationResult object with any validation errors or deprecation warnings.
```

bjmeuer (approved):
tested against my repository and works as expected.

kpbush30 (approved):
LGTM - tested against my repo and `cvp_instance_ip` and `cvp_instance_ips` and failed as expected in warning mode.

python-avd/pyavd/_schema/avddataconverter.py
ClausHolbechArista: added in https://github.com/aristanetworks/avd/pull/4562/commits/c3f8e8d816b27b5ba7515a00c4ceefb14db8751a

gmuloc (approved):

PR #4629, opened 2024-10-23
Feat(cv_deploy): Expose the details of the Workspace build validation errors

alexeygorbunov:
## Change Summary

Explicitly expose the details of the:
- Config Validation errors and warnings raised when part(s) of the proposed DC is rejected by EoS during validation via Config Session
- (descoped) ~Image Validation errors and warnings raised when part(s) of the proposed image bundle is rejected by EoS during validation~
- (descoped) ~Time taken by CV to process specific Workspace lifecycle stages~

## Related Issue(s)

Fixes aristanetworks/avd-internal#400

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- Expose under`cv_deploy_results.workspace.build_id` an ID of the latest build attempt. Needed as WS in Pending state maybe reused and may have multiple Build attempts (with each attempt producing different set of errors)
- Expose under `cv_deploy_results.workspace.build_results`:
  * Errors and warnings related to the validation on the proposed configuration.
  * (descoped) ~Errors and warnings related to the proposed image changes (not yet implemented in AVD).~
  * Details of the specific device associated with faced error/warning
  * (descoped) ~Time taken by different cv_workflow to execute specific activities on CV (will help identify if delays are related to the specific aspect of the workflow/CV)~
- Allow to:
  * Fully suppress returned build warnings
  * Suppress warnings based on custom regex patterns
  * Suppress warnings based on pre-defined use-cases (like enabling `portfast` on switchport)
- Extensive pytests

## How to test
 - pytest (`python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py`)

Example with default warning suppression settings (no suppression of warnings enabled):
```json
        "workspace": {
            "build_id": "req-0a4c4607-93ed-4291-b290-90c69eaef0b2",
            "build_results": [
                {
                    "config_validation": {
                        "errors": [
                            {
                                "configlet_name": "avd-A7D46613D44DE45D68D5B5C5CBA06B0D",
                                "error_msg": "> spanning-tree mode spine2 % Invalid input (at token 2: 'spine2')",
                                "line_num": 17
                            }
                        ],
                        "warnings": []
                    },
                    "device": {
                        "_exists_on_cv": true,
                        "hostname": "avd-ci-spine2",
                        "serial_number": "A7D46613D44DE45D68D5B5C5CBA06B0D",
                        "system_mac_address": "50:00:00:cb:38:c2"
                    },
                },
                {
                    "config_validation": {
                        "errors": [
                            {
                                "configlet_name": "avd-DCC816CEAC4BBD6319385043AD318362",
                                "error_msg": "> spanning-tree mode spine1 % Invalid input (at token 2: 'spine1')",
                                "line_num": 17
                            },
                            {
                                "configlet_name": "avd-DCC816CEAC4BBD6319385043AD318362",
                                "error_msg": ">    seqq 10 permit 10.20.0.0/24 eq 32 % Invalid input (at token 0: 'seqq')",
                                "line_num": 73
                            }
                        ],
                        "warnings": [
                            {
                                "configlet_name": "avd-DCC816CEAC4BBD6319385043AD318362",
                                "warning_msg": "! portfast should only be enabled on ports connected to a single host. Connecting hubs, concentrators, switches, bridges, etc. to this interface when portfast is enabled can cause temporary bridging loops. Use with CAUTION.",
                                "line_num": 41
                            },
                            {
                                "configlet_name": "avd-DCC816CEAC4BBD6319385043AD318362",
                                "warning_msg": "! /32 IPv4 address is not configured on the interface",
                                "line_num": 49
                            }
                        ]
                    },
                    "device": {
                        "_exists_on_cv": true,
                        "hostname": "avd-ci-spine1",
                        "serial_number": "DCC816CEAC4BBD6319385043AD318362",
                        "system_mac_address": "50:00:00:d7:ee:0b"
                    },
                },
                {
                    "config_validation": {
                        "errors": [],
                        "warnings": [
                            {
                                "configletName": "avd-13C20F1EDCCED2D85F6DB2FB9E3AC5B6",
                                "warning_msg": "! portfast should only be enabled on ports connected to a single host. Connecting hubs, concentrators, switches, bridges, etc. to this interface when portfast is enabled can cause temporary bridging loops. Use with CAUTION.",
                                "line_num": 59
                            }
                        ]
                    },
                    "device": {
                        "_exists_on_cv": true,
                        "hostname": "avd-ci-leaf1",
                        "serial_number": "13C20F1EDCCED2D85F6DB2FB9E3AC5B6",
                        "system_mac_address": "50:00:00:72:8b:31"
                    },
                }
            ],
            "build_warnings": true,
            "build_warnings_suppress_patterns": [],
            "build_warnings_suppress_portfast": false,
            "change_control_id": null,
            "description": "AVD-CV-DEPLOY-WORKSPACE-INVALID-CONFIGS-RX0B-CONVERGE",
            "force": false,
            "id": "ws-7c93e948-7676-4ce1-9667-7328e098643c",
            "name": "avd-cv-deploy-workspace-invalid-configs-rx0b-converge",
            "requested_state": "built",
            "state": "build failed"
        }
```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/cv_deploy/cc_false.yml
```diff
@@ -12,46 +12,54 @@
...
+    - name: "{{ test_id | upper }} Generate random string"
+      run_once: true
       ansible.builtin.set_fact:
         r: "{{ lookup('password', '/dev/null chars=ascii_lowercase,digits length=4') }}"
-      run_once: true
 
-    - name: Set facts
+    - name: "{{ test_id | upper }} Set deploy facts"
+      run_once: true
       ansible.builtin.set_fact:
```
ClausHolbechArista: Since you are now changing it completely to a pattern, we don't have to have this set_fact task at all. Just give the inputs on the cv_deploy task itself. Fewer tasks makes it easier to follow.

ansible_collections/arista/avd/molecule/cv_deploy/invalid_config.yml
```diff
@@ -0,0 +1,129 @@
...
+          ansible.builtin.debug:
+            msg: '{{ cv_deploy_results }}'
+
+        - name: "{{ test_id | upper }} Check CVP returns"
+          run_once: true
+          ansible.builtin.assert:
+            that:
+              # assert validate config build errors
+              - >
+                cv_deploy_results.workspace.build_errors | selectattr('config_validation', 'defined') | map(attribute='config_validation') | list | flatten |
```
ClausHolbechArista: Let's make the build_errors easier to pass. Instead of having to fish the relevant errors out of the list which could have different errors, I would rather have the list be of config validation errors directly. Like `cv_deploy_results.workspace.build_errors.config_validation`.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -87,3 +91,50 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
         return
 
     return
+
+
+async def process_workspace_build_errors(
+    workspace_id: str,
+    build_id: str,
+    cv_client: CVClient,
+) -> list[CVWSBuildError]:
```
ClausHolbechArista: This is returning a list of dicts. Please enable pyright (part of pylance) in your repo to assist with type checking.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -87,3 +91,50 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+
+    Returns:
+        List of CVWSBuildError objects describing observed WS build validation errors.
+    """
+    try:
+        ws_build_details = await cv_client.get_workspace_build_details(workspace_id=workspace_id, build_id=build_id)
+
+        ws_build_errors = [
+            {
+                "serial_number": response.value.key.device_id,
```
ClausHolbechArista: Let's think about the use cases. Maybe it will be better to have a list of config errors and a list of other errors. To make it easier to parse.
Seing the model you get from CV, I think I would just return `config_validation_results` as the list. An item per device with sub lists of warnings and errors.
Feel free to push back. Maybe the use of this will make it clear.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -73,6 +73,23 @@ class CVPathfinderMetadata:
     device: CVDevice | None = None
 
 
+@dataclass
+class CVWSBuildConfigValidationError:
```
ClausHolbechArista: We prefer not to have abbreviated variables. Also consider if this class is always an error or could it be a warning?
```suggestion
class CVWorkspaceBuildConfigValidationError:
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -73,6 +73,23 @@ class CVPathfinderMetadata:
...
+    configlet_name: str | None = None
+    """Name of the Studio-generated configlet which raised validation error(s)."""
+    error_msg: str | None = None
+    """EOS-returned error message."""
+    line_num: int | None = None
+    """Line number of the violating configuration line within the proposed configlet."""
+
+
+@dataclass
+class CVWSBuildError:
```
ClausHolbechArista: Since this could have warnings as well, we should consider renaming. Maybe also have a flag if we get it from CV if the device failed or succeeded.
```suggestion
class CVWorkspaceBuildResult:
    """Workspace build result for one device."""

```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -40,8 +40,12 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
 
     workspace_config = await cv_client.build_workspace(workspace_id=workspace.id)
     build_result, cv_workspace = await cv_client.wait_for_workspace_response(workspace_id=workspace.id, request_id=workspace_config.request_params.request_id)
+    workspace.build_id = cv_workspace.last_build_id
     if build_result.status != ResponseStatus.SUCCESS:
         workspace.state = "build failed"
+        ws_build_errors = await process_workspace_build_errors(workspace_id=workspace.id, build_id=workspace.build_id, cv_client=cv_client)
```
ClausHolbechArista: Even for a successful build we could get warnings. I think we should get these build results no matter what and then log out warnings and errors (which will be picked up correctly by ansible as well)

ClausHolbechArista: We could consider adding a knob to avoid logging the warnings.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/cv_deploy/cc_false.yml
alexeygorbunov: Addressed in `a9043ed8d Remove separate set_facts tasks`

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: (fixed) Converted this (commit 75c4de28a) to fully use pre-defined models

ansible_collections/arista/avd/molecule/cv_deploy/invalid_config.yml
alexeygorbunov: (fixed) Currently converted to the following model (as both config and image validation can have both errors and warnings), exposing per-device image and config items in a separate lists (config validation errors, config validation warnings, image validation errors, image validation warnings):
```json
                {
                    "config_validation": {
                        "errors": [
                            {
                                "configlet_name": <str>,
                                "error_code":  <str>,
                                "error_msg":  <str>,
                                "line_num": <int>
                            },
                            {
                               ...
                            }
                        ],
                        "warnings": [
                            {
                                "configlet_name":  <str>,
                                "error_code":  <str>,
                                "error_msg":  <str>,
                                "line_num": <int>
                            },
                            {
                                ...
                            }
                        ]
                    },
                    "image_validation": {
                        "errors": [
                            {
                                "sku": <str>,
                                "error_code":  <str>,
                                "error_msg":  <str>
                            },
                            {
                               ...
                            }
                       ],
                        "image_input_error": "",
                        "warnings": [
                            {
                                "sku": <str>,
                                "warning_code":  <str>,
                                "warning_msg":  <str>
                            },
                            {
                               ...
                            }
                      ]
                    },
                    "serial_number": "DCC816CEAC4BBD6319385043AD318362",
                    "stages_states": [
                        {
                            "stage": "BUILD_STAGE_IMAGE_VALIDATION",
                            "state": "SKIPPED"
                        },
                        {
                            "stage": "BUILD_STAGE_INPUT_VALIDATION",
                            "state": "SUCCESS"
                        },
                        {
                            "stage": "BUILD_STAGE_CONFIG_VALIDATION",
                            "state": "FAIL"
                        },
                        {
                            "stage": "BUILD_STAGE_CONFIGLET_BUILD",
                            "state": "SUCCESS"
                        }
                    ]
                }
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: (fixed) Tried to implement this model in 75c4de28a

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: (fixed) Changed behavior to always pull build details with ability to suppress:
* all warnings
* specific warnings based on a known behavior (like repeatable warnings regarding enabling `portfast` on a `switchport`)
* warnings matching custom string patterns controlled by the user

```yaml
# Fetch and expose Workspace build warnings.
# Suppress specific warnings based on pre-defined options or custom string(s).
cv_workspace_build_warnings:
  enabled: true
  suppress_patterns: []
  suppress_portfast: false
```

ansible_collections/arista/avd/molecule/cv_deploy/invalid_config.yml
ClausHolbechArista: perfect. This will also make it possible to later insert the config diff into the `config_validation` dict.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
```diff
@@ -32,6 +32,9 @@
       description: "{{ cv_workspace_description | arista.avd.default }}"
       requested_state: "{{ 'submitted' if cv_submit_workspace else 'built' }}"
       force: "{{ cv_submit_workspace_force }}"
+      build_warnings: "{{ cv_workspace_build_warnings.enabled | arista.avd.default(true) }}"
+      build_warnings_suppress_patterns: "{{ cv_workspace_build_warnings.suppress_patterns | arista.avd.default([]) }}"
+      build_warnings_suppress_portfast: "{{ cv_workspace_build_warnings.suppress_portfast | arista.avd.default(false) }}"
```
ClausHolbechArista: you don't need the default filter here since you already set them in the role defaults.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -87,3 +101,157 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+        ]
+
+    except Exception as e:
+        LOGGER.warning(
+            "process_workspace_build_details: Failed to fetch and process Workspace build results for workspace '%s' and build_id '%s'. Error: '%s'",
+            workspace.id,
+            workspace.build_id,
+            e,
+        )
+        return []
```
ClausHolbechArista: Let's to this try/except in the calling function instead of wrapping the full code here, making it more indented and harder to read.

ClausHolbechArista: Or even better make sure to try/except only the bits where you expect an issue that cannot be caught in a better way.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -87,3 +101,157 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+                    CVWorkspaceBuildStageState(
+                        stage=stage,
+                        state=getattr(state, "name", None),
+                    )
+                    for stage, state in getattr(getattr(getattr(response, "value", None), "build_stage_state", None), "values", {}).items()
+                ],
+                config_validation=CVWorkspaceBuildConfigValidationResult(
+                    errors=[
+                        CVWorkspaceBuildConfigValidationError(
+                            error_code=getattr(getattr(config_validation_error, "error_code", None), "name", None),
```
ClausHolbechArista: If we really need this, then please use get_v2 under our utils. It shoudl do exactly this.
It would be better to check the type of response early and then trust the API classes generated from the protos.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -87,3 +101,157 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
         return
 
     return
+
+
+async def process_workspace_build_details(
```
ClausHolbechArista: Please try to split this out in smaller nicer functions that take care of each part, and then call them from this main function.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -73,6 +73,131 @@ class CVPathfinderMetadata:
...
+            "TA_CV_INCOMPATIBLE",
+            "EOS_CV_INCOMPATIBLE",
+            "EOS_SUPPORT_NOT_INTRODUCED",
+            "EOS_SUPPORT_REMOVED",
+            "PHYSICAL_DEVICE_EOS_INCOMPATIBLE",
+            "TA_EMBEDDEDEXT_INCOMPATIBLE",
+            "DEVICE_EOS_2GB_INCOMPATIBLE",
+            "EOS_EXTENSION_VERSION_INCOMPATIBLE",
+            "ARCH_UNSUPPORTED",  # NEW CODE ADDED IN CVP BUT NOT TRACKED BY pyavd/_cv/api/arista/imagestatus/v1/__init__.py
+        ]
```
ClausHolbechArista: Lets try to grab this automatically from the imagestatus API so we don't need to maintain it here.
If CV implements something new, we may just need to regenerate the api wrappers. (Run the make file in the _cv/api directory)

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -73,6 +73,131 @@ class CVPathfinderMetadata:
...
+    """Error code of the returned error."""
+    error_msg: str | None = None
+    """EOS-returned error message."""
+
+
+@dataclass
+class CVWorkspaceBuildImageValidationWarning:
+    sku: str | None = None
+    """Name of the sku."""
+    warning_code: (
```
ClausHolbechArista: Try to grab these from the generated api code so we don't have to maintain it here. Otherwise just use a string.

ClausHolbechArista (changes requested):

ClausHolbechArista:
Moving to draft until comments have been addressed.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
alexeygorbunov: fixed, but had to fallback to plain-level variables now (vs dict), otherwise defining only one key (like 'enabled') when calling role would make other dict attributes undefined

was:
```
cv_workspace_build_warnings:
  enabled: true
  suppress_patterns: []
  suppress_portfast: false

```

now:
```
cv_workspace_build_warnings_enabled: true
cv_workspace_build_warnings_suppress_patterns: []
cv_workspace_build_warnings_suppress_portfast: false
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed. Removed that original large try/except as the only embedded call 'cv_client.get_workspace_build_details' already has try/except in it and the rest of the code is just a processing of the vars

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: WS build details may just return empty dict in case there are no errors, like

```
values.config_validation_result.errors: {}
```

but if actual errors were faced - it'll return nested dict (with list of ConfigError objects) that we need to loop through

```
values.config_validation_result.errors.values: [
    ConfigError
]
```

alexeygorbunov: fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixes. Split logic across 2 additional functions with smaller scope

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Fixed. Regenerated CV API wrappers as well (as there were updates to the things we use, like addition of ARCH_UNSUPPORTED under imagestatus ErrorCode)

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Fixed

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
ClausHolbechArista: Otherwise go back to dict and just remove it from `default/main.yml` and add back the arista.avd.default filters here. Up to you.

ansible_collections/arista/avd/molecule/cv_deploy/cc_false.yml
```diff
@@ -12,46 +12,50 @@
     structured_dir: "{{ playbook_dir }}/intended/structured_configs/test_configs"
     intended_tag_device: avd-ci-leaf1
     intended_tags: "{{ lookup('file', structured_dir ~ '/' ~ intended_tag_device ~ '.yml')| from_yaml }}"
+    test_id: "cc-false"
```
ClausHolbechArista: Should we also add tags with this name on all the tasks related to this test? To make it easier to just run some of them?

ansible_collections/arista/avd/molecule/cv_deploy/intended/configs/test_invalid_configs/avd-ci-core1.cfg
```diff
@@ -0,0 +1,69 @@
...
+   ip address 10.90.231.70/24
+!
+no ip routing
+no ip routing vrf MGMT
+!
+ip route vrf MGMT 0.0.0.0/0 10.90.231.1
+!
+ntp server vrf MGMT 1.pool.ntp.org iburst local-interface Management1
+!
+end
```
ClausHolbechArista: Please make sure all config files have a closing new-line. It seems like some do and some dont.

ansible_collections/arista/avd/molecule/cv_deploy/intended/structured_configs/test_invalid_configs/avd-ci-leaf2.yml
```diff
@@ -0,0 +1,2 @@
+hostname: avd-ci-leaf2
```
ClausHolbechArista: We can remove these files now, since they only contain the default values.

python-avd/pyavd/_cv/client/workspace.py
```diff
@@ -254,3 +259,39 @@ async def wait_for_workspace_response(
...
         except Exception as e:
             raise get_cv_client_exception(e, f"Workspace ID '{workspace_id}', Request ID '{request_id}") or e
+
+    async def get_workspace_build_details(
+        self: CVClient,
+        workspace_id: str,
+        build_id: str,
+        time: datetime | None = None,
+        timeout: float = DEFAULT_API_TIMEOUT,
+    ) -> WorkspaceBuildDetailsStreamResponse:
```
ClausHolbechArista: Please try to normalise the output a bit, so we don't return raw response objects. The resonses object is an async generator, so we need to process it inside this function to catch errors correctly. You could do `return [response.to_dict(Casing.SNAKE) async for response in reponses]`

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
alexeygorbunov: We can keep current flatten approach (used by most of the other cv_deploy's vars anyway) as long as you are OK with it

ansible_collections/arista/avd/molecule/cv_deploy/cc_false.yml
alexeygorbunov: Added. See commit 'PR comments - tag cv_deploy molecule play tasks for flexible scope control'

ansible_collections/arista/avd/molecule/cv_deploy/intended/configs/test_invalid_configs/avd-ci-core1.cfg
alexeygorbunov: Fixed in commit 'PR comments - end-of-file-fixer for avd-ci-core1.cfg'.
Easy to miss as pre-commit excludes these molecule files

ansible_collections/arista/avd/molecule/cv_deploy/intended/structured_configs/test_invalid_configs/avd-ci-leaf2.yml
alexeygorbunov: Fixed in commit `PR comments - Delete structure_config files containing only default settings`

python-avd/pyavd/_cv/client/workspace.py
alexeygorbunov: Fixed in 'PR Comments - normalize return of get_workspace_build_details'. Tested in CI environment with cv_deploy molecule scenario

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
alexeygorbunov: moved everything to pytest

ansible_collections/arista/avd/molecule/cv_deploy/cc_false.yml
alexeygorbunov: Moved everything to pytest

ansible_collections/arista/avd/molecule/cv_deploy/intended/configs/test_invalid_configs/avd-ci-core1.cfg
alexeygorbunov: Not applicable to this PR anymore

carl-baillargeon (changes requested):
Let's discuss internally if we should import the API classes from cloudvision-python instead.

alexeygorbunov:
> Let's discuss internally if we should import the API classes from cloudvision-python instead.

@carl-baillargeon just rechecked and no binding updates are needed. Removed all updates related to `_cv/api`

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -154,6 +154,19 @@
         description: Force submit the workspace even if some devices are not actively streaming to CloudVision.
         type: bool
         default: false
+      build_warnings:
```
ClausHolbechArista, line 157: I expect the list of suppression knobs will grow a lot. Can we do a dict like:

```yaml
build_warnings:
  enabled: <bool; default=true>
  suppress_patterns:
    - <str>
  suppress_portfast: <bool; default=false>
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -220,6 +221,12 @@ cv_submit_workspace: true
 # If set, configurations will not be validated for non-streaming devices.
 cv_submit_workspace_force: false
 
+# Fetch and expose Workspace build warnings.
+# Suppress specific warnings based on pre-defined options or custom regex pattern(s).
+cv_workspace_build_warnings_enabled: true
+cv_workspace_build_warnings_suppress_patterns: []
+cv_workspace_build_warnings_suppress_portfast: false
```
ClausHolbechArista, line 228: we can mirror the dict here.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +127,154 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+
+    Parameters:
+        workspace: Active Workspace.
+        static_suppression_patterns: Dictionary of pre-defined Regex string patterns to match undesired warnings.
+
+    Returns:
+        Unified list of compiled Regex patterns describing Workspace Build warnings that we would like to suppress.
+    """
+    # Deduplicate list of patterns from user
+    workspace_build_warning_suppress_list = list(set(workspace.build_warnings_suppress_patterns[:]))
```
ClausHolbechArista: I think we could just leave this as a set, and then remove some of the checks for the static patterns, so we don't have to check if it is there already but just add it (since the set will dedup)

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +127,154 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+        and (pattern := static_suppression_patterns.get("portfast"))
+        and pattern not in workspace_build_warning_suppress_list
+    ):
+        workspace_build_warning_suppress_list.append(pattern)
+
+    compiled_workspace_build_warnings_suppress_list: list[Pattern] = []
+    for proposed_regex_pattern in workspace_build_warning_suppress_list:
+        try:
+            compiled_workspace_build_warnings_suppress_list.append(re_compile(proposed_regex_pattern))
+        except Exception as e:  # noqa: PERF203
```
ClausHolbechArista: let's try to be more specific on the exception we catch here.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +127,154 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+    Parameters:
+        workspace: Active Workspace.
+        workspace_build_details: List of WorkspaceBuildDetails objects.
+        compiled_workspace_build_warnings_suppress_list: List of compiled Regex patterns for matching undesired warnings.
+        devices: List of CVDevice objects representing existing CV devices.
+
+    Returns:
+        List of CVWorkspaceBuildResult objects defining details of the Workspace build.
+    """
+    return [
```
ClausHolbechArista: please refactor this do not have as many nested comprehension. It makes if quite hard to read and follow.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +127,154 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+            or (
+                # Render build warnings if requested in inputs and warnings not matching suppress patterns are present
+                workspace.build_warnings
+                and workspace_build_details_item.config_validation_result.warnings
+                and (
+                    # Either no warnings suppression patterns are provided in inputs
+                    len(compiled_workspace_build_warnings_suppress_list) == 0
+                    # Or at least one warning message is not matched by the suppression patterns
+                    or {False}
+                    in [
```
ClausHolbechArista: I think we should use full_match instead of search so we don't have to bother with multiple returns. and then we can use `not any()` to stop on first match isntead of going through all of them.

ClausHolbechArista: we could also have a helper function to do the checks and maybe move the static ones in there, so we treat them special, to avoid having to do regex on things that are static strings.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Right now the role uses "flat" variable model (`cv_change_control_name`, `cv_change_control_description`, etc.) in all cases (except manifest where we have no defaults), including existing keys of the `workspace`

```yaml
  change_control:
    description: CloudVision Change Control to create for the deployment.
    type: dict
    suboptions:
      name:
        description: Optional name to use for the created Change Control. By default the name generated by CloudVision will be kept.
        type: str
      description:
        description: Optional description to use for the created Change Control.
        type: str
      requested_state:
        description: |-
          The requested state for the Change Control.

          - `pending approval` (default): Leave the Change Control in "pending approval" state.
          - `approved`: Approve the Change Control but do not start.
          - `running`: Approve and start the Change Control. Do not wait for the Change Control to be completed or failed.
          - `completed`: Approve and start the Change Control. Wait for the Change Control to be completed.
        type: str
        default: pending approval
        choices: ["pending approval", "approved", "running", "completed"]
```

# action plugin invocation
```yaml
  arista.avd.cv_workflow:
...
    change_control:
      name: "{{ cv_change_control_name | arista.avd.default }}"
      description: "{{ cv_change_control_description | arista.avd.default }}"
      requested_state: "{{ 'completed' if cv_run_change_control else 'pending approval' }}"
```

Shouldn't we stick to the same approach?

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: changed to `re.error` (should be the only one possible here)

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Changed to `set`

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
alexeygorbunov: Converted this to dict and added CVWorkspaceBuildWarnings dataclass

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Refactored this func

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Switched to `fullmatch`

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -154,6 +154,23 @@
...
+        description: Configuration for Workspace build warnings handling.
+        type: dict
+        suboptions:
+          enabled:
+            description: Fetch and expose Workspace build warnings.
+            type: bool
+            default: true
+          suppress_patterns:
+            description: Arbitrary list of the EoS CLI warning string patterns to suppress.
```
carl-baillargeon:
```suggestion
            description: Arbitrary list of the EOS CLI warning string patterns to suppress.
```

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -255,6 +272,10 @@
         #   id: <uuid or similar>
           requested_state: submitted
           force: true
+          # build_warnings:
+          #   enabled: true
+          #   suppress_patterns: []
```
carl-baillargeon: Can we add an example here?

ansible_collections/arista/avd/roles/cv_deploy/README.md
carl-baillargeon: I vote to stick with the same approach.

python-avd/pyavd/_cv/workflows/constants.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) 2025-2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+
+EOS_CLI_WARNINGS = {
+    # Text pattern to match EoS CLI warning regarding utilization of the spanning-tree portfast feature.
```
carl-baillargeon:
```suggestion
    # Text pattern to match EOS CLI warning regarding utilization of the spanning-tree portfast feature.
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+        static_suppression_patterns: Dictionary of pre-defined Regex string patterns to match undesired warnings.
+
+    Returns:
+        Unified list of compiled Regex patterns describing Workspace Build warnings that we would like to suppress.
+    """
+    # Deduplicate list of patterns from user
+    workspace_build_warnings_suppress_set = set(workspace.build_warnings.suppress_patterns[:])
+    # Add predefined static patterns if requested
+    if workspace.build_warnings.suppress_portfast and (pattern := static_suppression_patterns.get("portfast")):
+        workspace_build_warnings_suppress_set.add(pattern)
```
carl-baillargeon: EOS_CLI_WARNINGS is controlled by us so it is safe to do:
```suggestion
    if workspace.build_warnings.suppress_portfast:
        workspace_build_warnings_suppress_set.add(EOS_CLI_WARNINGS["portfast"])
```
and remove the `static_suppression_patterns` param from the function.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+
+    Parameters:
+        workspace: Active Workspace.
+        static_suppression_patterns: Dictionary of pre-defined Regex string patterns to match undesired warnings.
+
+    Returns:
+        Unified list of compiled Regex patterns describing Workspace Build warnings that we would like to suppress.
+    """
+    # Deduplicate list of patterns from user
+    workspace_build_warnings_suppress_set = set(workspace.build_warnings.suppress_patterns[:])
```
carl-baillargeon:
```suggestion
    workspace_build_warnings_suppress_set = set(workspace.build_warnings.suppress_patterns)
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+        CVDevice object matching the serial number.
+    """
+    # TODO: Can there be a case where no device is found here? Like if we apply Tag changes that lead to the config change for the device we have not targeted?
+    try:
+        return next(device for device in devices if device.serial_number == serial_number)
+    except StopIteration:
+        return CVDevice(hostname="Unknown Device (was not directly targeted by AVD deployment)", serial_number=serial_number)
+
+
+async def _produce_cvworkspace_build_result(
```
carl-baillargeon:
```suggestion
def _produce_cvworkspace_build_result(
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -132,6 +132,49 @@ class CVPathfinderMetadata:
...
+    config_validation: CVWorkspaceBuildConfigValidationResult
+    """Configuration validation results."""
+
+
+@dataclass
+class CVWorkspaceBuildWarnings:
+    enabled: bool = True
+    """Fetch and expose Workspace build warnings."""
+    suppress_patterns: list[str] = field(default_factory=list)
+    """Arbitrary list of the EoS CLI warning string patterns to suppress."""
```
```suggestion
    """Arbitrary list of the EOS CLI warning string patterns to suppress."""
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -155,6 +198,12 @@ class CVWorkspace:
     """The final state of the Workspace. Do not set this manually."""
     change_control_id: str | None = None
     """Do not set this manually."""
+    build_id: str | None = None
+    """last_build_id of the Workspace. Used to fetch build details related to the last Workspace build attempt."""
```
carl-baillargeon: Shoud we add a note that this should not be set manually (like change_control_id)?

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
```diff
@@ -0,0 +1,674 @@
...
+        assert not workspace.build_warnings.suppress_patterns
+        assert not workspace.build_warnings.suppress_portfast
+        assert not warnings
+        assert len(workspace.build_results) == 3
+        for build_result in workspace.build_results:
+            match build_result.device.hostname:
+                case "avd-ci-leaf1":
+                    assert not build_result.config_validation.errors
+                    assert len(build_result.config_validation.warnings) == 1
+                    assert CVWorkspaceBuildConfigValidationWarning(
```
carl-baillargeon: What are you expected to test here? This line constructs a dataclass instance and assert its truthiness, which always passes. It doesn't check the field values.

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
```diff
@@ -0,0 +1,674 @@
...
+                        CVWorkspaceBuildConfigValidationError(
+                            error_msg=">    seqq 10 permit 10.20.0.0/24 eq 32 % Invalid input (at token 0: 'seqq')",
+                            line_num=4,
+                            configlet_name="avd-DCC816CEAC4BBD6319385043AD318362",
+                        )
+                        in build_result.config_validation.errors
+                    )
+
+                    assert len(build_result.config_validation.warnings) == 2
+                    assert CVWorkspaceBuildConfigValidationWarning(
```
carl-baillargeon: Same here. This is not testing anything.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+
+def _prepare_build_warnings_suppress_patterns(
+    workspace: CVWorkspace,
+    static_suppression_patterns: dict[str, str],
+    warnings: list,
+) -> list[Pattern]:
+    """
+    Process static and custom inputs for Build Warnings Suppression and return unified list of Regex patterns.
+
+    Parameters:
```
carl-baillargeon, line 140: Missing `warnings` parameter.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+    Check if a warning message matches any of the suppression patterns.
+
+    Parameters:
+        warning_msg: Warning message to check.
+        compiled_workspace_build_warnings_suppress_list: List of compiled Regex patterns for matching undesired warnings.
+
+    Returns:
+        True if the warning message matches any suppression pattern, False otherwise.
+    """
+    return any(pattern.fullmatch(warning_msg) for pattern in compiled_workspace_build_warnings_suppress_list)
```
carl-baillargeon, line 179: This is a full match against user-provided patterns right? Should we consider using search which is more user-friendly? If not we should at least update Ansible docs so users know they need full-string patterns.

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
```diff
@@ -0,0 +1,674 @@
...
+            system_mac_address="50:00:00:cb:38:c2",
+            _exists_on_cv=True,
+            _streaming=True,
+        ),
+    )
+    workspace_id: str = MOCKED_WORKSPACE_ID
+    workspace_build_id: str = MOCKED_WORKSPACE_REQUEST_ID_BUILD_FAIL_CONFIG_VALIDATION["id"]
+
+    @pytest.mark.parametrize(("build_warnings"), [pytest.param(False, id="BUILD_WARNINGS_FALSE"), pytest.param(True, id="BUILD_WARNINGS_TRUE")])
+    async def test_clean_build(self, cv_client: CVClient, build_warnings: bool) -> None:
```
carl-baillargeon, line 118: This test parametrizes over True/False for build_warnings, but since the build has no warnings/errors, both paths produce the same result.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,260 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+        serial_number: Serial number to search for.
+
+    Returns:
+        CVDevice object matching the serial number.
+    """
+    # TODO: Can there be a case where no device is found here? Like if we apply Tag changes that lead to the config change for the device we have not targeted?
+    try:
+        return next(device for device in devices if device.serial_number == serial_number)
+    except StopIteration:
+        return CVDevice(hostname="Unknown Device (was not directly targeted by AVD deployment)", serial_number=serial_number)
```
carl-baillargeon: Building a `CVDevice` with a hardcoded hostname to fit it into the results feels wrong. I think a cleaner approach would be to log a warning and skip that build details item entirely in `_produce_cvworkspace_build_result`.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -155,6 +198,12 @@ class CVWorkspace:
...
+    build_warnings: CVWorkspaceBuildWarnings = field(default_factory=CVWorkspaceBuildWarnings)
+    """Fetch and expose Workspace build warnings."""
+    build_results: list[CVWorkspaceBuildResult] = field(default_factory=list)
+    """Details of Workspace build results."""
```
carl-baillargeon: I think the naming is confusing. `CVWorkspaceBuildWarnings` sounds like a list of warning messages from the build but from my understanding it's actually configuration/input knobs for enabling/suppressing warnings.

For `build_results: list[CVWorkspaceBuildResult]`, it sounds like the overall build outcome but it's actually per-device validation results (only for devices with issues).

I think we should go with something like:
```
build_warnings: CVWorkspaceBuildWarningsConfig
device_build_results: list[CVWorkspaceDeviceBuildResult]
```
...unless I am missing something.

carl-baillargeon (changes requested):

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
alexeygorbunov: Uncommented these lines and provided a custom suppress pattern as an example

python-avd/pyavd/_cv/workflows/constants.py
alexeygorbunov: Applied suggested change

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Applied suggested approach

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Added `warnings` param

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: True, no need to slice here. Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed, thanks!

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Added

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
alexeygorbunov: This is exactly what i was trying to test - that code produces no build warnings/errors when submitted designed config is clean.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: It was initially using search but i changed it to fullmatch based on the earlier feedback/comment. Updated Ansible side documentation to clarify the nature of matching logic

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Replaced it with warning-level log message and added test coverage for this use case

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: You are correct, descriptions are not accurately reflecting the purpose.. Applied your suggestions

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
alexeygorbunov: Yeah, i missed actual comparison here an in few other places. Should now be fixed everywhere

python-avd/tests/pyavd/cv/workflows/test_finalize_workspace_on_cv_invalid_config.py
alexeygorbunov: Should be fixed now

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
carl-baillargeon: I think we keep the comments for non-default values.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -114,3 +128,267 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient,
...
+        devices: List of CVDevice objects.
+        serial_number: Serial number to search for.
+
+    Returns:
+        CVDevice object matching the serial number.
+        None in case requested device is not in the list of the targeted devices.
+    """
+    # TODO: Can there be a case where no device is found here? Like if we apply Tag changes that lead to the config change for the device we have not targeted?
+    try:
+        return next(device for device in devices if device.serial_number == serial_number)
```
carl-baillargeon: You can do a one-liner:
```suggestion
        return next((device for device in devices if device.serial_number == serial_number), None)
```
With this I don't think the helper is needed anymore.

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
alexeygorbunov: Updated that example to keep options with default values - commented out

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Thanks! Applied suggestion

carl-baillargeon (approved):
LGTM if internal CI passes.

alexeygorbunov:
@carl-baillargeon CI succeeded twice (latest run: https://github.com/aristanetworks/avd-internal/actions/runs/23521595713)

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -201,7 +201,8 @@ By default the role will
 2. Push all configurations and tags.
 3. Unassign tags
 4. Build and submit the Workspace.
-5. Leave any created Change Control in `pending approval` state.
+5. Fetch and expose errors and warnings raised during the Workspace Build phase.
```
ClausHolbechArista, line 204: I am not sure where we write it, but I think we need to explain how it is exposed.

ClausHolbechArista (approved):

PR #4700, opened 2024-11-08
Refactor(eos_designs): Structured config output

ClausHolbechArista:
## Change Summary

Refactor eos_designs structured config logic to leverage classes to hold the generated structured config instead of dicts/lists etc.

## Proposed changes

- Create new base class StructuredConfigGenerator with structured config and custom structured config as extra attributes compared to AvdFacts.
- Refactor the render() function to merge into a single instance of structured_config.
- Refactor all custom structured config to merge into a class instances which will be merged onto structured_config at the end.
- Fix a few bugs in class merge logic.
- Improve development/compare.py to work for yaml.

## How to test

- No config changes.
- Compare structured config with
  ```sh
  # Running from repo root while having the branch checked out and devel fully up to date.
  GIT_EXTERNAL_DIFF=development/compare.py git diff devel --ext-diff -- **/structured_configs/*.yml
  ```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/bgp-peer-groups-1.cfg
```diff
@@ -136,7 +136,5 @@ router bgp 65001
       no neighbor EVPN-OVERLAY-PEERS activate
       neighbor IPv4-UNDERLAY-PEERS activate
       neighbor MLAG-IPv4-UNDERLAY-PEER activate
-   !
-   address-family vpn-ipv4
 !
```
ClausHolbechArista, line 139: This change is a bugfix. We should not render useless config. The change is caused by the new `_strip_empties` call where we remove all empty lists/dicts/None value from the output.

ansible_collections/arista/avd/tests/integration/targets/filter_hide_passwords/tasks/main.yml
```diff
@@ -2,5 +2,5 @@
 - name: Test hide_passwords filter
   ansible.builtin.assert:
     that:
-      - "{{ 'dummy' | arista.avd.hide_passwords(false) == 'dummy' }}"
-      - "{{ 'dummy' | arista.avd.hide_passwords(true) == '<removed>' }}"
+      - "'dummy' | arista.avd.hide_passwords(false) == 'dummy'"
```
ClausHolbechArista, line 5: Unrelated changes, but saw the warnings from ansible-test while troubleshooting other things.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -295,30 +293,6 @@ def _deepmerge(self, other: Self, list_merge: Literal["append", "replace"] = "ap
             # We merged into a "null" class, but since we now have proper data, we clear the flag.
             self._created_from_null = False
 
-    def _inherit(self, other: Self) -> None:
```
ClausHolbechArista, line 298: Removed since we never use it.

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -118,32 +108,42 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
...
+                # If there is any data produced by the template, convert and merge it on top of previous output.
+                if template_result_data:
+                    # Some templates return a list of dicts, others only return a dict. Here we normalize to list.
+                    if not isinstance(template_result_data, list):
+                        template_result_data = [template_result_data]
+
+                    try:
+                        merge(output, *template_result_data, list_merge=list_merge, schema=output_schema_tools.avdschema)
+                    except Exception as error:
+                        raise AnsibleActionFail(message=str(error)) from error
```
```suggestion
                if not template_result_data:
                    continue

               # There is data produced by the template, convert and merge it on top of previous output.
               # Some templates return a list of dicts, others only return a dict. Here we normalize to list.
               if not isinstance(template_result_data, list):
                   template_result_data = [template_result_data]

               try:
                   merge(output, *template_result_data, list_merge=list_merge, schema=output_schema_tools.avdschema)
               except Exception as error:
                   raise AnsibleActionFail(message=str(error)) from error
```

development/compare.py
```diff
@@ -80,6 +82,11 @@ def main() -> None:
     old = Path(args.old_file).read_text()
     new = Path(args.new_file).read_text()
 
+    if args.old_file.endswith("yml"):
+        old = yaml.dump(yaml.load(old, Loader=yaml.CSafeLoader), Dumper=yaml.CDumper, sort_keys=True, indent=2)
+    if args.new_file.endswith("yml"):
```
gmuloc: probably safe to add `yaml` as well?

python-avd/pyavd/_eos_designs/structured_config/__init__.py
```diff
@@ -86,36 +84,28 @@ def get_structured_config(
...
+    structured_config = EosCliConfigGen()
 
-        # All lists will be merged with "append" except for custom structured configuration where
-        # the default list merge is "append_rp" and can be overridden.
-        # TODO: Each dict entry can contain a list_merge key, which will be picked up by the merge function for all underlying lists.
-        if issubclass(cls, AvdStructuredConfigCustomStructuredConfiguration):
-            list_merge = get(module_vars, "custom_structured_configuration_list_merge", default="append_rp")
+    # Placeholder for custom structured configs added by the structured config generators.
+    # Will be applied last by AvdStructuredConfigCustomStructuredConfiguration.
+    # "root" are the legacy "struct_cfgs" to be applied at the root level.
```
gmuloc: are they really 'legacy'?

python-avd/pyavd/_eos_designs/structured_config/__init__.py
```diff
@@ -86,36 +84,28 @@ def get_structured_config(
...
+    # "nested" is one big structured config created from "structured_config" under various nested models.
+    # We need these variants because the order of application is important (root first, then nested).
+    #
+    # TODO: Figure out if we should support all the other strategies.
+    custom_structured_configs = StructCfgs(list_merge_strategy="replace" if inputs.custom_structured_configuration_list_merge == "replace" else "append")
```
gmuloc: changing default from `append_rp` to `append`?

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -19,15 +19,15 @@
...
+class AvdStructuredConfigBase(StructuredConfigGenerator, NtpMixin, SnmpServerMixin, RouterGeneralMixin):
     """
     The AvdStructuredConfig Class is imported by "get_structured_config" to render parts of the structured config.
 
     "get_structured_config" imports, instantiates and run the .render() method on the class.
     .render() runs all class methods not starting with _ and of type @cached property and inserts the returned data into
     a dict with the name of the method as key. This means that each key in the final dict corresponds to a method.
 
-    The Class uses AvdFacts, as the base class, to inherit the _hostvars, keys and other attributes.
+    The Class uses StructuredConfigGenerator, as the base class, to inherit the _hostvars, keys and other attributes.
```
gmuloc, line 30: are we still inheriting the `_hostvars`?

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
```diff
@@ -205,9 +206,22 @@ def _get_common_interface_cfg(self: AvdStructuredConfigCoreInterfacesAndL3Edge,
             "mtu": p2p_link._get("mtu", self.shared_utils.p2p_uplinks_mtu) if self.shared_utils.platform_settings.feature_support.per_interface_mtu else None,
             "service_profile": p2p_link._get("qos_profile", self.inputs.p2p_uplinks_qos_profile),
             "eos_cli": p2p_link.raw_eos_cli,
-            "struct_cfg": p2p_link.structured_config or None,
         }
 
+        if p2p_link.structured_config:
+            if str(interface_name := p2p_link_data["interface"]).lower().startswith("po"):
```
gmuloc: Just for the sake of noting this, this _could_ (very unlikely) with an input like
```
interface: p42
```
which is interpreted by EOS as port-channel but the config would end up being applied under ethernet_interfaces

is there a key we could use in the port-channel that would discriminate "for sure"?

python-avd/pyavd/_eos_designs/structured_config/custom_structured_configuration/__init__.py
```diff
@@ -3,135 +3,47 @@
...
-                },
-            },
-        ]
+        This method merges each custom structured config on top of self.structured_config.
 
-    def _router_bgp_vrfs(self) -> list:
-        if self._router_bgp is None:
-            return []
+        First strip all (None, {}, []) from regular structured_config. This is to avoid empty objects showing up in the output dict.
+        Next we merge in custom structured config from various sources including None, {}, [].
```
gmuloc, line 27: Let's mention keeping these None, {}, [] is to match current behavior and maybe add a note to revisit for AVD 6.0.0?

python-avd/pyavd/_eos_designs/structured_config/custom_structured_configuration/__init__.py
```diff
@@ -3,135 +3,47 @@
...
+        """
+        # Strip empties from regular structured config
+        self.structured_config._strip_empties()
```
gmuloc, line 30: Maybe the fact that this is happening here at the end should be mentioned somewhere else?

python-avd/pyavd/_eos_designs/structured_config/custom_structured_configuration/__init__.py
```diff
@@ -3,135 +3,47 @@
...
+        if struct_cfg := self.shared_utils.node_config.structured_config:
+            self.structured_config._deepmerge(struct_cfg, list_merge=self.custom_structured_configs.list_merge_strategy)
 
-        return [
-            {
-                "router_bgp": {
-                    "vrfs": struct_cfgs,
-                },
-            },
+        # Apply structured configs from base and vrfs.
```
gmuloc: I would change it to the following to make more sense and not fall out of sync if this changes (even if today these are the only two):

```suggestion
        # Apply structured configs from root.
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -157,7 +154,7 @@ def flow_tracking(self) -> dict | None:
             flow_tracking[tracker_type]["sample"] = global_settings.sample
 
         filtered_trackers = []
-        for tracker_name in configured_trackers:
+        for tracker_name in natural_sort(configured_trackers):
```
gmuloc, line 157: I think the comment line 152/153 can be removed:

```
            # asdict does not contain default values so we need to insert the default sample.
            # TODO: consider if asdict should include defaults.
```

python-avd/pyavd/_eos_designs/structured_config/metadata/cv_pathfinder.py
```diff
@@ -135,67 +135,73 @@ def _metadata_pathfinder_vtep_ips(self: AvdStructuredConfigMetadata) -> list:
...
+            avt_policy = avt_policies[vrf.policy]
+            metadata_vrf = {
+                "name": vrf.name,
+                "vni": self._get_vni_for_vrf_name(vrf.name),
+                "avts": [],
+            }
+            for profile in vrf.profiles:
+                if not profile.name:
+                    continue
+                lb_policy = load_balance_policies[self.shared_utils.generate_lb_policy_name(profile.name)]
```
gmuloc, line 174: so we ued to have `required: True` for this - right onw it feels like we could just end up with `Undefined` and then I think we should get AttributeErrors - is this being tested?

Same for `avt_policy = avt_policies[vrf.policy]` line 171

I suppose we handle this before in the code in the other modules so maybe we can never hit this but just poiniting out that it seems we are handling this less well in the refactor

python-avd/pyavd/_eos_designs/structured_config/network_services/struct_cfgs.py
```diff
@@ -22,26 +20,12 @@ class StructCfgsMixin(UtilsMixin):
...
-                        list_of_dicts=vrf_struct_cfgs,
-                        primary_key="vrf",
-                        new_dict=vrf_struct_cfg,
-                        context="Structured Config for VRF '{vrf['name']}'",
-                        context_keys=["vrf"],
-                    )
-
-        if vrf_struct_cfgs:
-            return [vrf_struct_cfg["struct_cfg"] for vrf_struct_cfg in vrf_struct_cfgs]
+            self.custom_structured_configs.root.extend(vrf.structured_config for vrf in tenant.vrfs if vrf.structured_config)
```
gmuloc, line 29: is this still going to check for duplicates across tenants?

gmuloc (commented):
Reviewed the python changes.
I ran the comparison and happy with all the changes except some for sflow but I think this requires rebasing as it is one of the latest PR we merged

python-avd/pyavd/_eos_designs/structured_config/__init__.py
gmuloc: `append` is the same as legacy `append_rp` in this case so thats fine

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: yes we are - e.g. still reading things from _hostvars, for instance facts

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
gmuloc:
```suggestion
            if str(interface_name := p2p_link_data["interface"]).lower().startswith("p"):
```

python-avd/pyavd/_eos_designs/structured_config/custom_structured_configuration/__init__.py
gmuloc: no - if the user puts it in custom_structured_config we keep this in

python-avd/pyavd/_eos_designs/structured_config/metadata/cv_pathfinder.py
gmuloc: it will raise a KeyError but this should never happen. If this happens it will be a bug to fix.

python-avd/pyavd/_eos_designs/structured_config/network_services/struct_cfgs.py
gmuloc: answer is no it is relaxing the previous reqs for the structured_config to be exactly the same across all tenants for the same VRF which was a restriction that was not really useful given that this struct_config is applied from the root and so any 2 VRFs can already overwrite each other in the current code.

python-avd/pyavd/_eos_designs/structured_config/__init__.py
```diff
@@ -41,7 +42,7 @@
     # Metadata must be after anything else that can generate structured config, since CV tags can consume from structured config.
     AvdStructuredConfigMetadata,
     # The Custom Structured Configuration module must be rendered last,
-    # since it parses all supported object looking for `struct_cfg`.
+    # since it applies the structured config snips gathered by the other generators.
```
gmuloc:
```suggestion
    # since it strips empties from all previously generated structured config and then
    # applies the custom structured config snips gathered by the other generators.
```

python-avd/pyavd/_eos_designs/structured_config/__init__.py
ClausHolbechArista: added some fancy stuff :)

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
```diff
@@ -24,7 +24,21 @@ class StructCfgs:
 
     root: list[EosCliConfigGen] = field(default_factory=list)
     nested: EosCliConfigGen = field(default_factory=EosCliConfigGen)
-    list_merge_strategy: Literal["append", "replace"] = "append"
+    list_merge_strategy: Literal["append_unique", "append", "replace", "keep", "prepend", "prepend_unique"] = "append_unique"
```
gmuloc, line 27: wondering if it is not worth make an enum type out of this at some point

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -161,32 +161,53 @@ def _filtered(self, function: Callable[[T_ItemType], bool]) -> Self:
...
+            (For AvdIndexedList this works the same as prepend_unique)
         """
         cls = type(self)
         if not isinstance(other, cls):
             msg = f"Unable to merge type '{type(other)}' into '{cls}'"
             raise TypeError(msg)
 
         if self._created_from_null or other._created_from_null:
-            # Overwrite all data from other and clear the flag.
+            # Clear the flag and set list_merge to replace so we overwrite with data from other below.
```
gmuloc:
```suggestion
            # Set the flag to the value of other and set list_merge to replace so we overwrite with data from other below.
```

gmuloc (commented):
two last comments - one we can decide to postpone otherwise LGTM and ready to merge.

I merged devel and sflow diff is gone from structured_conifg

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
ClausHolbechArista: Absolutely. But not Today :)

python-avd/pyavd/_eos_designs/structured_config/custom_structured_configuration/__init__.py
ClausHolbechArista: I think we added it together Yesterday

gmuloc: yeah yeah I forgot to resolve here as I added it in the other place! all good!

gmuloc (approved):
LGTM 🚢 :shipit: 🪇

carl-baillargeon (approved):
LGTM. Tested locally.

PR #4715, opened 2024-11-12
Feat(plugins): Set changed=true and print yellow updates when recompiling schemas/templates

### LINKED ISSUE #4709: validate_requirements: Set changed=true and print yellow updates when recompiling schemas/templates
### Enhancement summary

When validate_requirements detects we are running from source and have some changes in the templates and/or schemas it will recompile as needed. This should trigger yellow printouts like "Recompiling templates and/or schemas..." to make it visible to the user what is taking so long (it takes 10-15 secs).
I think it would also make sense if it sets the changed=True on the results dict.

others

### Use case example

N/A

N/A

_No response_

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

1/ set changed=true
2/ if any templates or schemas was recompiled print a yellow log.

## Related Issue(s)

Fixes #4709

## Component(s) name

`plugins`

## Proposed changes

cf summary

## How to test

remove the `.hash` files in `python-avd/pyavd/*/j2templates` and `python-avd/pyavd/*/schema/schema_fragments/` and run any molecule

No verbosity:

Screenshot 2024-11-12 at 15 28 13

Verbosity 1:

Screenshot 2024-11-12 at 15 27 43

### Repository Checklist


python-avd/schema_tools/check_schemas.py
```diff
@@ -15,14 +15,19 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def check_schemas() -> None:
+def check_schemas() -> bool:
```
gmuloc, line 18: let's split the functions as discussed on the call

python-avd/schema_tools/compile_templates.py
```diff
@@ -33,39 +33,79 @@ def compile_eos_designs_templates() -> None:
...
+        True if any templates changed, False otherwise
+    """
     if not RUNNING_FROM_SRC:
-        return
+        return False
 
     LOGGER.info("pyavd running from source detected, checking eos_cli_config_gen templates for any changes...")
     dir_path = EOS_CLI_CONFIG_GEN_JINJA2_TEMPLATE_PATH
 
     changed, new_hash = changed_hash(dir_path)
```
alexeygorbunov: As variable not used anymore
```suggestion
    changed, _ = changed_hash(dir_path)
```

alexeygorbunov (approved):
LGTM

python-avd/schema_tools/compile_templates.py
ClausHolbechArista: We are generally keeping the name for readability so I suggest `_new_hash` to not trigger linting issues.

python-avd/schema_tools/compile_templates.py
```diff
@@ -33,39 +33,79 @@ def compile_eos_designs_templates() -> None:
...
     LOGGER.info("pyavd running from source detected, checking eos_designs templates for any changes...")
     dir_path = EOS_DESIGNS_JINJA2_TEMPLATE_PATH
 
-    changed, new_hash = changed_hash(dir_path)
-    if changed:
-        LOGGER.info("Recompiling eos_designs templates...")
-        compile_eos_designs_templates()
-        with (dir_path / ".hash").open("w") as fd:
-            fd.write(new_hash)
+    changed, _ = changed_hash(dir_path)
```
ClausHolbechArista: Since we never use it, we can just remove the hash from the changed_hash()

ClausHolbechArista (changes requested):

ClausHolbechArista (approved):

PR #4768, opened 2024-11-28
Feat(plugins):  Add support for arista.avd.secure_hash to hash local_user passwords

### LINKED ISSUE #3883: Feat(plugins): Add local_users to a new `hash` filter
**NOTE: this is a hash function so decrypt is not needed** 

Hello AVD Developers,

Do you have any plans to extend filter with user type? This can be useful when setting "local_users" input variables, like:

```
local_users:
  - name: <str>
    sha512_password: "{{ text_password | arista.avd.encrypt(passwd_type='loca_user', hash_algorithm='sha512', key='') }}"
```

BR

_Originally posted by @PHipnarowicz in https://github.com/aristanetworks/avd/discussions/3874_

davidhayes9:
## Change Summary

Add support for a `arista.avd.secure_hash` filter to allow a local_user to hash their password. 

## TODO

- [x] Pending on #5379 

## Related Issue(s)

Fixes #3883

## Component(s) name

## Proposed changes

`secure_hash` does not accept any parameters other than the users password and returns a SHA512-Crypt password hash

```
local_users:
  - name: <str>
    sha512_password: "{{ 'clear_password' | secure_hash }}"
```

There is a dependancy on `passlib` and the filter been tested with `passlib==1.7.4` (latest version)

The filter  is named `secure_hash` as `hash` is the name of a Python built-in function but happy to hear other naming suggestions.

## How to test
Added test coverage in pytest and molecule 

### User Checklist

- N/A

### Repository Checklist


python-avd/pyproject.toml
```diff
@@ -57,6 +59,7 @@ ansible-collection = [
     "treelib>=1.5.5",
     # jsonschema is required for the arista.cvp Ansible collection
     "jsonschema>=3.2.0",
+    "passlib==1.7.4",
```
gmuloc: not needed if it is in pyavd already
```suggestion
```

python-avd/pyproject.toml
```diff
@@ -39,6 +40,7 @@ pytest = [
     "PyYAML>=6.0.0",
     "pydantic>=2.3.0",
     "referencing>=0.35.0",
+    "passlib==1.7.4",
```
gmuloc: Not needed 
```suggestion
```

python-avd/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "deepmerge>=1.1.0",
     "Jinja2>=3.0",
     "requests>=2.27.0",
+    "passlib==1.7.4",
```
gmuloc: please don't pin dependencies
```suggestion
    "passlib>=1.7.4",
```

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host1/local-users.yml
```diff
@@ -24,3 +24,13 @@ local_users:
...
 # yamllint disable-line rule:line-length
     secondary_ssh_key: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDHMTFuLHPz/prREZZIks0ca4btBIzEbvY6KRYGzhN7JCG5CTfre0Y9UCbNul7qNl7cxomQkh/0VjQNX6ecPd0HyOTKL2EK002ejNyvooUDarnglMWtjKIl40NgDR/GNSkvC3nEylvX1H7Rfmu38NCqiwIpWA8JFwgLCLvkWUoORxHhIIy8/vttLgMxr66HGlVAnRidf3VVCnlILm4gUpc3fR43EhvVoYByY3jEa/fypiS2nDP9K2fXtpXGrIHSbyMu4Mj3fnSdcqWysRF7Tqc6Kvet8ImS07fLcgpbdLp31ssF1rssbTnD1zWuAozvXpK1d+vFO4EfFr5yzkE2Q8lM0wPpdS4LBWQfJdWgi6t5XEXewWyTYfIDKCBOI2dECGtkDjme+PDNIL9IQiiYC2iXMmQrun9fsp8jicdw1svGef8Otdb4kmHXiQ3mAxTeHLgeYPfYyekKq/+dFMcAZT+sv0g24AHc4ulitfLRoGjxYHZLGg2KQpFfAn0aQKCd5vk= noname@hostmachine-asd-cl
+  - name: sha-user
+    privilege: 15
+    role: network-admin
+    sha512_password: "{{ 'randomstring' | secure_hash }}"
```
gmuloc: please move this to the host for inline Jinja `  host4_inline_jinja:`

ansible_collections/arista/avd/plugins/filter/secure_hash.py
```diff
@@ -0,0 +1,55 @@
...
+from ansible_collections.arista.avd.plugins.plugin_utils.pyavd_wrappers import RaiseOnUse, wrap_filter
+
+PLUGIN_NAME = "arista.avd.secure_hash"
+
+try:
+    from pyavd.j2filters import secure_hash
+except ImportError as e:
+    secure_hash = RaiseOnUse(
+        AnsibleFilterError(
+            f"The '{PLUGIN_NAME}' plugin requires the 'passlib' and 'pyavd' Python library. Got import error",
```
gmuloc:
```suggestion
            f"The '{PLUGIN_NAME}' plugin requires the 'pyavd' Python library. Got import error",
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,49 @@
...
+try:
+    from passlib.hash import sha512_crypt
+except ImportError as imp_exc:
+    raise AristaAvdError(imp_exc) from imp_exc
+
+
+def _get_password_hash(user_password: str) -> str:
+    """
+
+    Generate a SHA512-Crypt password hash with a random salt value for a local user.
```
gmuloc:
```suggestion
    """
    Generate a SHA512-Crypt password hash with a random salt value for a local user.
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,49 @@
...
+
+    Parameters:
+    ----------
+    user_password : str
+                    the clear text password provided by the user that will be hashed
+
```
gmuloc:
```suggestion
    user_password : str
        the clear text password provided by the user that will be hashed

```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,49 @@
...
+    Returns:
+    -------
+    str
+        the SHA512-Crypt password hash
```
gmuloc:
```suggestion
    str
        the SHA512-Crypt password hash
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,49 @@
...
+        try:
+            # setting the rounds parameter to 5000 to omit rounds from the hash string, similar to EOS implementation
+            return sha512_crypt.using(rounds=5000).hash(user_password)
+        except Exception as exc:
+            msg = "SHA512 password hashing failed - check the input parameters of arista.avd.secure_hash"
+            raise ValueError(msg) from exc
+
+    else:
+        msg = f"Password MUST be of type 'str' but is of type {type(user_password)}"
+        raise TypeError(msg)
```
gmuloc: We prefer to use this pattern rather than if / else (suggestion hand typed in Github so may need a round of linting)
```suggestion
    if not isinstance(user_password, str):
        msg = f"Password MUST be of type 'str' but is of type {type(user_password)}"
        raise TypeError(msg)

    try:
        # setting the rounds parameter to 5000 to omit rounds from the hash string, similar to EOS implementation
        return sha512_crypt.using(rounds=5000).hash(user_password)
    except Exception as exc:
        msg = "SHA512 password hashing failed - check the input parameters of arista.avd.secure_hash"
        raise ValueError(msg) from exc
```

python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -0,0 +1,39 @@
...
+
+from pyavd.j2filters import secure_hash
+
+sha512_regex = r"^\$6\$[A-Za-z0-9\.\\\/]{16}\$[A-Za-z0-9\.\\\/]{86}"
+
+
+INVALID_PASSWORDS = [
+    pytest.param(True, TypeError, "Password MUST be of type 'str' but is of type <class 'bool'>"),
+    pytest.param([], TypeError, "Password MUST be of type 'str' but is of type <class 'list'>"),
+    pytest.param({}, TypeError, "Password MUST be of type 'str' but is of type <class 'dict'>"),
```
gmuloc: I would say you can test only one to validate the functionality of isinstance but leave them there :)

gmuloc (changes requested):
Thanks for this PR and the implementation up to the tests! You have done all that was expected and thanks for this.

I have left a couple of comments of things to adjust.

The name in the referenced issue suggested by @ClausHolbechArista was `arista.avd.hash` not sure if we want to keep this one or use `secure_hash`

I have a slight concerned regarding not setting the salt as it means on every run of AVD the password will change (which is not going to make molecule happy among others) as it will be re-generated. It means what is produced won't be idempotent anymore (but yes it will work)

I am wondering if we should not add a `salt` option in the filter and tell people to generate one. And use it in the molecule scenario. Otherwise we need a different way to persist these (maybe one run and then people can take the output and remove the usage of the secure_hash)

Our stance until now also was that it was already supported by just using the builtin `password_hash` filter: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/password_hash_filter.html doing exactly this (offering the option to use salt)

davidhayes9:
> Thanks for this PR and the implementation up to the tests! You have done all that was expected and thanks for this.
> 
> I have left a couple of comments of things to adjust.

Thanks for reviewing and leaving the suggestions, i've pushed new changes and these should be resolve the suggestions.

> The name in the referenced issue suggested by @ClausHolbechArista was `arista.avd.hash` not sure if we want to keep this one or use `secure_hash`

I did originally start by using `arista.avd.hash` but it looks like the code format is to use the filter name as the main caller function. When I used hash for this I was getting an error (I think from one of the linters) that a built-in function name was being used. I can try moving it back to hash though and confirming. 

> I have a slight concerned regarding not setting the salt as it means on every run of AVD the password will change (which is not going to make molecule happy among others) as it will be re-generated. It means what is produced won't be idempotent anymore (but yes it will work)
> 
> I am wondering if we should not add a `salt` option in the filter and tell people to generate one. And use it in the molecule scenario. Otherwise we need a different way to persist these (maybe one run and then people can take the output and remove the usage of the secure_hash)
> 
> Our stance until now also was that it was already supported by just using the builtin `password_hash` filter: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/password_hash_filter.html doing exactly this (offering the option to use salt)

Yes good point about the password being re-generated each time. I can add an optional salt parameter similar to the password_hash

davidhayes9:
the linter errors when I use hash


```
ruff....................................................................................................................Failed
- hook id: ruff
- exit code: 1

ansible_collections/arista/avd/plugins/filter/hash.py:18:5: A001 Variable `hash` is shadowing a Python builtin
   |
16 |     from pyavd.j2filters import hash
17 | except ImportError as e:
18 |     hash = RaiseOnUse(
   |     ^^^^ A001
19 |         AnsibleFilterError(
20 |             f"The '{PLUGIN_NAME}' plugin requires the 'pyavd' Python library. Got import error",
   |

python-avd/pyavd/j2filters/hash.py:46:5: A001 Variable `hash` is shadowing a Python builtin
   |
46 | def hash(user_password: str) -> str:
   |     ^^^^ A001
47 |     return _get_password_hash(user_password)
   |

Found 2 errors.
```

Let me know what the preference is for the filter name.

python-avd/pyavd/j2filters/secure_hash.py
ClausHolbechArista: Please change this docstring to google docstring format.
Args: and Returns: (no lines and no types since we already have types in the signature)

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,47 @@
...
+    try:
+        # setting the rounds parameter to 5000 to omit rounds from the hash string, similar to EOS implementation
+        return sha512_crypt.using(rounds=5000).hash(user_password)
+    except Exception as exc:
+        msg = "SHA512 password hashing failed - check the input parameters of arista.avd.secure_hash"
+        raise ValueError(msg) from exc
+
+
+def secure_hash(user_password: str) -> str:
+    return _get_password_hash(user_password)
```
ClausHolbechArista: I don't think we need this function as long as we don't provide any options. You can just rename the main function.

Since we asume it is a user password we can either decide to rename this to user_password_hash or maybe better keep this function with an extra argument to specify the type of password, and then call a _user_password_hash function. (only supported option is the user password for now, but maybe others will come later).

@carlbuchmann @gmuloc comments?

python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -0,0 +1,39 @@
...
+class TestSecureHashFilter:
+    @pytest.mark.parametrize(("user_password", "expected_raise", "expected_raise_message"), INVALID_PASSWORDS)
+    def test_secure_hash_invalid(self, user_password: str, expected_raise: Exception, expected_raise_message: str) -> None:
+        with pytest.raises(expected_raise, match=expected_raise_message):
+            secure_hash(user_password)
+
+    @pytest.mark.parametrize(("user_password", "regex_of_hashed_password"), VALID_PASSWORDS)
+    def test_secure_hash_valid(self, user_password: str, regex_of_hashed_password: str) -> None:
+        hashed_password = secure_hash(user_password)
+        assert re.fullmatch(regex_of_hashed_password, hashed_password)
```
ClausHolbechArista: Are these password hashes not identical between runs? I would rather that we put the hash values in the expected test data instead of the same regex on all of them.

If they are not consistent we need to reconsider this completely, since it could lead to constant config changes for every AVD run and never have "compliant" configs.

davidhayes9: Ack, the way the filter is now there will be config changes when AVD is re-run so I will change this. I'm working on adding a salt argument to the filter after an above discussion with @gmuloc, this will allow a user to configure the salt value used when the hash is created and this will ensure the hashed password does not change when AVD is run again. 

Regarding this test, it is just to confirm the hashed password is of a valid format i.e. `$6$<salt-value>$<hashed-password>`. Once the salt arg is done, I'll create some tests using hardcoded salt passwords against the expected hashed password.

davidhayes9: added the support for the optional salt arg:

```
local_users:
  # Create sha512 password hash with random salt. Note: this will create a new hash each time it is run
  - name: admin
    sha512_password: "{{ 'password123' | arista.avd.secure_hash }}"

  # Create sha512 password hash with user defined salt value
  - name: cvpadmin
    sha512_password: "{{ 'securepassword' | arista.avd.secure_hash(salt='Yar49ahkzKddRVYS') }}"
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,95 @@
...
+        salt: Salt value to be used when creating password hash. A randomly generated salt will be used unless the user specifies one.
+
+    Returns:
+    -------
+        The SHA-512 password hash.
+
+    Raises:
+    ------
+        TypeError: If the password is not of type 'str'.
+        ValueError: If sha512_crypt fails for any reason.
```
gmuloc:
```suggestion
    Args:
        clear_password: The cleartext password provided by the user that will be hashed.
        salt: Salt value to be used when creating password hash. A randomly generated salt will be used unless the user specifies one.

    Returns:
        The SHA-512 password hash.

    Raises:
        TypeError: If the password is not of type 'str'.
        ValueError: If sha512_crypt fails for any reason.
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,95 @@
...
+        user_input: the user input cleartext that will be hashed.
+        output_type: the use case for the cleartext provided by the user.
+
+    Returns:
+    -------
+        The hash digest.
+
+    Raises:
+    ------
+        ValueError: if the output_type value provided by the user is not supported.
```
gmuloc:
```suggestion
    Args:
        user_input: the user input cleartext that will be hashed.
        output_type: the use case for the cleartext provided by the user.

    Returns:
        The hash digest.

    Raises:
        ValueError: if the output_type value provided by the user is not supported.
```

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host_inline_jinja/local-users.yml
```diff
@@ -0,0 +1,13 @@
...
+  - name: sha-user
+    privilege: 15
+    role: network-admin
+    sha512_password: "{{ 'randomstring' | arista.avd.secure_hash(salt='hcklJLJwToGeAS5u') }}"
+  - name: sha-user-the-second
+    privilege: 15
+    role: network-admin
+    sha512_password: "{{ clear_password | arista.avd.secure_hash(salt='Yar49ahkzKddRVYS', output_type='sha512_password') }}"
+
+clear_password: "arista"
```
gmuloc:
```suggestion

# Note: this clear_password should be in a vault for production use
clear_password: "arista"
```

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host6.md
```diff
@@ -4,8 +4,15 @@
 
 - [Management](#management)
   - [Management Interfaces](#management-interfaces)
+<<<<<<< HEAD:ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host6.md
```
gmuloc: hmmm this needs a fix

ansible_collections/arista/avd/plugins/filter/secure_hash.py
```diff
@@ -0,0 +1,92 @@
...
+        ),
+    )
+
+
+DOCUMENTATION = r"""
+---
+name: secure_hash
+collection: arista.avd
+author: Arista Ansible Team (@aristanetworks)
+version_added: "5.5.0"
```
gmuloc:
```suggestion
version_added: "5.2.0"
```

ansible_collections/arista/avd/plugins/filter/secure_hash.py
```diff
@@ -0,0 +1,92 @@
...
+positional: _input
+options:
+  _input:
+    description: The cleartext input to be hashed.
+    type: string
+    required: true
+  salt:
+    description: |-
+      The salt value to be used when creating a password hash.
+      The salt value can only contain the characters `./`, digits `0-9`, and uppercase or lowercase letters `A-Z a-z`.
```
gmuloc, line 48: is this base64 encoding?

ansible_collections/arista/avd/plugins/filter/secure_hash.py
```diff
@@ -0,0 +1,92 @@
...
+  - name: admin
+    sha512_password: "{{ 'password123' | arista.avd.secure_hash }}"
+"""
+
+
+RETURN = r"""
+---
+_value:
+  description:
+    - The hash digest.
```
gmuloc:
```suggestion
    - The hash digest
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Validate a given salt value.
+
+    Args:
+        salt: The salt value provided by the user.
+
+    Raises:
+        TypeError: If the salt is not of type 'str'.
+        ValueError: If the salt is greater than 16 characters.
+        ValueError: If the salt contains a character not matching ./0-9A-Za-z
```
gmuloc: Lets check if this is base64
```suggestion
        ValueError: If the salt contains a character not matching ./0-9A-Za-z
```

ansible_collections/arista/avd/plugins/filter/secure_hash.py
davidhayes9: its not base64 as plus(+) is not allowed, while dot(.) is allowed

python-avd/pyavd/j2filters/secure_hash.py
gmuloc: it is not - the specification for these valid characters seems to come from mkpasswd in UNIX (which would be what EOS used underneath) - so thats life.


https://stackoverflow.com/questions/69495640/which-characters-are-invalid-in-salt-for-sha512

davidhayes9: ya its slightly different than base64, mainly dot(.) is allowed instead of a plus(+). So a valid salt that contains a dot would fail a base64 check.

I believe mkpasswd uses the C crypt function from my research:
https://man7.org/linux/man-pages/man3/crypt.3.html

Python did implement this as part of its standard library but has since deprecated it since 3.11 - https://docs.python.org/3/library/crypt.html

[passlib](https://pypi.org/project/passlib/) was the best alternative for this that I could find (which I have used)

gmuloc: yes for passlib I was just looking for a "nicer" way to refer to this set of allowed characters but it seems there is not a common name

passlib calls it HASH64: https://github.com/glic3rinu/passlib/blob/master/passlib/utils/__init__.py#L1276 but thats not something commonly used it seems so lets keep it this way.

python-avd/pyavd/j2filters/secure_hash.py
carlbuchmann: Sorry, I completely missed this notification. I would keep this function with an extra argument to specify the type of password, and then call a `_user_password_hash` function. I agree that other options will likely come later.

davidhayes9: The extra arg has been done. There is an output_type arg that can take `sha512_password` for now. More values can be added in the future when there are more use cases. It will also default to`sha512_password` if the arg is left out. Below are some examples: 

```
local_users:
  # Create a sha512 password hash with a user defined salt value (recommended). The output_type will default to sha512_password.
  - name: cvpadmin
    sha512_password: "{{ 'securepassword' | arista.avd.secure_hash(salt='Yar49ahkzKddRVYS')}}"

    # Create a sha512 password hash with a user defined salt value and specifying the output_type as a sha512_password.
  - name: cvpuser
    sha512_password: "{{ 'newpassword' | arista.avd.secure_hash(salt='Kte5paJ3czRQczbk', output_type='sha512_password')}}"

  # Create a sha512 password hash with a random salt. Note: this will create a new hash each time AVD is run.
  - name: admin
    sha512_password: "{{ 'password123' | arista.avd.secure_hash }}"

```

python-avd/pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
     "deepmerge>=1.1.0",
     "Jinja2>=3.0",
     "requests>=2.27.0",
+    "passlib>=1.7.4",
```
alexeygorbunov: @gmuloc , are you able to install this `pyavd` from source without adding `passlib>=1.7.4` dependency under `[build-system]`?
I'm facing an issue as `passlib` is required for filters used in J2 templates that need to be compiled during the build

gmuloc:
on-hold until we bring some alternative to passlib

dgonzalez-arista:
as discussed with @gmuloc, it would be great to be able also to encrypt enable passwords, which are hashed using either md5 or sha512 (as part of this change or a separate one).

python-avd/pyproject.toml
```diff
@@ -55,6 +55,7 @@ requires = [
     "deepmerge>=1.1.0",
     "jinja2>=3.0",
     "jsonschema-rs>=0.24",
+    "pyavd-utils==6.0.0-dev3",
```
gmuloc:
```suggestion
    "pyavd-utils==6.0.0.dev3",
```

.github/workflows/pull-request-management.yml
```diff
@@ -116,7 +116,7 @@ jobs:
         working-directory: ./python-avd
         run: |
           pip install build
-          python -m build --wheel
+          python -m build --wheel -x
```
gmuloc:
```suggestion
          python -m build --wheel
```

ansible_collections/arista/avd/requirements-dev.txt
```diff
@@ -4,6 +4,7 @@
 anta==1.6.0
 cryptography>=43.0.0
 netaddr>=0.7.19
+passlib>=1.7.4
```
gmuloc:
```suggestion
```

tox.ini
```diff
@@ -33,6 +33,7 @@ commands =
     pytest {posargs}
 
 [testenv:coverage]
+skip_install=true
```
gmuloc:
```suggestion
```

tox.ini
```diff
@@ -46,6 +47,7 @@ commands =
     coverage combine
 
 [testenv:report]
+skip_install=true
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/docs/plugins/Filter_plugins/secure_hash.md
```diff
@@ -0,0 +1,62 @@
...
+# secure_hash
+
+!!! note
+    Always use the FQCN (Fully Qualified Collection Name) `arista.avd.secure_hash` when using this plugin.
+
+Generate an EOS supported hash digest.
+
+## Synopsis
+
+This filter generates an EOS supported hash digest. The algorithm for the hash is defined with the `output_type` key. `sha512_password` can be
```
gmuloc:
```suggestion
This filter generates an EOS supported hash digest. The algorithm for the hash is defined with the `hash_type` key. `sha512_password` can be
```

gmuloc (commented):
add setuptools in pyright so we don't get the warning anymore

.pre-commit-config.yaml
```diff
@@ -120,15 +120,16 @@ repos:
         name: PyRight static type checker
         types: [python]
         additional_dependencies:
-          - 'anta==1.6.0'
+          - 'PyYAML>=6.0.0'
```
gmuloc: lets use lower case for pyyaml (eveywhere)

gmuloc: (and add setuptool)

ansible_collections/arista/avd/docs/plugins/Filter_plugins/secure_hash.md
```diff
@@ -0,0 +1,62 @@
...
+used with the `output_type` key to create a SHA-512 password hash from a cleartext password for a local user. By default, the secure_hash filter will use
+sha512_password.
```
ClausHolbechArista:
```suggestion
`hash_type="sha512_password"`.
```

gmuloc (approved):
LGTM if CI passes.

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,88 @@
+# Copyright (c) 2023-2025 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+
+import re
+
+HASH_INPUT_TYPE = ["sha512_password"]
+
+
+def _validate_sha512_salt(salt: str) -> str:
```
ClausHolbechArista: Do we really need to do this here? We have validation inside the library.

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,88 @@
...
+        raise ValueError(msg)
+
+    if not re.fullmatch(r"[\.\/0-9A-Za-z]{1,16}", salt):
+        msg = "Salt value MUST only contain the characters ./0-9A-Za-z"
+        raise ValueError(msg)
+
+    return salt
+
+
+def _user_password_hash(clear_password: str, salt: str) -> str:
```
ClausHolbechArista: This should be called sha512_password

python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -0,0 +1,110 @@
...
+from __future__ import annotations
+
+import re
+from unittest.mock import patch
+
+import pytest
+
+from pyavd.j2filters import secure_hash
+
+sha512_regex = r"^\$6\$[A-Za-z0-9\.\/]{1,16}\$[A-Za-z0-9\.\/]{86}"
```
ClausHolbechArista: We are already testing all these things twice - in rust and in pyavd-utils. Let's do one good and one bad test and that's it.

.pre-commit-config.yaml
```diff
@@ -119,16 +119,21 @@ repos:
...
-          - 'pydantic>=2.3.0'
           - 'jsonschema-rs>=0.24'
+          - 'pyavd-utils==0.0.1.dev1'
+          - 'pydantic>=2.3.0'
+          - 'pytest>=8.2.0'
+          - 'pyyaml>=6.0.0'
           - 'referencing>=0.35.0'
           - 'ruff==0.11.11'
-          - 'pytest>=8.2.0'
+          - 'setuptools>=77.0'
```
ClausHolbechArista: it is complaining in the workflow triggered pyright. Not in pre-commit.

.github/workflows/pull-request-management.yml
```diff
@@ -136,7 +136,7 @@ jobs:
           python-version: "3.12"
       - name: Install pyavd and Ansible requirements
         run: |
-          pip install --group dev --upgrade
+          pip install --group dev setuptools>=77.0 --upgrade
```
ClausHolbechArista: could we not add it to the dev tools?

gmuloc: we could :p

python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -0,0 +1,66 @@
...
+        "$6$DxaFhAPPrrOzqgZV$pdiZUeB6SRwVsiTzW1jPQvBy3eP5DqJWjZ1Fd3mpO8E9tjJ/ntaiZx7CaIIkfYyOnzgV92AW7fFSWnQzzowzP.",
+    )
+]
+
+INVALID_HASH_HASH_TYPES = [
+    pytest.param(
+        "pass",
+        "",
+        ValueError,
+        "The hash_type key does not support the value ''. The value used with hash_type must be in \\['sha512_password'\\]",
```
ClausHolbechArista:
```suggestion
        "The hash_type key does not support the value ''. The value used with hash_type must be one of \\['sha512_password'\\]",
```

python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -0,0 +1,66 @@
...
+    ),
+    pytest.param(
+        "pass",
+        "user_p",
+        ValueError,
+        "The hash_type key does not support the value 'user_p'. The value used with hash_type must be in \\['sha512_password'\\]",
```
ClausHolbechArista:
```suggestion
        "The hash_type key does not support the value 'user_p'. The value used with hash_type must be one of \\['sha512_password'\\]",
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,58 @@
...
+        TypeError: If the password is not of type 'str'.
+        ValueError: If sha512_crypt fails for any reason.
+    """
+    # Importing inside the function to avoid neeedind pyavd-utils in the build dependencies.
+    from pyavd_utils.passwords import sha512_crypt  # noqa: PLC0415
+
+    if not isinstance(clear_password, str):
+        msg = f"Password MUST be of type 'str' but is of type {type(clear_password)}"
+        raise TypeError(msg)
+
```
ClausHolbechArista: don't we also handle this in rust?

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,58 @@
...
+        hash_type: the use case for the cleartext provided by the user.
+
+    Returns:
+        The hash digest.
+
+    Raises:
+        ValueError: if the hash_type value provided by the user is not supported.
+    """
+    if hash_type not in HASH_INPUT_TYPE:
+        msg = f"The hash_type key does not support the value '{hash_type}'. The value used with hash_type must be in {HASH_INPUT_TYPE}"
```
ClausHolbechArista:
```suggestion
        msg = f"The hash_type key does not support the value '{hash_type}'. The value used with hash_type must be one of {HASH_INPUT_TYPE}"
```

python-avd/pyavd/j2filters/secure_hash.py
gmuloc: rust would definitely complain but not sure it would make sense to the user:

```
 TypeError: argument 'password': 'bool' object cannot be converted to 'PyString'
```

so we may need to make a wrapper in pyavd-utils for this?

ClausHolbechArista: ok we keep it here for now

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2023-2025 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+
+
+HASH_INPUT_TYPE = ["sha512_password"]
+
+
+def sha512_hash(clear_password: str, salt: str) -> str:
```
ClausHolbechArista:
```suggestion
def sha512_password(clear_password: str, salt: str) -> str:
```

python-avd/pyavd/j2filters/secure_hash.py
```diff
@@ -0,0 +1,57 @@
...
+
+    try:
+        # setting the rounds parameter to 5000 to omit rounds from the hash string, similar to EOS implementation
+        return sha512_crypt(clear_password, salt)
+    except TypeError as exc:
+        msg = f"SHA-512 password hashing failed - check the input parameters of arista.avd.secure_hash: {exc}"
+        raise TypeError(msg) from exc
+    except Exception as exc:
+        msg = "SHA-512 password hashing failed - check the input parameters of arista.avd.secure_hash"
+        raise ValueError(msg) from exc
```
ClausHolbechArista:
```suggestion
    except Exception as exc:
        msg = f"SHA-512 password hashing failed - check the input parameters of arista.avd.secure_hash: {exc}"
        raise type(exc)(msg) from exc
```

ClausHolbechArista (approved):

PR #4795, opened 2024-12-09
Refactor(plugins): Improve schema models

ClausHolbechArista:
- Add a few extra methods to schema classes to improve code readability on future refactored code.
  - provide ways of updating or adding new items inline, without having to give the full class path.
  - `._update` on `AvdModel` and `append_new` on `AvdList` and `AvdIndexedList` gives type hints similar to `__init__()` of the item class. This means it is easy to provide the correct supported fields.
  - The new methods return the instance to make it easier to chain with other methods.


python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -208,20 +208,13 @@ def _get(self, name: str, default: Any = None) -> Any:
...
         return value
 
-    def _update(self, other: Self) -> None:
-        """Update instance by shallow merging the other instance in."""
-        cls = type(self)
-        if not isinstance(other, cls):
-            msg = f"Unable to merge type '{type(other)}' into '{cls}'"
-            raise TypeError(msg)
+    if TYPE_CHECKING:
+        _update: type[Self]
```
gmuloc, line 235: should it not be a `Callable[[type[Self]], None]`?
I feel like we are just saying `_update` is of `type[Self]` and it seems "not correct"

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -208,20 +208,13 @@ def _get(self, name: str, default: Any = None) -> Any:
...
+    else:
 
-        for field in cls._fields:
-            if new_value := other._get_defined_attr(field) is Undefined:
-                continue
-            old_value = self._get_defined_attr(field)
-            if old_value == new_value:
-                continue
-            setattr(self, field, new_value)
+        def _update(self, *args: Any, **kwargs: Any) -> None:
```
ClausHolbechArista:
```suggestion
        def _update(self, *args: Any, **kwargs: Any) -> Self:
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -208,20 +208,13 @@ def _get(self, name: str, default: Any = None) -> Any:
...
+            """Update instance with the given kwargs. Reuses __init__."""
+            self.__init__(*args, **kwargs)
```
ClausHolbechArista, line 240:
```suggestion
            self.__init__(*args, **kwargs)
            return self
```

gmuloc (dismissed):

emilarista (approved):
Tested, LGTM!

emilarista (approved):
Retested with same repos, still LGTM!

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -151,32 +180,49 @@ def _deepmerge(self, other: Self, list_merge: Literal["append", "replace"] = "ap
...
-                self[primary_key] = deepcopy(new_item)
+                self[primary_key] = new_item
                 continue
 
             # Existing item of same type, so deepmerge.
             self[primary_key]._deepmerge(new_item, list_merge=list_merge)
 
+            # Make sure the primary key is never lost / changed.
+            if getattr(self[primary_key], self._primary_key, None) != primary_key:
+                setattr(self[primary_key], self._primary_key, primary_key)
```
gmuloc: Is this check necessary? It feels like a bug in the `_deepmerge` code that it could happen no?

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -98,20 +98,43 @@ def get(self, index: int, default: T | UndefinedType = Undefined) -> T_ItemType
...
+
     def extend(self, items: Iterable[T_ItemType]) -> None:
         self._items.extend(items)
 
-    def _as_list(self, include_default_values: bool = False, strip_values: tuple = (None, [], {})) -> list:
+    def _strip_empties(self) -> None:
+        """In-place update the instance to remove data matching the given strip_values."""
+        if issubclass(self._item_type, AvdBase):
+            [item._strip_empties() for item in self._items]
+            self._items = [item for item in self._items if item]
```
gmuloc: This makes for one less list creation:
(verified in Python 3.10 and 3.13)
```suggestion
            self._items = [striped_item for item in self._items if (striped_items := item._strip_empties())]
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -164,39 +169,54 @@ def __bool__(self) -> bool:
...
         Filtered for nested None, {} and [] values.
         """
         as_dict = {}
-        for field, field_info in self._fields.items() or ():
-            if (value := self._get_defined_attr(field)) is Undefined:
+        for field, field_info in self._fields.items():
+            value = self._get_defined_attr(field)
+
+            if field == "_custom_data":
+                if isinstance(value, dict) and value:
```
gmuloc: Save one isinstance in some cases.. :)
```suggestion
                if value and isinstance(value, dict):
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -351,6 +393,20 @@ def _cast_as(self, new_type: type[T_AvdModel], ignore_extra_keys: bool = False)
...
             new_args[field] = value
-            continue
 
-        return new_type(**new_args)
+        new_instance = new_type(**new_args)
+
+        if self._created_from_null:
+            new_instance._created_from_null = True
+        if self._block_inheritance:
+            new_instance._block_inheritance = True
```
gmuloc: seems that we don't really need the if checks given that if they are False we set them to False as well

```suggestion
    # Pass along the flags
    new_instance._created_from_null = self._created_from_null
    new_instance._block_inheritance = self._block_inheritance
```

gmuloc (commented):
In general I feel like it would make sense to add some unittest for merging and inheriting with the flags for some classes to make sure we don't break this in further refactoring

python-avd/pyavd/_schema/models/avd_indexed_list.py
ClausHolbechArista: It happens when merging a null-value on top of an indexed list item.

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: Are you sure it is working? Remember strip_empties is not returning anything. It is doing inplace updates.

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: but costs a bool check which is not always "free" for these classes...

python-avd/pyavd/_schema/models/avd_list.py
gmuloc: good point - my scratch pad testing was working but I had a return.
Though we should not create the list line 122 then - could use map instead to get an iterator of item on which `_strip_empties` has been applied

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: can _custom_data ever be a class?

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: How can I use map when it is a method of the item itself? I would need lambda. I doubt that will be more efficient.

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc: hmm so it means that in `other` the `primary_key` is not set for a null-value? but then how does this can even have a primary_key in the first place?

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: not really. It was this or cast() to make type-checking work.

python-avd/pyavd/_schema/models/avd_indexed_list.py
ClausHolbechArista: Thinking over this again, we should probably remove the item if we merge a null on top.

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: hmm ok so isinstance is only present for the sake of type-checking?

Let's resolve this

python-avd/pyavd/_schema/models/avd_list.py
gmuloc: I think there would be options but lets not waste more time on this. It gave for nice readings .. :)

ClausHolbechArista: As discussed elsewhere we will stick with list comprehension.

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: I will refactor all these to use cast so it is more clear.

ClausHolbechArista: done

python-avd/pyavd/_schema/models/avd_indexed_list.py
ClausHolbechArista: Changed the merge to pop the value if we merge a null value on top.

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -174,11 +202,16 @@ def _cast_as(self, new_type: type[T_AvdList], ignore_extra_keys: bool = False) -
...
 
         if self._item_type != new_type._item_type:
             msg = f"Unable to cast '{cls}' as type '{new_type}' since they have incompatible item types."
             raise TypeError(msg)
 
-        return new_type(self._items)
+        new_instance = new_type(self._items)
+
+        if self._created_from_null:
+            new_instance._created_from_null = True
```
gmuloc:
```suggestion
        # Pass along the _created_from_null flag
        new_instance._created_from_null = self._created_from_null
```

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -195,4 +242,11 @@ def _cast_as(self, new_type: type[T_AvdIndexedList], ignore_extra_keys: bool = F
             msg = f"Unable to cast '{cls}' as type '{new_type}' since '{new_type}' is not an AvdIndexedList subclass."
             raise TypeError(msg)
 
-        return new_type([item._cast_as(new_type._item_type, ignore_extra_keys=ignore_extra_keys) for item in self])
+        new_instance = new_type([item._cast_as(new_type._item_type, ignore_extra_keys=ignore_extra_keys) for item in self])
+
+        if self._created_from_null:
+            new_instance._created_from_null = True
+        if self._block_inheritance:
+            new_instance._block_inheritance = True
```
gmuloc:
```suggestion
        # Pass along the internal flags
        new_instance._created_from_null = self._created_from_null
        new_instance._block_inheritance = self._block_inheritance
```

gmuloc (approved):

PR #4810, opened 2024-12-12
Feat(eos_designs): update feature support - 7358

colinmacgiolla:
## Change Summary
Update the platform settings for the 7358 as we now support BGP update_wait_install

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Default parameter update

### User Checklist

- N/A

### Repository Checklist


gmuloc:
Hi @colinmacgiolla thanks for the PR! Your description mentions

> as we _now_ support BGP update_wait_install

What does it mean from the point of view of support? Is it starting after some EOS version? (if so which one?) Essentially could introducing this change mean that suddenly we would start generated configuration that would break on already deployed routers because "they don't yet support" BGP update_wait_install?

ClausHolbechArista:
I think this will have to be postponed until 6.0 so we can warn users in the porting guide.

PR #4838, opened 2024-12-20
Feat(pyavd): Add _source to AvdBase type

gmuloc:
## Change Summary

Add `._source` to `AvdBase` to get the origin of the different variables at load time of the inputs. It is represented as an `InputPath`
 
## Component(s) name

`pyavd`

## Proposed changes

cf summary

## How to test

pytest + changes in EOS designs

### User Checklist

- [ ] maintain a table per field

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/connected-endpoints-monitor-session-connected-endpoint-acl.yml
```diff
@@ -29,4 +29,4 @@ servers:
 
 expected_error_message: >-
   Cannot set an ACL for both `session_settings` and `source_settings` under the monitor session 'DMF'
-  for servers[name=INDIVIDUAL_1].adapters[0].
+  for servers[0 (name=INDIVIDUAL_1)].adapters[0].
```
ClausHolbechArista: For indexed lists let's keep the old style of `[<primary_key>=<value>]`

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-connected-endpoint-parent-port-profile-does-not-exist.yml
```diff
@@ -21,4 +21,4 @@ servers:
         profile: PROFILE-1
 
 expected_error_message: >-
-  Profile 'THIS-PROFILE-DOES-NOT-EXIST' applied under port profile 'PROFILE-1' does not exist in `port_profiles`.
+  Profile 'THIS-PROFILE-DOES-NOT-EXIST' applied under port profile 'port_profiles[0 (profile=PROFILE-1)]' does not exist in 'port_profiles'.
```
ClausHolbechArista: Somehow we should have a reference to `parent_profile`. Either in the path or just in the text.

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-network-port-ptp-profile-does-not-exist.yml
```diff
@@ -26,4 +26,4 @@ network_ports:
       profile: THIS-PROFILE-DOES-NOT-EXIST
 
 expected_error_message: >-
-  PTP Profile 'THIS-PROFILE-DOES-NOT-EXIST' referenced under network_ports[0] does not exist in `ptp_profiles`.
+  PTP Profile 'THIS-PROFILE-DOES-NOT-EXIST' referenced under 'network_ports[0]' does not exist in 'ptp_profiles'.
```
ClausHolbechArista: "referenced under" here, "applied under" in other tests, and "defined under" in some as well. If we add the full path (including the `.profile` key maybe it should be "referenced in".

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -45,19 +46,33 @@ class AvdIndexedList(Sequence[T_AvdModel], Generic[T_PrimaryKey, T_AvdModel], Av
...
-        cls_items = cast(Iterable[T_AvdModel], (coerce_type(item, cls._item_type) for item in data))
-        return cls(cls_items)
+        cls_items = []
+        # Handle _DynamicKeys slightly differently
+        for index, item in enumerate(data):
+            # Need to retrieve the key from the data not from the model.
+            item_key = cls._item_type._field_to_key_map.get(cls._primary_key, cls._primary_key)
+            child_data_source = (
+                InputPath(item[item_key])
+                if cls.__name__.startswith("Dynamic")
```
ClausHolbechArista, line 69: This feels hacky.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -40,6 +41,8 @@ class AvdModel(AvdBase):
     """Map of field name to original dict key. Used when fields have the field_ prefix to get the original key."""
     _key_to_field_map: ClassVar[dict[str, str]] = {}
     """Map of dict key to field name. Used when the key is names with a reserved keyword or mixed case. E.g. `Vxlan1` or `as`."""
+    _field_source: ClassVar[dict[str, InputPath]] = {}
```
ClausHolbechArista: This should not be a classvar but an instance var. Having it as a class var would maintain a single dict across all instances.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -79,19 +84,34 @@ def _from_dict(cls: type[T_AvdModel], data: Mapping, keep_extra_keys: bool = Fal
...
+
+            cls_args[field] = coerce_type(data[key], cls._fields[field]["type"], data_source=child_data_source)
+            cls._field_source[field] = child_data_source
 
         if custom_data:
             cls_args["_custom_data"] = custom_data
 
-        return cls(**cls_args)
+        cls_instance = cls(**cls_args)
+        cls_instance._source = data_source
```
ClausHolbechArista: The _source could be set in the same way as `_custom_data` so you can just add it to the cls_args.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -79,19 +84,34 @@ def _from_dict(cls: type[T_AvdModel], data: Mapping, keep_extra_keys: bool = Fal
...
+        return cls_instance
 
     @classmethod
     def _get_field_name(cls, key: str) -> str | None:
         """Returns the field name for the given key. Returns None if the key is not matching a valid field."""
         field_name = cls._key_to_field_map.get(key, key)
         return field_name if field_name in cls._fields else None
 
+    def get_field_source(self, key: str) -> str:
```
ClausHolbechArista: Here you are not prefixing with underscore like we do for all the other methods. This could clash with a model field (if we ever had one with get_ :)

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -79,19 +84,34 @@ def _from_dict(cls: type[T_AvdModel], data: Mapping, keep_extra_keys: bool = Fal
...
+        """Returns the field source for the given key."""
+        # TODO: discuss if we should include the item path in error messages so that if the field come from a profile we
```
ClausHolbechArista, line 102: yeah we probably need something like `via` or similar. Also when we later add code sources or files.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -422,6 +442,8 @@ def _cast_as(self, new_type: type[T_AvdModel], ignore_extra_keys: bool = False)
         # Pass along the internal flags
         new_instance._created_from_null = self._created_from_null
         new_instance._block_inheritance = self._block_inheritance
+        # Copy the source
+        new_instance._source = self._source
```
ClausHolbechArista: I would have expected something similar on merge, inherit etc.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -79,19 +84,34 @@ def _from_dict(cls: type[T_AvdModel], data: Mapping, keep_extra_keys: bool = Fal
                 msg = f"Invalid key '{key}'. Not available on '{cls.__name__}'."
                 raise KeyError(msg)
 
-            cls_args[field] = coerce_type(data[key], cls._fields[field]["type"])
+            # Handle Dynamic Keys slightly differently
+            child_data_source = data_source if cls.__name__.startswith("Dynamic") and key == "value" else data_source.create_descendant(key)
+
+            cls_args[field] = coerce_type(data[key], cls._fields[field]["type"], data_source=child_data_source)
+            cls._field_source[field] = child_data_source
```
ClausHolbechArista: I don't think this would work well. Setting it on the class. A list of classes would break quickly.

ClausHolbechArista (changes requested):

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc: any better suggestion? :)

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: yeah will change - it is just that tehn ruff if sad when we use the function outside of the class :)

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: moved to the instance

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-network-port-ptp-profile-does-not-exist.yml
gmuloc: addressed

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: done

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: done

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc: Added a TODO for now - we need to refactor this forward

ClausHolbechArista:
moving to draft since I believe this one is not done yet. At least all tests are failing :)

gmuloc:
this should be revisited once we move more tracking into pyavd-utils

PR #4858, opened 2025-01-06
Feat(eos_designs): Accept auto as argument for rd_override

### LINKED ISSUE #4857: Feat (eos_designs): Accept auto as argument for rd_override
### Enhancement summary

The request is to allow the specific keyword `rd_override: auto` when set overrrides rd to auto.

eos_designs

### Use case example

When migrating existing DC deployments to AVD we find customer using `rd auto` which is currently not supported in AVD.

Existing rd_override supports two formats:
   - A single number will be used in the RD assigned number subfield (second part of the RD).
   - A full RD string with colon separator which will override the full RD.

Can we include a third format `auto`

_No response_

### Additional context

_No response_

### Contributing Guide

ghost:
## Change Summary

This PR adds feature to use the specific keyword `auto` when overriding rd to auto.

## Related Issue(s)

Fixes #4857 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
- Tests the `rd_overrride` to match the literal string `auto`.
- Return `auto` if matches

## Pending changes
- Need to update documenation to following

```   
  rd_override supports three formats:
    - A single number which will be used in the RD assigned number field instead of mac_vrf_id/mac_vrf_vni (see 'overlay_rd_type' for details).
    - A full RD string with colon separator which will override the full RD.
    - The literal string `auto` (This is not supported on a VRF level in EOS)
```

- Need to write tests

- [ ] I have read the [**CONTRIBUTING**](https://avd.sh/en/latest/docs/contribution/overview.html) document.


python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -330,7 +330,7 @@ def get_vrf_rd(self: AvdStructuredConfigNetworkServices, vrf: EosDesigns._Dynami
         rd_override = vrf.rd_override
 
         if rd_override is not None:
-            if ":" in rd_override:
+            if isinstance(rd_override, str) and (":" in rd_override or rd_override == "auto"):
```
ClausHolbechArista: We already know that the value is either a str or None.
```suggestion
            if ":" in rd_override or rd_override == "auto":
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -341,7 +341,7 @@ def get_vrf_rt(self: AvdStructuredConfigNetworkServices, vrf: EosDesigns._Dynami
         """Return a string with the route-target for one VRF."""
         rt_override = vrf.rt_override
 
-        if rt_override is not None and ":" in rt_override:
+        if isinstance(rt_override, str) and (":" in rt_override or rt_override == "auto"):
```
ClausHolbechArista: The check for None is much "cheaper" than isinstance, so when the type can only be None or str, we prefer that.
```suggestion
        if rt_override is not None and (":" in rt_override or rt_override == "auto"):
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -368,7 +368,7 @@ def get_vlan_aware_bundle_rd(
         admin_subfield = self.shared_utils.overlay_rd_type_vrf_admin_subfield if is_vrf else self.shared_utils.overlay_rd_type_admin_subfield
 
         if rd_override is not None:
-            if ":" in str(rd_override):
+            if isinstance(rd_override, str) and (":" in rd_override or rd_override == "auto"):
```
ClausHolbechArista:
```suggestion
            if ":" in rd_override or rd_override == "auto":
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -385,7 +385,7 @@ def get_vlan_aware_bundle_rt(
         rt_override: str | None = None,
     ) -> str:
         """Return a string with the route-target for one VLAN Aware Bundle."""
-        if rt_override is not None and ":" in str(rt_override):
+        if isinstance(rt_override, str) and (":" in rt_override or rt_override == "auto"):
```
ClausHolbechArista:
```suggestion
        if rt_override is not None and (":" in rt_override or rt_override == "auto"):
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -283,7 +283,7 @@ def get_vlan_rt(
         """Return a string with the route-target for one VLAN."""
         rt_override = default(vlan.rt_override, vlan.vni_override)
 
-        if isinstance(rt_override, str) and ":" in rt_override:
+        if isinstance(rt_override, str) and (":" in rt_override or rt_override == "auto"):
```
ClausHolbechArista: Here we need the isinstance because the value can be None, str, or int.

ClausHolbechArista (changes requested):

ClausHolbechArista:
Please add this to molecule tests and verify the outputs are as expected.

I would suggest modifying the file `ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANT_NETWORKS/Tenant_A.yml` and add the rt / rd override on VRF `Tenant_a_ERP_Zone` (line 78) and the SVI 122 below that.
If this also applies to EVPN bundles, we should also add testing in `ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/evpn_vlan_bundle.yml` where you should test it on one of the bundles (ex. `bundle6` in line 162). In the same file it would also be interesting to test on vlan 20 (line 29) since it will get it's own bundle using a different code path.

After updating the files mentioned you can rerun molecule with:
```sh
molecule converge -s eos_designs_unit_tests
```
and then inspect the changed files if you get _all_ and _only_ the changes you expect.

Moving this PR to draft while you work on this. Set it "ready for review" once you have addressed by comments and added test coverage.

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
ghost: I agree. Thanks

ghost:
I ran the above requested unit tests are following are my observations on setting rt and rd override:
### Case1: `evpn_vlan_aware_bundles: true`
1. setting on vrf: overrides as expected.
2. setting on a specific bundle in `evpn_vlan_bundles`: overrides as expected.
3. setting on a single vlan in `l2vlans`: overrides only if vlans are not auto bundles based on vlan name.
4. setting on a single svi in `vrfs.svis`: does not override if bundles are auto generated

### Case2: `evpn_vlan_aware_bundles: false`
1. setting on vrf: overrides as expected.
2. setting on a specific bundle in `evpn_vlan_bundles`: overrides as expected.
3. setting on a single vlan in `l2vlans`: overrides as expected.
4. setting on a single svi in `vrfs.svis`: overrides as expected.

In my opinion it does not make much sense to override these parameters when using vlan aware bundles or atleast when they are auto assigned.

If we agree. Please let me know what tests to add so that we can  merge this feature as soon as possilbe.

**NOTE:** DC1_FABRIC has `evpn_vlan_aware_bundles: true`. The following test will not validate our requirement.
>I would suggest modifying the file `ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANT_NETWORKS/Tenant_A.yml` and add the rt / rd override on VRF Tenant_a_ERP_Zone (line 78) and the SVI 122 below that.

Molecule Test commits on this branch -> https://github.com/rrajpuro-anet/avd/tree/rd-rt-override

ClausHolbechArista:
> If we agree. Please let me know what tests to add so that we can merge this feature as soon as possilbe.
> 

Thank you for going through the tests. The tests you have comitted cover the non-bundle case. If you could add the auto variant in some bundle test as well that would be good. You don't need to add on SVI for the bundle test, but you would need to cover the three other cases since the code paths are separate. Thanks.

ghost:
Hi @ClausHolbechArista, I have added tests for the requested scenarios. Please review.

ClausHolbechArista (approved):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-SVC3A.cfg
```diff
@@ -1120,9 +1120,9 @@ router bgp 65103
       redistribute connected
    !
    vrf Tenant_A_ERP_Zone
-      rd 65103:17
-      route-target import evpn 17:17
-      route-target export evpn 17:17
+      rd auto
```
ClausHolbechArista: This is not available on EOS. We cannot offer the posibility on VRF level.

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-SVC3A.cfg
```diff
@@ -1120,9 +1120,9 @@ router bgp 65103
...
+      route-target import evpn auto
+      route-target export evpn auto
```
gmuloc: testing on vEOS 4.32.0F - these commands are not accepted under VRF:


### auto under VRF

```
site3-leaf1(config-s-gm)#router bgp 666
! IP routing not enabled
site3-leaf1(config-s-gm-router-bgp)# vrf Tenant_A_ERP_Zone
! IP routing not enabled
site3-leaf1(config-s-gm-router-bgp-vrf-Tenant_A_ERP_Zone)#      rd auto
site3-leaf1(config-s-gm-router-bgp)#      route-target import evpn auto
% Invalid input
site3-leaf1(config-s-gm-router-bgp)#      route-target export evpn auto
% Invalid input
site3-leaf1(config-s-gm-router-bgp)#      router-id 192.168.255.12
site3-leaf1(config-s-gm-router-bgp)#      update wait-install
site3-leaf1(config-s-gm-router-bgp)#      neighbor 172.31.12.7 peer group MLAG-PEERS
site3-leaf1(config-s-gm-router-bgp)#show session-config diffs
--- system:/running-config
+++ session:/gm-session-config
+router bgp 666
+   router-id 192.168.255.12
+   update wait-install
+   rd auto
+   neighbor MLAG-PEERS peer group
+   neighbor 172.31.12.7 peer group MLAG-PEERS
+   !
+   vrf Tenant_A_ERP_Zone
```

### RD auto on vlan / vlan-aware-bundle

`rd auto` is configured under main router bgp
`route-target .* auto` is not accepted

### RT auto under VLAN / vlan-aware-bundle

`route-target import auto` under VLAN is not a valid commmand:

```
site3-leaf1(config-s-gm-macvrf-444)#route-target import  auto
% Incomplete command
site3-leaf1(config-s-gm-macvrf-444)#route-target import  auto ?
  <1-4294967295>       Autonomous system number
  <1-65535>.<0-65535>  AS Number in asdot notation
```

`route-target export auto` is accepted but can have some additional input

```
site3-leaf1(config-s-gm-macvrf-444)#route-target export auto ?
  <1-65535>  2 octets AS number
  <cr>

gmuloc (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-SVC3A.cfg
ghost: Correct. So we just have 3 cases for rd auto that is supported in EOS.
- directly under router bgp
- vlan
- vlan-aware-bundle

Do you suggest I disregard this PR and open a new one?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-SVC3A.cfg
ghost: Correct. Since EOS does not support rt auto and it does not make much sense to have this, I will be removing auto rt from this PR and only focus on rd auto

ClausHolbechArista (approved):

gmuloc (approved):

PR #4889, opened 2025-01-16
Feat(cv_deploy): Add verification of the duplicated devices

alexeygorbunov:
## Change Summary

Add new functionality to validate if `structured config` of the hosts processed by `cv_deploy` role contains overlapping `serial_number` or `metadata.system_mac_address` values.

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- Add additional logic to existing `deploy_to_cv` workflow to analyze `serial_number` and `metadata.system_mac_address` values (based on their structured_config files generated by `eos_designs`) of all targeted hosts. 
- Check for the presence of the duplicated `serial_number` or `metadata.system_mac_address` values.
- Raise an exception and terminate execution if:
   - two or more devices have the same `serial_number` (values of `metadata.system_mac_address` are not important in this case)
   - two or more devices have the same `metadata.system_mac_address` and have `serial_number` values unset
   - two or more targeted devices have the same `metadata.system_mac_address`, unique `serial_number` and `tolerate_duplicated_devices` is `False`
- Warn user (with log message and updated `cv_deploy_results.warnings`) if:
   - two or more targeted devices have the same `metadata.system_mac_address`, unique `serial_number` and `tolerate_duplicated_devices` is `True`

## How to test
- Run `cv_deploy` molecule test:
```bash
CVAAS_AAWG_CI='<CVaaS_AT>' molecule test -s cv_deploy
```
- Analyze results of the 3 new molecule workflows:
  - `no_duplicated_devices.yml` tests regular `cv_deploy` with `serial_number` and `metadata.system_mac_address` values being uniquely set per targeted device
  - `duplicated_devices_with_tolerance.yml` tests `cv_deploy` with overlapping `serial_number` and `metadata.system_mac_address` values and default value of `cv_tolerate_duplicated_devices` (`True`)

```bash
        "errors": [
            "('Duplicated devices found in inventory', [{'duplicated_serial_number': '13C20F1EDCCED2D85F6DB2FB9E3AC5B6', 'devices_with_duplicated_serial_number': [CVDevice(hostname='avd-ci-leaf1', serial_number='13C20F1EDCCED2D85F6DB2FB9E3AC5B6', system_mac_address='50:00:00:72:8b:31', _exists_on_cv=None), CVDevice(hostname='avd-ci-leaf2', serial_number='13C20F1EDCCED2D85F6DB2FB9E3AC5B6', system_mac_address='50:00:00:d5:5d:c0', _exists_on_cv=None)]}])"
        ],
        "failed": true,
        "warnings": [],
```

  - `duplicated_devices_without_tolerance.yml` tests `cv_deploy` with overlapping `serial_number` and `metadata.system_mac_address` values and non-default value of `cv_tolerate_duplicated_devices` (`False`)

```bash
        "errors": [
            "('Duplicated devices found in inventory', [{'duplicated_serial_number': '13C20F1EDCCED2D85F6DB2FB9E3AC5B6', 'devices_with_duplicated_serial_number': [CVDevice(hostname='avd-ci-leaf1', serial_number='13C20F1EDCCED2D85F6DB2FB9E3AC5B6', system_mac_address='50:00:00:72:8b:31', _exists_on_cv=None), CVDevice(hostname='avd-ci-leaf2', serial_number='13C20F1EDCCED2D85F6DB2FB9E3AC5B6', system_mac_address='50:00:00:d5:5d:c0', _exists_on_cv=None)]}], [{'duplicated_system_mac_address': '50:00:00:d7:ee:0b', 'devices_with_duplicated_system_mac_address': [CVDevice(hostname='avd-ci-spine1', serial_number='DCC816CEAC4BBD6319385043AD318362', system_mac_address='50:00:00:d7:ee:0b', _exists_on_cv=None), CVDevice(hostname='avd-ci-spine2', serial_number='A7D46613D44DE45D68D5B5C5CBA06B0D', system_mac_address='50:00:00:d7:ee:0b', _exists_on_cv=None)]}])"
        ],
        "failed": true,
        "warnings": [],
```

- Available pytest test covers 97%:

```bash
Name                                              Stmts   Miss  Cover   Missing
-------------------------------------------------------------------------------
python-avd/pyavd/_cv/workflows/verify_inputs.py      39      1    97%   13
-------------------------------------------------------------------------------
TOTAL                                                39      1    97%
```

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/docs/plugins/Modules_and_action_plugins/cv_workflow.md
```diff
@@ -87,6 +88,7 @@ The `arista.avd.cv_workflow` module is an Ansible Action Plugin providing the fo
         device_list: "{{ ansible_play_hosts }}"
         # strict_tags: false
         # skip_missing_devices: false
+        # tolerate_duplicated_devices: true
```
ClausHolbechArista: When would it be a good idea to tolerate duplicated devices? I know this would restrict existing deployments, but I see it more as a bug that we allow duplicate serials Today.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -26,6 +26,9 @@ async def verify_devices_on_cv(
     if not devices:
         return
 
+    # Check if structured config of the inventory devices contains overlapping `serial_number`s or `system_mac_address`es.
+    check_for_duplicated_devices(devices=devices, tolerate_duplicated_devices=tolerate_duplicated_devices, warnings=warnings)
```
ClausHolbechArista: Since this check is about our inputs, I think we should handle it outside of the verify_devices_on_cv function/file. Maybe a verify_inputs function called at first.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -178,3 +181,81 @@ def missing_devices_handler(*, missing_devices: list[CVDevice], skip_missing_dev
...
 
     return exception
+
+
+def check_for_duplicated_devices(*, devices: list[CVDevice], tolerate_duplicated_devices: bool, warnings: list[Exception]) -> None:
+    """Check for the duplicated `serial_number` or `metadata.system_mac_address` values in structured config of the inventory devices."""
+    duplicated_serial_number: dict[str, str | list[str]] = {}
+    duplicated_system_mac_address: dict[str, str | list[str]] = {}
+
+    for device in devices:
```
ClausHolbechArista: I think the logic below will be quite inefficient since it loops over the list of devices multiple times inside a loop over the devices.
I would suggest to use something like groupby to solve this in a more efficient way.

Take a look at `pyavd._utils.groupby_obj` which should be useful here.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -178,3 +181,81 @@ def missing_devices_handler(*, missing_devices: list[CVDevice], skip_missing_dev
...
+    *, duplicated_serial_number: dict[str, str | list[str]], duplicated_system_mac_address: dict[str, str | list[str]], tolerate_duplicated_devices: bool
+) -> Exception:
+    """
+    Handle input devices with duplicated `serial_number`s or `metadata.system_mac_address`es.
+
+      - Raise an exception if tolerate_duplicated_devices is set to False.
+      - Return an Exception if tolerate_duplicated_devices is set to True.
+    """
+    if duplicated_serial_number:
+        LOGGER.warning(
```
ClausHolbechArista: I would expect it to be an error if we do not tollerate duplicates.

python-avd/pyavd/_cv/client/exceptions.py
```diff
@@ -84,3 +84,7 @@ class CVMessageSizeExceeded(CVClientException):
     """Maximum GRPC message size"""
     size: int
     """Actual GRPC message size"""
+
+
+class CVDuplicatedDeviceInfo(CVClientException):
```
ClausHolbechArista:
```suggestion
class CVDuplicatedDevices(CVClientException):
```

ClausHolbechArista (dismissed):

ansible_collections/arista/avd/docs/plugins/Modules_and_action_plugins/cv_workflow.md
alexeygorbunov: Agree, can't think of any valid scenario where one would on purpose use the same serial/sys_mac on multiple devices.
Default `True` is more to have this change as a non-breaking change

python-avd/pyavd/_cv/client/exceptions.py
alexeygorbunov: Addressed in commit 'PR Review: rename class CVDuplicatedDeviceInfo into CVDuplicatedDevices'

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Logic got moved to verify_inputs.py

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Fixed

ansible_collections/arista/avd/docs/plugins/Modules_and_action_plugins/cv_workflow.md
ClausHolbechArista: I see it more as a fix, so I am not sure we need this knob. We are just catching invalid inputs earlier instead of waiting until the workspace is built or the devices are broken because we push config to the wrong device.

alexeygorbunov: Example usecase where we may want to have this knob present and allow duplicated items by default:
- multiple devices, all with both serial_number and system_mac_address set
- all devices have unique serial_numbers but some have overlapping system_mac_address
- this would work OK with AVD+ CV by default as serial_number is prioritized (if set) as an identity attribute (so duplicated system mac would just be ignored by AVD)
- blocking execution in such case (due to the duplicated system mac) would break a valid deployment (although with partially invalid inputs)

ClausHolbechArista: OK, let's change this PR then, to only check for duplicate serials and not have a toggle for it. This is really what is needed.

emilarista (approved):
Tested, LGTM. Error messages appear as expected for duplicated devices.

ansible_collections/arista/avd/docs/plugins/Modules_and_action_plugins/cv_workflow.md
alexeygorbunov: Duplicated system_mac_address may probably cause incorrect config to be pushed to the incorrect device as well. Potential scenario:
 - At least 2 devices with "serial_number" not set (in structured inputs) and with the same/duplicated "system_mac_address" set
 - Once we fetch `found_devices` (cv_client.get_inventory_devices) and start matching existing `devices` with `found_devices` - we'll update `.serial_number` attribute of both `devices` with real serial number (fetched from CV) matching that duplicated "system_mac_address" (address from the inputs)
 - At this point we'll have 2 devices with the same 'serial_number' and 'system_mac_address' but different hostnames and designed EOS config.
 - We can then end up with serial-based (for ID) containers/configlets on CV  holding actual EOS configuration from one of the two raw EoS config files.

ClausHolbechArista: OK, but then we should never accept duplicate system mac either. It does not make sense to ignore something that could explode easily later.

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
 cv_skip_missing_devices: true
 ```
 
+Presence of the same `serial_number` or `metadata.system_mac_address` values in structured configuration of multiple EOS devices may lead to the unexpected results (or even network outages) on the CloudVision side due to the possibility of pushing designed configuration of one device to another device.
+
+To eliminate this risk - role will always raise an error and will terminate it's execution prior to an attempt to update CloudVision in the following cases:
```
```suggestion
To eliminate this risk, this role will always raise an error and will terminate it's execution prior to an attempt to update CloudVision in the following cases:
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
...
+
+- Structured configuration files of two or more targeted devices have the same `serial_number` (values of `metadata.system_mac_address` are not important in this case)
```
ClausHolbechArista: Users don't know what metadata. is. So let's just refer to `system_mac_address`.
```suggestion
- Structured configuration files of two or more targeted devices have the same `serial_number` (values of `system_mac_address` are not important in this case)
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
...
+- Structured configuration files of two or more targeted devices have the same `metadata.system_mac_address` and have `serial_number` values unset
```
```suggestion
- Structured configuration files of two or more targeted devices have the same `system_mac_address` and have `serial_number` values unset
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
...
+
+By default - role will not raise/terminate and will only warn user (with log message and returned `cv_deploy_results.warnings`) about present inconsistency in input structured configuration files in the following case:
```
```suggestion
By default, this role will warn the user about inconsistencies in the structured configuration files in the following case:
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
 cv_skip_missing_devices: true
 ```
 
+Presence of the same `serial_number` or `metadata.system_mac_address` values in structured configuration of multiple EOS devices may lead to the unexpected results (or even network outages) on the CloudVision side due to the possibility of pushing designed configuration of one device to another device.
```
ClausHolbechArista: Let's put all this into it's own subsection `Structured configuration validation` under the `Role Behavior` section.

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
...
+To eliminate this risk - role will always raise an error and will terminate it's execution prior to an attempt to update CloudVision in the following cases:
+
+- Structured configuration files of two or more targeted devices have the same `serial_number` (values of `metadata.system_mac_address` are not important in this case)
+- Structured configuration files of two or more targeted devices have the same `metadata.system_mac_address` and have `serial_number` values unset
+
+By default - role will not raise/terminate and will only warn user (with log message and returned `cv_deploy_results.warnings`) about present inconsistency in input structured configuration files in the following case:
+
+- Structured configuration files of two or more targeted devices have the same `metadata.system_mac_address` but unique `serial_number` values
+
+This later usecase by default should not lead to an unexpected results on a CloudVision (as role overwrites duplicated `metadata.system_mac_address` values based on the real values fetched from the CloudVision using devices' unique `serial_number`s).
```
```suggestion
Having duplicate `system_mac_address` but unique `serial_number` will not lead to unexpected results on CloudVision as the `serial_number` takes precedence.
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -185,6 +185,25 @@ It is possible to ignore other missing devices by simply skipping them and conti
...
+
+Setting role's variable `cv_tolerate_duplicated_devices` to `false` will change that later behavior and will cause role to raise an error even if structured configuration files of two or more targeted devices contain the same `metadata.system_mac_address` but unique `serial_number` values.
```
ClausHolbechArista: Suggesting a rename, since the toggle is limited to this corner case.
```suggestion
To force an error to always be raised in case of duplicate `system_mac_address`, set the `cv_strict_system_mac_address` to `true`
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
ClausHolbechArista: same elsewhere.

ansible_collections/arista/avd/roles/cv_deploy/README.md

ansible_collections/arista/avd/roles/cv_deploy/README.md

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: fixed

ansible_collections/arista/avd/roles/cv_deploy/README.md

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Done

ansible_collections/arista/avd/roles/cv_deploy/README.md

ansible_collections/arista/avd/roles/cv_deploy/README.md

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,133 @@
...
+            if devices_with_set_serial_number := [device for device in devices_with_current_system_mac_address if device.serial_number is not None]:
+                duplicated_system_mac_address_set_serial_number.append(
+                    {
+                        "duplicated_system_mac_address": current_system_mac_address,
+                        "devices_with_duplicated_system_mac_address": devices_with_set_serial_number,
+                    }
+                )
+
+    if duplicated_serial_number or duplicated_system_mac_address_unset_serial_number or duplicated_system_mac_address_set_serial_number:
+        warnings.append(
```
ClausHolbechArista: Let's add it inside the function where we also log the warning. So just no return value.

alexeygorbunov: Fixed

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -245,6 +245,27 @@ cv_register_detailed_results: false
 cv_workspace_build_timeout: 300
 ```
 
+##### Structured configuration validation
+
+Presence of the same `serial_number` or `system_mac_address` values in structured configuration of multiple EOS devices may lead to the unexpected results (or even network outages) on the CloudVision side due to the possibility of pushing designed configuration of one device to another device.
+
+To eliminate this risk, this role will always raise an error and will terminate it's execution prior to an attempt to update CloudVision in the following cases:
```
gmuloc:
```suggestion
To eliminate this risk, this role will always raise an error and will terminate its execution before updating CloudVision in the following cases:
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -245,6 +245,27 @@ cv_register_detailed_results: false
...
+
+- Structured configuration files of two or more targeted devices have the same `serial_number` (values of `system_mac_address` are not important in this case)
+- Structured configuration files of two or more targeted devices have the same `system_mac_address` and have `serial_number` values unset
```
gmuloc:
```suggestion
- Structured configuration files of two or more targeted devices have the same `serial_number` (values of `system_mac_address` are not important in this case).
- Structured configuration files of two or more targeted devices have the same `system_mac_address` and have `serial_number` values unset.
```

docs/plugins/Modules_and_action_plugins/cv_workflow.md
```diff
@@ -40,6 +40,7 @@ The `arista.avd.cv_workflow` module is an Ansible Action Plugin providing the fo
 | <samp>device_list</samp> | list | True | None |  | List of devices to deploy. The names are used to find AVD structured configuration and EOS configuration files. |
 | <samp>strict_tags</samp> | bool | optional | False |  | If `true` other tags associated with the devices will get removed. Otherwise other tags will be left as-is. |
 | <samp>skip_missing_devices</samp> | bool | optional | False |  | If `true` anything that can be deployed will get deployed. Otherwise the Workspace will be abandoned on any issue. |
+| <samp>strict_system_mac_address</samp> | bool | optional | False |  | Controls behavior if input structured configuration contains devices with duplicated system_mac_address but unique serial_number values. |
```
gmuloc: Maybe we need to say warning/error behavior otherwise just reading this I am a bit in difficulty to understand what it does

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,133 @@
...
+    # Set object to track IDs of unique CVDevice objects
+    unique_device_ids: set[str] = set()
+    # List object to hold unique CVDevice objects from original `devices`
+    unique_devices: list[CVDevice] = []
+
+    # Deduplicate CVDevice objects as original `devices` list may contain duplicated items
+    for device in devices:
+        if (device_id := id(device)) not in unique_device_ids:
+            unique_device_ids.add(device_id)
+            unique_devices.append(device)
```
gmuloc:
```suggestion
    device_dict = {id(device): device for devices in devices}
```

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Group devices based on <CVDevice>.serial_number as long as it's not None
+    devices_grouped_by_serial_number = groupby_obj(
+        list_of_objects=[device for device in unique_devices if device.serial_number is not None], attr="serial_number"
+    )
+
+    # Group devices based on <CVDevice>.system_mac_address as long as it's not None
+    devices_grouped_by_system_mac_address = groupby_obj(
+        list_of_objects=[device for device in unique_devices if device.system_mac_address is not None], attr="system_mac_address"
+    )
```
gmuloc: Lots of iterations going on here maybe we can try sonething like this (not 100% tested) reusing the device dict from above
```
    # Create a two layer dict dict[sn][system_mac]=list[Device]
    sn_system_mac = defaultdict(lambda: defaultdict(list))
    for device in device_dict.values():
      sn_system_mac[device.serial_number][device.system_mac_address].append(device)
    
    duplicate_sn_devices = []
    duplicate_mac_no_sn = []
    duplicate_mac_with_sn = []
    for sn, value in sn_system_mac.items():
        # Duplicate SN
        dup_devices = list(chain.from_iterable(a.values()))     
            if len(dup_devices):
                duplicate_sn_devices.append((sn, dup_devices))
        # Duplicate system mac - SN not None:
        for system_mac, devices in value.items():
          if len(devices) > 1:
            if sn is None:
              duplicate_mac_no_sn.append((system_mac, devices))
            else:
              duplicate_mac_with_sn.append((system_mac, devices))
    duplicated_devices_handler(...)
```

ansible_collections/arista/avd/roles/cv_deploy/README.md

ansible_collections/arista/avd/roles/cv_deploy/README.md

python-avd/pyavd/_cv/workflows/verify_inputs.py
alexeygorbunov: Hey @gmuloc ! Just wanted to confirm (as suggested data format is not matching what is needed) - you are suggesting to use comprehension instead, right?

Would this be OK?

```python
    # Deduplicate CVDevice objects as original `devices` list may contain duplicated items
    unique_devices = list({id(device): device for device in devices}.values())
```

instead of

```python
    # Set object to track IDs of unique CVDevice objects
    unique_device_ids: set[str] = set()
    # List object to hold unique CVDevice objects from original `devices`
    unique_devices: list[CVDevice] = []

    # Deduplicate CVDevice objects as original `devices` list may contain duplicated items
    for device in devices:
        if (device_id := id(device)) not in unique_device_ids:
            unique_device_ids.add(device_id)
            unique_devices.append(device)
```

comprehension approach indeed shows faster execution (0.257s to process 1M unique CVDevice instances VS 0.680s needed by the original approach)

python-avd/pyavd/_cv/workflows/verify_inputs.py
alexeygorbunov: Iteration over the single `sn_system_mac[device.serial_number][device.system_mac_address]` may probably miss some use cases as in this case we are looking for overlapping system_mac_addresses only within set of devices with overlapping serial_number.

I tried to modify your suggested approach (available as `def identify_duplicated_devices_new` inside `./python-avd/pyavd/_cv/workflows/verify_inputs.py`) so that its output matches original logic (available as `def identify_duplicated_devices` inside `./python-avd/pyavd/_cv/workflows/verify_inputs.py`).

I as well updated `pytest` with an extensive input scenario covering multiple different combination of CVDevices that we may see.
While both old and new approach right now produce exact same `return` - new approach seems to require more processing time:

  - old approach require `1.883s` to process 1M CVDevices
    ```
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.886    0.886    1.883    1.883 verify_inputs.py:46(identify_duplicated_devices)
    ```

  - new approach require `5.076s` to process 1M CVDevices
    ```
    ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    2.201    2.201    5.076    5.076 verify_inputs.py:105(identify_duplicated_devices_new)
    ```

docs/plugins/Modules_and_action_plugins/cv_workflow.md
alexeygorbunov: Tried to make it more clear:
```diff
- Controls behavior if input structured configuration contains devices with duplicated system_mac_address but unique serial_number values.
+ If `true`, raise an exception if the input data contains devices with a duplicated system_mac_address but unique serial_number values.<br>Otherwise, just issue a warning..
```

python-avd/pyavd/_cv/workflows/verify_inputs.py
ClausHolbechArista: He is suggesting a nested comprehension. Outer is list taking values from an inner dict comprehension. Very fancy and efficient :)

gmuloc: so you had two data structures - one with a set of device IDs and one with a list of unique device

my suggestion was to use list comprehension and to have a single dict structure doing a mapping of

```
dict = {device_id: device} 
```

This way if you need the ids only: `dict.keys()` and if you need the actual objects: `dict.values()`

python-avd/pyavd/_cv/workflows/verify_inputs.py
gmuloc: > Iteration over the single sn_system_mac[device.serial_number][device.system_mac_address] may probably miss some use cases as in this case we are looking for overlapping system_mac_addresses only within set of devices with overlapping serial_number.

Hmm indeed my logic was broken let me take a look at the new code

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -60,6 +60,12 @@
     description: If `true` anything that can be deployed will get deployed. Otherwise the Workspace will be abandoned on any issue.
     type: bool
     default: false
+  strict_system_mac_address:
+    description: |-
+      If `true`, raise an exception if the input structured configuration contains devices with a duplicated system_mac_address but unique serial_number values.
```
gmuloc:
```suggestion
      If `true`, raise an exception if the input data contains devices with a duplicated system_mac_address but unique serial_number values.
```

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,203 @@
...
+    Return tuple containing:
+      - List of dictionaries with information about CVDevices with overlapping serial_number.
+      - List of dictionaries with information about CVDevices with overlapping system_mac_address and unset serial_number.
+      - List of dictionaries with information about CVDevices with overlapping system_mac_address and set serial_number.
+    """
+    devices_grouped_by_serial_number = defaultdict(lambda: defaultdict(list))
+    devices_grouped_by_system_mac_address = defaultdict(lambda: defaultdict(list))
+    for device in devices:
+        devices_grouped_by_serial_number[device.serial_number][device.system_mac_address].append(device)
+        devices_grouped_by_system_mac_address[device.system_mac_address][device.serial_number].append(device)
```
gmuloc: yeah so you have iterables and these are big lists so less efficient and we cannot do what I wanted here alas (because I missed a point in the logic)

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,203 @@
...
+    devices_grouped_by_system_mac_address = groupby_obj(
+        list_of_objects=[device for device in unique_devices if device.system_mac_address is not None], attr="system_mac_address"
+    )
+
+    # Populate list of CVDevice with duplicated serial_number values
+    for current_serial_number, device_iterator_object in devices_grouped_by_serial_number:
+        if len(devices_with_current_serial_number := list(device_iterator_object)) > 1:
+            duplicated_serial_number.append(
+                {"duplicated_serial_number": current_serial_number, "devices_with_duplicated_serial_number": devices_with_current_serial_number}
+            )
```
gmuloc: I think you can build this with.
In anycase you don't need the keys "duplicated_serial_number" / "devices_with_duplcyaed_serial_number" as we know what they are in the code so a simple dict being
`duplicated_sn = {sn: devices}` is enough rather than holding maybe 100 times the strings.

```suggestion
    from itertools import filterfalse
    # we can replace lambda with a helper function that takes a tuple (string, list) and returns true or False)
    duplicated_sn = {sn: devices for sn, devices in filterfalse(lambda group: len(group[1]==1, devices_grouped_by_serial_number)}
```

gmuloc: hmm scratch the suggestion we return iterators inside the iterator in groupby so not sure `devices` will hold anything in the dict need to test, thats why you need the list!

The removing keys from the dict is still valid I think, this is just carrying extra info

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,203 @@
...
+
+    # Populate list of CVDevice with duplicated system_mac_address values
+    for current_system_mac_address, device_iterator_object in devices_grouped_by_system_mac_address:
+        if len(devices_with_current_system_mac_address := list(device_iterator_object)) > 1:
+            if devices_with_unset_serial_number := [device for device in devices_with_current_system_mac_address if device.serial_number is None]:
```
gmuloc: what if there is one device with SN (say `D_A` and one without SN (`D_B) with the same mac address `M`? Then you end up with


```
duplicated_system_mac_address_unset_serial_number =[
    {
        "duplicated_system_mac_address": "M",
        "devices_with_duplicated_system_mac_address": ["D_B"],
    }
]

duplicated_system_mac_address_set_serial_number = [
    {
        "duplicated_system_mac_address": "M",
        "devices_with_duplicated_system_mac_address": ["D_A"],
    }
]
```

and then if `strict_system_mac_address` is `False` then only D_B will be listed in duplicated devices it seems?

python-avd/pyavd/_cv/workflows/verify_inputs.py
gmuloc: and of course we want  best perf :)

python-avd/pyavd/_cv/workflows/verify_inputs.py
alexeygorbunov: Great catch! This is indeed a case that was not yet covered. IF at least one devices with duplicated mac has unset serial - this is a potential issues (even if all other devices with the same mac have unique serial set). I changed logic to align with this approach.

python-avd/pyavd/_cv/workflows/verify_inputs.py
alexeygorbunov: Removed those extra "informative" keys and converted to dicts

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,132 @@
...
+    return duplicated_serial_number, duplicated_system_mac_address_unset_or_mixed_serial_number, duplicated_system_mac_address_set_serial_number
+
+
+def duplicated_devices_handler(
+    *,
+    duplicated_serial_number: dict[str, list[CVDevice]],
+    duplicated_system_mac_address_unset_or_mixed_serial_number: dict[str, list[CVDevice]],
+    duplicated_system_mac_address_set_serial_number: dict[str, list[CVDevice]],
+    strict_system_mac_address: bool,
+    warnings: list[Exception],
```
gmuloc:
```suggestion
    duplicated_serial_number: dict[str, list[CVDevice]],
    duplicated_system_mac_address_unset_or_mixed_serial_number: dict[str, list[CVDevice]],
    duplicated_system_mac_address_set_serial_number: dict[str, list[CVDevice]],
    warnings: list[Exception],
    *,
    strict_system_mac_address: bool,
```

python-avd/pyavd/_cv/workflows/verify_inputs.py
gmuloc: list is better because you don;t actually needs the ids resolving!

python-avd/pyavd/_cv/workflows/verify_inputs.py
alexeygorbunov: Fixed in all added functions

alexeygorbunov: FIxed

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Group devices based on <CVDevice>.serial_number as long as it's not None
+    devices_grouped_by_serial_number = groupby_obj(
+        list_of_objects=[device for device in unique_devices if device.serial_number is not None], attr="serial_number", skip_singles=True
+    )
+
+    # Group devices based on <CVDevice>.system_mac_address as long as it's not None
+    devices_grouped_by_system_mac_address = groupby_obj(
+        list_of_objects=[device for device in unique_devices if device.system_mac_address is not None], attr="system_mac_address", skip_singles=True
+    )
```
gmuloc, line 62: so coming back to this - groupby_obj will created twice a sorted copy of the devices for each call. For a big number (1M) it should be quite significant.
we could optimize this by sorting only once and calling group_by directly here rather than groupby_obj.

gmuloc: yeah this would fail as well because the sorting needs to be with the correct key:)


here is a suggestion NOT using nested default dicts - maybe we can benchmark perf:


```
    devices_by_serial = collections.defaultdict(list)
    devices_by_mac = collections.defaultdict(list)

    # filling up the default dicts
    for device in unique_devices:
        if device.serial_number is not None:
            devices_by_serial[device.serial_number].append(device)
        if device.system_mac_address is not None:
            devices_by_mac[device.system_mac_address].append(device)

    for serial, device_list in devices_by_serial.items():
        if len(device_list) > 1:
            duplicated_devices.serial_number[serial] = device_list

    for mac, device_list in devices_by_mac.items():
        if len(device_list) > 1:
            # Check if *all* devices found for this MAC address have a serial number set.
            if all(d.serial_number is not None for d in device_list):
                duplicated_devices.system_mac_address.set_serial_number[mac] = device_list
            else:
                # If at least one device lacks a serial number, categorize it differently.
                duplicated_devices.system_mac_address.unset_or_mixed_serial_number[mac] = device_list

    return duplicated_devices
```

I know this removes the iterators but given that `sorted` will create a copy of the list twice anyways we should end up using the same space.

alexeygorbunov: Thanks @gmuloc ! I added both proposed approaches and compared (processing of 1M items) with an original approach:

 - original approach: 11000008 function calls in 1.828 seconds
 - proposal A: 10000005 function calls in 4.973 seconds
 - proposal B (Latest): 4000005 function calls in 2.556 seconds

Despite less number of function calls in latest proposal - it's still takes more time to process input data.

Details:
```bash
tests/pyavd/workflows/test_verify_inputs.py::test_identify_duplicated_devices[IDENTIFY_DUPLICATED_DEVICES_FULL_INVENTORY_ORIGINAL_FUNCTION]          11000008 function calls in 1.828 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <string>:2(__init__)
        2    0.542    0.271    1.454    0.727 groupby.py:21(groupby_obj)
  4000000    0.345    0.000    0.572    0.000 groupby.py:24(getkey)
        1    0.314    0.314    1.828    1.828 verify_inputs.py:42(identify_duplicated_devices)
  4000000    0.227    0.000    0.227    0.000 {built-in method builtins.getattr}
  1000000    0.060    0.000    0.060    0.000 {built-in method builtins.id}
  2000000    0.096    0.000    0.096    0.000 {built-in method builtins.len}
        2    0.244    0.122    0.518    0.259 {built-in method builtins.sorted}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        1    0.000    0.000    0.000    0.000 {method 'values' of 'dict' objects}


PASSED
tests/pyavd/workflows/test_verify_inputs.py::test_identify_duplicated_devices[IDENTIFY_DUPLICATED_DEVICES_FULL_INVENTORY_CANDIDATE_A]          10000005 function calls in 4.973 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <string>:2(__init__)
        1    2.151    2.151    4.973    4.973 verify_inputs.py:83(identify_duplicated_devices_candidate_a)
  1000000    0.153    0.000    0.153    0.000 verify_inputs.py:94(<lambda>)
  1000000    2.083    0.000    2.083    0.000 verify_inputs.py:95(<lambda>)
  2000000    0.109    0.000    0.109    0.000 {built-in method builtins.len}
  2000000    0.140    0.000    0.140    0.000 {built-in method from_iterable}
  2000000    0.173    0.000    0.173    0.000 {method 'append' of 'list' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        2    0.000    0.000    0.000    0.000 {method 'items' of 'dict' objects}
  2000000    0.164    0.000    0.164    0.000 {method 'values' of 'dict' objects}


PASSED
tests/pyavd/workflows/test_verify_inputs.py::test_identify_duplicated_devices[IDENTIFY_DUPLICATED_DEVICES_FULL_INVENTORY_CANDIDATE_B]          4000005 function calls in 2.556 seconds

   Ordered by: standard name

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    0.000    0.000 <string>:2(__init__)
        1    2.309    2.309    2.556    2.556 verify_inputs.py:118(identify_duplicated_devices_candidate_b)
  2000000    0.099    0.000    0.099    0.000 {built-in method builtins.len}
  2000000    0.148    0.000    0.148    0.000 {method 'append' of 'list' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
        2    0.000    0.000    0.000    0.000 {method 'items' of 'dict' objects}


PASSED
```

gmuloc: ok well I guess there is not much more to be done there - thanks for checking @alexeygorbunov !

gmuloc: let's clean up the code and merge the method with the 2 groupby_obj!

gmuloc: the code is already cleaned up! approving :)

gmuloc (dismissed):
LGTM - thanks for testing the refactorings!

gmuloc (approved):

PR #4897, opened 2025-01-20
Refactor(eos_designs): Improve structured_config object duplication checks with opt-in

ClausHolbechArista:
## Change Summary

Improve structured_config object duplication checks with opt-in

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Check for duplicate objects across all structured config modules - today we only check within one module
  - Like only finding duplicate interfaces if assigned for different connected endpoints
  - With this we also catch if uplink is the same as a connected endpoint
- This may lead to issues for people "misusing" various protocols to build their configs, so the feature is introduced behind the knob `avd_eos_designs_enforce_duplication_checks_across_all_models: true`
- The duplication check is moved to the schema-based classes where performance is improved significantly.
  - Now using hash lookup instead of looping over all the items for each added item.
- Refactored connected endpoints `ethernet_interfaces` and `port_channel_interfaces` to use the new classes.
  - Fix missing default value for port-channel LACP fallback timer in schema.

## How to test

- Added molecule coverage to negative test with the knob enabled.
- Tested performance for the refactor and seeing ~14% performance improvement for this change alone. Expecting many improvements as we refactor the rest.

### User Checklist

- N/A

### Repository Checklist


emilarista (approved):
Tested with some different conflict combos between servers, uplinks and core_interfaces. The new knob works as expected. LGTM!

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_adapter_config.schema.yml
```diff
@@ -428,7 +422,8 @@ $defs:
                 type: int
                 convert_types:
                   - str
-                description: Timeout in seconds. EOS default is 90 seconds.
+                description: Timeout in seconds.
+                default: 90
```
gmuloc, line 426: note to self: is it what the code was doing?

python-avd/pyavd/_eos_designs/structured_config/__init__.py
```diff
@@ -36,6 +36,10 @@
     AvdStructuredConfigNetworkServices,
     AvdStructuredConfigConnectedEndpoints,
     AvdStructuredConfigInbandManagement,
+    # The classes below this has the property ignore_avd_eos_designs_enforce_duplication_checks_across_all_models = True
```
gmuloc:
```suggestion
    # The classes below this have the property ignore_avd_eos_designs_enforce_duplication_checks_across_all_models = True
```

python-avd/pyavd/_eos_designs/structured_config/__init__.py
```diff
@@ -36,6 +36,10 @@
...
+    # This lets the classes inspect structured config, and since their output will not at a risk of generating duplicate
```
gmuloc:
```suggestion
    # This lets the classes inspect structured config, and since their output is not at risk of generating duplicate
```

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
```diff
@@ -65,8 +84,45 @@ def render(self) -> None:
         In-place update the structured_config by deepmerging the rendered dict over the structured_config object.
 
         This method is bridging the gap between older classes which returns builtin types on all methods,
-        and refactored classes which returns AVD schema class instances. The _from_dict will automatically convert as needed.
+        and refactored classes which inplace updates the self.structured_config.
         """
+        # This knob makes us keep the legacy behavior of maintaining a single structured config object per module and merging them.
+        # Without it set we will just inplace update the same structured_config, which means any duplication checks will be enforced across all modules.
```
gmuloc:
```suggestion
        # Without it set, we will just in-place update the same structured_config, which means any duplication checks will be enforced across all modules.
```

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
```diff
@@ -65,8 +84,45 @@ def render(self) -> None:
...
+            self, "ignore_avd_eos_designs_enforce_duplication_checks_across_all_models", False
+        ):
+            self._complete_structured_config = self.structured_config
+            self.structured_config = EosCliConfigGen()
+
+        # In-place update self.structured_config by calling all the refactored methods marked with @structured_config_contributor
+        self.render_structured_config()
+
+        # The render method on AvdFacts class will only execute methods with @cached_property not starting with _.
+        # These are the legacy methods which once refactored to use the @structured_config_contributor decorator instead.
```
gmuloc:
```suggestion
        # These are the legacy methods which will be refactored to use the @structured_config_contributor decorator instead.
```

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
```diff
@@ -65,8 +84,45 @@ def render(self) -> None:
...
+
+        Each method will in-place update self.structured_config.
+        """
+        for method in self.structured_config_methods():
+            method(self)
+
+    @classmethod
+    def structured_config_methods(cls) -> list[Callable[[Self], None]]:
+        """Return the list of methods decorated with 'structured_config_contributor'."""
+        return [method for key in cls._keys() if hasattr(method := getattr(cls, key), "_is_structured_config_contributor")]
```
gmuloc: is it `hasattr` or do we want to check the value is `True` (I know it is not really needed for now so we can postpone it to a later potential usage)

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -406,4 +406,20 @@ def _compare(self, other: Self, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+                continue
+
+            field_type = field_info["type"]
+            if not issubclass(field_type, AvdModel) or not isinstance(other_value, field_type):
+                # Difference and not a nested model.
+                return False
+
+            value = cast(AvdModel, value)
+            if not value._compare(other_value):
+                return False
```
gmuloc: so is this a fix? we were not casting to AvdModel before

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
```diff
@@ -50,6 +56,10 @@ def ethernet_interfaces(self: AvdStructuredConfigConnectedEndpoints) -> list | N
             )
             network_port_as_adapter._context = network_port._context
             for ethernet_interface_name in range_expand(network_port.switch_ports):
+                # Skip the interface if it was already created by some other feature like connected endpoints or uplinks etc.
```
gmuloc, line 59: uplinks only if the cross module check is enabled but in the long run the comment is true :)

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
```diff
@@ -58,83 +68,69 @@ def ethernet_interfaces(self: AvdStructuredConfigConnectedEndpoints) -> list | N
...
+            )
 
     def _get_ethernet_interface_cfg(
         self: AvdStructuredConfigConnectedEndpoints,
         adapter: EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem.ConnectedEndpointsItem.AdaptersItem,
         node_index: int,
         connected_endpoint: EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem.ConnectedEndpointsItem,
-    ) -> dict:
+    ) -> EosCliConfigGen.EthernetInterfacesItem:
         """Return structured_config for one ethernet_interface."""
```
gmuloc: wondering if this docstring still makes sense

also given that we are having many people to do the refactor should we take the opportunity to write "better" docstrings with Args ?

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
```diff
@@ -146,115 +133,114 @@ def _get_port_channel_interface_cfg(
...
+                    ),
+                    native_vlan_tag=adapter.native_vlan_tag,
+                    native_vlan=adapter.native_vlan,
+                )
+
+            elif adapter.mode == "trunk phone":
+                port_channel_interface.switchport.trunk._update(
+                    native_vlan_tag=adapter.native_vlan_tag,
+                    native_vlan=adapter.native_vlan,
+                )
```
gmuloc: we used to set `groups` for trunk phone and we are not here. Just a note to make sure this is covered in molecule

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
```diff
@@ -58,83 +68,69 @@ def ethernet_interfaces(self: AvdStructuredConfigConnectedEndpoints) -> list | N
...
+                groups=self._get_adapter_trunk_groups(adapter, connected_endpoint, output_type=EosCliConfigGen.EthernetInterfacesItem.Switchport.Trunk.Groups),
+                native_vlan_tag=adapter.native_vlan_tag,
+                native_vlan=adapter.native_vlan,
+            )
+
+        elif adapter.mode == "trunk phone":
+            ethernet_interface.switchport.trunk._update(
+                native_vlan_tag=adapter.native_vlan_tag,
+                native_vlan=adapter.native_vlan,
+            )
```
gmuloc: trunk groups should be able to have groups according to the previous code and the _get_adapter_trunk_groups. We are probably missing this in molecule

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
gmuloc: same a trunk groups

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
```diff
@@ -146,115 +133,114 @@ def _get_port_channel_interface_cfg(
...
+            )
+            or None,
+            shutdown=not (adapter.port_channel.enabled if adapter.port_channel.enabled is not None else True),
+            mtu=adapter.mtu if self.shared_utils.platform_settings.feature_support.per_interface_mtu else None,
+            storm_control=self._get_adapter_storm_control(adapter, output_type=EosCliConfigGen.PortChannelInterfacesItem.StormControl),
+            service_profile=adapter.qos_profile,
+            link_tracking_groups=self._get_adapter_link_tracking_groups(adapter, output_type=EosCliConfigGen.PortChannelInterfacesItem.LinkTrackingGroups),
+            ptp=self._get_adapter_ptp(adapter, output_type=EosCliConfigGen.PortChannelInterfacesItem.Ptp),
+            sflow=self._get_adapter_sflow(adapter, output_type=EosCliConfigGen.PortChannelInterfacesItem.Sflow),
+            flow_tracker=self._get_adapter_flowtracker(adapter, output_type=EosCliConfigGen.PortChannelInterfacesItem.FlowTracker),
```
gmuloc: would these move back to shared utils down the line when we have merged everything?

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_adapter_config.schema.yml
gmuloc: yes it is - we just did not document it!

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
ClausHolbechArista: With the strong hints I don't see much value in spending time on adding args and return for things like this. They would end up being very generic basically repeating the name of the key and the type.

ClausHolbechArista: In this case the doc string is still kind of correct, since it returns structured config data. At some point we will rename the class to `StructuredConfig` instead of `EosCliConfigGen`.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
ClausHolbechArista: Maybe I could write a new wrapper in shared utils that takes the output_type as argument like this.

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
ClausHolbechArista: I can change it to check for boolean value defaulting to False.

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: We never used this before, so yes it is a fix. It makes sure that we compare the content of nested models instead of just checking if they are in fact the same (same id). We may have to reimplement the __eq__ at some point, although it will mean getting a performance penalty.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
ClausHolbechArista: Right. I don't think groups should actually be supported for that case, but it would be a behavioral change, so I will change the section above instead.

ClausHolbechArista: done

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
ClausHolbechArista: reverted the behavioral change. Not sure if the trunk group config would actually work on eos...

ClausHolbechArista (commented):
Address review comments

python-avd/pyavd/_eos_designs/structured_config/structured_config_generator.py
ClausHolbechArista: done

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/port_channel_interfaces.py
ClausHolbechArista: done please take another look.

gmuloc (approved):

PR #4901, opened 2025-01-21
Refactor(plugins): Add support for default_value to natural_sort

ClausHolbechArista:
## Change Summary

Add support for default_value to natural_sort

## Component(s) name

`arista.avd.natural_sort` J2 filter and python util.

## Proposed changes

- Adding `default_value` kwarg to natural_sort, so if missing the key, we will use this value instead for sorting.
- Useful in eos_cli_config_gen templates to enforce same sorting as on EOS without requiring default values.

## How to test

- Added pytest coverage

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/j2filters/natural_sort.py
+++ b/python-avd/pyavd/j2filters/natural_sort.py
@@ -5,40 +5,23 @@ from __future__ import annotations
 
 import re
 from collections.abc import Iterable, Mapping
+from functools import partial
 from typing import Any
 
 from jinja2.runtime import Undefined
 from jinja2.utils import Namespace
 
 
-def convert(text: str, ignore_case: bool) -> int | str:
-    """Converts the input string to be sorted.
-
-    Converts the string to an integer if it is a digit, otherwise converts
-    it to lower case if ignore_case is True.
-
-    Args:
-    -----
-        text (str): Input string.
-        ignore_case (bool): If ignore_case is True, strings are applied lower() function.
-
-    Returns:
-    -------
-        int | str: Converted string.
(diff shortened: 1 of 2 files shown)


python-avd/pyavd/j2filters/natural_sort.py
```diff
@@ -5,40 +5,23 @@
...
-
-def natural_sort(iterable: Iterable | None, sort_key: str | None = None, *, strict: bool = True, ignore_case: bool = True) -> list:
-    """Sorts an iterable in a natural (alphanumeric) order.
+    Sorts an iterable in a natural (alphanumeric) order.
 
     Args:
         iterable: Input iterable.
         sort_key: Key to sort by, defaults to None.
-        strict: If strict is True, raise an error is the sort_key is missing.
+        strict: If strict is True, raise an error is the sort_key is missing and no default value is given.
```
gmuloc: old typo fixing
```suggestion
        strict: If strict is True, raise an error if the sort_key is missing and no default value is given.
```

gmuloc (approved):

Vibhu-gslab (approved):

MaheshGSLAB (approved):
We can move the `pattern = r"(\d+)"` outside method at the top to precompile the regex pattern but not getting much difference while execution. 
LGTM

ClausHolbechArista:
> We can move the `pattern = r"(\d+)"` outside method at the top to precompile the regex pattern but not getting much difference while execution. LGTM

Good idea. I have added that. Tests are still passing so merging when CI is done.

PR #4938, opened 2025-01-28
Refactor: Change mixin classes to use `Protocol`

ClausHolbechArista:
Change structure of classes with Mixins to use "Protocol" as documented in PEP and other sources. This resolves many type checking issues in PyRight and MyPy.


python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/__init__.py
```diff
@@ -33,3 +38,18 @@ def render(self) -> None:
             return super().render()
 
         return None
+
+
+class AvdStructuredConfigConnectedEndpoints(StructuredConfigGenerator, AvdStructuredConfigConnectedEndpointsProtocol):
+    """
+    Protocol for the AvdStructuredConfig Class which is imported by "get_structured_config" to render parts of the structured config.
```
gmuloc: not protocol

ClausHolbechArista: fixed

pyproject.toml
```diff
@@ -27,7 +27,7 @@ ignore = [
   "TD002",   # We don't have require authors in TODO
   "TD003",   # We don't have an issue link for all TODOs today
   "FIX002",  # Line contains TODO - ignoring for ruff for now
-  "F821",    # Disable undefined-name until resolution of #10451
+  # "F821",    # Disable undefined-name until resolution of #10451
```
ClausHolbechArista:
```suggestion
```

gmuloc (approved):
🍏

PR #4955, opened 2025-01-31
Refactor(eos_designs): structured_config for network_services vlans

ClausHolbechArista:
Refactor eos_designs structured_config for network_services vlans to use output classes

PR #4957, opened 2025-01-31
Refactor(eos_designs): structured_config for network_services vlan_interfaces

ClausHolbechArista:
Depends on #4955, otherwise negative tests will fail (because of execution reordering of refactored classes)

Refactor eos_designs structured_config for network_services vlan_interfaces to use output classes

### DIFF
--- a/python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
+++ b/python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Protocol
 
 from pyavd._eos_cli_config_gen.schema import EosCliConfigGen
 from pyavd._errors import AristaAvdError, AristaAvdMissingVariableError
-from pyavd._utils import default, get, get_ip_from_ip_prefix, get_item, strip_empties_from_dict
+from pyavd._utils import Undefined, default, get, get_ip_from_ip_prefix, get_item, strip_empties_from_dict
 from pyavd.api.interface_descriptions import InterfaceDescriptionData
 from pyavd.j2filters import natural_sort, range_expand
 
@@ -229,7 +229,7 @@ class UtilsMixin(Protocol):
                     if svi.id not in vlans:
                         continue
 
-                    interfaces.append(self._get_l2_as_subint(link, svi, vrf))
+                    interfaces.append(self._get_l2_as_subint(link, svi, vrf)._as_dict())
 
         # If we have the main interface covered, we can just exclude it from the list and return as main interface.
         # Otherwise we return an almost empty dict as the main interface since it was already covered by the calling function.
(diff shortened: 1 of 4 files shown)


python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -250,60 +250,63 @@ def _get_l2_as_subint(
...
+            subinterface.flow_tracker._update(**flowtracker)
 
         if svi.structured_config:
             self.custom_structured_configs.nested.ethernet_interfaces.obtain(interface_name)._deepmerge(
                 svi.structured_config._cast_as(EosCliConfigGen.EthernetInterfacesItem, ignore_extra_keys=True),
                 list_merge=self.custom_structured_configs.list_merge_strategy,
             )
 
-        if (mtu := subinterface["mtu"]) is not None and subinterface["mtu"] > self.shared_utils.p2p_uplinks_mtu:
+        if subinterface.mtu and self.shared_utils.p2p_uplinks_mtu and subinterface.mtu > self.shared_utils.p2p_uplinks_mtu:
```
gmuloc, line 288: can `self.shared_utils.p2p_uplinks_mtu` be not set? According to our doc we have a default value of 9214 so this check should not be necessary.

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -250,60 +250,63 @@ def _get_l2_as_subint(
...
+            peer_type=link["peer_type"],
+            description=default(svi.description, svi.name),
+            shutdown=not default(svi.enabled, False),  # noqa: FBT003
+            switchport=EosCliConfigGen.EthernetInterfacesItem.Switchport(enabled=False) if is_native else Undefined,
+            encapsulation_dot1q=EosCliConfigGen.EthernetInterfacesItem.EncapsulationDot1q(vlan=svi.id) if not is_native else Undefined,
+            vrf=vrf.name if vrf.name != "default" else None,
+            ip_address=svi.ip_address,
+            ipv6_address=svi.ipv6_address,
+            ipv6_enable=svi.ipv6_enable,
+            mtu=svi.mtu if self.shared_utils.platform_settings.feature_support.per_interface_mtu else None,
```
gmuloc, line 274: should we use Undefined instead of None where possible? (same for vrf)

```suggestion
            mtu=svi.mtu if self.shared_utils.platform_settings.feature_support.per_interface_mtu else Undefined,
```

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -250,60 +250,63 @@ def _get_l2_as_subint(
...
+            eos_cli=svi.raw_eos_cli,
+        )
+
+        if flowtracker := link.get("flow_tracker"):
+            ## TODO: When link has been refactored to a class this should be changed.
```
gmuloc:
```suggestion
            # TODO: When link has been refactored to a class this should be changed.
```

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
gmuloc: not needed for values only for classes - anyways things are striped at the end

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
ClausHolbechArista: for the sake of type checker since it does not understand the platform setting support of per interface mtu

gmuloc (approved):

PR #4961, opened 2025-01-31
Refactor(eos_designs): structured_config for network_services router_bgp

ClausHolbechArista:
Refactor(eos_designs): structured_config for network_services router_bgp


python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -236,3 +236,19 @@ def _cast_as(self, new_type: type[T_AvdList], ignore_extra_keys: bool = False) -
...
+        if not isinstance(other, cls):
+            msg = f"Unable to compare '{cls}' with a '{type(other)}' class."
+            raise TypeError(msg)
+
+        if len(self) != len(other):
+            return False
+
+        if issubclass(self._item_type, AvdBase):
+            items = cast(list[AvdBase], self._items)
+            other_items = cast(list[AvdBase], other._items)
```
gmuloc: are we not losing info here? (if two subclass have the same fields we could end up with a wrong comparison)

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -93,66 +94,48 @@ def _router_bgp_peer_groups(self: AvdStructuredConfigNetworkServicesProtocol) ->
...
         Covers these areas:
         - vrfs for all VRFs.
         - neighbors and address_family_ipv4/6 for VRF default.
         """
         if not self.shared_utils.network_services_l3:
-            return {}
+            return
 
-        router_bgp = {"vrfs": []}
+        # For VRF default the bgp_vrf will var will be set to the global router_bgp for some settings.
```
```suggestion
        # For VRF default the bgp_vrf will var will be set to the global router_bgp for some settings.
```
```suggestion
        # For VRF default the bgp_vrf variables will be set to the global router_bgp for some settings.
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -164,124 +147,87 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> dict:
...
                         route_map = f"RM-{vrf_name}-{peer_ip}-SET-NEXT-HOP-OUT"
-                        bgp_peer_dict["route_map_out"] = route_map
-                        if bgp_peer_dict.get("default_originate") is not None:
-                            bgp_peer_dict["default_originate"].setdefault("route_map", route_map)
-
-                        bgp_peer_dict.pop("set_ipv4_next_hop", None)
-                        bgp_peer_dict.pop("set_ipv6_next_hop", None)
+                        bgp_peer_config.route_map_out = route_map
+                        if bgp_peer_config.default_originate and not bgp_peer_config.default_originate.route_map:
+                            bgp_peer_config.default_originate.route_map = route_map
```
gmuloc, line 201: seems to be a change in behavior here - it looks like before we would have overwritten the existing route-map and now we don't

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -164,124 +147,87 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> dict:
...
-
-                    append_if_not_duplicate(
-                        list_of_dicts=bgp_vrf.setdefault("neighbors", []),
-                        primary_key="ip_address",
-                        new_dict={"ip_address": peer_ip, **bgp_peer_dict},
-                        context="BGP peer defined under VRFs",
-                        context_keys=["ip_address"],
-                    )
+                    # TODO: Figure out how to fix type checking. It looses track of the bgp_peer_config even though it was derived from bgp_vrf.NeighborsItem.
+                    bgp_vrf.neighbors.append(bgp_peer_config)
```
gmuloc, line 204: seems diffficult I get two diagnostic message - typing is able to infer that bgp_peer_config can be either / or NeighborsItem but so can bgp_vrf and so it complains of the cross assignment issue

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -164,124 +147,87 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> dict:
...
-                bgp_vrf = strip_empties_from_dict(bgp_vrf)
+                if bgp_vrf.neighbors and self.inputs.bgp_update_wait_install and self.shared_utils.platform_settings.feature_support.bgp_update_wait_install:
+                    bgp_vrf.updates.wait_install = True
 
                 # Skip adding the VRF if we have no config.
                 if not bgp_vrf:
                     continue
-
-                if vrf_name == "default":
+                if isinstance(bgp_vrf, EosCliConfigGen.RouterBgp):
```
gmuloc: is isinstance faster than checking the name in that situation?

python-avd/pyavd/_eos_designs/structured_config/network_services/vlan_interfaces.py
```diff
@@ -90,48 +82,52 @@ def _get_vlan_interface_config_for_svi(
...
         # Only set Anycast GW if VARP is not set
-        if not vlan_interface_config.ip_virtual_router_addresses:
-            vlan_interface_config.ip_address_virtual = svi.ip_address_virtual
-            vlan_interface_config.ip_address_virtual_secondaries.extend(svi.ip_address_virtual_secondaries)
-            self._check_virtual_router_mac_address(vlan_interface_config, ("ip_address_virtual", "ip_address_virtual_secondaries"))
-
-        if default(svi.evpn_l3_multicast.enabled, getattr(vrf, "_evpn_l3_multicast_enabled", False)) is True:
+        else:
+            # TODO: Move pim_source_interface_needed under the if svi.ip_address_virtual.
+            # Historicly we checked for presence of ip_address_virtual even if None.
```
gmuloc:
```suggestion
            # Historically we checked for presence of ip_address_virtual even if None.
```

python-avd/pyavd/_schema/models/avd_base.py
```diff
@@ -31,13 +55,30 @@ class AvdBase(ABC):
...
-    _block_inheritance: bool = False
+    _block_inheritance: bool
     """Flag to block inheriting further if we at some point inherited from a class with _created_from_null set."""
 
+    _internal_data_instance: InternalData
+    """Placeholder for Internal data used for storing internal context on data objects, without affecting other logic."""
+
+    def __init__(self) -> None:
+        self._created_from_null = False
+        self._block_inheritance = False
```
gmuloc: maybe lets add a comment that we are moving these to init because slots can't get default values

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -236,3 +236,17 @@ def _cast_as(self, new_type: type[T_AvdList], ignore_extra_keys: bool = False) -
...
+
+    def _compare(self, other: object) -> bool:
+        if not isinstance(other, type(self)):
+            return False
+
+        if len(self) != len(other):
+            return False
+
+        if self._created_from_null != other._created_from_null:
+            return False
```
gmuloc, line 248: we could revert these two operations as the second one seems less expensive and we could fail faster (same for indexed list)

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -236,3 +236,17 @@ def _cast_as(self, new_type: type[T_AvdList], ignore_extra_keys: bool = False) -
...
+
+        items = cast(list[AvdBase], self._items)
+        other_items = cast(list[AvdBase], other._items)
+        return all(item == other_items[index] for index, item in enumerate(items))
```
gmuloc, line 252: 1. are we expecting the lists to be sorted?
2. not sure why we use `==` here and not `_compare` as in indexed list

gmuloc (commented):
left to review router_bgp and avd_models

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista:
```suggestion
        # For VRF default the bgp_vrf variable will be set to the global router_bgp for some settings.
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista: Exactly. Same here. So I may need to move it out in a function with a type var.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista: Absolutely not :) But maybe it was to help the type checker. I will try again

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: But the likelyhood of _created_from_null is very low.

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: In indexed list we offer the to ignore some keys on the append method. So we need to use the _compare to bring those over. Here the items can be other lists or models, so we use __eq__ since that would work the same on all and is simple to understand.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista: we used setdefault before.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
ClausHolbechArista: it was to help the type checker. Now I have two cast() instead

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: well lists are ordered. So if they are not having the same order, they are not the same.

python-avd/pyavd/_schema/models/avd_list.py
ClausHolbechArista: This is just cast to help the type checker accept the _compare.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
gmuloc: yeah I am getting this in other places - lets park this for now

python-avd/pyavd/_schema/models/avd_list.py
gmuloc: ack so not so much perf to shave there

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -111,20 +126,19 @@ def __init__(self, **kwargs: Any) -> None:
 
         Only walking the given kwargs improves performance compared to having named kwargs.
 
-        This method is typically overridden when TYPE_HINTING is True, to provider proper suggestions and type hints for the arguments.
+        This method is typically overridden when TYPE_CHECKING is True, to provider proper suggestions and type hints for the arguments.
```
gmuloc:
```suggestion
        This method is typically overridden when TYPE_CHECKING is True, to provide proper suggestions and type hints for the arguments.
```

gmuloc (approved):

carlbuchmann (approved):
LGTM - reviewed molecule artifacts.
We will be improving error messages over time, for now they convey to correct errors to users.

PR #4962, opened 2025-01-31
Refactor(eos_designs): Refactor network_service vxlan_interfaces

gmuloc:
## Change Summary

TODO:
- [x] adjust negative unit tests
- [x] address TODO with @ClausHolbechArista 

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,98 @@ class VxlanInterfaceMixin(Protocol):
...
+            vxlan.flood_vteps = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.FloodVteps(natural_sort(unique(common)))
 
         if self.shared_utils.overlay_cvx:
-            vxlan["controller_client"] = {"enabled": True}
-
-        vlans: list[dict] = []
-        vrfs: list[dict] = []
-        # vnis is a list of dicts only used for duplication checks across multiple types of objects all having "vni" as a key.
-        vnis: list[dict] = []
+            vxlan.controller_client = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.ControllerClient(enabled=True)
```
ClausHolbechArista:
```suggestion
            vxlan.controller_client.enabled = True
```

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,98 @@ class VxlanInterfaceMixin(Protocol):
...
-            vxlan["source_interface"] = self.shared_utils.vtep_loopback
+            vxlan.source_interface = self.shared_utils.vtep_loopback
 
         if self.shared_utils.mlag_l3 and self.shared_utils.network_services_l3 and self.shared_utils.overlay_evpn:
-            vxlan["virtual_router_encapsulation_mac_address"] = "mlag-system-id"
+            vxlan.virtual_router_encapsulation_mac_address = "mlag-system-id"
 
         if self.shared_utils.overlay_her and not self.inputs.overlay_her_flood_list_per_vni and (common := self._overlay_her_flood_lists.get("common")):
-            vxlan["flood_vteps"] = natural_sort(unique(common))
+            vxlan.flood_vteps = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.FloodVteps(natural_sort(unique(common)))
```
ClausHolbechArista:
```suggestion
            vxlan.flood_vteps.extend(natural_sort(unique(common)))
```

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,98 @@ class VxlanInterfaceMixin(Protocol):
...
         all Network Services deployed on this device.
         """
         if not (self.shared_utils.overlay_vtep or self.shared_utils.is_wan_router):
-            return None
+            return
 
-        vxlan = {
-            "udp_port": 4789,
-        }
+        vxlan = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan(udp_port=4789)
```
ClausHolbechArista: Any reason you want to keep a copy of all the data here instead of just inserting directly into self.structured_config everywhere? (By inserting directly we will not overwrite config contributed by other modules some time in the future.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,98 @@ class VxlanInterfaceMixin(Protocol):
...
         if self.shared_utils.overlay_cvx:
-            vxlan["controller_client"] = {"enabled": True}
-
-        vlans: list[dict] = []
-        vrfs: list[dict] = []
-        # vnis is a list of dicts only used for duplication checks across multiple types of objects all having "vni" as a key.
-        vnis: list[dict] = []
+            vxlan.controller_client = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.ControllerClient(enabled=True)
 
+        # TODO: maybe this could be made a class attribute?
```
ClausHolbechArista: It makes sense to make it a class attribute if you set structured config directly in various helper functions.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,98 @@ class VxlanInterfaceMixin(Protocol):
...
-            "vxlan1": {
-                "description": f"{self.shared_utils.hostname}_VTEP",
-                "vxlan": vxlan,
-            },
-        }
-
-    def _get_vxlan_interface_config_for_vrf(
+                new_vrf = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VrfsItem(name=vrf.name, vni=vrf.wan_vni)
+                self._check_for_duplicate(new_vrf, vnis)
+                vxlan.vrfs.append(new_vrf)
```
ClausHolbechArista: We could just append to the final list everywhere and then have a single function that parses all the lists once and check for duplicate VNIs in the end?

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -305,6 +236,20 @@ def _overlay_her_flood_lists(self: AvdStructuredConfigNetworkServicesProtocol) -
 
         return overlay_her_flood_lists
 
-    @cached_property
-    def _multi_vtep(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
-        return self.shared_utils.mlag is True and self.shared_utils.evpn_multicast is True
+    def _check_for_duplicate(
```
ClausHolbechArista: Let's redesign so we only call this once at the end, where it parses self.structured_config looking for duplicate VNIs between vlans (and soon vlan_ranges) and VRFs.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -305,6 +228,29 @@ def _overlay_her_flood_lists(self: AvdStructuredConfigNetworkServicesProtocol) -
...
+        Raise an exception if the passed vlan or VRF object VNI is already in the vnis dictionary.
+
+        If no duplicate is found, add the vlan or vrf object to the vnis dictionary
+        """
+        vnis: dict[int, EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VrfsItem | EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VlansItem] = {}
+        for vlan in self.structured_config.vxlan_interface.vxlan1.vxlan.vlans:
+            if vlan.vni in vnis:
+                msg = (
+                    "Found duplicate objects with conflicting data while generating configuration for VXLAN VNIs. "
+                    f"VLAN {vlan._as_dict()} conflicts with VLAN {vnis[vlan.vni]._as_dict()}."
```
gmuloc: this could be a VRF from a previous tenant and not always a VLAN

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/duplicate-vrfs-id-conflict.yml
```diff
@@ -44,8 +44,4 @@ tenants:
             ip_address_virtual: 10.0.100.1/24
 
 expected_error_message: >-
-  Found duplicate objects with conflicting data while generating configuration for BGP VRFs defined under network services.
-  {'name': 'VRF1', 'rd': '192.168.255.101:1', 'route_targets': {'import': [{'address_family': 'evpn', 'route_targets': ['1:1']}],
-  'export': [{'address_family': 'evpn', 'route_targets': ['1:1']}]}} conflicts with {'name': 'VRF1', 'rd': '192.168.255.101:2',
-  'route_targets': {'import': [{'address_family': 'evpn', 'route_targets': ['2:2']}], 'export': [{'address_family': 'evpn',
-  'route_targets': ['2:2']}]}}.
+  Found duplicate objects with conflicting data while generating configuration for Vrfs. {'name': 'VRF1', 'vni': 1} conflicts with {'name': 'VRF1', 'vni': 2}.
```
ClausHolbechArista: This happens because the order of rendering changes, so we catch the duplicate under VXLAN VRFs instead of under BGP VRFs.

ClausHolbechArista: try to make vrf_vni unique and keep vrf_id the same.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -181,80 +122,62 @@ def _get_vxlan_interface_config_for_vrf(
             # NOTE: this can never be None here, it would be caught previously in the code
             vrf_id: int = default(vrf.vrf_id, vrf.vrf_vni)
 
-            vrf_data = {"name": vrf_name, "vni": vni}
+            new_vrf = EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VrfsItem(name=vrf.name, vni=vni)
```
ClausHolbechArista: let's call it vxlan_vrf for consistency and better understanding. Thanks.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -305,6 +228,29 @@ def _overlay_her_flood_lists(self: AvdStructuredConfigNetworkServicesProtocol) -
...
+            if vlan.vni in vnis:
+                msg = (
+                    "Found duplicate objects with conflicting data while generating configuration for VXLAN VNIs. "
+                    f"VLAN {vlan._as_dict()} conflicts with VLAN {vnis[vlan.vni]._as_dict()}."
+                )
+                raise AristaAvdInvalidInputsError(msg)
+            vnis[vlan.vni] = vlan
+        for vrf in self.structured_config.vxlan_interface.vxlan1.vxlan.vrfs:
+            if vrf.vni in vnis:
+                conflicting_object_type = "VLAN" if vnis[vrf.vni].__class__.__name__ == "VlansItem" else "VRF"
```
ClausHolbechArista: Let's do isinstance check (like someone told me to do :p)

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/duplicate-vrfs-id-conflict.yml
```diff
@@ -35,6 +37,8 @@ tenants:
     mac_vrf_vni_base: 10000
     vrfs:
       - name: VRF1
+        # Making VNI the same and ID different
+        vrf_vni: 1
```
ClausHolbechArista, line 41: Should be opposite. The title is ID conflict, so ID should be the same. Not VNI.

ClausHolbechArista: nvm. my bad.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,91 @@ class VxlanInterfaceMixin(Protocol):
...
     @cached_property
-    def vxlan_interface(self: AvdStructuredConfigNetworkServicesProtocol) -> dict | None:
+    def _multi_vtep(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
+        return self.shared_utils.mlag is True and self.shared_utils.evpn_multicast is True
+
+    @structured_config_contributor
+    def vxlan_interface(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
         """
-        Returns structured config for vxlan_interface.
+        Update structured config for vxlan_interface.
```
ClausHolbechArista: We have used "Set structured ...." in other places. Also for methods, I think we mostly use update_<> when we pass an object that is inplace updated. Maybe we need a naming guide...

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -305,6 +228,30 @@ def _overlay_her_flood_lists(self: AvdStructuredConfigNetworkServicesProtocol) -
...
+
+        If no duplicate is found, add the vlan or vrf object to the vnis dictionary
+        """
+        vnis: dict[int, EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VrfsItem | EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VlansItem] = {}
+        for vlan in self.structured_config.vxlan_interface.vxlan1.vxlan.vlans:
+            if vlan.vni in vnis:
+                conflicting_object_type = "VLAN" if isinstance(vnis[vlan.vni], EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VlansItem) else "VRF"
+                msg = (
+                    "Found duplicate objects with conflicting data while generating configuration for VXLAN VNIs. "
+                    f"VLAN {vlan._as_dict()} conflicts with {conflicting_object_type} {vnis[vlan.vni]._as_dict()}."
```
ClausHolbechArista: Maybe we could make these messages more clear, instead of dumping the full objects.
```
Found conflicting VNI 321 between VLAN 123 ('MY_VLAN_NAME') in Tenant 'A' and VRF 'foo' in Tenant 'B'.
```
The first tenant part should only be printed if the second is different. I think we have (or had) this logic somewhere.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -24,152 +26,91 @@ class VxlanInterfaceMixin(Protocol):
...
-                    list_of_dicts=vrfs,
-                    primary_key="name",
-                    new_dict=vrf_data,
-                    context="VXLAN VNIs for VRFs",
-                    context_keys=["name", "vni"],
-                )
-
-        if vlans:
-            vxlan["vlans"] = vlans
+                vxlan.vrfs.append_new(name=vrf.name, vni=vrf.wan_vni)
```
ClausHolbechArista: It is a bit confusing that we also add VRFs here - especially since it could be the same VRFs as added in the other logic. So we could easily end up with conflicts, since the other one takes more variables into account. Can we combine them? (Optional since it already works like this, but when we do gateway this will more likely break)

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -305,6 +228,30 @@ def _overlay_her_flood_lists(self: AvdStructuredConfigNetworkServicesProtocol) -
...
+        self: AvdStructuredConfigNetworkServicesProtocol,
+    ) -> None:
+        """
+        Raise an exception if the passed vlan or VRF object VNI is already in the vnis dictionary.
+
+        If no duplicate is found, add the vlan or vrf object to the vnis dictionary
+        """
+        vnis: dict[int, EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VrfsItem | EosCliConfigGen.VxlanInterface.Vxlan1.Vxlan.VlansItem] = {}
+        for vlan in self.structured_config.vxlan_interface.vxlan1.vxlan.vlans:
+            if vlan.vni in vnis:
```
ClausHolbechArista: I think it will be more efficient to use defaultdict and append the object and then walk the list in the end looking for more than one. That way you avoid iterating over the growing list every time

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
gmuloc: at this time I don't hve any info about the tenant when I go through this

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
gmuloc: this is only for WAN servers to pick up WAN VRFs that may have been omitted - need to check if I can do soemthing in filtered tenants for these. probably yes (may be a leftover from some time ago)

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
gmuloc: fixed but this is bringing some new changes in negative unit test where we were not consifering tenant before (which was, granted, less helpful)

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -3,19 +3,37 @@
...
+    vni: int
+    """The VNI."""
+    source_type: Literal["L2VLAN", "VRF", "SVI"]
+    """The source type of the VNI."""
+    name: str
+    """The VRF name or the VLAN ID as a string."""
+    tenant: str = field(compare=False)
+    """The tenant name."""
+
+    def __repr__(self) -> str:
```
ClausHolbechArista, line 33: nice!

ClausHolbechArista (approved):

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
ClausHolbechArista: I will close this comment, since the behavior is the same as Today. Approved the PR.

Vibhu-gslab (approved):

PR #4979, opened 2025-02-05
Fix(eos_cli_config_gen): Fix the errdisable documentation J2 expects recovery.interval to be always set

### LINKED ISSUE #4971: eos_cli_config_gen: errdisable documentation J2 expects recovery.interval to be always set
### Issue Summary

Documentation template `python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2` unconditionally accesses `errdisable.recovery.interval` without trying to confirm it's existence or to set a default value (defined in schema but not implemented in J2):

```
{%         if errdisable.recovery.interval is arista.avd.defined %}
{%         endif %}
{%         if errdisable.recovery.causes is defined %}
|  Detect Cause | Enabled | Interval |
| ------------- | ------- | -------- |
{%             for cause in errdisable.recovery.causes | arista.avd.natural_sort %}
{%                 if cause == 'arp-inspection' %}
| {{ cause }} | True | {{ errdisable.recovery.interval }} |
{%                 elif cause == 'bpduguard' %}
...
``` 

eos_cli_config_gen

### How do you run AVD ?

PyAVD

### Steps to reproduce

```shell
Use input data model with `errdisable.recovery.causes` but without `errdisable.recovery.interval`


errdisable:
  recovery:
    causes:
      - speed-misconfigured
```

### Relevant log output

```shell
File "/usr/local/lib/python3.12/site-packages/pyavd/_eos_cli_config_gen/j2templates/compiled_templates/tmpl_c671a478831eaa86b3ae485f99407794782d2c7e.py", line 117, in top-level template code
    for event in gen:
  File "/usr/local/lib/python3.12/site-packages/pyavd/_eos_cli_config_gen/j2templates/compiled_templates/tmpl_a1a80e0d32b19177c16af5276463801f42022e9e.py", line 63, in top-level template code
    yield '| '
jinja2.exceptions.UndefinedError: 'ansible.parsing.yaml.objects.AnsibleMapping object' has no attribute 'interval'
fatal: [bluespine -> localhost]: FAILED! => {
    "changed": false,
    "msg": "Error during plugin execution: 'ansible.parsing.yaml.objects.AnsibleMapping object' has no attribute 'interval'"
```

### Contributing Guide

MaheshGSLAB:
## Change Summary

Documentation template python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2 unconditionally accesses errdisable.recovery.interval without trying to confirm it's existence or to set a default value (defined in schema but not implemented in J2).
Fix to use default as - if not defined.

## Related Issue(s)

Fixes #4971 

## Component(s) name

`eos_cli_config_gen`

## Proposed changes
Updated the j2 logic in errdisable.j2 for documentation to fix the recovery.interval to be always set

## How to test
Run molecule with new tests

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11612,24 +11612,26 @@ ip nat synchronization
 | xcvr-overheat | True |
 | xcvr-power-unsupported | True |
 
-|  Detect Cause | Enabled | Interval |
-| ------------- | ------- | -------- |
-| arp-inspection | True | 300 |
-| bpduguard | True | 300 |
-| dot1x | True | 300 |
-| hitless-reload-down | True | 300 |
-| lacp-rate-limit | True | 300 |
-| link-flap | True | 300 |
-| no-internal-vlan | True | 300 |
-| portchannelguard | True | 300 |
-| portsec | True | 300 |
-| speed-misconfigured | True | 300 |
-| tapagg | True | 300 |
-| uplink-failure-detection | True | 300 |
-| xcvr-misconfigured | True | 300 |
-| xcvr-overheat | True | 300 |
-| xcvr-power-unsupported | True | 300 |
-| xcvr-unsupported | True | 300 |
+|  Detect Cause | Enabled |
+| ------------- | ------- |
+| arp-inspection | True |
+| bpduguard | True |
+| dot1x | True |
+| hitless-reload-down | True |
+| lacp-rate-limit | True |
+| link-flap | True |
+| no-internal-vlan | True |
+| portchannelguard | True |
+| portsec | True |
+| speed-misconfigured | True |
+| tapagg | True |
+| uplink-failure-detection | True |
+| xcvr-misconfigured | True |
+| xcvr-overheat | True |
+| xcvr-power-unsupported | True |
+| xcvr-unsupported | True |
+
+Errdisable recovery timer interval: 300
 
 ```eos
 !
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11631,7 +11631,7 @@ ip nat synchronization
 | xcvr-power-unsupported | True |
 | xcvr-unsupported | True |
 
-Errdisable recovery timer interval: 300
+Errdisable recovery timer interval: 300 seconds
 
 ```eos
 !
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11601,8 +11601,10 @@ ip nat synchronization
 
 ### Errdisable Summary
 
-|  Detect Cause | Detection Enabled | Recovery Enabled |
-| ------------- | ----------------- | ---------------- |
+Errdisable recovery timer interval: 300 seconds
+
+|  Cause | Detection Enabled | Recovery Enabled |
+| ------ | ----------------- | ---------------- |
 | acl | True | False |
 | arp-inspection | True | True |
 | bpduguard | - | True |
@@ -11621,8 +11623,6 @@ ip nat synchronization
 | xcvr-overheat | True | True |
 | xcvr-power-unsupported | True | True |
 | xcvr-unsupported | - | True |
-
-Errdisable recovery timer interval: 300 seconds
 
 ```eos
 !
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11601,6 +11601,8 @@ ip nat synchronization
 
 ### Errdisable Summary
 
+Errdisable recovery timer interval: 300 seconds
+
 |  Detect Cause | Enabled |
 | ------------- | ------- |
 | acl | True |
@@ -11630,8 +11632,6 @@ ip nat synchronization
 | xcvr-overheat | True |
 | xcvr-power-unsupported | True |
 | xcvr-unsupported | True |
-
-Errdisable recovery timer interval: 300 seconds
 
 ```eos
 !
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11605,13 +11605,13 @@ Errdisable recovery timer interval: 300 seconds
 
 |  Cause | Detection Enabled | Recovery Enabled |
 | ------ | ----------------- | ---------------- |
-| acl | True | False |
+| acl | True | - |
 | arp-inspection | True | True |
 | bpduguard | - | True |
 | dot1x | True | True |
 | hitless-reload-down | - | True |
 | lacp-rate-limit | - | True |
-| link-change | True | False |
+| link-change | True | - |
 | link-flap | - | True |
 | no-internal-vlan | - | True |
 | portchannelguard | - | True |
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
@@ -11601,37 +11601,28 @@ ip nat synchronization
 
 ### Errdisable Summary
 
+|  Detect Cause | Detection Enabled | Recovery Enabled |
+| ------------- | ----------------- | ---------------- |
+| acl | True | False |
+| arp-inspection | True | True |
+| bpduguard | False | True |
+| dot1x | True | True |
+| hitless-reload-down | False | True |
+| lacp-rate-limit | False | True |
+| link-change | True | False |
+| link-flap | False | True |
+| no-internal-vlan | False | True |
+| portchannelguard | False | True |
+| portsec | False | True |
+| speed-misconfigured | False | True |
+| tapagg | True | True |
+| uplink-failure-detection | False | True |
+| xcvr-misconfigured | True | True |
+| xcvr-overheat | True | True |
+| xcvr-power-unsupported | True | True |
+| xcvr-unsupported | False | True |
+
 Errdisable recovery timer interval: 300 seconds
-
-|  Detect Cause | Enabled |
-| ------------- | ------- |
-| acl | True |
-| arp-inspection | True |
-| dot1x | True |
-| link-change | True |
-| tapagg | True |
-| xcvr-misconfigured | True |
-| xcvr-overheat | True |
-| xcvr-power-unsupported | True |
-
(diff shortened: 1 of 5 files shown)


python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
```diff
@@ -9,73 +9,22 @@
 ## Errdisable
 
 ### Errdisable Summary
+{%     if errdisable.detect.causes is defined %}
```
MaheshGSLAB:
```suggestion
{%     if errdisable.detect.causes is arista.avd.defined %}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
```diff
@@ -9,73 +9,22 @@
...
-{%                 elif cause == 'xcvr-overheat' %}
-| {{ cause }} | True |
-{%                 elif cause == 'xcvr-power-unsupported' %}
-| {{ cause }} | True |
-{%                 endif %}
-{%             endfor %}
-{%         endif %}
+{%         endfor %}
+{%     endif %}
+{%     if errdisable.recovery.causes is defined %}
```
MaheshGSLAB:
```suggestion
{%     if errdisable.recovery.causes is arista.avd.defined %}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
Vibhu-gslab:
```suggestion
{%     if errdisable.detect.causes is arista.avd.defined %}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
Vibhu-gslab:
```suggestion
{%     if errdisable.recovery.causes is arista.avd.defined %}
```

Vibhu-gslab (dismissed):

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host2/errdisable.yml
```diff
@@ -0,0 +1,11 @@
...
+### Errdisable ###
+
+errdisable:
+  recovery:
+    causes:
+      - bpduguard
+      - xcvr-overheat
+      - xcvr-power-unsupported
+      - xcvr-unsupported
+      - arp-inspection
```
Vibhu-gslab, line 6: Can we add an example with `interval`

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
```diff
@@ -9,73 +9,22 @@
...
-| {{ cause }} | True | {{ errdisable.recovery.interval }} |
-{%                 elif cause == 'xcvr-unsupported' %}
-| {{ cause }} | True | {{ errdisable.recovery.interval }} |
-{%                 endif %}
-{%             endfor %}
-{%         endif %}
+{%         set interval = errdisable.recovery.interval | arista.avd.default("-") %}
+{%         for cause in errdisable.recovery.causes | arista.avd.natural_sort %}
+| {{ cause }} | True | {{ interval }} |
+{%         endfor %}
```
Vibhu-gslab: If `interval` is defined but `causes` are not this logic would not work.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host2/errdisable.yml
Vibhu-gslab: Also for only `interval`

MaheshGSLAB: With interval is already there in host1.

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
MaheshGSLAB: But then we may not need it if cause is not defined. Else we need to update the documentation having table of only cause and status. Interval should be documented as one liner not for each cause

Vibhu-gslab: @ClausHolbechArista @gmuloc

gmuloc:
```
+errdisable recovery cause bpduguard interval 46
+errdisable recovery cause tapagg interval 555
+errdisable recovery interval 1000
```

so EOS cli supports inteval for each cause but our model does not it seems?

gmuloc: I would think that this "global" interval value: `errdisable.recovery.interval` should be outside of the table if set and then we should update the model to support per recovery cause interval to render in the table (so the column would make sense)

MaheshGSLAB: In this PR i just updated the recovery interval as one liner message. For each cause may be good add it in new PR as feature request(https://github.com/aristanetworks/avd/issues/4980) not a fix.

gmuloc (approved):

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host2/errdisable.yml
```diff
@@ -0,0 +1,11 @@
+---
+### Errdisable ###
+
+errdisable:
+  recovery:
+    causes:
+      - bpduguard
+      - xcvr-overheat
+      - xcvr-power-unsupported
+      - xcvr-unsupported
```
alexeygorbunov: As none of the `xcvr-*` causes are supported by vEOS or cEOS (even with guards disabled) and all of them are already being tested with host1 - i'm wondering if we could avoid those here and use instead those that are not platform-dependent (just to limit the number of "ci-problematic" commands and configlets)

MaheshGSLAB: updated

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
```diff
@@ -11601,6 +11601,8 @@ ip nat synchronization
 
 ### Errdisable Summary
 
+Errdisable recovery timer interval: 300 seconds
+
```
ClausHolbechArista, line 11612: Since we are redesigning this, I think we should combine to a single table with `Cause`, `Detection Enabled`, `Recovery Enabled` (and later `Recovery Interval`).

MaheshGSLAB: Update the table. So for both detect and recovery valid values may not be always same hence i combined both lists with unique elements and check the status of each cause in list or detect/recovery. Only issue i see here is that if a cause is not supported in any of the list even after we are adding status as False. Let me know if that is fine.

ClausHolbechArista: we should render `N/A` for the unsupported. And IMO render `-` for those that are not set.

ClausHolbechArista: hmm I see why the False. I think '-' is better because EOS might have them enabled by default.

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
```diff
@@ -9,73 +9,20 @@
...
-{%                 elif cause == 'xcvr-overheat' %}
-| {{ cause }} | True | {{ errdisable.recovery.interval }} |
-{%                 elif cause == 'xcvr-power-unsupported' %}
-| {{ cause }} | True | {{ errdisable.recovery.interval }} |
-{%                 elif cause == 'xcvr-unsupported' %}
-| {{ cause }} | True | {{ errdisable.recovery.interval }} |
-{%                 endif %}
-{%             endfor %}
-{%         endif %}
+Errdisable recovery timer interval: {{ errdisable.recovery.interval }} seconds
```
ClausHolbechArista: Above the table please.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
alexeygorbunov: "-" would make more sense for Detect column as all causes are enabled for detection by default (so even if we enable them in AVD-generated config - it doesn't really change much on EoS side) and if we don't have them enabled in input vars this doesn't really mean that this detection cause is actually disabled/False on EoS.
But all recovery causes seem to be disabled by default. So if we don't have recovery explicitly enabled in the config - we can more or less be sure that it is not enabled on EoS side as well.

```jinja2
{%             set detect_status = True if cause in (errdisable.detect.causes | arista.avd.default([])) else '-' %}
{%             set recovery_status = cause in errdisable.recovery.causes | arista.avd.default([]) %}
```

MaheshGSLAB: Updated the above logic but as suggested earlier not using one liner if else logic in j2 template.

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
MaheshGSLAB: Updated

ansible_collections/arista/avd/molecule/eos_cli_config_gen/documentation/devices/host1.md
ClausHolbechArista: We do not wish to track EOS' logic and defaults. So we render `-` when we don't have a value available, and whatever value is given when available.

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/errdisable.j2
```diff
@@ -9,73 +9,24 @@
...
-{%         endif %}
+|  Cause | Detection Enabled | Recovery Enabled |
+| ------ | ----------------- | ---------------- |
+{%         for cause in combined_causes | arista.avd.natural_sort %}
+{%             if cause in errdisable.detect.causes | arista.avd.default([]) %}
+{%                 set detect_status = True %}
+{%             else %}
+{%                 set detect_status = "-" %}
+{%             endif %}
+{%             set recovery_status = cause in errdisable.recovery.causes | arista.avd.default([]) %}
```
ClausHolbechArista: This should have the same `True` or `-`

MaheshGSLAB: Updated

gmuloc (approved):
LGTM

ClausHolbechArista (approved):

PR #4990, opened 2025-02-07
Feat(cv_deploy): Expose list of targeted inactive devices if Workspace submission failed due to inactive devices

### LINKED ISSUE #4038: Improve error message on cv_deploy when devices are not streaming
### Enhancement summary

As discussed on GHC, opening request to improve error message for failed submission of workspaces when devices are not streaming

others

### Use case example

Devices are not streaming to CVP/CVaaS and configuration to other devices have to be pushed. The message seems to be referring to devices that do not exist in CVP, but the cause to fail the workspace is that some devices are not streaming:
fatal: [<device_name>-> localhost]: FAILED! => {
    "changed": true,
    "errors": [
        "Failed to submit workspace ws-600bac17-5593-41dd-a8db-95eced2ffae4: Response(status=ResponseStatus.FAIL, message='some devices are inactive', code=ResponseCode.INACTIVE_DEVICES_EXIST)"
    ]

Better error message

_No response_

### Additional context

_No response_

### Contributing Guide

alexeygorbunov:
## Change Summary

Verify streaming status of all devices targeted by `cv_deploy` and raise (and update Errors) or update Warnings (depending on the `force` parameter of the Workspace) if any inactive device caused Workspace submission failure.

## Related Issue(s)

Fixes #4038

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- `cv_client.get_inventory_devices` in `verify_devices_in_cloudvision_inventory` already fetches actual state of all targeted devices including their streaming status. Add additional attribute `_streaming` to describe streaming state of `CVDevice`s.
- If Workspace submission succeeded, was `forced` and `inactive` devices were present prior to initiating submission - log warning message and append it to result.warnings as well:

```
TASK [arista.avd.cv_deploy : Deploy device configurations and tags to CloudVision] ***
[WARNING]: [pyavd] - Inactive devices present: [CVDevice(hostname='avd-ci-
core1', serial_number='20C292B489214DF32F9506C242A722FF',
system_mac_address='50:00:00:a1:33:1a', _exists_on_cv=True, _streaming=False)]
[WARNING]: Inactive devices present: [CVDevice(hostname='avd-ci-core1',
serial_number='20C292B489214DF32F9506C242A722FF',
system_mac_address='50:00:00:a1:33:1a', _exists_on_cv=True, _streaming=False)]

        "errors": [],
        "failed": false,
        "warnings": [
            "Inactive devices present: [CVDevice(hostname='avd-ci-core1', serial_number='20C292B489214DF32F9506C242A722FF', system_mac_address='50:00:00:a1:33:1a', _exists_on_cv=True, _streaming=False)]"
        ],
        "workspace": {
            "force": true,
            "requested_state": "submitted",
            "state": "submitted"
        }
```

- If Workspace submission failed, was not `forced`, `submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST`  and `inactive` devices were present prior to initiating submission - log warning message and raise exception:

```
TASK [arista.avd.cv_deploy : Deploy device configurations and tags to CloudVision] ***
[WARNING]: [pyavd] - Failed to submit CloudVision Workspace due to the presence
of inactive devices. Use force to override. Inactive devices:
[CVDevice(hostname='avd-ci-core1',
serial_number='20C292B489214DF32F9506C242A722FF',
system_mac_address='50:00:00:a1:33:1a', _exists_on_cv=True, _streaming=False)].

        "errors": [
            "Failed to submit CloudVision Workspace due to the presence of inactive devices. Use force to override. Inactive devices: [CVDevice(hostname='avd-ci-core1', serial_number='20C292B489214DF32F9506C242A722FF', system_mac_address='50:00:00:a1:33:1a', _exists_on_cv=True, _streaming=False)]."
        ],
        "failed": true,
        "warnings": [],
        "workspace": {
            "force": false,
            "requested_state": "submitted",
            "state": "submit failed"
        }
```

- If Workspace submission failed, was not `forced`, `submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST`  and but all devices were active prior to initiating submission - log warning message and raise exception without specifying actual devices (they might have change their streaming status between device verification and Workspace submission phases):

```
                msg = (
                    "Failed to submit CloudVision Workspace due to the presence of inactive devices. "
                    "Use force to override. Exact list of inactive devices is unknown."
                )
                LOGGER.warning(msg)
                raise CVWorkspaceSubmitFailedInactiveDevices(msg)
```

- In any other failure cases - raise generic submission exception 

## How to test
- Initiate `cv_deploy` molecule tests targeting CI environment
- Check outputs of the `cv_deploy with cv_submit_workspace_force = true` test
- Check outputs of the `cv_deploy with cv_submit_workspace_force = false` test
- `pytest`

```
Name                                              Stmts   Miss Branch BrPart  Cover   Missing
---------------------------------------------------------------------------------------------
pyavd/_cv/workflows/create_workspace_on_cv.py        17      0      2      0   100%
pyavd/_cv/workflows/finalize_workspace_on_cv.py      62      2     24      2    95%   102, 119
```

### User Checklist
- N/A

### Repository Checklist


ClausHolbechArista:
We should only warn if requested state for the workspace will not need to submit it. It can be useful to do builds and preview the changes even if devices are not streaming.

alexeygorbunov:
> We should only warn if requested state for the workspace will not need to submit it. It can be useful to do builds and preview the changes even if devices are not streaming.

fixed

```
    existing_devices = await verify_devices_in_cloudvision_inventory(
        ....
        raise_if_inactive=((not workspace.force) and workspace.requested_state == "submitted"),
```

python-avd/pyavd/_cv/client/exceptions.py
```diff
@@ -84,3 +84,11 @@ class CVMessageSizeExceeded(CVClientException):
     """Maximum GRPC message size"""
     size: int
     """Actual GRPC message size"""
+
+
+class CVWorkspaceSubmitInactiveDevices(CVClientException):
+    """Can not submit unforced Workspace due to the presence of inactive devices."""
```
ClausHolbechArista:
```suggestion
    """Unable to submit Workspace due to the presence of inactive devices. Use force to override."""
```

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -99,6 +120,23 @@ async def verify_devices_in_cloudvision_inventory(
         device.serial_number = found_device_dict_by_hostname[device.hostname].key.device_id
         device.system_mac_address = found_device_dict_by_hostname[device.hostname].system_mac_address
 
+    if verify_streaming:
```
ClausHolbechArista: If feels like we are building more lists and iterating over everything again and again.

Could be just set the streaming flag during the inspection above?

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -39,6 +54,9 @@ async def verify_devices_in_cloudvision_inventory(
     skip_missing_devices: bool,
     warnings: list[Exception],
     cv_client: CVClient,
+    verify_streaming: bool = False,
+    raise_if_inactive: bool = False,
+    inactive_exception_class: type = CVInactiveDevices,
```
ClausHolbechArista: I don't think we need all these extra options on this function. If you just gather the streaming state in here, you could move the invocation of the verify to the workspace submission logic where everything would make more sense. It would also mean we would not raise early (before even trying to build the workspace).

ClausHolbechArista (changes requested):

python-avd/pyavd/_cv/client/exceptions.py
alexeygorbunov: Done

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: You mean to extend `CVDevice` class to add `streaming` attribute?

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Just to confirm, after moving this streaming check over to submission phase - should we still raise (and avoid initiating submission) is we discover inactive devices targeted to configuration update when WS submission is not forced? Or should we just let submission to be initiated and let if fail (although we know in advance that it will fail)?

ClausHolbechArista: I know I said let's not try, but what if the device started streaming meanwhile. I think we should just submit and catch the error and provide a nice message suggesting the force option.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ClausHolbechArista: Yeah that would be my suggestion. But if we decide to just submit no matter what, we don't even need to record this, but just set the force flag or not as given.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: I think the challenge here is that in contrast to the failing build (where we can fetch details of actual ws_id + build_id pair showing what exactly failed for which device) - failure of the WS submission does not provide these details (`subscribe` to `WorkspaceServiceStub` just returns iterator of `Workspace` objects and they have no details about what exactly happened/failed, just an overall `code` and `status`, not telling exactly which devices were not streaming).
If we assume that streaming status can change from `inactive` to `active` between "verifying devices" step and step to submit WS - then it's probably fair to say that streaming status can change between failing WS submission and re-fetching device inventory (streaming) (if we decide to react to the failed submission postfactum). Maybe a possible way to accurately identify actual streaming state of the devices during WS submission is by subscribing to `DeviceServiceStub` somehow for the whole duration of the submission phase and then correlating potential changes with exact WS submission timestamp. But not sure if this is not an overkill.
Maybe we can then indeed just fetch streaming status at the initial "verification" phase and then if WS submission fails - just assume that streaming didn't change between verification and submission phases

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -6,31 +6,56 @@
...
+    unique_existing_device_tuples: list[tuple[str | None, str | None, str | None]] = []
+    existing_deduplicated_devices: list[CVDevice] = []
+    for existing_device in existing_devices:
+        if (
+            unique_existing_device_tuple := (existing_device.serial_number, existing_device.system_mac_address, existing_device.hostname)
+        ) not in unique_existing_device_tuples:
+            unique_existing_device_tuples.append(unique_existing_device_tuple)
+            existing_deduplicated_devices.append(existing_device)
+
+    return existing_deduplicated_devices
```
ClausHolbechArista: Could we not reuse (or adjust + reuse) something in verify_devices... to avoid running over everything again?

alexeygorbunov: Changed it to form a deduplicated list of `existing_devices` right when we update `CVDevice`s based on the results of the `get_inventory_devices` call

alexeygorbunov:
@ClausHolbechArista `pytest` coverage based on the "static" mocked API calls is ready.

python-avd/pyavd/_cv/client/exceptions.py
```diff
@@ -66,7 +66,11 @@ class CVWorkspaceBuildFailed(CVClientException):
 
 
 class CVWorkspaceSubmitFailed(CVClientException):
-    """Build of CloudVision Workspace failed."""
+    """Failed to submit CloudVision Workspace."""
+
+
+class CVWorkspaceSubmitFailedInactiveDevices(CVClientException):
+    """Failed to submit CloudVision Workspace due to the presence of inactive devices. Use force to override."""
```
ClausHolbechArista: "Use force to override" is not clear enough. Can we refer to the option that should be set, or somehow make it more user friendly without writing a whole essay? Remember this will show up in both cv_workspace and cv_deploy where inputs may be different.

python-avd/pyavd/_cv/client/exceptions.py
```diff
@@ -88,3 +92,7 @@ class CVMessageSizeExceeded(CVClientException):
 
 class CVDuplicatedDevices(CVClientException):
     """Device inputs contain duplicated serial_number or system_mac_address."""
+
+
+class CVInactiveDevices(CVClientException):
```
ClausHolbechArista: I think we need to redesign the warnings to not inherit from exception. Is this the only one or do we have others?

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -98,12 +113,12 @@ async def verify_devices_in_cloudvision_inventory(
         device._exists_on_cv = True
         device.serial_number = found_device_dict_by_hostname[device.hostname].key.device_id
         device.system_mac_address = found_device_dict_by_hostname[device.hostname].system_mac_address
+        # Update streaming status
+        device._streaming = found_device_dict_by_hostname[device.hostname].streaming_status == StreamingStatus.ACTIVE
+        existing_devices.append(device)
 
     # Now we know which devices are on CV, so we can dig deeper and check for them in I&T Studio
     # If a device is found, we will ensure hostname is correct and if not, update the hostname.
-    # If a device is not found, we will set _exist_on_cv back to False.
```
ClausHolbechArista, line 104: Why remove this line?

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -98,12 +113,12 @@ async def verify_devices_in_cloudvision_inventory(
...
-    existing_devices = [device for device in devices if device._exists_on_cv]
-    # Using set to only include a device once.
```
ClausHolbechArista, line 106: Why remove this line?

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,11 +65,43 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
             workspace.state = "submit failed"
+            # Unforced Workspace submission failed due to inactive devices.
+            if submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST and not workspace.force:
+                # Usecase where some of the devices that we targeted were known to be inactive prior to Workspace submission
+                if inactive_devices:
+                    msg = (
+                        f"Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                        f"Use force to override. Inactive devices: {inactive_devices}."
+                    )
+                    LOGGER.warning(msg)
```
ClausHolbechArista: warning and raise seems wrong together. I think these should be LOGGER.error

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,11 +65,43 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+                # Usecase where all devices were actively streaming prior to Workspace submission
+                msg = (
+                    "Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                    "Use force to override. Exact list of inactive devices is unknown."
+                )
+                LOGGER.warning(msg)
+                raise CVWorkspaceSubmitFailedInactiveDevices(msg)
+
+            # If Workspace submission failed for any other reason and known inactive devices were present - append information to warnings.
+            if inactive_devices:
```
ClausHolbechArista: You have this twice for success or failure case. I you move it above the if for submit_result.status you can just have it once.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,11 +65,43 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
             workspace_id=workspace.id,
             request_id=workspace_config.request_params.request_id,
         )
+        # Form a list of known inactive existing devices
+        inactive_devices = [device for device in devices if not device._streaming]
         if submit_result.status != ResponseStatus.SUCCESS:
             workspace.state = "submit failed"
+            # Unforced Workspace submission failed due to inactive devices.
+            if submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST and not workspace.force:
```
ClausHolbechArista: I don't think we need to check for workspace.force here. It would be the only case where INACTIVE_DEVICES_EXIST would block a workspace submission.

python-avd/tests/pyavd/cv/mockery.py
```diff
@@ -104,6 +120,102 @@ async def playback_unary_stream(
...
+    if not recording_file.exists():
+        raise FileNotFoundError(recording_file, "for request", request)
+    recording = json.loads(recording_file.read_text())
+    if recorded_exception := get_v2(recording, "raise"):
+        raise GRPCError(Status[recorded_exception["status"]], recorded_exception["message"])
+    for message_as_dict in recording["payload"]:
+        yield response_type().from_dict(message_as_dict)
+
+
+async def mocked_cv_client_build_workspace(
```
ClausHolbechArista: why do we need these here? The mock stuff above should work for the regular methods. By having these we no longer test the production methods.

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +130,585 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+    """
+    if hostnames:
+        return [CVDevice(item) for item in hostnames]
+    if device_count:
+        return [CVDevice(str(item), str(item), str(item)) for item in range(device_count)]
+    return [CVDevice(str(item), str(item), str(item)) for item in range(1000000)]
+
+
+@dataclass
+class DataUnderTest:
```
ClausHolbechArista: This seems overly complex, so very hard to add and maintain tests.

I would prefer if you just use the constants in the test directly.

Also instead of feeding so many expected* things, it is much simpler to write separate tests for each expected outcome.

ClausHolbechArista (changes requested):

python-avd/pyavd/_cv/client/exceptions.py
alexeygorbunov: Rephrased to "Failed to submit CloudVision Workspace due to the presence of inactive devices. Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` (if using action plugin `cv_workflow`) to override."

python-avd/pyavd/_cv/client/exceptions.py
alexeygorbunov: We do pass Exceptions into the warnings in few places (but not too often). Example:

https://github.com/aristanetworks/avd/blob/51c7ba60d30d2f006d5aaf601fadfa2311a3a744/python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py#L117

Removed `CVInactiveDevices` and references to it.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: I'm just not seeing described functionality implemented anywhere after this line (inside current function or `verify_devices_in_topology_studio` is not resetting `_exist_on_cv`)

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: i'm appending devices to `existing_devices` as a part of the `for device in devices` loop where we already loop through all device and validate their presence (and streaming status) on CV

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/tests/pyavd/cv/mockery.py
alexeygorbunov: Both original `build_workspace` and `submit_workspace` methods form `request` using randomly-generated `request_id` (using `request_id=f"req-{uuid4()}"`). This leads to the different sha hashsum of the request for every call, although we need to point them all to the same hash (file with the same mocked api response)

alexeygorbunov: These new mocked methods read pre-defined request_ids that are passed using temp `_workspace_build_id` attribute of the cv_client: `request_id=self._workspace_build_id`.

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: I had to use dataclass to bypass limitation of 15 parameters that we can pass to the function (pre-commit is failing otherwise).
I'll ping you to clarify `separate tests for each expected outcome. `

python-avd/tests/pyavd/cv/mockery.py
ClausHolbechArista: You can update the expected data before comparing instead. So we only test production code.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Fixed, added _streaming attribute

alexeygorbunov:
Example of the latest run: https://github.com/aristanetworks/avd-internal/actions/runs/17166357449/job/48707595398

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ClausHolbechArista: I think assuming the state did not change during build is okay for now.

ankudinov (approved):
Tested that on staging. LGTM once comments from Claus are resolved.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,9 +65,35 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
             workspace.state = "submit failed"
-            LOGGER.info("finalize_workspace_on_cv: %s", workspace)
+            # Unforced Workspace submission failed due to inactive devices.
+            if submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST:
+                # Usecase where some of the devices that we targeted were known to be inactive prior to Workspace submission
+                if inactive_devices:
+                    msg = (
+                        f"Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                        f"Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
+                        f"(if using action plugin `cv_workflow`) to override. Inactive devices: {inactive_devices}."
```
ClausHolbechArista:
```suggestion
                        f"Failed to submit CloudVision Workspace due to the presence of inactive devices: {inactive_devices}. "
                        f"Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
                        f"(if using action plugin `cv_workflow`) to override."
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,9 +65,35 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+                    )
+                    LOGGER.error(msg)
+                    raise CVWorkspaceSubmitFailedInactiveDevices(msg)
+                # Usecase where all devices were actively streaming prior to Workspace submission
+                msg = (
+                    "Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                    "Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
+                    "(if using action plugin `cv_workflow`) to override. Exact list of inactive devices is unknown."
```
ClausHolbechArista:
```suggestion
                    "Failed to submit CloudVision Workspace due to the presence of inactive devices. "
                    "Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
                    "(if using action plugin `cv_workflow`) to override."
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -42,7 +42,7 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
     build_result, cv_workspace = await cv_client.wait_for_workspace_response(workspace_id=workspace.id, request_id=workspace_config.request_params.request_id)
     if build_result.status != ResponseStatus.SUCCESS:
         workspace.state = "build failed"
-        LOGGER.info("finalize_workspace_on_cv: %s", workspace)
+        LOGGER.error("finalize_workspace_on_cv: %s", workspace)
```
ClausHolbechArista: We grab the errors here and print in Ansible as red. I am not sure we want that for this internal data structure. Better to keep the info log since it is more for troubleshooting/debugging.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,9 +65,35 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+            # Unforced Workspace submission failed due to inactive devices.
+            if submit_result.code == ResponseCode.INACTIVE_DEVICES_EXIST:
+                # Usecase where some of the devices that we targeted were known to be inactive prior to Workspace submission
+                if inactive_devices:
+                    msg = (
+                        f"Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                        f"Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
+                        f"(if using action plugin `cv_workflow`) to override. Inactive devices: {inactive_devices}."
+                    )
+                    LOGGER.error(msg)
```
ClausHolbechArista: I think both logging and raising may lead to duplicate messages when running in Ansible. I'd prefer to just keep it as raise.

ClausHolbechArista: same elsewhere. Feel free to test first to confirm the behavior.

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,9 +65,35 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
             workspace_id=workspace.id,
             request_id=workspace_config.request_params.request_id,
         )
+        # Form a list of known inactive existing devices
+        if inactive_devices := [device for device in devices if not device._streaming]:
```
ClausHolbechArista: The print-out of inactive devices is not very user-friendly. Let's do a string instead.

```suggestion
        if inactive_devices := [f"{device.hostname}({device.serial_number})" for device in devices if not device._streaming]:
```

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,9 +65,35 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
...
+                msg = (
+                    "Failed to submit CloudVision Workspace due to the presence of inactive devices. "
+                    "Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` "
+                    "(if using action plugin `cv_workflow`) to override. Exact list of inactive devices is unknown."
+                )
+                LOGGER.error(msg)
+                raise CVWorkspaceSubmitFailedInactiveDevices(msg)
+
+            # If Workspace submission failed for any other reason - raise general exception.
+            LOGGER.error("finalize_workspace_on_cv: %s", workspace)
```
ClausHolbechArista: info

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -26,51 +43,88 @@
...
+ExpectedExceptionContext = AbstractContextManager[pytest.ExceptionInfo | None]
 
-@pytest_asyncio.fixture
-async def cv_client() -> AsyncGenerator[CVClient, None]:
-    """
-    Instance of CVClient.
 
-    If CV_ACCESS_TOKEN environment variable is set, but RECORDING environment variable is not set,
-    this will return a proper instance of CVClient connected to CloudVision with the token.
+def _mocked_cvdevices(hostnames: list[str] | None = None, device_count: int | None = None) -> list[CVDevice]:
```
ClausHolbechArista: In general these tests I think these tests focuses too much on reuse over readability. It is really hard for me to figure out where and how to add a test, or to decode why a test may be failing.

I would prefer if we could rewrite this to be many small tests just doing and asserting on the things we are testing instead of having these helper functions (especially the deploy_to_cv_core_logic) moving it away from the actual test.

But we can look into this and discuss further seperately, so I will not hold this PR back.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ClausHolbechArista: Take the easy route right now, so just take the streaming state gathered during inventory verification and use that to print the list. But don't assume it will fail.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -51,14 +57,16 @@ async def verify_devices_in_cloudvision_inventory(
...
     Skip checks for devices where _exists_on_cv is already filled out on the device.
 
-    Returns a list of CVDevice objects found to exist on CloudVision.
+    Populate current streaming status for all existing devices.
+
+    Returns deduplicated list of CVDevice objects found on CloudVision.
     """
     # Using set to only include a device once.
-    device_tuples = {
+    device_tuples = [
```
gmuloc: comments above is referring to set but you are changing to list - any reason to change from set to list?

python-avd/tests/pyavd/cv/mockery.py
```diff
@@ -65,12 +69,15 @@ async def recording_unary_stream(
...
-        messages_as_json.append(message.to_json(indent=4))
-        yield message
 
-    result = f"[{', '.join(messages_as_json)}]"
-    recording_file.write_text(result)
+    try:
+        async for message in self._org_unary_stream(route, request, response_type, timeout=timeout, deadline=deadline, metadata=metadata):
+            messages_as_json.append(message.to_json(indent=4))
+            yield message
+    finally:
```
gmuloc: so I understand you added this to make sure we capture the messages even if something goes awry - but if for any reason there is a transietn Exception - would we not record incomplete data?

python-avd/tests/pyavd/cv/mockery.py
```diff
@@ -104,6 +111,43 @@ async def playback_unary_stream(
         yield response_type().from_pydict(message_as_dict)
 
 
+async def playback_static_recording_unary_unary(
```
gmuloc, line 133: what is the difference between a static recording and a recording?

python-avd/tests/pyavd/cv/client/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Instance of CVClient.
+
+    If CV_ACCESS_TOKEN environment variable is set, but RECORDING environment variable is not set,
+    this will return a proper instance of CVClient connected to CloudVision with the token.
+
+    If CV_ACCESS_TOKEN environment variable is set, but RECORDING environment variable is set,
+    this will return an instance of CVClient connected to CloudVision with the token where all API calls will be recorded.
+
+    Otherwise this will return an instance of CVClient where API calls are mocked using previously recorded API messages.
```
gmuloc, line 53: whats it the point of `static_recording`?

python-avd/tests/pyavd/cv/client/conftest.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    static_recording = get_v2(request, "param.static_recording")
```
gmuloc: is it needed here? it seems it could be called much later in the funciton

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cv_client", [{"static_recording": True}], ids=["CV_CLIENT_STATIC_RECORDINGS"], indirect=True)
+async def test_finalize_workspace_on_cv_built_state(caplog: pytest.LogCaptureFixture, cv_client: CVClient) -> None:
+    """
+    Test Workspace in built state.
+
+    Build request:
+        WorkspaceConfigSetRequest(value=WorkspaceConfig(key=WorkspaceKey(workspace_id='ws-cbf7c7ea-a57c-481d-b96b-97c12856395e'),
+        request=Request.START_BUILD, request_params=RequestParams(request_id='req-914310f3-08dd-4239-bd42-6d78bf781229')))
+    Recodred build response:
```
gmuloc:
```suggestion
    Recorded build response:
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+        # bd1b5fdaa11249efe21fa9479c729168b06cda69.json
+        pytest.param("ws-cbf7c7ea-a57c-481d-b96b-97c128560000", "pending", does_not_raise(), id="PENDING"),
+        # mocked API response: tests/pyavd/cv/mocked_api_recordings/arista.workspace.v1.WorkspaceService/GetOne/www.cv-prod-us-central1-c.arista.io/\\
+        # e3c8d23b2dffba4c050956c45d0bda0124500f00.json
+        pytest.param("ws-cbf7c7ea-a57c-481d-b96b-97c128560001", None, pytest.raises(CVResourceInvalidState), id="ROLLED_BACK"),
+    ],
+)
+async def test_create_existing_workspace_on_cv(
+    caplog: pytest.LogCaptureFixture, cv_client: CVClient, workspace_id: str, workspace_state: str | None, expected_exception: ExpectedExceptionContext
+) -> None:
```
gmuloc: please add some docstrings in this test as the amount of parameter makes it sometime hard to follow

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cv_client", [{"static_recording": True}], ids=["CV_CLIENT_STATIC_RECORDINGS"], indirect=True)
+async def test_finalize_workspace_on_cv_abandoned_state(caplog: pytest.LogCaptureFixture, cv_client: CVClient) -> None:
+    """
+    Test Workspace in abandoned state.
+
+    Build request:
+        WorkspaceConfigSetRequest(value=WorkspaceConfig(key=WorkspaceKey(workspace_id='ws-cbf7c7ea-a57c-481d-b96b-97c12856395e'),
+        request=Request.START_BUILD, request_params=RequestParams(request_id='req-914310f3-08dd-4239-bd42-6d78bf781229')))
+    Recodred build response:
```
gmuloc:
```suggestion
    Recorded build response:
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+    """
+    workspace_id: str = MOCKED_WORKSPACE_ID
+    workspace_build_id: str = MOCKED_WORKSPACE_REQUEST_ID_BUILD_SUCCESS["id"]
+    workspace_abandon_id: str = MOCKED_WORKSPACE_REQUEST_ID_ABANDON["id"]
+    workspace_requested_state: str = "abandoned"
+    workspace_state: str = "abandoned"
+
+    with (
+        caplog.at_level(INFO),
+        patch("pyavd._cv.client.workspace.uuid4", side_effect=[workspace_build_id.removeprefix("req-"), workspace_abandon_id.removeprefix("req-")]),
```
gmuloc: Note to self:
seems here we need to remove a prefix because the constant has it because it is needed in other context

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+    ):
+        workspace = CVWorkspace(id=workspace_id, requested_state=workspace_requested_state)
+        await finalize_workspace_on_cv(workspace, cv_client, _mocked_cvdevices(hostnames=["avd-ci-leaf1"]), [])
```
gmuloc: is is ok that we hardcode the device on a recorded test?

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+    with caplog.at_level(INFO), patch("pyavd._cv.client.workspace.uuid4", side_effect=[workspace_build_id.removeprefix("req-")]):
+        workspace = CVWorkspace(id=workspace_id, requested_state=workspace_requested_state)
+        await finalize_workspace_on_cv(workspace, cv_client, _mocked_cvdevices(hostnames=["avd-ci-leaf1"]), [])
+
+    assert workspace.state == workspace_state
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cv_client", [{"static_recording": True}], ids=["CV_CLIENT_STATIC_RECORDINGS"], indirect=True)
+async def test_finalize_workspace_on_cv_abandoned_state(caplog: pytest.LogCaptureFixture, cv_client: CVClient) -> None:
```
gmuloc: is it correct to understand that this test can only be run with the static recording and cannot be updated with "newer" recordings if say the format changed in the future (we would need to record new requests and write new tests for these?)

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+        request=Request.ABANDON, request_params=RequestParams(request_id='req-b65374c1-4333-4c68-9b09-d753e8560609')))
+    Recorded abandon responses:
+        tests/pyavd/cv/mocked_api_recordings/arista.workspace.v1.WorkspaceConfigService/Set/www.cv-prod-us-central1-c.arista.io/
+        c3455eeb927146c3ba4e5fbb3d51b959fc84da17.json
+    """
+    workspace_id: str = MOCKED_WORKSPACE_ID
+    workspace_build_id: str = MOCKED_WORKSPACE_REQUEST_ID_BUILD_SUCCESS["id"]
+    workspace_abandon_id: str = MOCKED_WORKSPACE_REQUEST_ID_ABANDON["id"]
+    workspace_requested_state: str = "abandoned"
+    workspace_state: str = "abandoned"
```
gmuloc:
```suggestion
    expected_workspace_state: str = "abandoned"
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+@pytest.mark.asyncio
+@pytest.mark.parametrize("cv_client", [{"static_recording": True}], ids=["CV_CLIENT_STATIC_RECORDINGS"], indirect=True)
+async def test_finalize_workspace_on_cv_deleted_state(caplog: pytest.LogCaptureFixture, cv_client: CVClient) -> None:
+    """
+    Test Workspace in deleted state.
+
+    Build request:
+        WorkspaceConfigSetRequest(value=WorkspaceConfig(key=WorkspaceKey(workspace_id='ws-cbf7c7ea-a57c-481d-b96b-97c12856395e'),
+        request=Request.START_BUILD, request_params=RequestParams(request_id='req-914310f3-08dd-4239-bd42-6d78bf781229')))
+    Recodred build response:
```
gmuloc:
```suggestion
    Recorded build response:
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+    logs_patterns: str,
+    expected_exception: ExpectedExceptionContext,
+) -> None:
+    """
+    Test Workspace with failing build.
+
+    Build request:
+        WorkspaceConfigSetRequest(value=WorkspaceConfig(key=WorkspaceKey(workspace_id='ws-cbf7c7ea-a57c-481d-b96b-97c12856395e'),
+        request=Request.START_BUILD, request_params=RequestParams(request_id='req-914310f3-08dd-4239-bd42-6d78b0000000')))
+    Recodred build response:
```
gmuloc:
```suggestion
    Recorded build response:
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -85,3 +139,811 @@ async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
...
+        request=Request.SUBMIT, request_params=RequestParams(request_id='req-b8f4e511-58de-4afe-99f0-b75abf980131')))
+    Recorded submit responses:
+        tests/pyavd/cv/mocked_api_recordings/arista.workspace.v1.WorkspaceConfigService/Set/www.cv-prod-us-central1-c.arista.io/
+        ba83e98eab07691e8b079958618ab2973822bfe8.json
+    """
+    with patch(
+        "pyavd._cv.client.workspace.uuid4",
+        side_effect=[workspace["build_request_id"]["id"].removeprefix("req-"), workspace["submit_request_id"]["id"].removeprefix("req-")],
+    ):
+        await _deploy_to_cv_core_logic(**locals())
```
gmuloc: I feel this is a bit too much of dark magic.

1/ not every person knows what locals contain
2/ it may contain other stuff down the line (difficult pattern to copy)
3/ this is really tough to follow
4/ we are not coupling the test function inputs with the underlying core one,
I would vote to have the 5 extra lines where we pass only the param we want

gmuloc (commented):
mostly questions and couple of typos. I won't be blocking the PR for these but I feel like the test suite is going to be tough to maintain for the team

ClausHolbechArista (changes requested):

python-avd/tests/pyavd/cv/mockery.py
gmuloc:
```suggestion
    finally:
        # TODO: Investigate whether it can lead to incomplete data
```

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Few reasons:
1.  static typing does not match targeted `cv_client.get_inventory_devices` (as it expects list): https://github.com/aristanetworks/avd/blob/8cdff312ae2c19103f0af292bf03e52bfac3a270/python-avd/pyavd/_cv/client/inventory.py#L28
2. Using set here break logic that generates hash (to have consistent filenames) based on the `DeviceStreamRequest` (as it is using `partial_eq_filter` which is filled by looping through the iterable with devices). When devices are ina a form of set - order is different (not stable) and therefore we can not get consistent Request string and ending up with non-matching hashes

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Sure, changed back to `info`

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: I implemented the following dedup inside `deploy_to_cv.py::deploy_to_cv` to dedup this list early in a workflow:

```
            # Form deduplicated list of targeted CVDevices
            devices = list(
                {
                    id(device): device
                    for device in (
                        [tag.device for tag in device_tags if tag.device is not None]
                        + [tag.device for tag in interface_tags if tag.device is not None]
                        + [config.device for config in configs if config.device is not None]
                    )
                }.values()
            )
```

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: It should be this way right now

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Using `set()` later in the workflow breaks hashing of the request string because looping through set of devices (when appending items to partial_eq_filter) gives unexpected order

python-avd/tests/pyavd/cv/mockery.py
alexeygorbunov: i solved this by pathing `uuid` class imported into `pyavd._cv.client.workspace` so that it returns predefined values (based on what test function needs). Example:

```python
    with caplog.at_level(INFO), patch("pyavd._cv.client.workspace.uuid4", side_effect=[workspace_build_id.removeprefix("req-")]):
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: This dataclass thing was removed. Now just grouped params with similar purpose.

python-avd/tests/pyavd/cv/mockery.py
alexeygorbunov: General recordings are those that are natively received from CV and dumped as is. Static recordings are the same recordings but wrapped into `payload` and `raise` that we further use (rely on) during the tests (like in mockery.py::playback_static_recording_unary_stream). These static recordings are not updated/impacted by `RECORDING` evn var and must be changed manually (if ever needed).

python-avd/tests/pyavd/cv/client/conftest.py
alexeygorbunov: Static recordings wrapped into `payload` and with optional `raise` that we use as an input for the defined pytests.
Regular recordings at this point are still used by few earlier pytests but those should probably be migrated to the static ones as well. Regular recordings will be needed when if responces of existing calls change or new requests/responces need to be captured dynamically from the CV

python-avd/tests/pyavd/cv/client/conftest.py
alexeygorbunov: Fixed. Moved to the line where it's actually used

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: All fixed

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Done

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Fixed

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: `req-` is a static prefix added when `RequestParams` based on random `uuid` https://github.com/aristanetworks/avd/blob/8cdff312ae2c19103f0af292bf03e52bfac3a270/python-avd/pyavd/_cv/client/workspace.py#L135

We are removing `req-` just to fetch a consistent uuid value which is later used as a hash function input

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Device have to be statically predefined as they are used in `DeviceStreamRequest` which is later passed into the `hash` function to generate unique and consistent filename

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: We can always update static recordings if there changes on the CV side. We'll just need to run a live call towards CV and record it. These pytests were mostly created for offline testing which can be ran as many times as needed and in parallel.
For all live tests we have molecule scenarios right now. Maybe we could extend them to capture actual responses for all requests that we use and notify us where those responses change (so that we don't learn it when users start complaining)? Or we can as well add some live recording tests in pytest and run them nightly to update and "compare" regular recordings

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Changed

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: FIxed

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Fixed

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Agreed. Especially that fact that extra variables in locals would break it on a receiving function side.

Tried to improve it by the following:

```python
# globally expose expected params
_DEPLOY_TO_CV_CORE_LOGIC_PARAMS = inspect.signature(_deploy_to_cv_core_logic).parameters.keys()

# in a caller function - only take what is expected:
        await _deploy_to_cv_core_logic(**{param: value for param, value in locals().items() if param in _DEPLOY_TO_CV_CORE_LOGIC_PARAMS})
```

python-avd/tests/pyavd/cv/mockery.py
alexeygorbunov: If exception is raised within `try` - it will still be re-raised after `finally` but we will at least dump response(s) what were received prior/during the exception. This should help with understanding of what has failed.

alexeygorbunov:
Latest run of the `cv_deploy` CI: https://github.com/aristanetworks/avd-internal/actions/runs/17317129009/job/49162117186

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
gmuloc: ok so we need to remove the comment line 64 which is now incorrect.
It feels like we could end up with multiple time the same device but it may be caught someplace else (which was the goal of the set according to the comment).

for 1. there is only one function calling get_inventory_devices which is this one so we could also have changed the typehint there.

2. is a very good reason though it feels it is impacting only testing right? (the hash of the file is used for the testing if I am not wrong?)

python-avd/tests/pyavd/cv/mockery.py
gmuloc:
```suggestion
# General recordings are those that are natively received from CV and dumped as is.
# Static recordings are the same recordings but wrapped into 'payload' and 'raise'
# that are used during the tests like in mockery.py::playback_static_recording_unary_stream.
# These static recordings are not updated/impacted by RECORDING env var and must be changed manually (if ever needed).
async def playback_static_recording_unary_unary(
```

python-avd/tests/pyavd/cv/client/test_cv_client.py
gmuloc: lets maybe keep track of this to make sure our tests don't become obsolete - it can be a criteria when we redesign the tests as we discussed yesterday

python-avd/tests/pyavd/cv/client/test_cv_client.py
gmuloc: hehe forgot to remove this comment sorry!

python-avd/tests/pyavd/cv/client/test_cv_client.py
gmuloc: does it imply that if we were to record new answers we need to have the same device?

python-avd/tests/pyavd/cv/client/test_cv_client.py
ClausHolbechArista: this is just worse... Just write the kwargs out like we do everywhere else.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ClausHolbechArista: so we are changing the production code to make our testing pass? :D

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ClausHolbechArista: We should not change the production code to satisfy the needs of our testing infra. Then we need to change the way we test.

gmuloc (commented):
Screenshot 2025-08-29 at 11 47 22
Screenshot 2025-08-29 at 11 45 13


tested latest version and behavior looks good

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -51,14 +57,15 @@ async def verify_devices_in_cloudvision_inventory(
 
     Skip checks for devices where _exists_on_cv is already filled out on the device.
 
-    Returns a list of CVDevice objects found to exist on CloudVision.
+    Populate current streaming status for all existing devices.
+
+    Returns deduplicated list of CVDevice objects found on CloudVision.
     """
-    # Using set to only include a device once.
```
alexeygorbunov, line 56: CVDevices are now deduplicated at the early stage of the deploy_to_cv module

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Returned back set() and adjusted scope of the pytest

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
alexeygorbunov: Rolled back all changes inspired by failing pytests

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: If we want new record (to check new path/request or look for new fields) - we don't really have to use the same device. Having the same request and same set of devices is only important for replaying

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Removed all this extra logic.

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -26,62 +46,771 @@
 CV_TOKEN = environ.get("CV_ACCESS_TOKEN")
 RECORDING = environ.get("RECORDING")
 
+ExpectedExceptionContext = AbstractContextManager[pytest.ExceptionInfo | None]
+
+
+@pytest.mark.asyncio
+async def test_get_inventory_devices(cv_client: CVClient) -> None:
```
carl-baillargeon: Can we move these in their relevant test modules for clarity?

`get_inventory_devices` tests should go to `client/test_inventory.py`, etc.

python-avd/tests/pyavd/cv/client/test_cv_client.py
```diff
@@ -26,62 +46,771 @@
...
+@pytest.mark.asyncio
+async def test_get_inventory_devices_with_filter(cv_client: CVClient) -> None:
+    result = await cv_client.get_inventory_devices([(None, None, "avd-ci-spine1")])
+    assert len(result) == 1
+    assert hasattr(result[0], "hostname")
+    assert result[0].hostname == "avd-ci-spine1"
+
+
+@pytest.mark.asyncio
+async def test_verify_devices_on_cv_no_devices(cv_client: CVClient) -> None:
```
carl-baillargeon: These are workflow tests so they should be in their own modules under `workflow/` tests package.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -147,11 +147,16 @@ async def deploy_to_cv(
...
 
-            # Form the list of targeted CVDevices (list may contain duplicated items)
-            devices = (
-                [tag.device for tag in device_tags if tag.device is not None]
-                + [tag.device for tag in interface_tags if tag.device is not None]
-                + [config.device for config in configs if config.device is not None]
+            # Form deduplicated list of targeted CVDevices
+            devices = list(
+                {
+                    id(device): device
```
carl-baillargeon, line 157: If we are deduplicating here, do we still need to dedup in `verify_device_inputs`?

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -161,7 +166,7 @@ async def deploy_to_cv(
                 # Depending on skip_missing_devices we will raise or skip missing devices.
                 # Since verify_devices will silently return if _exists_on_cv is already set,
                 # we can just send all the items even if we have duplicate device objects.
```
carl-baillargeon: This is inaccurate now that we are deduping above right?

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -161,7 +166,7 @@ async def deploy_to_cv(
...
-                await verify_devices_on_cv(
+                existing_devices = await verify_devices_on_cv(
```
carl-baillargeon: Do we absolutely have to change the function signature here? Can't we just pass `devices` from line 151 to `finalize_workspace_on_cv` below and build the list of inactive devices by checking both _streaming and _exists_on_cv properties?

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Fixed. Moved old inventory tests to the separate test module

python-avd/tests/pyavd/cv/client/test_cv_client.py
alexeygorbunov: Fixed. Moved all tests to the proper test modules

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
alexeygorbunov: Yes, `verify_device_inputs` is still required as it catches use cases  where the same serial/mac is used for multiple different devices. This would not be caught in `deploy_to_cv.py`

alexeygorbunov: Right, should not be needed. Removed the following

```
    # Deduplicate CVDevice objects as original `devices` list may contain duplicated items
    unique_devices = list({id(device): device for device in devices}.values())
```

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
alexeygorbunov: Yes, at this point there shouldn't be any duplicated devices. Removed the following

```
                # Since verify_devices will silently return if _exists_on_cv is already set,
                # we can just send all the items even if we have duplicate device objects.
```

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
alexeygorbunov: Indeed, not really needed and is just a waste of cycles. Should be fixed now

python-avd/pyavd/_cv/workflows/finalize_workspace_on_cv.py
```diff
@@ -65,8 +65,31 @@ async def finalize_workspace_on_cv(workspace: CVWorkspace, cv_client: CVClient)
             workspace_id=workspace.id,
             request_id=workspace_config.request_params.request_id,
         )
+        # Form a list of known inactive existing devices
+        if inactive_devices := [f"{device.hostname} ({device.serial_number})" for device in devices if not device._streaming]:
```
carl-baillargeon: Last comment before I approve.

Do we need to also check if the device exists on CV here?

Let's be extra clear on the condition here to avoid confusion with typing (since `device._streaming` can be None):
```suggestion
        if inactive_devices := [f"{device.hostname} ({device.serial_number})" for device in devices if device._streaming is False]:
```
See if we need to check `device._exists_on_cv` as well.

Thanks

alexeygorbunov: Thanks for catching! Yes, since we are not returning an explicit list of existing devices (and use original `devices` instead) - streaming status can now be None as well and needs a more precise matching.
Fixed now.

carl-baillargeon (approved):
LGTM.

python-avd/pyavd/_cv/client/exceptions.py
```diff
@@ -29,7 +29,15 @@ class CVWorkspaceBuildFailed(CVClientException):
...
-    """Build of CloudVision Workspace failed."""
+    """Failed to submit CloudVision Workspace."""
+
+
+class CVWorkspaceSubmitFailedInactiveDevices(CVClientException):
+    """
+    Failed to submit CloudVision Workspace due to the presence of inactive devices.
+
+    Use `cv_submit_workspace_force: true` (if using role `cv_deploy`) or `workspace['force']: true` (if using action plugin `cv_workflow`) to override.
+    """
```
ClausHolbechArista:
```suggestion
    """Failed to submit CloudVision Workspace due to the presence of inactive devices."""
```

ClausHolbechArista (approved):

PR #5009, opened 2025-02-10
Feat(eos_designs): Add support for RSS interface profile for select platforms

### LINKED ISSUE #4983: Support configuration of Receive Side Scaling (RSS) interface profiles for select SFE based platforms
### Enhancement summary

Allow configuration of RSS interface profiles on SFE (Software Forwarding Engine) based hardware platforms such as
AWE-5310, AWE-7230R
AWE-5510, AWE-7250R

This feature is not supported on other SFE based hardware platforms or virtualized hardware platforms such as KVM, CloudEOS on AWS, etc.

eos_cli_config_gen, eos_designs

### Use case example

Target CLI below:
Configure a list of RSS profiles
Allow selection of one such profile to be applied to the system

===
platform sfe interface
   interface profile amsProf1
   !
   profile amsProf1
      interface Ethernet1/1
         rx-queue count 5
         rx-queue worker 1-3,6,9
      !
      interface Ethernet1/10
         rx-queue count 2
      !
      interface Ethernet1/16
   !
   profile amsProf2
      interface Ethernet1/5
         rx-queue count 3
         rx-queue worker 5
         rx-queue mode exclusive
!
===

Allow end-user to configure one or more RSS profiles and hint which profile needs to be applied

Did not explore this.
We need to add support on both eos_designs and eos_cli_config_gen roles for this.

### Additional context

_No response_

### Contributing Guide

ashenoy-arista:
## Change Summary

Build schema for eos_designs role to help generate structured config for RSS interface profile.
RSS interface profile is only applicable to certain SFE (Software Forwarding Engine) hardware platforms.
We would like to have toggle/knob under  `platform_settings` -> `feature_support` schema that end-user can enable for supported hardware platforms to allow generation of relevant structured config.

## Related Issue(s)

Fixes #4983 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Propose adding `rx_queue` dictionary type under L3 interface and member eth interfaces of L3 Port-Channel interface.
The keys within this dictionary are `count`, `workers` and `mode`.
Add a knob under  `platform_settings` -> `feature_support` schema. When enabled, underlying logic would generate structured config for RSS interface profile for devices capable of supporting this feature. Such profile would include sections for each interface for which `rx_queue` dictionary type has been specified/populated in yml.
There will be only one such RSS interface profile generated and same profile would be set as one to apply for the system.

## How to test

Added molecule test  `molecule/eos_designs_unit_tests/inventory/host_vars/sfe-interface-profile-options.yml` to validate the generated structured config and EOS cli config.
Also added a few negative tests to validate invalid configuration inputs.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -113,6 +113,10 @@ keys:
                 Do not advertise reachability to a prefix until that prefix has been installed in hardware.
                 This will eliminate any temporary black holes due to a BGP speaker advertising reachability to a prefix that may not yet be installed into the forwarding plane.
                 Can be overridden by setting "bgp_update_wait_for_convergence" host/group_vars.
+            build_rss_profile:
```
ClausHolbechArista:
```suggestion
            platform_sfe_interface_profile:
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -113,6 +113,10 @@ keys:
...
+              type: bool
+              default: false
+              description: Build RSS interface profile for supported platforms.
```
ClausHolbechArista:
```suggestion
              description: Support for Platform SFE Interface Profiles used for Receive Side Scaling (RSS).
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_port_channels.schema.yml
```diff
@@ -68,6 +68,30 @@ $defs:
                 type: str
                 description: |-
                   Speed should be set in the format `<interface_speed>` or `forced <interface_speed>` or `auto <interface_speed>`.
+              rx_queue:
```
ClausHolbechArista, line 71: Did someone test that this works for port-channel interfaces?

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_interfaces.schema.yml
```diff
@@ -169,6 +169,30 @@ $defs:
                     description: |-
                       Number range to use for Tunnel interfaces to an internet-exit service provider using this local interface.
                       Examples: '1-3' or '100,200,300'
+        rx_queue:
+          type: dict
+          description: |-
+            Receive queue parameters for Receive Side Scaling (RSS) profile for this interface.
```
ClausHolbechArista:
```suggestion
          description: |-
            Receive queue parameters for Receive Side Scaling (RSS) profile for this interface.
            This setting is ignored unless the `platform_sfe_interface_profile` is set as `true` under `platform_settings.feature_support` for the `platform` set on this device.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_port_channels.schema.yml
```diff
@@ -68,6 +68,30 @@ $defs:
...
+                      Number of receive queues.
+                      The maximum value is platform dependent.
+                  worker:
+                    type: str
+                    description: |-
+                      Worker ids specified as combination of range and/or comma separated values
+                      such as 0-4,7.
+                  mode:
+                    type: str
+                    description: Mode applicable to the workers. Default mode is 'shared'.
```
ClausHolbechArista: Set the default key instead of writing it in description. (for eos_designs we can have default :) )

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ClausHolbechArista: We can implement an automatic max if we wish? So you can enforce the max depending on the platform. It would mean you should change the key under feature_support to be a dict like:

```yaml
...
  feature_support:
    platform_sfe_interface_profile:
      supported: <bool; default=false>
      max_rx_queues: <int; default=??>  # Either we have a default or we can require this key.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -294,6 +298,7 @@ keys:
           bgp_update_wait_install: false
           interface_storm_control: false
           queue_monitor_length_notify: false
+          build_rss_profile: false
```
ClausHolbechArista: this should be enabled for these platforms right?

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ClausHolbechArista: If you decide to add this max per platform, we probably need to split out the default platforms depending on the max.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_port_channels.schema.yml
ashenoy-arista: Checked with team. 
RSS profile is supported only on physical NICs/interfaces.
Logical interfaces such as Tunnel/ sub-interface/port-channel do not apply here.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ashenoy-arista: Thank you for the suggestion.
Have added `platform_sfe_interface_profile` as dictionary and created additional keys

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ashenoy-arista: Have set `supported` to true under respective supported platforms
Had to split existing batch of AWE platforms into 2 groups and then configured applicable values for `max_rx_queues` for each group.
Please see if this schema looks fine.
My understanding is that `supported` when set to `true` will serve as hint to `eos_designs` logic to validate schema inputs and generate structured config, correct?

ashenoy-arista:
Initial logic implementation in eos_designs to validate input yml and generate structured config for RSS profiles.
Molecule unit test has been added. 
Negative unit test cases to be added next.

ashenoy-arista:
Updated few newly added input yml schema keys. 
Added unit tests under `eos_designs_unit_tests` and `eos_designs_negative_unit_tests`.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -507,7 +507,13 @@ def platform(self) -> dict | None:
...
             msg = "For AutoVPN RRs and Pathfinders, 'data_plane_cpu_allocation_max' must be set"
             raise AristaAvdInvalidInputsError(msg)
-
+        # populate RSS interface profile for SFE platform (if supported)
+        if self.shared_utils.is_rss_profile_supported:
+            sfe_rss_profiles = self.shared_utils.get_rss_profiles
+            if sfe_rss_profiles:
+                if "sfe" not in platform:
+                    platform["sfe"] = {}
+                platform["sfe"]["interface"] = sfe_rss_profiles
```
```suggestion
        if self.shared_utils.is_rss_profile_supported and (sfe_rss_profiles := self.shared_utils.get_rss_profiles):
            platform.setdefault("sfe", {})["interface"] = sfe_rss_profiles
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -322,3 +322,127 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+        if l3_member_interface._get("rx_queue"):
+            # validate rx_queue 'count' when specified
+            if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.max_rss_queues:
+                msg = f"rx-queue count for interface '{l3_member_interface.name}' exceeds maximum supported '{self.max_rss_queues}' for this platform."
+                raise AristaAvdInvalidInputsError(msg)
+            intf_with_rss_settings["name"] = l3_member_interface.name
+            rx_queue_worker_cum = ""
+            for _worker in l3_member_interface.rx_queue.worker:
+                # validate worker ids for format and permissible values
+                is_range = "-" in _worker
```
ClausHolbechArista: Please simplify all this with our filter `pyavd.j2filters.range_expand` which returns a list of strings with the expanded range. It returns a list with a single item if the input is not a range.

ClausHolbechArista: Similarly we also have a compress filter that takes a list of ints and create a range string.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/shared_utils/misc.py
ashenoy-arista: Thanks for the pointers.
Updated logic with `range_expand` and `list_compress` to build required worker string

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
         l3_bgp_neighbors = self.get_l3_generic_interface_bgp_neighbors(self.l3_interfaces)
         l3_bgp_neighbors.extend(self.get_l3_generic_interface_bgp_neighbors(self.node_config.l3_port_channels))
         return l3_bgp_neighbors
+
+    @cached_property
+    def is_rss_profile_supported(self: SharedUtilsProtocol) -> bool:
+        """Returns bool indicating whether Receive Side Scaling (RSS) profile is supported for this platform."""
+        return self.platform_settings.feature_support.platform_sfe_interface_profile.supported
+
+    @cached_property
```
ClausHolbechArista, line 373: let's use "SFE interface profile" instead of "RSS profile" everywhere. This could be used for other things than RSS in the future.

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+                        new_dict=intf_with_rss,
+                        context="Port-Channel member interface defined with RSS settings",
+                        context_keys=["name"],
+                        ignore_same_dict=False,
+                    )
+        if rss_profile_member_interfaces:
+            return rss_profile_member_interfaces
+        return None
+
+    def build_interface_with_rx_queue_settings(
```
ClausHolbechArista: I think all these helper methods could live in a `UtilsMixin` for the base structured config module. Remember to prefix them with underscore if they are cached properties. (They are not shared when they are not used across multiple modules).

You can also choose to split out the `platform` part into it's own mixin and keep the helpers there if that is better?

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -113,6 +113,21 @@ keys:
...
+              description: Support for Platform SFE Interface Profiles used for Receive Side Scaling (RSS).
+              keys:
+                supported:
+                  type: bool
+                  default: false
+                  description: Capability flag for generation of RSS profiles.
+                max_rx_queues:
+                  type: int
+                  required: true
+                  default: 6
```
ClausHolbechArista: required and default should not be combined.

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+            for _worker in l3_member_interface.rx_queue.worker:
+                for worker_id in range_expand(_worker):
+                    if int(worker_id) >= self.max_rss_queues:
+                        msg = (
+                            f"One or more worker ids within '{_worker}' under 'rx_queue' for interface '{l3_member_interface.name}' "
+                            f"equal or exceed maximum supported '{self.max_rss_queues}' for this platform."
+                        )
+                        raise AristaAvdInvalidInputsError(msg)
+                    rx_queue_worker_cum_set.add(int(worker_id))
+            rx_queue_worker_cum = list_compress(sorted(rx_queue_worker_cum_set))
```
ClausHolbechArista: it is getting sorted inside list_compress, but we need it to be a list.
```suggestion
            rx_queue_workers_range = list_compress(list(rx_queue_workers))
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+    ) -> dict | None:
+        """Returns interface dictionary with RSS settings."""
+        intf_with_rss_settings = {}
+        if l3_member_interface._get("rx_queue"):
+            # validate rx_queue 'count' when specified
+            if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.max_rss_queues:
+                msg = f"'rx_queue' count for interface '{l3_member_interface.name}' exceeds maximum supported '{self.max_rss_queues}' for this platform."
+                raise AristaAvdInvalidInputsError(msg)
+            intf_with_rss_settings["name"] = l3_member_interface.name
+            rx_queue_worker_cum_set = set()
```
ClausHolbechArista:
```suggestion
            rx_queue_workers = set()
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+            for _worker in l3_member_interface.rx_queue.worker:
+                for worker_id in range_expand(_worker):
+                    if int(worker_id) >= self.max_rss_queues:
+                        msg = (
+                            f"One or more worker ids within '{_worker}' under 'rx_queue' for interface '{l3_member_interface.name}' "
+                            f"equal or exceed maximum supported '{self.max_rss_queues}' for this platform."
+                        )
+                        raise AristaAvdInvalidInputsError(msg)
+                    rx_queue_worker_cum_set.add(int(worker_id))
```
ClausHolbechArista:
```suggestion
                    rx_queue_workers.add(int(worker_id))
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -364,3 +364,107 @@ def l3_bgp_neighbors(self: SharedUtilsProtocol) -> list:
...
+            rx_queue_worker_cum = list_compress(sorted(rx_queue_worker_cum_set))
+            rx_queue = {
+                "count": l3_member_interface.rx_queue.count,
+                "worker": rx_queue_worker_cum,
```
ClausHolbechArista:
```suggestion
                "worker": rx_queue_workers_range,
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
ashenoy-arista: Made changes to names of methods/attributes to keep this generic

python-avd/pyavd/_eos_designs/shared_utils/misc.py
ashenoy-arista: Added helper methods to PlatformMixin class
Tried moving existing platform() method from structured_config/base/__init__.py to this class.
But it was not generating `platform` key in yml files for structured config (regardless of decorator `cached_property` or `structured_config_contributor`)
Hence did not move platform() cached property to PlatformMixin class.
Please let me know if this method can be moved and what should be the decorator.

ashenoy-arista: Have successfully moved `platform` part into its own mixin class and able to generate `platform` key using `structured_config_contributor` decorator for this method in the newly added class.
Thank you for the pointers.

ashenoy-arista: Also made changes to comments/method names/attributes to avoid referring to `RSS` for such profile.
Instead we refer to this as SFE interface profile.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_interfaces.schema.yml
```diff
@@ -169,6 +169,36 @@ $defs:
                     description: |-
                       Number range to use for Tunnel interfaces to an internet-exit service provider using this local interface.
                       Examples: '1-3' or '100,200,300'
+        rx_queue:
```
gmuloc, line 190: Maybe we can use a def / ref here to avoid repeating this?

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Mixin Class used to generate structured config for one key.
+
+    Class should only be used as Mixin to a AvdStructuredConfig class or other Mixins.
+    """
+
+    @structured_config_contributor
+    def platform(self: AvdStructuredConfigBaseProtocol) -> None:
+        """
+        Platform set based on platform data-model, using various platform related information.
```
gmuloc:
```suggestion
        Set platform keys in structured config based on platform data-model, using various platform-related information.
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+
+        platform set is based on the following:
+
```
gmuloc:
```suggestion
        The following settings are used: 

```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        # populate interface profile for SFE platform (if supported)
+        if self.shared_utils.is_sfe_interface_profile_supported and (sfe_member_interfaces_for_profile := self._get_sfe_interface_profile_member_interfaces):
+            default_sfe_interface_profile_name = "Default_Interface_Profile"
+            # build single ProfilesItem and append to Profiles
+            profiles_item = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem(
+                name=default_sfe_interface_profile_name,
+                interfaces=sfe_member_interfaces_for_profile,
+            )
+            self.structured_config.platform.sfe.interface.profiles.append(profiles_item)
```
gmuloc: we can just use append_new as you don't modify afterwards
```suggestion
            default_sfe_interface_profile_name = "Default_Interface_Profile"
            # build single ProfilesItem and append to Profiles
            self.structured_config.platform.sfe.interface.profiles.append_new(
                            name=default_sfe_interface_profile_name,
                interfaces=sfe_member_interfaces_for_profile,
            )
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+            # indicate above profile as the one to apply for the device
+            self.structured_config.platform.sfe.interface.interface_profile = default_sfe_interface_profile_name
+
+    @cached_property
+    def _get_sfe_interface_profile_member_interfaces(
+        self: AvdStructuredConfigBaseProtocol,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.Interfaces | None:
+        """Returns list of eligible SFE interfaces with profile settings for structured config."""
+        # Iterate thru all L3 interfaces checking for those with 'rx_queue' config
```
gmuloc:
```suggestion
        # Iterate through all L3 interfaces checking for those with 'rx_queue' config
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        # Also iterate thru member interfaces of all L3 Port-Channels with 'rx_queue' config
```
gmuloc:
```suggestion
        # Also iterate through member interfaces of all L3 Port-Channels with 'rx_queue' config
```

gmuloc: same for others

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+                interfaces=sfe_member_interfaces_for_profile,
+            )
+            self.structured_config.platform.sfe.interface.profiles.append(profiles_item)
+            # indicate above profile as the one to apply for the device
+            self.structured_config.platform.sfe.interface.interface_profile = default_sfe_interface_profile_name
+
+    @cached_property
+    def _get_sfe_interface_profile_member_interfaces(
+        self: AvdStructuredConfigBaseProtocol,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.Interfaces | None:
```
gmuloc: can you return None? seems like you just return an empty object

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+            return sfe_profile_member_interfaces
+        return None
+
+    def _build_interface_with_rx_queue_settings(
+        self: AvdStructuredConfigBaseProtocol,
+        l3_member_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
+        | EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3PortChannelsItem.MemberInterfacesItem,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem | None:
+        """Returns one SFE interface with profile settings for structured config."""
+        if l3_member_interface._get("rx_queue"):
```
gmuloc: We prefer returning fast + you should be able to use it like this:
```suggestion
        if not l3_member_interface.rx_queue:
            return
        [...] code
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+            intf_profile_settings = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem()
+            # validate rx_queue 'count' when specified
+            if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
```
gmuloc:
```suggestion
            if l3_member_interface.rx_queue.count and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+                            f"equal or exceed maximum supported '{self.shared_utils.max_rx_queues}' for this platform."
+                        )
+                        raise AristaAvdInvalidInputsError(msg)
+                    rx_queue_workers.add(int(worker_id))
+            rx_queue_workers_range = list_compress(list(rx_queue_workers))
+            intf_profile_settings.rx_queue.count = l3_member_interface.rx_queue.count
+            if rx_queue_workers_range:
+                intf_profile_settings.rx_queue.worker = rx_queue_workers_range
+            if l3_member_interface.rx_queue._get("mode"):
+                intf_profile_settings.rx_queue.mode = l3_member_interface.rx_queue.mode
```
gmuloc: I think you can just do this as we will strip empty values
```suggestion
            intf_profile_settings.rx_queue.mode = l3_member_interface.rx_queue.mode
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+            rx_queue_workers = set()
+            for _worker in l3_member_interface.rx_queue.worker:
+                for worker_id in range_expand(_worker):
+                    if int(worker_id) >= self.shared_utils.max_rx_queues:
+                        msg = (
+                            f"One or more worker ids within '{_worker}' under 'rx_queue' for interface '{l3_member_interface.name}' "
+                            f"equal or exceed maximum supported '{self.shared_utils.max_rx_queues}' for this platform."
+                        )
+                        raise AristaAvdInvalidInputsError(msg)
+                    rx_queue_workers.add(int(worker_id))
```
gmuloc: I think you can create a helper like _validate_rx_queue_worker that just raise if any is wrong - you don't need the set as compress would take care of this

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,129 @@
...
+        return None
+
+    def _build_interface_with_rx_queue_settings(
+        self: AvdStructuredConfigBaseProtocol,
+        l3_member_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
+        | EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3PortChannelsItem.MemberInterfacesItem,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem | None:
+        """Returns one SFE interface with profile settings for structured config."""
+        if l3_member_interface._get("rx_queue"):
+            intf_profile_settings = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem()
```
gmuloc: don't create it here but after the validations I think that will help making it cleaner

so 1/ check count 2/ validate the rx_queue_workers (maybe it returns to you the compress list) 3/ create the object with all the values

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ClausHolbechArista: I think compress will need it to be an int, so IMO since we have to go over them it makes sense to dedup with a set.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type_l3_interfaces.schema.yml
ashenoy-arista: Good suggestion.
Added `defs_rx_queue_settings.schema.yml` under schema_fragments for the same
We reference this in `defs_node_type_l3_interfaces.schema.yml` and `defs_node_type_l3_port_channels.schema.yml`

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Updated logic to return None when `self.shared_utils.is_sfe_interface_profile_supported` is false instead of empty object.

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Tried this, we need the conditional check `if l3_member_interface.rx_queue._get("mode")` since `mode` has default value = "shared" as per schema definition.
If we dont include conditional check above, we will *always* set `mode` as "shared" in each member interface with "rx_queue" set. We intend to set value for `mode` attribute only when it is explicitly set in input yml.
Hence preserving this conditional check for `mode`.

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Have added helper wrapper for validation _validate_rx_queue_worker () as suggested.
Have maintained existing logic that adds int(worker_id) to `set` to populate unique worker ids across any overlapping ranges(if any). Next we call list_compress() which works on list of int worker_ids retrieved from `set`.
Let me know if any change is needed here.

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Thank you for the suggestion. Cleaned up this logic

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ClausHolbechArista: Then please remove the default value from the schema if that is really the behavior you are looking for.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_rx_queue_settings.schema.yml
```diff
@@ -0,0 +1,38 @@
...
+          type: str
+          convert_types:
+            - int
+          description: |-
+            Worker ids specified as values or range of values such as 0-4 or 7.
+            Valid values are between 0 and one less than maximum value determined by `platform_sfe_interface_profile.max_rx_queues` under `platform_settings.feature_support` for the `platform` set on this device.
+      mode:
+        type: str
+        description: Mode applicable to the workers.
+        default: shared
```
ClausHolbechArista: Let's remove this so we don't have to dance around it in the code.
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+        l3_member_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
+        | EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3PortChannelsItem.MemberInterfacesItem,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem | None:
+        """Returns one SFE interface with profile settings for structured config."""
+        if not l3_member_interface._get("rx_queue"):
+            # specified interface does not have explicit "rx_queue" key or sub-keys specified
+            # exclude such interface from profile being built
+            return None
+        # validate rx_queue 'count' when specified
+        if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
```
ClausHolbechArista:
```suggestion
        if l3_member_interface.rx_queue.count and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+            return sfe_profile_member_interfaces._natural_sorted()
+        return None
+
+    def _build_interface_with_rx_queue_settings(
+        self: AvdStructuredConfigBaseProtocol,
+        l3_member_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
+        | EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3PortChannelsItem.MemberInterfacesItem,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem | None:
+        """Returns one SFE interface with profile settings for structured config."""
+        if not l3_member_interface._get("rx_queue"):
```
ClausHolbechArista:
```suggestion
        if not l3_member_interface.rx_queue:
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+        # validate rx_queue 'count' when specified
+        if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
+            msg = (
+                f"'rx_queue' count for interface '{l3_member_interface.name}' exceeds maximum supported '{self.shared_utils.max_rx_queues}' for this platform."
+            )
+            raise AristaAvdInvalidInputsError(msg)
+
+        rx_queue_workers = set()
+        for _worker in l3_member_interface.rx_queue.worker:
+            for worker_id in range_expand(_worker):
```
ClausHolbechArista: Remove the int() in mutliple places and do it once here instead.
```suggestion
            for worker_id in map(int, range_expand(_worker)):
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+            return None
+        # validate rx_queue 'count' when specified
+        if l3_member_interface.rx_queue._get("count") and l3_member_interface.rx_queue.count > self.shared_utils.max_rx_queues:
+            msg = (
+                f"'rx_queue' count for interface '{l3_member_interface.name}' exceeds maximum supported '{self.shared_utils.max_rx_queues}' for this platform."
+            )
+            raise AristaAvdInvalidInputsError(msg)
+
+        rx_queue_workers = set()
+        for _worker in l3_member_interface.rx_queue.worker:
```
ClausHolbechArista: Maybe rx_queue.worker should be plural instead since we have multiple workers?

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ClausHolbechArista: FYI Range expand supports a list as input, and it will expand both the list and the string ranges inside recursively.

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,132 @@
...
+            intf_profile_settings.rx_queue.mode = l3_member_interface.rx_queue.mode
+        return intf_profile_settings
+
+    def _validate_rx_queue_worker(self: AvdStructuredConfigBaseProtocol, worker_id: str, worker_range: str, l3_member_interface_name: str) -> None:
+        if int(worker_id) >= self.shared_utils.max_rx_queues:
+            msg = (
+                f"One or more worker ids within '{worker_range}' under 'rx_queue' for interface '{l3_member_interface_name}' "
+                f"equal or exceed maximum supported '{self.shared_utils.max_rx_queues}' for this platform."
+            )
+            raise AristaAvdInvalidInputsError(msg)
```
ClausHolbechArista: I think it is enough to just tell which worker number was above the limit and not care about how it was given.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Made the suggested changes
Without conditional check we see `rx_queue.worker` set as empty string '' when `worker` is not specified in input yml
Wondering why strip empty not able to remove empty string set for `worker` for the generated structured config? 
```
if rx_queue_workers_range:
            intf_profile_settings.rx_queue.worker = rx_queue_workers_range
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Added separate wrapper for validating rx_queue.count - _validate_rx_queue_count() to keep it consistent

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Made change to schema for this `rx_queue.workers`

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+            self.structured_config.platform.sfe.data_plane_cpu_allocation_max = cpu_max_allocation
+        elif self.shared_utils.is_wan_server:
+            # For AutoVPN Route Reflectors and Pathfinders, running on CloudEOS, setting
+            # this value is required for the solution to work.
+            msg = "For AutoVPN RRs and Pathfinders, 'data_plane_cpu_allocation_max' must be set"
+            raise AristaAvdInvalidInputsError(msg)
+
+        # populate interface profile for SFE platform (if supported)
+        if self.shared_utils.is_sfe_interface_profile_supported and (sfe_member_interfaces_for_profile := self._get_sfe_interface_profile_member_interfaces):
+            default_sfe_interface_profile_name = "Default_Interface_Profile"
```
gmuloc: we should use upper case here to be consistent with how we define AVD objects

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @cached_property
+    def _get_sfe_interface_profile_member_interfaces(
+        self: AvdStructuredConfigBaseProtocol,
+    ) -> EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.Interfaces | None:
+        """Returns list of eligible SFE interfaces with profile settings for structured config."""
+        # Iterate through all L3 interfaces checking for those with 'rx_queue' config
+        # Also iterate through member interfaces of all L3 Port-Channels with 'rx_queue' config
+        if not self.shared_utils.is_sfe_interface_profile_supported:
+            return None
```
gmuloc: I dont think this check is needed as it should never be called if this is not true (the way it is used today, you can specify in docstring that the caller is in charge of checking this)

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+        sfe_profile_member_interfaces = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.Interfaces()
```
gmuloc, line 77: I think if you return an empty list for this then it should be considered as false and you can avoid returning None (need to be checked)

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+                rx_queue_workers.add(worker_id)
+        rx_queue_workers_range = list_compress(list(rx_queue_workers))
+
+        intf_profile_settings = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem()
+        intf_profile_settings.name = l3_member_interface.name
+        intf_profile_settings.rx_queue.count = l3_member_interface.rx_queue.count
+        if rx_queue_workers_range:
+            intf_profile_settings.rx_queue.worker = rx_queue_workers_range
+        intf_profile_settings.rx_queue.mode = l3_member_interface.rx_queue.mode
+        return intf_profile_settings
```
gmuloc: Can it not be created like this?
```suggestion
        return EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem(
            name = l3_member_interface.name.
            rx_queue.count = l3_member_interface.rx_queue.count
            rx_queue = EosCliConfigGen.Platform.Sfe.Interface.ProfilesItem.InterfacesItem.RxQueue(
                worker = rx_queue_workers_range
                mode = l3_member_interface.rx_queue.mode
            )
        )
```

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,133 @@
...
+                if intf_profile_settings is not None:
+                    sfe_profile_member_interfaces.append(intf_profile_settings)
+
+        if sfe_profile_member_interfaces:
+            return sfe_profile_member_interfaces._natural_sorted()
+        return None
+
+    def _build_interface_with_rx_queue_settings(
+        self: AvdStructuredConfigBaseProtocol,
+        l3_member_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
```
gmuloc: do we want to call this l3_member_interface? it feels like it should just be "l3_ethernet_interface` maybe?

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
ashenoy-arista: Resolving as per conversation with Claus.
Empty strings are not stripped by helper utility wrapper for attributes of structured config

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,139 @@
...
+        elif self.shared_utils.is_wan_server:
+            # For AutoVPN Route Reflectors and Pathfinders, running on CloudEOS, setting
+            # this value is required for the solution to work.
+            msg = "For AutoVPN RRs and Pathfinders, 'data_plane_cpu_allocation_max' must be set"
+            raise AristaAvdInvalidInputsError(msg)
+
+        # populate interface profile for SFE platform (if supported)
+        if self.shared_utils.is_sfe_interface_profile_supported and (sfe_member_interfaces_for_profile := self._get_sfe_interface_profile_member_interfaces):
+            # build single ProfilesItem and append to Profiles
+            default_sfe_interface_profile_name = "DEFAULT-INTERFACE-PROFILE"
```
ClausHolbechArista: @gmuloc any strong feelings on naming here?

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
```diff
@@ -0,0 +1,139 @@
...
+            return None
+        # validate rx_queue 'count' when specified
+        self._validate_rx_queue_count(l3_ethernet_interface.rx_queue.count, l3_ethernet_interface.name)
+
+        rx_queue_workers = set()
+        for _worker in l3_ethernet_interface.rx_queue.workers:
+            for worker_id in map(int, range_expand(_worker)):
+                self._validate_rx_queue_worker(worker_id, l3_ethernet_interface.name)
+                rx_queue_workers.add(worker_id)
+        rx_queue_workers_range = list_compress(list(rx_queue_workers))
```
ClausHolbechArista:
```suggestion
        if rx_queue_workers := set(map(int, range_expand(l3_ethernet_interface.rx_queue.workers)))
            self._validate_rx_queue_worker(max(rx_queue_workers), l3_ethernet_interface.name)
        rx_queue_workers_range = list_compress(list(rx_queue_workers))
```

ClausHolbechArista (commented):
Only a minor optimization and a question for Guillaume. Other than that it LGTM!

python-avd/pyavd/_eos_designs/structured_config/base/platform_mixin.py
gmuloc: nope - I guess we could put some AVD in there in order to make it clear that it does not come from EOS (like the default SSL `ARISTA_DEFAULT_PROFILE`)

ashenoy-arista: Renamed profile as "AVD-DEFAULT-INTERFACE-PROFILE"

ashenoy-arista:
The review suggestions have been addressed.
Kindly approve/merge this change to 'devel' branch
Thank you

gmuloc (approved):

ClausHolbechArista (approved):

PR #5025, opened 2025-02-12
Fix(eos_designs): Wrong duplicate detection between SVIs and L2VLANs

gmuloc:
## Change Summary

The duplicate detection is using SVI / L2VLAN field in the match. This was not caught by our CI.

## Related Issue(s)

Fixes #5022

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- [x] Add failing test in molecule reproducing the issue (first commit)
- [x] Ignore the source_type in data class for comparison and use the real type instead.

## How to test

molecule passes

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #5029, opened 2025-02-12
Fix(eos_designs): Wrong duplicate detection between SVIs and L2VLANs (#5025)

ClausHolbechArista:
Cherry pick of Fix(eos_designs): Wrong duplicate detection between SVIs and L2VLANs (#5025)


carlbuchmann (approved):
LGTM

PR #5048, opened 2025-02-14
CI: Simplify schema tooling constants

ClausHolbechArista:
Group schema constants into a class per schema instead of multiple dicts.
Preparation for adding more schemas.

### DIFF
--- a/python-avd/schema_tools/constants.py
+++ b/python-avd/schema_tools/constants.py
@@ -25,7 +25,7 @@ class SchemaPaths:
     docs_path: Path | None = None
 
 
-# Remember to also updated PICKLED_SCHEMAS in pyavd/_schema/constants.py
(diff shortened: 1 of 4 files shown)


python-avd/schema_tools/constants.py
```diff
@@ -3,47 +3,46 @@
...
+@dataclass(frozen=True)
+class SchemaPaths:
+    yaml_file: Path
+    pickled_schema: Path
+    fragments_dir: Path | None = None
+    python_class: Path | None = None
+    docs_path: Path | None = None
+
+
+# Remember to also updated PICKLED_SCHEMAS in pyavd/_schema/constants.py
```
gmuloc:
```suggestion
# Remember to also update PICKLED_SCHEMAS in pyavd/_schema/constants.py
```

gmuloc (approved):

PR #5054, opened 2025-02-18
Refactor(eos_designs): Structured config classes for underlay ethernet_interfaces

ClausHolbechArista:
## Change Summary

Refactor(eos_designs): Structured config classes for underlay ethernet_interfaces

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Refactor eos_designs underlay ethernet interfaces to classes
- Moving a few helper functions only used for ethernet interfaces into that file
- Refactoring how ACLs are applied to remove several layers of overhead.

## How to test

- No changes to output configs
- A bit of reordering of ipv4_acls in structured config.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -39,173 +39,155 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
+                    ethernet_interface.mac_security.profile = mac_security["profile"]
+
+                if link_tracking_groups := link.get("link_tracking_groups"):
+                    for link_tracking_group in link_tracking_groups:
+                        ethernet_interface.link_tracking_groups.append_new(
+                            name=link_tracking_group["name"],
+                            direction=link_tracking_group["direction"],
+                        )
+                if sflow := link.get("sflow"):
+                    ethernet_interface.sflow.enable = sflow["enable"]
```
gmuloc, line 74: for all these we are sure that the link.foo will always have the keys or could we end up with KeyError in some corner cases?

gmuloc: looking at the code it looks like yes for all 3 but would be nice indeed to have a model

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -39,173 +39,155 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
-
-                        if len(ospf_keys) > 0:
-                            ethernet_interface["ospf_authentication"] = "message-digest"
-                            ethernet_interface["ospf_message_digest_keys"] = ospf_keys
-                        else:
-                            msg = "'underlay_ospf_authentication.enabled' is True but no message-digest keys with both key and ID are defined."
-                            raise AristaAvdError(msg)
+                            )
 
                 if self.shared_utils.underlay_isis is True:
```
gmuloc, line 124: could this become `elif` or can we have both underlays be true?

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -39,173 +39,155 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
                 if structured_config := link.get("structured_config"):
                     self.custom_structured_configs.nested.ethernet_interfaces.obtain(link["interface"])._deepmerge(
                         EosCliConfigGen.EthernetInterfacesItem._from_dict(structured_config), list_merge=self.custom_structured_configs.list_merge_strategy
                     )
 
+                self.structured_config.ethernet_interfaces.append(ethernet_interface)
+
             # L2 interface
             elif link["type"] == "underlay_l2":
+                ethernet_subinterfaces = ()
```
gmuloc: this probably needs a type hint

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -39,173 +39,155 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
+                        flow_tracker=self.shared_utils.new_get_flow_tracker(
+                            link.get("flow_tracking"), output_type=EosCliConfigGen.EthernetInterfacesItem.FlowTracker
+                        ),
+                    )
+                    if link_tracking_groups := link.get("link_tracking_groups"):
+                        for link_tracking_group in link_tracking_groups:
+                            ethernet_interface.link_tracking_groups.append_new(
+                                name=link_tracking_group["name"],
+                                direction=link_tracking_group["direction"],
+                            )
```
gmuloc, line 186: worth alittle helper as we do this in two places

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -39,173 +39,155 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
-            append_if_not_duplicate(
-                list_of_dicts=ethernet_interfaces,
-                primary_key="name",
-                new_dict=ethernet_interface,
-                context="Ethernet Interfaces defined for underlay",
-                context_keys=["name", "peer", "peer_interface"],
-            )
+                self.structured_config.ethernet_interfaces.append(ethernet_interface)
+                if ethernet_subinterfaces:
+                    self.structured_config.ethernet_interfaces.extend(ethernet_subinterfaces)
```
gmuloc: it looks like it could even be moved closer to where you are getting them line 160

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
```diff
@@ -220,44 +202,29 @@ def ethernet_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> list | Non
...
-                        ethernet_subinterface.update({"ip_address": f"{subinterface['ip_address']}/{subinterface['prefix_length']}"})
-
-                    ethernet_subinterface = {key: value for key, value in ethernet_subinterface.items() if value is not None}
-                    append_if_not_duplicate(
-                        list_of_dicts=ethernet_interfaces,
-                        primary_key="name",
-                        new_dict=ethernet_subinterface,
-                        context="Ethernet sub-interfaces defined for underlay",
-                        context_keys=["name", "peer", "peer_interface"],
+                        description=description or None,
```
gmuloc, line 211: do we need `or None` because description can be `""`?

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -236,119 +188,102 @@ def _get_l3_port_channel_cfg(
             raise AristaAvdError(msg)
         return strip_empties_from_dict(interface)
 
+    # These overloads are just here to help the type checker enforce that input type x gives output type y
```
gmuloc, line 191: nice

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -236,119 +188,102 @@ def _get_l3_port_channel_cfg(
...
         l3_generic_interface: EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem
         | EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3PortChannelsItem,
-    ) -> dict:
+    ) -> EosCliConfigGen.EthernetInterfacesItem | EosCliConfigGen.PortChannelInterfacesItem:
         """Returns common structured_configuration for L3 interface or L3 Port-Channel."""
         # variables being set for constructing appropriate validation error
         if isinstance(l3_generic_interface, EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem.L3InterfacesItem):
-            node_type_in_schema = "l3_interfaces"
+            interface = EosCliConfigGen.EthernetInterfacesItem()
+            schema_key = "l3_interfaces"
```
gmuloc, line 213: would be cool to get this from interface!

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: Model will come. This PR is focusing on structured config.

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: It is just a dummy which will be replaced if we have anything to generate.

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: Plan is to replace `link` with a model that resembles eos_cli_config_gen (or maybe even _is_ eos_cli_config_gen) so we can just take it directly.

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: It was put here because of ordering. So main interface is added before the subinterfaces

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: correct, since the user can give any string (even empty) but EOS reacts badly on empty strings.

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
ClausHolbechArista: Once we have source, this is not needed.

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: I don't *think* both can be true, but I wanted to keep all behavior the same here.

python-avd/pyavd/_eos_designs/structured_config/underlay/ethernet_interfaces.py
ClausHolbechArista: later

emilarista (approved):
Tested, no config diffs seen. LGTM!

gmuloc (approved):

PR #5066, opened 2025-02-19
CI: Add coverage for ansible-test units and integration

ClausHolbechArista:
Add coverage for ansible-test units and integration


ClausHolbechArista:
@gmuloc said go go :)

PR #5068, opened 2025-02-19
Fix(eos_designs): Better error message when no ip address configured on a l3_interface on wan_router

gmuloc:
## Change Summary

The code was failing badly in shared_utils when trying to retrieve the interface public IP

## Related Issue(s)

Found while working on #4838 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Ignore silently the missing IP if not on a WAN server and let it fail later.

## How to test

molecule test added

### Repository Checklist


python-avd/pyavd/_eos_designs/shared_utils/wan.py
```diff
@@ -214,17 +214,14 @@ def get_public_ip_for_wan_interface(
...
-        If not a WAN route-server this returns public IP and if not found then the interface IP without a mask.
+        If not a WAN route-server this returns public IP and if not found then the interface IP without a mask or None if no ip is set.
 
         For WAN route-servers we try to find the IP under wan_route_servers.path_groups.interfaces.
         If not found we look for the public_ip and then the ip_address under the interface.
         If there is no public_ip and if ip_address is "dhcp" we raise an error.
         """
-        if not self.is_wan_server:
-            return default(interface.public_ip, get_ip_from_ip_prefix(interface.ip_address))
-
```
ClausHolbechArista, line 227: Removing this means we will run through a lot of extra logic for every non wan server device, just to fall through to the end.

gmuloc: yes but we need the extra logic:

if there is a public_ip we want it
if there is no ip_address or it is dhcp we don't want to call get_ip_from_prefix

and otherwise we want get_ip_from_ip_prefix


the only difference is taht for a wan_server we should raise if we have no public IP or if it is dhcp whereas for a wan interface, we don't need to for dhcp and even if we are missing the ip_address (though we would raise later as per the test if there is no ip_address at all)

ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #5072, opened 2025-02-20
Fix(eos_designs): Do not fail documentation task when missing structured configuration

### LINKED ISSUE #5060: Fabric documentation fails on runs with --limit
Fabric documentation fails on runs with --limit since the structured config files are missing.

Suggestion: Ignore the missing structured config and exclude the device from the fabric doc.

---

### Discussed in https://github.com/aristanetworks/avd/discussions/4886

<div type='discussions-op-text'>

<sup>Originally posted by **chrisgerdener** January 16, 2025</sup>
Hi everyone,
We updated to version 5 yesterday and noticed that it is no longer possible to only render part of the fabric with the new version. For example previously we were able to only create configs for some groups of our fabric, but now the fabric documentation steps fails if doing that because it cannot find structured config files for those groups that are not selected.

Example:

FABRIC:
    GroupA:
    GroupB:
    
Previously, when running a config build only on GroupA, AVD would only build and document switches defined under that group. Now it also tries to document GroupB

Am I missing a setting for the documentation step or is this a bug/our previous usage was not intended?</div>

ClausHolbechArista:
## Change Summary

Fix(eos_designs): Do not fail documentation task when missing structured configuration

## Related Issue(s)

Fixes #5060

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Missed a return from the function after warning about missing structured config file.

## How to test

- Manually
- Added an ansible integration test

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -110,12 +110,22 @@ def main(self, validated_args: dict, task_vars: dict, result: dict) -> dict:
...
+        if missing:
+            LOGGER.warning("Could not find structured config files for '%s'. The documentation may be incomplete.", ",".join(missing))
+
+        return structured_configs
 
     def read_one_structured_config(self, device: str, structured_config_dir: str, structured_config_suffix: str) -> dict:
         path = Path(structured_config_dir, f"{device}.{structured_config_suffix}")
         if not path.exists():
-            logging.warning("Could not find structured config file for '%s'. The documentation may be incomplete.", device)
+            return {}
```
gmuloc, line 128: wow we did not have a return! 🤡

gmuloc (approved):

carlbuchmann:
I would rather skip this task as it is incomplete, and some of the regenerated info is incorrect. I'm concerned as it could generate misleading information.

Suggested message: ```[SKIPPED]: Could not find structured config files for [device list]'. Fabric documentation generation was skipped!```

carlbuchmann (changes requested):
I would rather skip this task as it is incomplete, and some of the regenerated info is incorrect. I'm concerned as it could generate misleading information.

Suggested message: ```[SKIPPED]: Could not find structured config files for [device list]'. Fabric documentation generation was skipped!```

carlbuchmann (approved):
Approve as to keep existing behavior.

PR #5078, opened 2025-02-20
CI: Run code testing workflow on pushes to devel

ClausHolbechArista:
CI: Run code testing workflow on pushes to devel


ClausHolbechArista:
@gmuloc maybe we need to limit all jobs here to only run on our repo?

PR #5117, opened 2025-02-28
Refactor(eos_designs): Refactor eos_designs structured_config code for router_bgp(overlay)

MaheshGSLAB:
## Change Summary

Refactor(eos_designs): Refactor eos_designs structured_config code for router_bgp(overlay)

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Refactor(eos_designs): Refactor eos_designs structured_config code for router_bgp(overlay)

## How to test
Run Molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
+++ b/python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
@@ -89,7 +89,7 @@ class RouterBgpMixin(Protocol):
         )
 
     def _set_peer_groups(self: AvdStructuredConfigOverlayProtocol) -> None:
-        peer_groups = EosCliConfigGen.RouterBgp.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.peer_groups
         if self.shared_utils.overlay_routing_protocol == "ebgp":
             # EVPN OVERLAY peer group
             ebgp_overlay_peer_group = self._generate_base_peer_group("evpn", "evpn_overlay_peers")
@@ -168,11 +168,9 @@ class RouterBgpMixin(Protocol):
             )
             peer_groups.append(ipvpn_gateway_peer_group)
 
-        self.structured_config.router_bgp.peer_groups = peer_groups
-
     def _set_address_family_ipv4(self: AvdStructuredConfigOverlayProtocol) -> None:
         """Deactivate the relevant peer_groups in address_family_ipv4."""
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_ipv4.peer_groups
         if self.shared_utils.is_wan_router:
             peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_overlay_peers.name, activate=False)
             if self._is_wan_server_with_peers:
@@ -197,10 +195,8 @@ class RouterBgpMixin(Protocol):
         if self.shared_utils.overlay_ipvpn_gateway is True:
             peer_groups.append_new(name=self.inputs.bgp_peer_groups.ipvpn_gateway_peers.name, activate=False)
 
-        self.structured_config.router_bgp.address_family_ipv4.peer_groups = peer_groups
-
     def _set_address_family_evpn(self: AvdStructuredConfigOverlayProtocol) -> None:
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyEvpn.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_evpn.peer_groups
         overlay_peer_group = EosCliConfigGen.RouterBgp.AddressFamilyEvpn.PeerGroupsItem()
         if self.shared_utils.is_wan_router:
             wan_overlay_peer_group = EosCliConfigGen.RouterBgp.AddressFamilyEvpn.PeerGroupsItem(
@@ -285,27 +281,24 @@ class RouterBgpMixin(Protocol):
                 encapsulation=self.inputs.wan_encapsulation,
             )
 
-        self.structured_config.router_bgp.address_family_evpn.peer_groups = peer_groups
-
     def _set_address_family_ipv4_sr_te(self: AvdStructuredConfigOverlayProtocol) -> None:
         """Set the structured config for IPv4 SR-TE address family."""
         if not self.shared_utils.is_cv_pathfinder_router:
             return
 
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyIpv4SrTe.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_ipv4_sr_te.peer_groups
 
         peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_overlay_peers.name, activate=True)
 
         if self._is_wan_server_with_peers:
             peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_rr_overlay_peers.name, activate=True)
-        self.structured_config.router_bgp.address_family_ipv4_sr_te.peer_groups = peer_groups
 
     def _set_address_family_link_state(self: AvdStructuredConfigOverlayProtocol) -> None:
         """Set the structured config for link-state address family."""
         if not self.shared_utils.is_cv_pathfinder_router:
             return
 
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyLinkState.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_link_state.peer_groups
         peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_overlay_peers.name, activate=True)
 
         if self.shared_utils.is_cv_pathfinder_server:
@@ -317,20 +310,17 @@ class RouterBgpMixin(Protocol):
         if self._is_wan_server_with_peers:
             peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_rr_overlay_peers.name, activate=True)
 
-        self.structured_config.router_bgp.address_family_link_state.peer_groups = peer_groups
-
     def _set_address_family_path_selection(self: AvdStructuredConfigOverlayProtocol) -> None:
         if not self.shared_utils.is_wan_router:
             return
 
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyPathSelection.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_path_selection.peer_groups
         peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_overlay_peers.name, activate=True)
 
         self.structured_config.router_bgp.address_family_path_selection.bgp.additional_paths._update(receive=True, send="any")
 
         if self._is_wan_server_with_peers:
             peer_groups.append_new(name=self.inputs.bgp_peer_groups.wan_rr_overlay_peers.name, activate=True)
-        self.structured_config.router_bgp.address_family_path_selection.peer_groups = peer_groups
 
     def _set_address_family_rtc(self: AvdStructuredConfigOverlayProtocol) -> None:
         """
@@ -340,7 +330,7 @@ class RouterBgpMixin(Protocol):
         """
         if not self.inputs.evpn_overlay_bgp_rtc:
             return
-        peer_groups = EosCliConfigGen.RouterBgp.AddressFamilyRtc.PeerGroups()
+        peer_groups = self.structured_config.router_bgp.address_family_rtc.peer_groups
         peer_groups.append_new(name=self.inputs.bgp_peer_groups.evpn_overlay_peers.name)
         if self.shared_utils.overlay_evpn_vxlan is True:
             peer_groups.obtain(self.inputs.bgp_peer_groups.evpn_overlay_peers.name).activate = True
@@ -365,8 +355,6 @@ class RouterBgpMixin(Protocol):
 
             if self.shared_utils.overlay_evpn_vxlan is True and (self.shared_utils.evpn_role == "server" or self.shared_utils.mpls_overlay_role == "server"):
                 peer_groups.obtain(self.inputs.bgp_peer_groups.evpn_overlay_peers.name).default_route_target.only = True
-
-        self.structured_config.router_bgp.address_family_rtc.peer_groups = peer_groups
 
     def _set_address_family_vpn_ipvx(self: AvdStructuredConfigOverlayProtocol, version: int) -> None:
         if version not in [4, 6]:
@@ -422,7 +410,7 @@ class RouterBgpMixin(Protocol):
         return neighbor
 
     def _set_neighbors(self: AvdStructuredConfigOverlayProtocol) -> None:
-        neighbors = EosCliConfigGen.RouterBgp.Neighbors()
+        neighbors = self.structured_config.router_bgp.neighbors
         if self.shared_utils.overlay_routing_protocol == "ebgp":
             for route_server, data in natural_sort(self._evpn_route_servers.items()):
                 neighbor = self._create_neighbor(
@@ -565,8 +553,6 @@ class RouterBgpMixin(Protocol):
                 neighbor.ebgp_multihop = self.inputs.evpn_ebgp_gateway_multihop
 
             neighbors.append(neighbor)
-
-        self.structured_config.router_bgp.neighbors = neighbors
 
     def _ip_in_listen_ranges(
         self: AvdStructuredConfigOverlayProtocol, source_ip: str, listen_range_prefixes: EosDesigns.BgpPeerGroups.WanOverlayPeers.ListenRangePrefixes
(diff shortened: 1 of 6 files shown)


python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -646,27 +562,15 @@ def _neighbors(self: AvdStructuredConfigOverlayProtocol) -> list | None:
...
-                neighbor["ebgp_multihop"] = self.inputs.evpn_ebgp_gateway_multihop
+                neighbor.ebgp_multihop = self.inputs.evpn_ebgp_gateway_multihop
 
             neighbors.append(neighbor)
 
-        if neighbors:
-            return neighbors
-
-        return None
+        self.structured_config.router_bgp.neighbors = neighbors
```
ClausHolbechArista: Please use append or append new on `self.structured_config.router_bgp.neighbors` above where you create them instead of resetting the full list. You can set `neighbors = self.structured_config.router_bgp.neighbors` at the top to keep it short.

ClausHolbechArista: same for peer groups

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
MaheshGSLAB: updated all over

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,361 @@ class RouterBgpMixin(Protocol):
...
+    def _set_peer_groups(self: AvdStructuredConfigOverlayProtocol) -> None:
+        peer_groups = self.structured_config.router_bgp.peer_groups
         if self.shared_utils.overlay_routing_protocol == "ebgp":
             # EVPN OVERLAY peer group
-            ebgp_peer_group = {
-                **self._generate_base_peer_group("evpn", "evpn_overlay_peers"),
-                "ebgp_multihop": self.inputs.evpn_ebgp_multihop,
-            }
+            ebgp_overlay_peer_group = self._generate_base_peer_group("evpn", "evpn_overlay_peers")
+            ebgp_overlay_peer_group.ebgp_multihop = self.inputs.evpn_ebgp_multihop
```
laxmikantchintakindi, line 96: can we add this in _generate_base_peer_group?

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,361 @@ class RouterBgpMixin(Protocol):
...
-
-    def _address_family_evpn(self: AvdStructuredConfigOverlayProtocol) -> dict | None:
-        address_family_evpn = {}
+            peer_groups.append_new(name=self.inputs.bgp_peer_groups.ipvpn_gateway_peers.name, activate=False)
 
-        peer_groups = []
-
-        overlay_peer_group = {}
+    def _set_address_family_evpn(self: AvdStructuredConfigOverlayProtocol) -> None:
+        peer_groups = self.structured_config.router_bgp.address_family_evpn.peer_groups
```
laxmikantchintakindi, line 199: may be use evpn = self.structured_config.router_bgp.address_family_evpn so that we can update self.structured_config.router_bgp.address_family_evpn.neighbor_default.next_hop_self_received_evpn_routes, self.structured_config.router_bgp.address_family_evpn.neighbor_default.encapsulation, self.structured_config.router_bgp.address_family_evpn.evpn_hostflap_detection in short.

laxmikantchintakindi, line 198: May be we can call this
`struct_config = self.structured_config.router_bgp.address_family_evpn` and use this variable in this method.

ClausHolbechArista: I like that the variable is called the same as in the schema. Otherwise it becomed harder to track.

Vibhu-gslab (approved):

laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,361 @@ class RouterBgpMixin(Protocol):
...
-                peer_group_config["route_reflector_client"] = True
-            peer_group_config["bfd_timers"] = self.inputs.bgp_peer_groups.wan_overlay_peers.bfd_timers._as_dict(include_default_values=True)
-            peer_groups.append(
-                {
-                    **self._generate_base_peer_group("wan", "wan_overlay_peers", update_source=self.shared_utils.vtep_loopback),
-                    **peer_group_config,
-                },
+                wan_overlay_peer_group.route_reflector_client = True
+
+            bfd_timers_dict = self.inputs.bgp_peer_groups.wan_overlay_peers.bfd_timers._as_dict(include_default_values=True)
```
ClausHolbechArista: please avoid _as_dict. It should be possible to mape the values correctly.

ClausHolbechArista: same elsewhere

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,359 @@ class RouterBgpMixin(Protocol):
...
-                        "route_reflector_client": True,
-                    },
+                wan_rr_overlay_peer_group._update(
+                    remote_as=self.shared_utils.bgp_as,
+                    ttl_maximum_hops=self.inputs.bgp_peer_groups.wan_rr_overlay_peers.ttl_maximum_hops,
+                    route_reflector_client=True,
+                )
+
+                bfd_timers_item = self.inputs.bgp_peer_groups.wan_rr_overlay_peers.bfd_timers
+                wan_rr_overlay_peer_group.bfd_timers._update(
```
ClausHolbechArista, line 157:
```suggestion
                # We have to apply the attributes individually to get the defaults from the input class.
                wan_rr_overlay_peer_group.bfd_timers._update(
```

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,360 @@ class RouterBgpMixin(Protocol):
...
 
-    def _address_family_evpn(self: AvdStructuredConfigOverlayProtocol) -> dict | None:
-        address_family_evpn = {}
-
-        peer_groups = []
-
-        overlay_peer_group = {}
+    def _set_address_family_evpn(self: AvdStructuredConfigOverlayProtocol) -> None:
+        peer_groups = self.structured_config.router_bgp.address_family_evpn.peer_groups
+        overlay_peer_group = EosCliConfigGen.RouterBgp.AddressFamilyEvpn.PeerGroupsItem()
```
ClausHolbechArista: Let's move this one down below the wan-route section (so ~10 lines down), since it is not used until then.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,360 @@ class RouterBgpMixin(Protocol):
...
-                    },
+                wan_overlay_peer_group._update(
+                    route_map_in="RM-EVPN-SOO-IN",
+                    route_map_out="RM-EVPN-SOO-OUT",
                 )
             peer_groups.append(wan_overlay_peer_group)
 
         if self.shared_utils.overlay_evpn_vxlan is True:
-            overlay_peer_group = {"name": self.inputs.bgp_peer_groups.evpn_overlay_peers.name, "activate": True}
+            overlay_peer_group._update(name=self.inputs.bgp_peer_groups.evpn_overlay_peers.name, activate=True)
```
ClausHolbechArista: Actually this could also be moved down below the ebgp section since don't need it until the ibgp section.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,360 @@ class RouterBgpMixin(Protocol):
...
+                overlay_peer_group._update(
+                    route_map_in="RM-EVPN-SOO-IN",
+                    route_map_out="RM-EVPN-SOO-OUT",
                 )
-        if overlay_peer_group:
-            peer_groups.append(overlay_peer_group)
 
-        if peer_groups:
-            address_family_evpn["peer_groups"] = peer_groups
+        peer_groups.append(overlay_peer_group)
```
ClausHolbechArista: I think we wish to keep the condition. since there is no need to add the class item if we have not filled out anything.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,360 @@ class RouterBgpMixin(Protocol):
...
-        address_family_rtc = {}
-
-        peer_groups = []
-        evpn_overlay_peers = {"name": self.inputs.bgp_peer_groups.evpn_overlay_peers.name}
+            return
+        peer_groups = self.structured_config.router_bgp.address_family_rtc.peer_groups
+        peer_groups.append_new(name=self.inputs.bgp_peer_groups.evpn_overlay_peers.name)
         if self.shared_utils.overlay_evpn_vxlan is True:
-            evpn_overlay_peers["activate"] = True
+            peer_groups.obtain(self.inputs.bgp_peer_groups.evpn_overlay_peers.name).activate = True
```
ClausHolbechArista: We should not use obtain here. Instead move the if above the append_new and set activate. But it seems weird that we would not always activate the peer-group if we add it. So IMO we should only add the peer-group here if we also intend to activate it.

Please also avoid the obtain below. Maybe just create new peer-group items and append them after you have set the attributes you need.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,360 @@ class RouterBgpMixin(Protocol):
...
 
         if (version == 4 and self.shared_utils.overlay_vpn_ipv4 is not True) or (version == 6 and self.shared_utils.overlay_vpn_ipv6 is not True):
-            return None
-
-        address_family_vpn_ipvx = {}
+            return
 
-        if self.shared_utils.overlay_ler is True or self.shared_utils.overlay_ipvpn_gateway is True:
-            address_family_vpn_ipvx["neighbor_default_encapsulation_mpls_next_hop_self"] = {"source_interface": "Loopback0"}
+        af_vpn = getattr(self.structured_config.router_bgp, f"address_family_vpn_ipv{version}")
```
ClausHolbechArista: We loose all type safety with this. Could we just do a ternary option like `af_vpn = self.structured.config.router_bgp.address_family_vpn_ipv4 if version==4 else self.structured.config.router_bgp.address_family_vpn_ipv6`. That would give us a union type on the af_vpn and ensure that everything we do below will work with both.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
ClausHolbechArista: Still need to remove obtain from the other address families. I like the logic of evpn now, so use the same pattern.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -483,29 +393,28 @@ def _create_neighbor(
...
 
         return neighbor
 
-    def _neighbors(self: AvdStructuredConfigOverlayProtocol) -> list | None:
-        neighbors = []
-
+    def _set_neighbors(self: AvdStructuredConfigOverlayProtocol) -> None:
+        neighbors = self.structured_config.router_bgp.neighbors
         if self.shared_utils.overlay_routing_protocol == "ebgp":
             for route_server, data in natural_sort(self._evpn_route_servers.items()):
```
ClausHolbechArista, line 420: We need to refactor `self._evpn_route_servers` to provide properly typed data instead of a dict where we loose all tracking. For now you can just update the return hint to `dict[str, [dict[str, str]]`, but I would like to get something better.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,365 @@ class RouterBgpMixin(Protocol):
...
-        address_family_rtc["peer_groups"] = peer_groups
-
-        return address_family_rtc
+                peer_groups.obtain(self.inputs.bgp_peer_groups.evpn_overlay_peers.name).default_route_target.only = True
 
-    def _address_family_vpn_ipvx(self: AvdStructuredConfigOverlayProtocol, version: int) -> dict | None:
+    def _set_address_family_vpn_ipvx(self: AvdStructuredConfigOverlayProtocol, version: int) -> None:
         if version not in [4, 6]:
             msg = "_address_family_vpn_ipvx should be called with version 4 or 6 only"
             raise AristaAvdError(msg)
```
ClausHolbechArista:
```suggestion
    def _set_address_family_vpn_ipvx(self: AvdStructuredConfigOverlayProtocol, version: Literal[4, 6]) -> None:
```

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Updated the hint.
will raise an issue as well.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: https://github.com/aristanetworks/avd/issues/5208

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -24,457 +26,365 @@ class RouterBgpMixin(Protocol):
...
-            address_family_link_state["path_selection"] = {"roles": {"consumer": True, "propagator": True}}
-            address_family_link_state["peer_groups"][0].update(
-                {
-                    "missing_policy": {
-                        "direction_out_action": "deny",
-                    },
-                },
-            )
+            self.structured_config.router_bgp.address_family_link_state.path_selection.roles._update(consumer=True, propagator=True)
+            peer_groups.obtain(self.inputs.bgp_peer_groups.wan_overlay_peers.name).missing_policy.direction_out_action = "deny"
```
ClausHolbechArista: Still an obtain here. Please just create the peer group object instead of append_new above and then update it here and append below.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
Vibhu-gslab: Fixed! 
Apologies for missing out on this one.

ClausHolbechArista (approved):
LGTM. Thank you.

PR #5119, opened 2025-03-03
Refactor(eos_designs): Refactor eos_designs structured_config code for spanning_tree

MaheshGSLAB:
## Change Summary

Refactor eos_designs structured_config code for spanning_tree

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Refactor eos_designs structured_config code for spanning_tree

## How to test
Run Molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
+++ b/python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
@@ -390,13 +390,11 @@ class AvdStructuredConfigBaseProtocol(NtpMixin, SnmpServerMixin, RouterGeneralMi
         """spanning_tree set based on spanning_tree_root_super, spanning_tree_mode and spanning_tree_priority."""
         if not self.shared_utils.network_services_l2:
             self.structured_config.spanning_tree.mode = "none"
-
-        spanning_tree_root_super = self.shared_utils.node_config.spanning_tree_root_super
-        spanning_tree_mode = self.shared_utils.node_config.spanning_tree_mode
-        if spanning_tree_root_super is not True and spanning_tree_mode is None:
             return
 
-        if spanning_tree_root_super is True:
+        spanning_tree_mode = self.shared_utils.node_config.spanning_tree_mode
+
+        if self.shared_utils.node_config.spanning_tree_root_super is True:
             self.structured_config.spanning_tree.root_super = True
 
         if spanning_tree_mode is not None:
--- a/python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
+++ b/python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
@@ -385,31 +385,28 @@ class AvdStructuredConfigBaseProtocol(NtpMixin, SnmpServerMixin, RouterGeneralMi
             }
         return None
 
-    @cached_property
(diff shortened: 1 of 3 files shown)


Vibhu-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -385,31 +385,28 @@ def interface_defaults(self) -> dict | None:
...
         """spanning_tree set based on spanning_tree_root_super, spanning_tree_mode and spanning_tree_priority."""
         if not self.shared_utils.network_services_l2:
-            return {"mode": "none"}
+            self.structured_config.spanning_tree.mode = "none"
 
         spanning_tree_root_super = self.shared_utils.node_config.spanning_tree_root_super
         spanning_tree_mode = self.shared_utils.node_config.spanning_tree_mode
         if spanning_tree_root_super is not True and spanning_tree_mode is None:
-            return None
+            return
```
gmuloc: I think you don't need this `if` anymore and so probably you can used directly the shared_utils in the two below if statements

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -385,31 +385,28 @@ def interface_defaults(self) -> dict | None:
...
         return None
 
-    @cached_property
-    def spanning_tree(self) -> dict | None:
+    @structured_config_contributor
+    def spanning_tree(self) -> None:
         """spanning_tree set based on spanning_tree_root_super, spanning_tree_mode and spanning_tree_priority."""
         if not self.shared_utils.network_services_l2:
-            return {"mode": "none"}
+            self.structured_config.spanning_tree.mode = "none"
```
ClausHolbechArista, line 392: the old code returned here, so you should add a return after setting the none. Please ensure we have testing of this case. I would expect all spines to hit this path.

ClausHolbechArista (dismissed):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
MaheshGSLAB: spanning_tree_mode used multiple times hence keeping and removed the spanning_tree_root_super variable

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
MaheshGSLAB: Updated the code.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: it is used inside a block that says " if spanning_tee_mode is not None" which is redundant with this check

gmuloc: hmm actually I am not seeing the last change so I amy be commenting the wrong stuff :)

gmuloc (approved):

PR #5123, opened 2025-03-03
Refactor(eos_designs): Refactor eos_designs structured_config code for flows init

MaheshGSLAB:
## Change Summary

Refactor eos_designs structured_config code for flow init

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Refactor eos_designs structured_config code for flow init

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -91,29 +91,16 @@ def sflow(self) -> dict | None:
...
-                        "destination": destination.destination,
-                        "port": destination.port,
-                    },
-                )
-                sflow_vrfs[vrf]["source_interface"] = source_interface
-
-        # convert sflow_vrfs dict into list and insert into sflow
-        if sflow_vrfs:
-            sflow["vrfs"] = [{"name": vrf_name, **vrf} for vrf_name, vrf in sflow_vrfs.items()]
+                vrf_item = sflow_vrfs.obtain(vrf)
```
Vibhu-gslab: I think it would be good if we specify `.obtain(name=vrf)` here, to make it easy to read

Vibhu-gslab: Or we can rename `vrf` to `vrf_name`

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,60 @@ def _enable_sflow(self) -> bool:
...
-                tracker["table_size"] = table_size
-            if (mpls := tracker_settings.sampled.record_export.mpls) is not None:
-                tracker["record_export"]["mpls"] = mpls
-
-        return tracker
+    def resolve_flow_tracker_by_type(
+        self, tracker_settings: EosDesigns.FlowTrackingSettings.TrackersItem
+    ) -> EosCliConfigGen.FlowTracking.Hardware.TrackersItem | EosCliConfigGen.FlowTracking.Sampled.TrackersItem:
+        tracker_type = self.shared_utils.flow_tracking_type
+        if tracker_type == "sampled":
```
Vibhu-gslab: since we are using it only once may be we can check the condition directly.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,60 @@ def _enable_sflow(self) -> bool:
...
-        if tracker_type == "sampled":
-            # asdict does not contain default values so we need to insert the default sample.
-            # TODO: consider if asdict should include defaults.
-            flow_tracking[tracker_type]["sample"] = global_settings.sample
-
-        filtered_trackers = []
+        structured_config = self.structured_config.flow_tracking
+        flow_tracking_settings = self.inputs.flow_tracking_settings
+
+        if tracker_type == "hardware":
```
Vibhu-gslab: Same for this one as well.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,60 @@ def _enable_sflow(self) -> bool:
...
+            structured_config.hardware = flow_tracking_settings.hardware._cast_as(EosCliConfigGen.FlowTracking.Hardware)
+            structured_config.hardware.shutdown = False
+            tracker_list = structured_config.hardware.trackers
+        else:
+            structured_config.sampled.sample = flow_tracking_settings.sampled.sample
```
laxmikantchintakindi:
```suggestion
           structured_config.sampled._update(sample = flow_tracking_settings.sampled.sample, shutdown=False)
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,60 @@ def _enable_sflow(self) -> bool:
...
+            structured_config.sampled.encapsulation._update(
+                ipv4_ipv6=flow_tracking_settings.sampled.encapsulation.ipv4_ipv6,
+                mpls=flow_tracking_settings.sampled.encapsulation.mpls,
+            )
+            structured_config.sampled.hardware_offload._update(
+                ipv4=flow_tracking_settings.sampled.hardware_offload.ipv4,
+                ipv6=flow_tracking_settings.sampled.hardware_offload.ipv6,
+                threshold_minimum=flow_tracking_settings.sampled.hardware_offload.threshold_minimum,
+            )
+            structured_config.sampled.shutdown = False
```
laxmikantchintakindi:
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: changed to vrf_name

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: updated

laxmikantchintakindi (approved):

Vibhu-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -47,73 +47,60 @@ def sflow(self) -> dict | None:
...
         # At this point we have at least one interface with sFlow enabled
         # and at least one destination.
-        sflow = {"run": True, "polling_interval": self.inputs.sflow_settings.polling_interval, "sample": self.inputs.sflow_settings.sample.rate}
-
-        # Using a temporary dict for VRFs
-        sflow_vrfs = {}
+        self.structured_config.sflow._update(
+            run=True, polling_interval=self.inputs.sflow_settings.polling_interval, sample=self.inputs.sflow_settings.sample.rate
+        )
+        sflow_vrfs = EosCliConfigGen.Sflow.Vrfs()
```
gmuloc: @MaheshGSLAB looking at the code I am wondering if you even need the intermediate `sflow_vrfs` - the only thing you do is doing obtain in it to add some destinations which can be done on the `self.structured_config.sflow.vrfs` directly I believe

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,57 @@ def _enable_sflow(self) -> bool:
...
+        structured_config = self.structured_config.flow_tracking
+        flow_tracking_settings = self.inputs.flow_tracking_settings
 
-        filtered_trackers = []
+        if self.shared_utils.flow_tracking_type == "hardware":
+            structured_config.hardware = flow_tracking_settings.hardware._cast_as(EosCliConfigGen.FlowTracking.Hardware)
+            structured_config.hardware.shutdown = False
+            tracker_list = structured_config.hardware.trackers
+        else:
+            structured_config.sampled._update(sample=flow_tracking_settings.sampled.sample, shutdown=False)
```
gmuloc: any reason to not use _case_as for the sampled case?

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,57 @@ def _enable_sflow(self) -> bool:
...
+            tracker_list = structured_config.sampled.trackers
+
+        default_tracker = next(iter(EosDesigns.FlowTrackingSettings().trackers))
         for tracker_name in natural_sort(configured_trackers):
             """
             We allow overriding the default flow tracker name, so if user has configured a tracker
             with the default tracker name, then we just use that, if not, we create a default config
             """
-            default_tracker = next(iter(EosDesigns.FlowTrackingSettings().trackers))
             if tracker_name not in self.inputs.flow_tracking_settings.trackers:
```
gmuloc: here we are using a mix of `self.inputs.flow_tracking_settigns` and your shorter version of `flow_tracking_settings`. Wondering if we should not align this.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,44 +111,57 @@ def _enable_sflow(self) -> bool:
...
-        flow_tracking[tracker_type] = global_settings._as_dict()
-        if tracker_type == "sampled":
-            # asdict does not contain default values so we need to insert the default sample.
-            # TODO: consider if asdict should include defaults.
-            flow_tracking[tracker_type]["sample"] = global_settings.sample
+        structured_config = self.structured_config.flow_tracking
+        flow_tracking_settings = self.inputs.flow_tracking_settings
 
-        filtered_trackers = []
+        if self.shared_utils.flow_tracking_type == "hardware":
```
gmuloc: looking at this if /else here and in the tracking resolution, it almost feel like it would be easier to follow if we had one sub-function that does all the hardware and one that does all the sampled config and we just call the appropriate ones. The only inconvenient would be the repeat of the error message if the tracker name is part of the configured one but I feel like we can make this check early on and then carry on.

Thoughts?

MaheshGSLAB: spilt the methods, let me know if this looks fine.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: Yes there is some default values and we decided to not use _cast_as for such cases.
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: Unable to cast '<class 'pyavd._eos_designs.schema.EosDesigns.FlowTrackingSettings.Sampled'>' as type '<class 'pyavd._eos_cli_config_gen.schema.EosCliConfigGen.FlowTracking.Sampled'>' since the field 'sample' is incompatible. Value 50000
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: Yes you are right, updated the code

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: ah yeah .. the comparison in cast_as. thanks!

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
MaheshGSLAB: @gmuloc Revert the changes as discussed

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +110,95 @@ def _enable_sflow(self) -> bool:
...
+        all_interfaces = chain(
+            self.structured_config.ethernet_interfaces, self.structured_config.port_channel_interfaces, self.structured_config.dps_interfaces
+        )
 
-            filtered_trackers.append(self.resolve_flow_tracker_by_type(tracker))
+        trackers: set[str] = {interface.flow_tracker.hardware for interface in all_interfaces if interface.flow_tracker.hardware}
+        if not trackers:
+            return
+
+        self.structured_config.flow_tracking.hardware = self.inputs.flow_tracking_settings.hardware._cast_as(EosCliConfigGen.FlowTracking.Hardware)
```
ClausHolbechArista: Can we try to change the schema to use a $ref when we also use cast_as? Otherwise a drift in schemas might lead to bad issues.

ClausHolbechArista: Actually it seems like we only support `hardware.record.format_ipfix_standard_timestamps_counters` so that should be faster to just set directly.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +110,95 @@ def _enable_sflow(self) -> bool:
...
+                name=config.name,
+                record_export=record_export,
+                exporters=config.exporters._cast_as(EosCliConfigGen.FlowTracking.Sampled.TrackersItem.Exporters),
+                table_size=config.sampled.table_size,
+            )
+
+    def _get_tracker_input_config(
+        self, default_tracker: EosDesigns.FlowTrackingSettings.TrackersItem, tracker_name: str
+    ) -> EosDesigns.FlowTrackingSettings.TrackersItem:
+        """Retrieves inputs for the given tracker_name.
```
ClausHolbechArista:
```suggestion
        """
        Retrieves inputs for the given tracker_name.
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +110,95 @@ def _enable_sflow(self) -> bool:
...
-            "exporters": tracker_settings.exporters._as_list(),
-        }
-        if self.shared_utils.flow_tracking_type == "sampled":
-            if (table_size := tracker_settings.sampled.table_size) is not None:
-                tracker["table_size"] = table_size
-            if (mpls := tracker_settings.sampled.record_export.mpls) is not None:
-                tracker["record_export"]["mpls"] = mpls
+    @structured_config_contributor
+    def flow_tracking(self) -> None:
+        """Set the structured config for flow_tracking if any interface is enabled for flow-tracking.
```
ClausHolbechArista:
```suggestion
        """
        Set the structured config for flow_tracking if any interface is enabled for flow-tracking.
```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +110,95 @@ def _enable_sflow(self) -> bool:
...
+
+        We allow overriding the default flow tracker name, so if user has configured a tracker
+        with the default tracker name, then we just use that, if not, we create a default config
         """
-        all_interfaces = chain(
-            self.structured_config.ethernet_interfaces, self.structured_config.port_channel_interfaces, self.structured_config.dps_interfaces
-        )
-        if self.shared_utils.flow_tracking_type == "hardware":
-            return {interface.flow_tracker.hardware for interface in all_interfaces if interface.flow_tracker.hardware}
+        if (tracker_settings := self.inputs.flow_tracking_settings.trackers.get(tracker_name)) != Undefined:
```
ClausHolbechArista: IMO a `if tracker_name in self.inputs.flow_tracking_settings.trackers` is simpler. and then return the item with `[tracker_name]`. Two cheap operations and simpler to read.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +110,95 @@ def _enable_sflow(self) -> bool:
...
+            hardware_offload=EosCliConfigGen.FlowTracking.Sampled.HardwareOffload(
+                ipv4=self.inputs.flow_tracking_settings.sampled.hardware_offload.ipv4,
+                ipv6=self.inputs.flow_tracking_settings.sampled.hardware_offload.ipv6,
+                threshold_minimum=self.inputs.flow_tracking_settings.sampled.hardware_offload.threshold_minimum,
+            ),
+        )
 
-        return flow_tracking
+        # Validate and configure trackers
+        default_tracker = next(iter(EosDesigns.FlowTrackingSettings().trackers))
```
ClausHolbechArista: I think it would make sense to do this in the helper function after seing the tracker was not there

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: it means potentially many next/iter

gmuloc: changed

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/flow-tracking-missing-tracker.yml
```diff
@@ -0,0 +1,23 @@
...
+servers:
+  - name: test
+    adapters:
+      - switches: [flow-tracking-missing-tracker]
+        switch_ports: [Ethernet1]
+        flow_tracking:
+          enabled: true
+
+expected_error_message: >-
+  'I-DO-NOT-EXIST' is being used for one of the interfaces, but is not configured in 'self.inputs.flow_tracking_settings'.
```
ClausHolbechArista:
```suggestion
  The flow tracker 'I-DO-NOT-EXIST' is being used for at least one interface, but is not configured in 'self.inputs.flow_tracking_settings'.
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/FLOW_TRACKING_TESTS_SPINES_2.yml
```diff
@@ -20,41 +20,35 @@ flow_tracking_settings:
...
+        on_interval: 300331
+      exporters:
+        - name: ayush_exporter
+          collector:
+            host: 127.0.0.1
+          local_interface: Loopback0
+          template_interval: 40000
 
 # False by default
 fabric_flow_tracking:
```
ClausHolbechArista, line 45: why all these changes to tests? It makes it much harder to verify the refactor is correct.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +109,96 @@ def _enable_sflow(self) -> bool:
...
+                threshold_minimum=self.inputs.flow_tracking_settings.sampled.hardware_offload.threshold_minimum,
+            ),
+        )
+
+        # Validate and configure trackers
+        for tracker_name in natural_sort(trackers):
+            config = self._get_tracker_input_config(tracker_name)
+
+            # Need to handle mpls specifically
+            record_export = config.record_export._cast_as(EosCliConfigGen.FlowTracking.Sampled.TrackersItem.RecordExport)
```
ClausHolbechArista, line 177: This is one boolean, so let us avoid cast_as

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +109,96 @@ def _enable_sflow(self) -> bool:
...
+            record_export.mpls = config.sampled.record_export.mpls
 
-        return flow_tracking
+            self.structured_config.flow_tracking.sampled.trackers.append_new(
+                name=config.name,
+                record_export=record_export,
+                exporters=config.exporters._cast_as(EosCliConfigGen.FlowTracking.Sampled.TrackersItem.Exporters),
```
ClausHolbechArista, line 183: Could we update the exporters schema to use a pure $ref instead of repeating it from eos_cli_config_gen? Effectively it should be the same, and then the classes would be cli gen classes already.

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
```diff
@@ -124,70 +109,96 @@ def _enable_sflow(self) -> bool:
...
+        self.structured_config.flow_tracking.hardware.shutdown = False
+
+        # Validate and configure trackers
+        for tracker_name in natural_sort(trackers):
+            config = self._get_tracker_input_config(tracker_name)
+            self.structured_config.flow_tracking.hardware.trackers.append_new(
+                name=config.name,
+                record_export=config.record_export._cast_as(EosCliConfigGen.FlowTracking.Hardware.TrackersItem.RecordExport),
+                exporters=config.exporters._cast_as(EosCliConfigGen.FlowTracking.Hardware.TrackersItem.Exporters),
+            )
```
ClausHolbechArista, line 146: Please check if we can modify the schemas to be $ref to cli gen so we don't need this

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/FLOW_TRACKING_TESTS_SPINES_2.yml
gmuloc: we are not testing this today so we don't know if the refactor is correct

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: this is not only one boolean:


```
flow_tracking:
  sampled:
    sample: 10000
    trackers:
    - record_export:
        on_inactive_timeout: 50001
        on_interval: 300332
      name: FLOW-TRACKER
      exporters:
      - name: ayush_exporter
        collector:
          host: 127.0.0.2
        local_interface: Loopback0
        template_interval: 40002
    - table_size: 4331
      record_export:
        mpls: true. <<<<<<<<<<<<<< this is added in the schema on top of a def
        on_inactive_timeout: 50000
        on_interval: 300331
 ```

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: discussed on call we can't because we would need to ref a defs which we can't today

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: same as above..

python-avd/pyavd/_eos_designs/structured_config/flows/__init__.py
gmuloc: so discussed we cna't do that

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/FLOW_TRACKING_TESTS_SPINES_2.yml
gmuloc: so as discussed base on Code Coverage report this changes is to be able to use the default tracker without redefining it.
The removal of uplinks / endpoints and mlag is because this group_vars file is only for spine 2 which does not have any uplink, endpoin and mlag so this is just noise that makes understandting what we test harder

ClausHolbechArista (approved):

PR #5127, opened 2025-03-04
Refactor(eos_designs): Refactor eos_designs structured_config code for models in base/init.py and utils.py

Shivani-gslab:
## Change Summary

Refactor eos_designs structured_config code for models in base/init.py

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


gmuloc:
@Shivani-gslab for the hotspots please add # NOSONAR to tell it to ignore the `/tmp`

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -261,175 +239,156 @@ def daemon_terminattr(self) -> dict | None:
...
-            return {
-                "mtu": self.shared_utils.default_interface_mtu,
-            }
-        return None
+            self.structured_config.interface_defaults.mtu = self.shared_utils.default_interface_mtu
 
-    @cached_property
-    def spanning_tree(self) -> dict | None:
+    @structured_config_contributor
+    def spanning_tree(self) -> None:
```
MaheshGSLAB, line 325: Will be addressed in PR 5119

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -73,63 +74,46 @@ def router_bgp(self) -> dict | None:
...
-                    "graceful_restart": {
-                        "enabled": True,
-                        "restart_time": self.inputs.bgp_graceful_restart.restart_time,
-                    },
-                },
-            )
+            self.structured_config.router_bgp.graceful_restart._update(enabled=True, restart_time=self.inputs.bgp_graceful_restart.restart_time)
 
-        l3_interfaces_neighbors = []
+        l3_interfaces_neighbors = EosCliConfigGen.RouterBgp.Neighbors()
```
ClausHolbechArista: you can just append directly on self.structured_config.router_bgp.neighbors below, and also activate the neighbor under ipv4 at the same time.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -164,64 +148,61 @@ def ipv6_static_routes(self) -> None:
...
-        if self.ip_routing_ipv6_interfaces is True:
-            return None
-        return True
+        if self.structured_config.ip_routing_ipv6_interfaces is True:
+            return
 
-    @cached_property
-    def ipv6_unicast_routing(self) -> bool | None:
+        self.structured_config.ip_routing = True
+
```
ClausHolbechArista: Let's combine these into one function `ip_routing` and then do the if self.inputs.underlay_rfc5549 and an `else` to set the ip_routing

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -164,64 +148,61 @@ def ipv6_static_routes(self) -> None:
...
 
-        return router_multicast
-
-    @cached_property
-    def hardware_counters(self) -> dict | None:
+    @structured_config_contributor
+    def hardware_counters(self) -> None:
         """hardware_counters set based on hardware_counters.features variable."""
-        return self.inputs.hardware_counters._as_dict() or None
+        self.structured_config.hardware_counters = self.inputs.hardware_counters._cast_as(EosCliConfigGen.HardwareCounters)
```
ClausHolbechArista: I don't think you need the cast_as here. The schema is a pure $ref, so the class should already be eos_cli_config_gen.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -39,28 +39,29 @@ class AvdStructuredConfigBaseProtocol(NtpMixin, SnmpServerMixin, RouterGeneralMi
...
-        return self.inputs.is_deployed
+    @structured_config_contributor
+    def is_deployed(self) -> None:
+        self.structured_config.is_deployed = self.inputs.is_deployed
 
+    # TODO: reafactor with structured_config_contributor
     @cached_property
     def serial_number(self) -> str | None:
         """serial_number variable set based on serial_number fact."""
         return self.shared_utils.serial_number
```
Vibhu-gslab: Why aren't we doing in this PR?

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -261,129 +239,113 @@ def daemon_terminattr(self) -> dict | None:
...
-        return self.inputs.event_handlers._as_list() or None
+        self.structured_config.event_handlers = self.inputs.event_handlers._cast_as(EosCliConfigGen.EventHandlers)
 
-    @cached_property
-    def load_interval(self) -> dict | None:
+    @structured_config_contributor
+    def load_interval(self) -> None:
         """load_interval set based on load_interval_default variable."""
-        return self.inputs.load_interval._as_dict() or None
+        self.structured_config.load_interval = self.inputs.load_interval._cast_as(EosCliConfigGen.LoadInterval)
```
ClausHolbechArista: I think this is the correct class already. Try without cast_as

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -261,129 +239,113 @@ def daemon_terminattr(self) -> dict | None:
...
-        return self.inputs.event_monitor._as_dict() or None
+        self.structured_config.event_monitor = self.inputs.event_monitor._cast_as(EosCliConfigGen.EventMonitor)
 
-    @cached_property
-    def event_handlers(self) -> list | None:
+    @structured_config_contributor
+    def event_handlers(self) -> None:
         """event_handlers set based on event_handlers data-model."""
-        return self.inputs.event_handlers._as_list() or None
+        self.structured_config.event_handlers = self.inputs.event_handlers._cast_as(EosCliConfigGen.EventHandlers)
```
ClausHolbechArista: I think this is the correct class already. Try with cast_as

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -261,129 +239,113 @@ def daemon_terminattr(self) -> dict | None:
...
-        return not self.shared_utils.is_wan_router
+        self.structured_config.transceiver_qsfp_default_mode_4x10 = not self.shared_utils.is_wan_router
 
-    @cached_property
-    def event_monitor(self) -> dict | None:
+    @structured_config_contributor
+    def event_monitor(self) -> None:
         """event_monitor set based on event_monitor data-model."""
-        return self.inputs.event_monitor._as_dict() or None
+        self.structured_config.event_monitor = self.inputs.event_monitor._cast_as(EosCliConfigGen.EventMonitor)
```
ClausHolbechArista: I think this is the correct class already. Try with cast_as

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -406,25 +368,25 @@ def spanning_tree(self) -> None:
...
+    @structured_config_contributor
+    def local_users(self) -> None:
         """local_users set based on local_users data model."""
         if not self.inputs.local_users:
-            return None
+            return
 
-        return [user._as_dict() for user in self.inputs.local_users._natural_sorted()]
+        for user in self.inputs.local_users._natural_sorted():
+            self.structured_config.local_users.append(user)
```
ClausHolbechArista: I think this is the correct class already. Try this:
```suggestion
        self.structured_config.local_users = self.inputs.local_users.natural_sorted()
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -435,34 +397,28 @@ def vrfs(self) -> None:
...
-                interface_settings.update(
-                    {
-                        "ipv6_enable": True,
-                        "ipv6_address": self.shared_utils.node_config.ipv6_mgmt_ip,
-                        "ipv6_gateway": self.shared_utils.ipv6_mgmt_gateway,
-                    },
-                )
+                interface_settings.ipv6_enable = True
+                interface_settings.ipv6_address = self.shared_utils.node_config.ipv6_mgmt_ip
+                interface_settings.ipv6_gateway = self.shared_utils.ipv6_mgmt_gateway
```
ClausHolbechArista: use ._update() here

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -73,63 +74,46 @@ def router_bgp(self) -> dict | None:
...
-                    "address_family_ipv4": {
-                        "neighbors": [{"ip_address": neighbor["ip_address"], "activate": True} for neighbor in l3_interfaces_neighbors],
-                    },
-                }
+            l3_interfaces_neighbors.append_new(
+                ip_address=neighbor_info["ip_address"],
+                remote_as=neighbor_info["remote_as"],
+                description=neighbor_info["description"] if neighbor_info["description"] else None,
+                route_map_in=get(neighbor_info, "route_map_in"),
+                route_map_out=get(neighbor_info, "route_map_out"),
```
Vibhu-gslab: Have we missed rcf_in and rcf_out?

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -164,64 +148,61 @@ def ipv6_static_routes(self) -> None:
...
 
-    @cached_property
-    def ip_routing(self) -> bool | None:
+    @structured_config_contributor
+    def ip_routing_ipv6_interfaces(self) -> None:
+        """ip_routing_ipv6_interfaces set based on underlay_rfc5549 variable."""
+        if not self.shared_utils.underlay_router and not self.shared_utils.node_config.always_configure_ip_routing:
+            return
+
+        if self.inputs.underlay_rfc5549:
```
Vibhu-gslab, line 162: let's move this check first as this is just checking inputs and would return immediately without having to check for the more complex ones above

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: can not find it in structured_config  -
Screenshot 2025-03-11 at 3 29 28 PM

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: can not find these attributes inside `self.structured_config.router_bgp.neighbors`

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +69,28 @@ def _build_source_interfaces(
             # Check for duplicate VRF
             # inband_mgmt_vrf returns None in case of VRF "default", but here we want the "default" VRF name to have proper duplicate detection.
             inband_mgmt_vrf = self.shared_utils.inband_mgmt_vrf or "default"
-            if [source_interface for source_interface in source_interfaces if source_interface.get("vrf", "default") == inband_mgmt_vrf]:
+            if [source_interface for source_interface in source_interfaces if (getattr(source_interface, "vrf", "default") or "default") == inband_mgmt_vrf]:
```
laxmikantchintakindi:
```suggestion
           if [source_interface for source_interface in source_interfaces if source_interface.vrf or "default" == inband_mgmt_vrf]:
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: Done

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: Updated schema to fix this

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: Updated schema to fix this.

MaheshGSLAB: The rcf_in and rcf_out are not valid for neighbors as per EOS so we dont need this.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
laxmikantchintakindi: i think source_interface.vrf or "default" works. You have give only source_interface.vrf.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: reeverted it

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/serial_number.schema.yml
```diff
@@ -0,0 +1,15 @@
...
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  serial_number:
+    description: |
+      Serial Number of the device.
+      Used only for documentation and deployment purposes. It is used by the 'eos_config_deploy_cvp' role.
```
ClausHolbechArista:
```suggestion
      Used only for documentation and deployment purposes. It is used by the 'eos_config_deploy_cvp' and 'cv_deploy' roles.
```

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
```diff
@@ -21,18 +22,17 @@ class LinkTrackingGroupsMixin(Protocol):
...
+            link_tracking_groups = EosCliConfigGen.LinkTrackingGroups()
             default_recovery_delay = default(self.platform_settings.reload_delay.mlag, 300)
             if len(self.node_config.link_tracking.groups) > 0:
                 for lt_group in self.node_config.link_tracking.groups:
-                    lt_group_dict = lt_group._as_dict(include_default_values=True)
-                    lt_group_dict["recovery_delay"] = default(lt_group.recovery_delay, default_recovery_delay)
-                    link_tracking_groups.append(lt_group_dict)
+                    lt_group_item = lt_group._cast_as(EosCliConfigGen.LinkTrackingGroupsItem, ignore_extra_keys=True)
+                    lt_group_item.recovery_delay = default(lt_group.recovery_delay, default_recovery_delay)
+                    link_tracking_groups.append(lt_group_item)
```
ClausHolbechArista: There are only two keys in the link tracking group so I would prefer to avoid the cast_as since we also override the recovery_delay anyway.
```suggestion
                    link_tracking_groups.append_new(
                        name=lt_group.name,
                        links_minimum=lt_group.links_minimum,
                        recovery_delay=default(name=lt_group.recovery_delay, default_recovery_delay),
                    )
```

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
```diff
@@ -21,18 +22,17 @@ class LinkTrackingGroupsMixin(Protocol):
...
             else:
-                link_tracking_groups.append({"name": "LT_GROUP1", "recovery_delay": default_recovery_delay})
-
-            return strip_empties_from_list(link_tracking_groups)
+                link_tracking_groups.append_new(name="LT_GROUP1", recovery_delay=default_recovery_delay)
```
ClausHolbechArista: Please check coverage if this ever runs. I think our schema classes will add the default dict from the schema so we don't need to implement it here.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +69,28 @@ def _build_source_interfaces(
             # Check for duplicate VRF
             # inband_mgmt_vrf returns None in case of VRF "default", but here we want the "default" VRF name to have proper duplicate detection.
             inband_mgmt_vrf = self.shared_utils.inband_mgmt_vrf or "default"
-            if [source_interface for source_interface in source_interfaces if source_interface.get("vrf", "default") == inband_mgmt_vrf]:
+            if [source_interface for source_interface in source_interfaces if source_interface.vrf or "default" == inband_mgmt_vrf]:  # noqa: SIM300
```
ClausHolbechArista:
```suggestion
            if [source_interface for source_interface in source_interfaces if source_interface.vrf or inband_mgmt_vrf == "default"]:
```

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +69,28 @@ def _build_source_interfaces(
...
 
+        redistribute_route = EosCliConfigGen.RouterBgp.Redistribute()
         if (self.shared_utils.overlay_routing_protocol != "none" or self.shared_utils.is_wan_router) and self.inputs.underlay_filter_redistribute_connected:
             # Use route-map for redistribution
-            return {"connected": {"enabled": True, "route_map": "RM-CONN-2-BGP"}}
+            redistribute_route.connected._update(enabled=True, route_map="RM-CONN-2-BGP")
+            return redistribute_route
 
-        return {"connected": {"enabled": True}}
+        redistribute_route.connected.enabled = True
```
ClausHolbechArista: Both cases have connected.enabled.True, so you could set that above the if and just set the route-map under the if.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
```diff
@@ -228,7 +228,7 @@ def _get_adapter_link_tracking_groups(
             return Undefined
 
         output = output_type()
-        output.append_new(name=adapter.link_tracking.name or self.shared_utils.link_tracking_groups[0]["name"], direction="downstream")
+        output.append_new(name=adapter.link_tracking.name or self.shared_utils.link_tracking_groups._as_list()[0]["name"], direction="downstream")
```
ClausHolbechArista: Do you need to convert it to a list? Would it not work with `self.shared_utils.link_tracking_groups[0].name ?

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
Shivani-gslab: it gives type checking error -
```
Argument of type "Literal[0]" cannot be assigned to parameter "key" of type "str" in function "__getitem__"
  "Literal[0]" is not assignable to "str"
```

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
Shivani-gslab: It has been added in this PR - https://github.com/aristanetworks/avd/pull/4804/files#diff-9b3d7d19b581987980885c90bbdb67794dfff72cde4fb7202250159ca30f842fR111

ClausHolbechArista: It must have been added by mistake or wrongly. It should be rendered under the address family. Not globally for the neighbor.

Shivani-gslab: It does not impact the address-family config, so we can remove it.

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
Shivani-gslab: the `else` part  is uncovered
https://sonarcloud.io/component_measures?metric=new_coverage&selected=aristanetworks_avd%3Apython-avd%2Fpyavd%2F_eos_designs%2Fshared_utils%2Flink_tracking_groups.py&view=list&pullRequest=5127&id=aristanetworks_avd

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -20,27 +33,29 @@ class UtilsMixin(Protocol):
...
                 msg = f"Unable to configure {error_context} source-interface since 'mgmt_ip' or 'ipv6_mgmt_ip' are not set."
                 raise AristaAvdInvalidInputsError(msg)
 
             # mgmt_interface is always set (defaults to "Management1") so no need for error handling missing interface.
-            source_interface = {"name": self.shared_utils.mgmt_interface}
-            if self.inputs.mgmt_interface_vrf not in [None, "default"]:
-                source_interface["vrf"] = self.inputs.mgmt_interface_vrf
-            source_interfaces.append(source_interface)
+            source_interfaces.append_new(
+                name=self.shared_utils.mgmt_interface, vrf=self.inputs.mgmt_interface_vrf if self.inputs.mgmt_interface_vrf not in [None, "default"] else None
```
Vibhu-gslab: I think if we only check for `default` that would be fine.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,28 @@ def _build_source_interfaces(
...
                 msg = f"Unable to configure multiple {error_context} source-interfaces for the same VRF '{inband_mgmt_vrf}'."
                 raise AristaAvdError(msg)
 
-            source_interface = {"name": self.shared_utils.inband_mgmt_interface}
-            if self.shared_utils.inband_mgmt_vrf not in [None, "default"]:
-                source_interface["vrf"] = self.shared_utils.inband_mgmt_vrf
-            source_interfaces.append(source_interface)
+            source_interfaces.append_new(
+                name=self.shared_utils.inband_mgmt_interface,
+                vrf=self.shared_utils.inband_mgmt_vrf if self.shared_utils.inband_mgmt_vrf not in [None, "default"] else None,
```
Vibhu-gslab: Same for this

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,28 @@ def _build_source_interfaces(
...
+            )
 
         return source_interfaces
 
     @cached_property
-    def _router_bgp_redistribute_routes(self: AvdStructuredConfigBaseProtocol) -> dict | None:
+    def _router_bgp_redistribute_routes(self: AvdStructuredConfigBaseProtocol) -> EosCliConfigGen.RouterBgp.Redistribute | None:
         """Return structured config for router_bgp.redistribute."""
```
Vibhu-gslab: should this be `Return object of EosCliConfigGen.RouterBgp.Redistribute`?

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
Shivani-gslab: Update the doc string

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,28 @@ def _build_source_interfaces(
...
+        """Return redistribute route settings for router bgp."""
         if not (self.shared_utils.underlay_bgp or self.shared_utils.is_wan_router or self.shared_utils.l3_bgp_neighbors):
             return None
 
+        redistribute_route = EosCliConfigGen.RouterBgp.Redistribute()
+        redistribute_route.connected.enabled = True
         if (self.shared_utils.overlay_routing_protocol != "none" or self.shared_utils.is_wan_router) and self.inputs.underlay_filter_redistribute_connected:
             # Use route-map for redistribution
-            return {"connected": {"enabled": True, "route_map": "RM-CONN-2-BGP"}}
+            redistribute_route.connected._update(route_map="RM-CONN-2-BGP")
```
laxmikantchintakindi:
```suggestion
            redistribute_route.connected.route_map = "RM-CONN-2-BGP"
```

laxmikantchintakindi: May be it is even better to update enabled and router_map in this line.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -471,77 +399,69 @@ def management_security(self) -> None:
...
 
-    @cached_property
-    def link_tracking_groups(self) -> list | None:
-        """link_tracking_groups."""
-        return self.shared_utils.link_tracking_groups
+    @structured_config_contributor
+    def link_tracking_groups(self) -> None:
+        """Set link_tracking_groups."""
+        if link_tracking_groups := self.shared_utils.link_tracking_groups:
+            self.structured_config.link_tracking_groups = link_tracking_groups
```
Vibhu-gslab, line 435:
```suggestion
        self.structured_config.link_tracking_groups = self.shared_utils.link_tracking_groups
```

Shivani-gslab: it throws error - 
```
fatal: [snmp-settings-1 -> localhost]: FAILED! => {"changed": false, "msg": "'NoneType' object has no attribute '_strip_empties'"}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'NoneType' object has no attribute '_strip_empties'
```

laxmikantchintakindi: if link_tracking_groups:=self.shared_utils.link_tracking_groups:
    self.structured_config.link_tracking_groups.extend(link_tracking_groups)
would be accurate.

laxmikantchintakindi: may be we can return LinkTrackingGroups or [](currently None) self.shared_utils. link_tracking_groups.

Shivani-gslab: not sure why the suggestion from vibhu is not handling `None type`, it handles it when we take it from inputs @ClausHolbechArista @gmuloc

ClausHolbechArista: The fields holding list classes cannot be set to `None`. I would expect a type warning to also show up somewhere here, since the shared_utils would have `| None`

Vibhu-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
ClausHolbechArista: Let's set a default_name above like `default_name = next(iter(self.shared_utils.link_tracking_groups)).name`. It would raise if the list is empty, but so would the current code. It should never be empty according to the schema, so I think we can keep this.

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
```diff
@@ -21,18 +22,17 @@ class LinkTrackingGroupsMixin(Protocol):
     """
 
     @cached_property
-    def link_tracking_groups(self: SharedUtilsProtocol) -> list | None:
+    def link_tracking_groups(self: SharedUtilsProtocol) -> EosCliConfigGen.LinkTrackingGroups | None:
         if self.node_config.link_tracking.enabled:
```
ClausHolbechArista: Reverse the logic - needs indentation fix below.
```suggestion
        if not self.node_config.link_tracking.enabled:
            return None
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/serial_number.schema.yml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2023-2024 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  serial_number:
+    description: |
```
ClausHolbechArista:
```suggestion
    description: |-
```

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
```diff
@@ -21,18 +22,17 @@ class LinkTrackingGroupsMixin(Protocol):
...
-            link_tracking_groups = []
+            link_tracking_groups = EosCliConfigGen.LinkTrackingGroups()
             default_recovery_delay = default(self.platform_settings.reload_delay.mlag, 300)
             if len(self.node_config.link_tracking.groups) > 0:
```
ClausHolbechArista: This if is not needed, since we would just not loop if the list is empty.

laxmikantchintakindi: removed the if condition.

python-avd/pyavd/_eos_designs/shared_utils/link_tracking_groups.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,28 @@ def _build_source_interfaces(
...
         if not (self.shared_utils.underlay_bgp or self.shared_utils.is_wan_router or self.shared_utils.l3_bgp_neighbors):
             return None
 
+        redistribute_route = EosCliConfigGen.RouterBgp.Redistribute()
+        redistribute_route.connected.enabled = True
         if (self.shared_utils.overlay_routing_protocol != "none" or self.shared_utils.is_wan_router) and self.inputs.underlay_filter_redistribute_connected:
             # Use route-map for redistribution
-            return {"connected": {"enabled": True, "route_map": "RM-CONN-2-BGP"}}
+            redistribute_route.connected.route_map = "RM-CONN-2-BGP"
+            return redistribute_route
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/base/snmp_server.py
```diff
@@ -191,9 +191,13 @@ def _snmp_local_interfaces(self: AvdStructuredConfigBaseProtocol, source_interfa
         if not source_interfaces_inputs:
             # Empty dict or None
             return
-        local_interfaces = self._build_source_interfaces(source_interfaces_inputs.mgmt_interface, source_interfaces_inputs.inband_mgmt_interface, "SNMP")
-        for local_interface in local_interfaces:
-            self.structured_config.snmp_server.local_interfaces.append(EosCliConfigGen.SnmpServer.LocalInterfacesItem(**local_interface))
+        local_interfaces = self._build_source_interfaces(
+            source_interfaces_inputs.mgmt_interface,
+            source_interfaces_inputs.inband_mgmt_interface,
+            "SNMP",
```
ClausHolbechArista: just for readability
```suggestion
            error_context="SNMP",
```

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
laxmikantchintakindi: Fixed.

python-avd/pyavd/_eos_designs/structured_config/base/snmp_server.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,27 @@ def _build_source_interfaces(
...
+                name=self.shared_utils.inband_mgmt_interface,
+                vrf=self.shared_utils.inband_mgmt_vrf if self.shared_utils.inband_mgmt_vrf != "default" else None,
+            )
 
         return source_interfaces
 
     @cached_property
-    def _router_bgp_redistribute_routes(self: AvdStructuredConfigBaseProtocol) -> dict | None:
-        """Return structured config for router_bgp.redistribute."""
+    def _router_bgp_redistribute_routes(self: AvdStructuredConfigBaseProtocol) -> EosCliConfigGen.RouterBgp.Redistribute | None:
```
laxmikantchintakindi: can we move this one time used cached property to router bgp as a method to update router bgp struct config?

laxmikantchintakindi: @gmuloc @ClausHolbechArista

gmuloc: lets remove the cached_property. If it does not impact readability too much you can do the check directly in the function otherwise keep a helper function`_set_router_bgp_redistribute_routes`

laxmikantchintakindi: Moved the code block from utils to directly in router_bgp method. It is quite easy to read.

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/serial_number.schema.yml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2023-2024 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  serial_number:
+    description: |-
```
ClausHolbechArista, line 12:
```suggestion
    documentation_settings:
      table: metadata
    description: |-
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -74,63 +74,44 @@ def router_bgp(self) -> dict | None:
...
-                {
-                    "neighbors": l3_interfaces_neighbors,
-                    "address_family_ipv4": {
-                        "neighbors": [{"ip_address": neighbor["ip_address"], "activate": True} for neighbor in l3_interfaces_neighbors],
-                    },
-                }
+            self.structured_config.router_bgp.neighbors.append_new(
+                ip_address=neighbor_info["ip_address"],
+                remote_as=neighbor_info["remote_as"],
+                description=neighbor_info["description"] if neighbor_info["description"] else None,
```
```suggestion
                description=neighbor_info["description"] or None,
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -262,129 +217,113 @@ def daemon_terminattr(self) -> dict | None:
...
+            smashexcludes=self.inputs.terminattr_smashexcludes,
+            ingestexclude=self.inputs.terminattr_ingestexclude,
+            disable_aaa=self.inputs.terminattr_disable_aaa,
+        )
 
-    @cached_property
-    def vlan_internal_order(self) -> dict | None:
+    @structured_config_contributor
+    def vlan_internal_order(self) -> None:
         """vlan_internal_order set based on internal_vlan_order data-model."""
```
```suggestion
        """
        vlan_internal_order set based on internal_vlan_order data-model.
        
        TODO: Add platform_setting to control this.
        """
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -564,154 +484,153 @@ def ptp(self) -> dict | None:
...
+            mode=self.shared_utils.node_config.ptp.mode,
+            mode_one_step=self.shared_utils.node_config.ptp.mode_one_step or None,  # Historic output is without false
+            forward_unicast=self.shared_utils.node_config.ptp.forward_unicast or None,  # Historic output is without false
+            clock_identity=default(self.shared_utils.node_config.ptp.clock_identity, default_clock_identity),
+            priority1=priority1,
+            priority2=priority2,
+            ttl=self.shared_utils.node_config.ptp.ttl,
+            domain=default(self.shared_utils.node_config.ptp.domain, default_ptp_domain),
+        )
+        monitor = self.shared_utils.node_config.ptp.monitor._as_dict(include_default_values=True)
```
ClausHolbechArista: Please avoid this _as_dict.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -564,154 +484,153 @@ def ptp(self) -> dict | None:
...
+                self.structured_config.prefix_lists.append(pfx_list)
 
-            if (prefix_list_out := get(neighbor, "ipv4_prefix_list_out")) and prefix_list_out not in prefix_lists_in_use:
-                pfx_list = self._get_prefix_list(prefix_list_out)._as_dict()
-                prefix_lists.append(pfx_list)
-                prefix_lists_in_use.add(prefix_list_out)
-
-        return prefix_lists or None
+            if (prefix_list_out := get(neighbor, "ipv4_prefix_list_out")) and prefix_list_out:
+                pfx_list = self._get_prefix_list(prefix_list_out)._cast_as(EosCliConfigGen.PrefixListsItem)
```
ClausHolbechArista: You could move the cast to the helper function instead.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -564,154 +484,153 @@ def ptp(self) -> dict | None:
...
-                ]
+                route_maps_item = EosCliConfigGen.RouteMapsItem(name=neighbor["route_map_in"])
+                sequence_number = EosCliConfigGen.RouteMapsItem.SequenceNumbersItem(
+                    sequence=10, type="permit", match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match([f"ip address prefix-list {prefix_list_in}"])
+                )
+
                 # set no advertise is set only for WAN neighbors, which will also have prefix_list_in
                 if neighbor.get("set_no_advertise"):
-                    sequence_numbers[0]["set"] = ["community no-advertise additive"]
+                    sequence_number.set = EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Set(["community no-advertise additive"])
```
```suggestion
                    sequence_number.set.append("community no-advertise additive")
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -564,154 +484,153 @@ def ptp(self) -> dict | None:
...
             msg = f"ipv4_prefix_list_catalog[name={name}]"
             raise AristaAvdMissingVariableError(msg)
         return self.inputs.ipv4_prefix_list_catalog[name]
 
-    @cached_property
-    def route_maps(self) -> list | None:
-        route_maps = []
+    @structured_config_contributor
+    def route_maps(self) -> None:
         for neighbor in self.shared_utils.l3_bgp_neighbors:
```
ClausHolbechArista: Could this helper in shared_utils return class instances instead of dicts?

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -564,154 +484,153 @@ def ptp(self) -> dict | None:
...
-            route_maps.append({"name": neighbor["route_map_out"], "sequence_numbers": sequence_numbers})
+                route_maps_item.sequence_numbers.append_new(
+                    sequence=10,
+                    type="deny",
+                )
 
-        return route_maps or None
+            self.structured_config.route_maps.append(route_maps_item)
 
     @cached_property
```
ClausHolbechArista: Also change this decorator.

ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/tables/serial-number.md
```diff
@@ -0,0 +1,18 @@
+<!--
+  ~ Copyright (c) 2025 Arista Networks, Inc.
+  ~ Use of this source code is governed by the Apache License 2.0
+  ~ that can be found in the LICENSE file.
+  -->
+=== "Table"
+
+    | Variable | Type | Required | Default | Value Restrictions | Description |
```
gmuloc: should it be in its own table? I don't see this table included in the main doc so I feel like we would lose this

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc:
```suggestion
    @structured_config_contributor
```

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,13 @@ def _build_source_interfaces(
             # Check for duplicate VRF
             # inband_mgmt_vrf returns None in case of VRF "default", but here we want the "default" VRF name to have proper duplicate detection.
             inband_mgmt_vrf = self.shared_utils.inband_mgmt_vrf or "default"
-            if [source_interface for source_interface in source_interfaces if source_interface.get("vrf", "default") == inband_mgmt_vrf]:
+            if [source_interface for source_interface in source_interfaces if source_interface.vrf or inband_mgmt_vrf == "default"]:
```
gmuloc: I don't think this logic change is correct (though it is not failing any test so need to dig deeper):
* beforehand we were checking is source_interface.vrf (defaulting to default) was equal to inband_mgmt_vrf 
* now we check if there is a source_interface.vrf (any value) OR if the inband_mgmt_vrf is default

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -51,25 +66,13 @@ def _build_source_interfaces(
...
                 msg = f"Unable to configure multiple {error_context} source-interfaces for the same VRF '{inband_mgmt_vrf}'."
                 raise AristaAvdError(msg)
 
-            source_interface = {"name": self.shared_utils.inband_mgmt_interface}
-            if self.shared_utils.inband_mgmt_vrf not in [None, "default"]:
-                source_interface["vrf"] = self.shared_utils.inband_mgmt_vrf
-            source_interfaces.append(source_interface)
+            source_interfaces.append_new(
+                name=self.shared_utils.inband_mgmt_interface,
+                vrf=self.shared_utils.inband_mgmt_vrf if self.shared_utils.inband_mgmt_vrf != "default" else None,
```
gmuloc: so this is valid only because if `self.shared_utils.inband_mgmt_vrf` returns `None` we also inject `None` in vrf but it is not evident when reading this I think. I am partial to rewrite this to:

```suggestion
                vrf=self.shared_utils.inband_mgmt_vrf if self.shared_utils.inband_mgmt_vrf not in [None, "default"] else None,
```

for the sake of clarity. @ClausHolbechArista thoughts?

ansible_collections/arista/avd/roles/eos_cli_config_gen/docs/tables/serial-number.md
gmuloc: addressed in one of Claus comments

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: replaced with cast_as

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
ClausHolbechArista: I am ok with the current change. None will be the result for None so hey...

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
ClausHolbechArista: I think there is missing a paranthesis pair around the or.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: or not :)

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
gmuloc: typing is unhappy

gmuloc: only for basedpyright

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
gmuloc: fixed

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -276,94 +277,150 @@ def _get_ipv4_acl_field_with_substitution(field_value: str, replacements: dict[s
...
+
+        neighbor = EosCliConfigGen.RouterBgp.NeighborsItem(
+            ip_address=interface.peer_ip,
+            remote_as=interface.bgp.peer_as,
+            description=description,
+        )
+
+        # TODO: Remove when reviewing in Github
+        # This used to check for "or set_no_advertise" which was equivalent to "is_wan_interface"
+        # Above we raise if is_wan_interface but the In prefix list is not set so the OR was redundant.
```
gmuloc: To be removed by another reviewer as I wrote this ;)
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/base/utils.py
```diff
@@ -18,6 +18,7 @@
         "T_Source_Interfaces",
         EosCliConfigGen.IpHttpClientSourceInterfaces,
         EosCliConfigGen.IpDomainLookup.SourceInterfaces,
+        # Notice special handling for SSH since VRF is not optional there.
```
gmuloc: should this be removed now?

gmuloc:
```suggestion
```

Shivani-gslab: fixed

ClausHolbechArista (approved):

gmuloc (approved):

PR #5215, opened 2025-03-25
Refactor(eos_designs): Better handling of index error p2p_links nodes and other fields in core_interface

### LINKED ISSUE #4727: Better handling of index error p2p_links nodes and other fields in core_interface / l3_edge schema
```
  p2p_links:
    # Pe_P
    - nodes: [pe11-sav366, p11-brv350, pe11-sav366, p12-brv350]
      interfaces: [Ethernet43/1, Ethernet1/1, Ethernet44/1, Ethernet1/1]
      profile: pe_p_profile
      id: 1
```

This is accepted but fails sadly 

```
The full traceback is:
  File "/home/arista/.ansible/collections/ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py", line 100, in run
    output = get_structured_config(
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/structured_config/__init__.py", line 96, in get_structured_config
    results = eos_designs_module.render()
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/__init__.py", line 40, in render
    result_list.append(super().render())
                       ^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/avdfacts.py", line 64, in render
    return {key: getattr(self, key) for key in self.keys() if getattr(self, key) is not None}
                                                              ^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/functools.py", line 995, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/ethernet_interfaces.py", line 30, in ethernet_interfaces
    for p2p_link in self._filtered_p2p_links:
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/functools.py", line 995, in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py", line 67, in _filtered_p2p_links
    [p2p_link.update({"data": self._get_p2p_data(p2p_link)}) for p2p_link in p2p_links]
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arista/.local/share/pipx/venvs/pyavd/lib/python3.12/site-packages/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py", line 146, in _get_p2p_data
    "ip": ip[index],
          ~~^^^^^^^
IndexError: list index out of range
fatal: [p22-qzd516 -> localhost]: FAILED! => {
    "changed": false,
    "msg": "list index out of range"
}
```

First idea was to limit the llength in the list

Vibhu-gslab:
## Related Issue(s)

Fixes #4727 

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
```diff
@@ -119,10 +119,16 @@ def _get_p2p_data(self: AvdStructuredConfigCoreInterfacesAndL3EdgeProtocol, p2p_
         # Set descriptions or fallback to list with None values
         descriptions = p2p_link.descriptions or [None, None]
 
+        try:
+            ip_idx = ip[index]
+        except IndexError as exc:
+            msg = "IP_pools cannot allocate more than 2 ips but it can be done manually"
+            raise AristaAvdError(msg) from exc
+
```
gmuloc: Hmm I think it is too late here - we should do the logic in the `_resolve_p2p_ips` function.
Though if we have more than 2 nodes the peer index is also broken sadly

@ClausHolbechArista I have seen people use this model with more then 2 nodes in a same link happily but I am wondering if we should remove this possibility in 6.0. We should really have had `max_length` on node to be 2 in the schema but I feel it is breaking


@Vibhu-gslab the error message should be that if you need 3 ips then you need to give them (and make sure all lists are the right size..)

```
p2p_links:
  - nodes: [a, b, c]
    ip: [IP_a, IP_b, IP_c]
```

in your input on the link

gmuloc: @Vibhu-gslab after discussion with @ClausHolbechArista lets do the following:

1. the error message catching any `IndexError` (including the potential one for description line 135) should just output an error message stating `p2p_links model is intended to work for only two devices per entry.`)
2. Add a TODO in the schema for 6.0 stating we want to add a min_length of 2 and max_length of 2 to the lists (description, ip, nodes and interfaces)

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
```diff
@@ -259,7 +265,11 @@ def _update_common_interface_cfg(
                 )
 
         if p2p_link.ip:
-            interface.ip_address = p2p_link.ip[index]
+            try:
+                interface.ip_address = p2p_link.ip[index]
```
gmuloc: here we are trying to read again from p2p_link.ip but it feels we could read from p2p_link_data instead so we have already gone through the process of checking for the index.

If you do this, it means that we can catch the index issue only in the `_get_p2p_data` method which should limit the scope of where we try/catch.



(Makes me wonder if we could not further remove the need for p2p_link_data but that is not the purpose of this PR)

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
gmuloc: group in the try/catch block the checks for ip, peer_ip and description

Rename `ip` to `ips` as really this is a list:

```
 # Set ips or fallback to list with None values
 ips = p2p_link.ip or [None, None]
```

and then 

```
        try:
            ip = ips[index]
            peer_ip = ips[peer_index]
            description = descriptions[index] 
        except IndexError as exc:
            msg = "IP_pools cannot allocate more than 2 ips but it can be done manually"
            raise AristaAvdError(msg) from exc
```

and use these variables in `data` below.

Vibhu-gslab: Fixed!

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/utils.py
Vibhu-gslab: Getting it from p2p_link_data

gmuloc (approved):

Shivani-gslab (approved):
LGTM

PR #5254, opened 2025-04-04
Test(pyavd): Improve testing of schema validation

ClausHolbechArista:
Improve pytest of python-based schema validation.

Also clean up unused metaschema definitions. (Not used of our schemas and not implemented in validators)


gmuloc (commented):
need to remove .coverage

gmuloc (approved):
need to remove .coverage

PR #5278, opened 2025-04-08
Feat(anta_runner): Add ANTA DNS test

carl-baillargeon:
## Change Summary

- VerifyDNSServers

## Component(s) name

`arista.avd.anta_runner`

## Proposed changes
- Need `ip_name_servers` data model in the structured config.

## How to test
1- Configure DNS servers via `name_servers` in eos_designs or `ip_name_servers` in eos_cli_config_gen.
2- Generate the structured configurations
3- Run ANTA --> `ansible-playbook playbook/fabric-validate.yml`

### Repository Checklist


python-avd/pyavd/_anta/input_factories/services.py
```diff
@@ -0,0 +1,32 @@
...
+        ]
+        return [VerifyDNSServers.Input(dns_servers=natural_sort(dns_servers, sort_key="server_address"))] if dns_servers else None
+
+
+class VerifyDNSLookupInputFactory(AntaTestInputFactory):
+    """Input factory class for the `VerifyDNSLookup` test."""
+
+    def create(self) -> list[VerifyDNSLookup.Input] | None:
+        """Create a list of inputs for the `VerifyDNSLookup` test."""
+        # TODO: Fetch domain_names from the structured config?
```
MaheshGSLAB: We have dns_domain  key in structured config, may need to append it in domain_names

carl-baillargeon: I don't think so. The resolvability of `dns_domain` can vary too much depending on the deployment:

- It could be an internal-only domain so an external server will likely fail to resolve it
- Could be a placeholder, invalid, or an internal domain name that doesn't have records for the base domain itself

I think using a known, stable public FQDN (like www.arista.io) is the safest and most informative as it relies on the full DNS path.

Users can skip the test if their configured DNS servers cannot resolve public FQDN's (until we make `domain_names` configurable).

ClausHolbechArista: I don't think we should hardcode a dns name like this. Also in most cases I know of, the devices would not be able to look up public names, but only some internal resource. So IMO it should be a test input. And if not set, we don't run the test (to avoid having a default).

ansible_collections/arista/avd/molecule/anta_runner/anta/reports/filtered_run/anta_report.json
```diff
@@ -307,6 +307,28 @@
         "messages": [],
         "custom_field": "Generated by AVD"
     },
+    {
+        "name": "dc1-leaf1a",
+        "test": "VerifyDNSLookup",
+        "categories": [
+            "services"
+        ],
+        "description": "Verifies the DNS name to IP address resolution.",
```
ClausHolbechArista: Something seems off here, but someone with better English skills should take a look :)
```suggestion
        "description": "Verifies DNS name to IP address resolution.",
```

carl-baillargeon: :) We'll find something better and fix it on the ANTA side.

python-avd/pyavd/_anta/input_factories/services.py
carl-baillargeon: @ClausHolbechArista Maybe we can check the TerminAttr config for potential CV URLs?

carl-baillargeon: As discussed live we will remove this test for now.

carl-baillargeon: Fixed.

MaheshGSLAB (approved):

ClausHolbechArista (approved):

PR #5311, opened 2025-04-18
Refactor(eos_designs): Improve connected_endpoints/monitor_sessions

Vibhu-gslab:
## Change Summary

Add pytest coverage for network_services/ip_access_lists (Coverage: 100%)

## Related Issue(s)

Fixes 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Test was already added just fixed the `if` condition

## How to test
CI

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/monitor_sessions.py
+++ b/python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/monitor_sessions.py
@@ -39,7 +39,7 @@ class MonitorSessionsMixin(Protocol):
             for session_config in session_configs_list[1:]:
                 merged_settings._deepmerge(session_config)
 
-            if merged_settings.session_settings.access_group:
+            if merged_settings.session_settings._get("access_group"):
                 for session in session_configs_list:
                     if session.source_settings.access_group:
                         msg = (


Shivani-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/monitor_sessions.py
```diff
@@ -39,7 +39,7 @@ def monitor_sessions(self: AvdStructuredConfigConnectedEndpointsProtocol) -> Non
             for session_config in session_configs_list[1:]:
                 merged_settings._deepmerge(session_config)
 
-            if merged_settings.session_settings.access_group:
+            if merged_settings.session_settings._get("access_group"):
```
ClausHolbechArista, line 42: This is a behavioral change, where we loose the default access-group. Why do we need this?

ClausHolbechArista: We should test that only setting a source_settings ACL will not trigger the error with the old code.

ClausHolbechArista: you are correct. _get will avoid inserting the access-group object into the data, otherwise triggering the boolean check to be true in line 63. So your change is good.

ClausHolbechArista (approved):

PR #5317, opened 2025-04-21
Refactor(eos_designs): Improve pytest coverage for network_services/route_maps.py

MaheshGSLAB:
## Change Summary

Improve the pytest coverage for network_services/route_maps.py

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Improve the pytest coverage for network_services/route_maps.py. Coverage is 100% now

## How to test
Run tox

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -81,9 +81,6 @@ def _route_maps_vrf_default(self: AvdStructuredConfigNetworkServicesProtocol) ->
         self._redistribute_static_to_bgp_route_map()
 
     def _route_maps_vrf_default_check(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
-        if not self._vrf_default_evpn:
```
MaheshGSLAB, line 84: In network_services/router_bgp.py the method _route_maps_vrf_default_check is call and with that we are already checking it self._vrf_default_evpn so this condition is useless and never met.

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -158,8 +155,7 @@ def _evpn_export_vrf_default_route_map(self: AvdStructuredConfigNetworkServicesP
                     type="permit",
                     match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ip address prefix-list PL-STATIC-VRF-DEFAULT"]),
                 )
-        if not sequence_numbers:
```
MaheshGSLAB, line 161: There will be always sequence number because to be reach this method one of the if condition always required

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -188,9 +184,6 @@ def _bgp_underlay_peers_route_map(self: AvdStructuredConfigNetworkServicesProtoc
                 match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ip address prefix-list PL-STATIC-VRF-DEFAULT"]),
             )
 
-        if not sequence_numbers:
```
MaheshGSLAB, line 191: There will be always sequence number because to reach this method one of the if condition always required

Shivani-gslab (approved):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/missing-vrf-default-ipv4-subnets.yml
ClausHolbechArista, line 1: Please find a better name. This seems like a negative test.
Maybe `tenant-default-vrf-static-routes`

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
ClausHolbechArista: What if I have vrf default without any subnets or static routes? or if I only have ipv6?

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
MaheshGSLAB: This condition, defined in the `_evpn_export_vrf_default_route_map` method, is guaranteed to be true because the preceding checks in the _route_maps_vrf_default method ensure the host is either a WAN router, has a default VRF with a subnet, or has a default VRF with a static route. Otherwise, the code would have returned earlier.

MaheshGSLAB: Also i believe we can optimize this module later on because we are iterating the filtered_tenants more then 2 times

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/missing-vrf-default-ipv4-subnets.yml
MaheshGSLAB: Updated

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
gmuloc: I agree with this statement based on where the function is being called from today.

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
gmuloc: I agree with Mahesh regarding the fact that right now the function is only called when one of the three conditions (cf line 62) is true and so there will always be one sequence number.

That being said the code was built like this before a split I think and in case we are tented to reuse the method in another context, maybe we should mention it in the docstring?

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -81,9 +81,6 @@ def _route_maps_vrf_default(self: AvdStructuredConfigNetworkServicesProtocol) ->
         self._redistribute_static_to_bgp_route_map()
 
     def _route_maps_vrf_default_check(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
```
ClausHolbechArista, line 83:
```suggestion
    def _route_maps_vrf_default_check(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
        """This should only be called when self._vrf_default_evpn is true."""
```

ClausHolbechArista: similar for the others where we rely on other logic

MaheshGSLAB: updated the doc string

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/tenant-default-vrf-static-routes.yml
```diff
@@ -0,0 +1,29 @@
...
+    vrfs:
+      - name: default
+        vrf_id: 111
+        svis:
+          - id: 111
+            name: svi_with_default_vrf
+            tags: [fabric]
+        static_routes:
+          - destination_address_prefix: 10.0.0.0/8
+            gateway: 10.2.10.100
```
laxmikantchintakindi:
```suggestion
          - prefix: 10.0.0.0/8
            next_hop: 10.2.10.100
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/underlay-filter-redistribute-connected-with-route-map.yml
```diff
@@ -0,0 +1,29 @@
+---
+type: l3leaf
+
+underlay_filter_redistribute_connected: false
```
laxmikantchintakindi: Is it possible to use this key in tenant-default-vrf-static-routes instead of this new host?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/tenant-default-vrf-static-routes.yml
laxmikantchintakindi: This will be correct after rebasing with devel.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/underlay-filter-redistribute-connected-with-route-map.yml
MaheshGSLAB: If we add underlay_filter_redistribute_connected in the tenant-default-vrf-static-routes then the rest of code will not cover because it return with this input  and we will loose the coverage so kept two different hosts.

laxmikantchintakindi (approved):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/tenant-default-vrf-with-loopbacks.yml
```diff
@@ -1,6 +1,8 @@
 ---
 type: l3leaf
 
+underlay_filter_redistribute_connected: false
```
ClausHolbechArista: I feel like updating this will make us loose the `true` case where we have loopbacks in default vrf with evpn.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-with-loopbacks.cfg
ClausHolbechArista, line 1: I don't like these changes, since we are removing tests. Where is this tested now?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/underlay-filter-redistribute-connected-with-route-map.cfg
```diff
@@ -0,0 +1,99 @@
...
+   neighbor EVPN-OVERLAY-PEERS update-source Loopback0
+   neighbor EVPN-OVERLAY-PEERS bfd
+   neighbor EVPN-OVERLAY-PEERS ebgp-multihop 3
+   neighbor EVPN-OVERLAY-PEERS send-community
+   neighbor EVPN-OVERLAY-PEERS maximum-routes 0
+   neighbor IPv4-UNDERLAY-PEERS peer group
+   neighbor IPv4-UNDERLAY-PEERS route-map RM-BGP-UNDERLAY-PEERS-OUT out
+   neighbor IPv4-UNDERLAY-PEERS send-community
+   neighbor IPv4-UNDERLAY-PEERS maximum-routes 12000
+   redistribute connected
```
ClausHolbechArista: The name of the host indicates that we should redistribute connected with a route-map. But we have the opposite here.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/UPLINK_LAN_TESTS.yml
```diff
@@ -61,6 +61,13 @@ tenants: # dynamic_key: network_services
   - name: TEST
     mac_vrf_vni_base: 10000
     vrfs:
+      - name: default
```
ClausHolbechArista: adding this changes a lot of logic in the wan case. Do we still have tests without this elsewhere? Maybe just apply this to only one of the devices.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/tenant-default-vrf-with-loopbacks.yml
MaheshGSLAB: revert this changes and added new host for this test

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-with-loopbacks.cfg
MaheshGSLAB: revert the changes

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/underlay-filter-redistribute-connected-with-route-map.cfg
MaheshGSLAB: updated the host name

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/UPLINK_LAN_TESTS.yml
MaheshGSLAB: Added new host for this test and revert the changes from here

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/underlay-filter-redistribute-connected-vrf-default.yml
ClausHolbechArista, line 1: Maybe just call this `underlay-filter-redistribute-connected-false-svi-loopback` if that is what we test. I am not sure how the different tests here differentiate, so maybe make sure the naming is more clear and/or add some comments about what we are testing.

MaheshGSLAB: Updated the host name as underlay-filter-redistribute-connected-false-svi. 
Also added the comment in host file as:

Test the scenario where underlay_filter_redistribute_connected is false within the _route_maps_vrf_default_check function in network_services/route_maps.py

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/underlay-filter-redistribute-connected-false-svi.yml
```diff
@@ -0,0 +1,36 @@
+---
+
+# Test the underlay_filter_redistribute_connected: false for _route_maps_vrf_default_check in network services route_maps
```
MaheshGSLAB:
```suggestion
# Test the scenario where underlay_filter_redistribute_connected is false within the _route_maps_vrf_default_check function in network_services/route_maps.py
```

ClausHolbechArista (approved):

PR #5326, opened 2025-04-25
Feat(cv_deploy): New decorator class to handle various gRPC errors for CVClient

alexeygorbunov:
## Change Summary

Introduce decorator class for CVClient async methods to handle different GRPCErrors returned by CV (UNAVAILABLE, RESOURCE_EXHAUSTED, etc.).

## Related Issue(s)

Fixes [Issue](https://github.com/aristanetworks/avd-internal/issues/213)

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- Migrate logic from existing decorator function addressing gRPC messaging size
- Add logic to handle UNAVAILABLE errors by retransmitting with exponential delay
- Add extensive pytest coverage

## How to test
- `cv_deploy` molecule tests (to confirm overall operability of the role)
- `pytest` (python-avd/tests/pyavd/cv/client/test_async_decorators.py) to test all possible edge cases

```
Name                                             Stmts   Miss Branch BrPart  Cover   Missing
--------------------------------------------------------------------------------------------
pyavd/_cv/client/async_decorators.py               148      2     50      2    98%   244->204, 306-314
```

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +154,52 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
             raise LookupError(msg)
 
         return wrapper_cv_version
+
+
+def grpc_unavailable_handler(max_retries: int = 5, initial_delay: int = 1, factor: int = 2) -> Callable:
```
ClausHolbechArista: I would prefer if we could combine all our various needs into a single decorator. Ex. we could move the recasting of errors here, so we can simplify each method.

alexeygorbunov: Fixed. Merged all functionality (excluding version limiting) into a single decorator class.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,205 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+                return await self.func(*call_args, **call_kwargs)
+            except Exception as e:  # noqa: PERF203
+                if isinstance(e, GRPCError):
+                    status = e.status
+
+                    if status == Status.NOT_FOUND:
+                        raise CVResourceNotFound(*e.args)
+
+                    if status == Status.CANCELLED:
+                        raise CVTimeoutError(*e.args)
```
ClausHolbechArista: let's use match syntax

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,205 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+
+    async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs: dict) -> Any:
+        """Executes a single call to self.func with retry logic for gRPC UNAVAILABLE."""
+        func_name = self.func.__name__
+
+        for attempt in range(1, self.max_retries + 2):
+            try:
+                return await self.func(*call_args, **call_kwargs)
+            except Exception as e:  # noqa: PERF203
+                if isinstance(e, GRPCError):
```
ClausHolbechArista: let's negate this check so we don't have to scroll many lines to figure out the other outcome.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,205 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+        self.func_signature: Signature
+
+    def __call__(self, func: Callable) -> Callable:
+        self.func = func
+        self.func_signature = signature(func)
+
+        if self.list_field:
+            return_annotation = (
+                list
+                if type(self.func_signature.return_annotation) is str and self.func_signature.return_annotation.startswith("list")
```
ClausHolbechArista: use isinstance instead of type()

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Done

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Done

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Done. Switched to "match" here as well

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,206 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+                len(list_value),
+            )
+            # Use case where ratio is too high leading to the chuck_size being calculated as zero
+            if chunk_size == 0 and len(list_value) > 0:
+                chunk_size = 1
+
+            planned_attempts_qty = int((len(list_value) / chunk_size) + (1 if len(list_value) % chunk_size else 0))
+
+            if chunk_size >= len(list_value) and len(list_value) > 0:
+                LOGGER.exception(
```
ClausHolbechArista: I don't think we use exception anywhere else. We need to see how this is handled when running from Ansible.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,206 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+        list_field: str | None = None,
+        min_items_for_splitting_attempt: int = 2,
+    ) -> None:
+        self.max_retries: int = max_retries
+        self.initial_delay: int = initial_delay
+        self.factor: int = factor
+        self.list_field: str | None = list_field
+        self.min_items_for_splitting_attempt = max(2, min_items_for_splitting_attempt)
+        self.func: Callable
+        self.func_signature: Signature
```
ClausHolbechArista: Please move type hints outside of __init__. So on the class do `max_retries: int` etc

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,206 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+
+        bound_arguments = self.func_signature.bind(*original_call_args, **original_call_kwargs)
+        current_arguments_dict = bound_arguments.arguments
+
+        if self.list_field not in current_arguments_dict:
+            msg = (
+                f"{self.__class__.__name__} decorator is unable to find the list_field '{self.list_field}' "
+                f"in the given arguments to '{func_name}'. Found: '{list(current_arguments_dict.keys())}'."
+            )
+            raise KeyError(msg)
```
ClausHolbechArista: This check could be done in the __call__ when first applying the decorator. Check the signature has the field and that the field is a list (and not optional).

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +102,206 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+
+        list_value: list = current_arguments_dict[self.list_field]
+        if not isinstance(list_value, list):
+            msg = (
+                f"{self.__class__.__name__} decorator expected the value of the list_field '{self.list_field}' for function '{func_name}' "
+                f"to be a list. Got '{type(list_value)}'."
+            )
+            raise TypeError(msg)
+
+        LOGGER.info("%s: Preparing call for '%s' for list_field '%s' with %s item(s).", self.__class__.__name__, func_name, self.list_field, len(list_value))
```
ClausHolbechArista: Maybe messages like this is a bit much for info? We could move them to .debug.

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -103,14 +105,11 @@ async def get_studio(
...
-                # If we get here it means we got an entry with "removed: True" so no need to look further.
-                msg = "The studio was deleted in the workspace."
-                raise CVResourceNotFound(msg, f"Studio ID '{studio_id}, Workspace ID '{workspace_id}'")  # noqa: TRY301 TODO: Improve error handling
-        except Exception as e:
-            raise get_cv_client_exception(e, f"Studio ID '{studio_id}, Workspace ID '{workspace_id}'") or e
+        responses = client.get_all(request, metadata=self._metadata, timeout=timeout)
+        async for _response in responses:
+            # If we get here it means we got an entry with "removed: True" so no need to look further.
+            msg = "The studio was deleted in the workspace."
+            raise GRPCError(status=Status.NOT_FOUND, message=f"{msg}. Studio ID '{studio_id}, Workspace ID '{workspace_id}'")
```
ClausHolbechArista: I would prefer not to raise GRPCErrors from something that is not really a GRPC error. We could still raise CVResourceNotFound, or create a more specific error type if we like.

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -162,27 +159,24 @@ async def get_studio_inputs(
...
-            async for response in responses:
-                if response.value.inputs is None:
-                    continue
-
-                self._set_value_from_path(
-                    path=response.value.key.path.values,
-                    data=studio_inputs,
-                    value=json.loads(response.value.inputs),
-                )
+        # We use get_all since inputs can be larger than the maximum message size.
```
ClausHolbechArista, line 164:
```suggestion

        # We use get_all since inputs can be larger than the maximum message size.
```

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -322,8 +305,7 @@ async def get_studio_inputs_with_path(
         try:
             response = await client.get_one(request, metadata=self._metadata, timeout=timeout)
         except Exception as e:  # pylint: disable=broad-exception-caught
```
ClausHolbechArista: Since you are only interested in this type, we can improve this:
```suggestion
        except GRPCError as e:
```
And then just check the status.not_found.

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -272,8 +260,7 @@ async def get_studio_inputs_with_path(
...
```
ClausHolbechArista:
```suggestion
        except GRPCError as e:
```

python-avd/tests/pyavd/cv/client/test_async_decorators.py
```diff
@@ -96,14 +141,371 @@ async def test_msg_size_handler_invalid_fuction_return_type() -> None:
...
+    with pytest.raises(KeyError, match="GRPCRequestHandler decorator is unable to find the list_field .+"):
+        await GRPCRequestHandler(list_field="_field")(function_with_wrong_arg)(["foo", "bar"])
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    (
+        "failures",
+        "async_sleep_calls",
+        "logs_qty",
```
ClausHolbechArista: Can this not be found from the list of logs?

python-avd/tests/pyavd/cv/client/test_async_decorators.py
```diff
@@ -96,14 +141,371 @@ async def test_msg_size_handler_invalid_fuction_return_type() -> None:
...
+        "log_patterns",
+        "inner_exception",
```
ClausHolbechArista: Now this is the same on all, so you can remove it from inputs

python-avd/tests/pyavd/cv/client/test_async_decorators.py
```diff
@@ -57,18 +74,46 @@ async def version_limited_method(self) -> tuple[str, str]:  # noqa: F811
...
-            e.max_size = max_accepted_len
-            e.size = len(field)
-            raise e
+            raise GRPCError(status=Status.RESOURCE_EXHAUSTED, message=f"grpc: received message larger than max ({len(field)} vs. {max_accepted_len})")
 
         # return list with len of fields for this execution.
         return [len(field)]
 
+    @GRPCRequestHandler()
+    async def msgsize_unlimited_grpc_method(
```
ClausHolbechArista: Maybe just create different methods that always fail, raise or succeed instead of giving inputs like this. Would simplify everything else.

python-avd/tests/pyavd/cv/client/test_async_decorators.py
```diff
@@ -96,14 +141,371 @@ async def test_msg_size_handler_invalid_fuction_return_type() -> None:
...
+            id="SEVEN_GRPC_STATUS_UNAVAILABLE_FAILURES",
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    ("grpc_method", "extra_args"),
+    [
+        pytest.param("msgsize_unlimited_grpc_method", {}, id="UNLIMITED_SIZE_GRPC_METHOD"),
+        pytest.param("msgsize_limited_grpc_method", {"field": [0]}, id="LIMITED_SIZE_GRPC_METHOD"),
+    ],
```
ClausHolbechArista: IMO this is too "clever". Can we simplify this by just calling both in the test and have different asserts for each. Alternatively you can write different tests for each and then set the list of params as a constant that can be reused.

ClausHolbechArista (commented):
Looking good. Only minor things. The test things can be postponed to later if you prefer.

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: `ruff` doesn't allow to use `error` and forces `exception`:

```
ruff....................................................................................................................Failed
- hook id: ruff
- exit code: 1

python-avd/pyavd/_cv/client/async_decorators.py:272:17: TRY400 Use `logging.exception` instead of `logging.error`
    |
271 |               if chunk_size >= len(list_value) and len(list_value) > 0:
272 | /                 LOGGER.error(
273 | |                     "%s: Cannot split list_field '%s' for '%s' any further. Item count %s, calculated chunk_size %s.",
274 | |                     self.__class__.__name__,
275 | |                     self.list_field,
276 | |                     func_name,
277 | |                     len(list_value),
278 | |                     chunk_size,
279 | |                 )
    | |_________________^ TRY400
280 |                   raise
    |
    = help: Replace with `exception`

Found 1 error.
```

I see it already used in one place only: https://github.com/aristanetworks/avd/blob/8893c7377e8faf7123c5b518a99e8d8d77d20ae1/ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py#L99

I'm not seeing `error` level used anywhere as well. Should i fall back to emitting logs with `warning` level even if it's related to the faced error/exception?

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Done

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Done (partially). Moved to __call__ (of the decorator class) the validation of 'list_field' parameter presence  in decorated function. Also added to the __call__ the annotation validation for this parameter.

> and that the field is a list (and not optional)

This may be an issue as some of the currently decorated methods actually have this field annotated as a UnionType:

```
    @GRPCRequestHandler(list_field="configlet_ids")
    async def get_configlets(
        self: CVClientProtocol,
        workspace_id: str,
        configlet_ids: list[str] | None = None,     <<<<<<<<
        time: datetime | None = None,
        timeout: float = DEFAULT_API_TIMEOUT,
    ) -> list[Configlet]:
```

Therefore for annotation validation i included support for UnionType as long as it includes list in it.


At the same time i kept original type validation of the actual value of the list_field inside _execute_with_splitting as i believe these are two different problem vectors.

Added pytest coverage for these use cases

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Changed to `debug`

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Sure! Returned this back to `CVResourceNotFound`. In the decorator - re-raising if error is an instance of the base `CVClientException`

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Logic below needs to be adjusted then. Would this be OK?

```diff
         client = InputsServiceStub(self._channel)
         try:
             response = await client.get_one(request, metadata=self._metadata, timeout=timeout)
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            if isinstance(e, GRPCError) and e.status == Status.NOT_FOUND and workspace_id != "":
+        except GRPCError as e:  # pylint: disable=broad-exception-caught
+            if e.status == Status.NOT_FOUND and workspace_id != "":
                 # Ignore this error, since it simply means we have to check if inputs got deleted in this workspace or fetch from mainline as last resort.
                 pass
             else:
                 raise
+        except Exception as e:
+            msg = f"Unable to get Inputs '{input_path}' of Studio '{studio_id}' in workspace '{workspace_id}'."
+            raise CVClientException(msg) from e
         else:
             # We get here if no exception was raised.
             # We only get a response if the inputs are set/changed in the workspace.
```

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Done

```diff
         client = InputsServiceStub(self._channel)
         try:
             response = await client.get_one(request, metadata=self._metadata, timeout=timeout)
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            if isinstance(e, GRPCError) and e.status == Status.NOT_FOUND:
+        except GRPCError as e:
+            if e.status == Status.NOT_FOUND:
                 # Ignore this error, since it simply means we no inputs are in the studio so we will return the default value.
                 return default_value
             raise
+        except Exception as e:
+            msg = f"Unable to get Inputs '{input_path}' of Studio '{studio_id}' in workspace '{workspace_id}'."
+            raise CVClientException(msg) from e
 
         if response.value.inputs is not None:
             return json.loads(response.value.inputs)
```

python-avd/tests/pyavd/cv/client/test_async_decorators.py
alexeygorbunov: Removed this from things to assert. I was just trying to follow the suggestion to not calculate expected results inside test and to instead provide it as an input

python-avd/tests/pyavd/cv/client/test_async_decorators.py
alexeygorbunov: Removed

python-avd/tests/pyavd/cv/client/test_async_decorators.py
alexeygorbunov: Split both limited and unlimited into success/failure/exception methods

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: if it is an error we handle, we should not emit anything or maximum a warning. If it is an error we don't handle, we should raise and stop. In the case for the decorator we could give up and just warn + pass the original request along without splitting.

python-avd/pyavd/_cv/client/studio.py
ClausHolbechArista: why do you need the broad exception here? Shouldn't that be handled by the decorator?

python-avd/tests/pyavd/cv/client/test_async_decorators.py
ClausHolbechArista: right but do you need to assert the length when you provide the actual list?

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +104,238 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+        self.initial_delay = initial_delay
+        self.factor = factor
+        self.list_field = list_field
+        self.min_items_for_splitting_attempt = max(2, min_items_for_splitting_attempt)
+
+    def __call__(self, func: Callable) -> Callable:
+        self.func = func
+        self.func_signature = signature(func)
+
+        def _is_list_annotation(annotation: Any, strict: bool = False) -> tuple[bool, Any]:
```
ClausHolbechArista: move this to the class level as a @staticmethod, so we don't redefine it every time.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -150,3 +104,238 @@ async def wrapper_cv_version(*args: Any, **kwargs: Any) -> list:
...
+                else annotation
+            )
+
+            return _string_based_annotation is list or get_origin(annotation) is list, _string_based_annotation
+
+        if self.list_field:
+            if not (return_annotation := _is_list_annotation(self.func_signature.return_annotation, strict=True))[0]:
+                msg = (
+                    f"GRPCRequestHandler decorator is unable to bind to the function '{func.__name__}'. "
+                    f"Expected a return type of 'list'. Got '{return_annotation[1]}'."
```
ClausHolbechArista:
```suggestion
                    f"GRPCRequestHandler decorator is unable to bind to the function '{func.__name__}' with the 'list_field' argument. "
                    f"Expected a return type of 'list'. Got '{return_annotation[1]}'."
```

python-avd/pyavd/_cv/client/async_decorators.py
ClausHolbechArista: if you wish to raise, you should build an error type that covers the info you need and avoid logging it as well.

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -271,13 +261,15 @@ async def get_studio_inputs_with_path(
         client = InputsServiceStub(self._channel)
         try:
             response = await client.get_one(request, metadata=self._metadata, timeout=timeout)
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            e = get_cv_client_exception(e, f"Studio ID '{studio_id}, Workspace ID '{workspace_id}', Path '{input_path}'") or e
-            if isinstance(e, CVResourceNotFound) and workspace_id != "":
+        except GRPCError as e:  # pylint: disable=broad-exception-caught
```
ClausHolbechArista:
```suggestion
        except GRPCError as e:
```

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -271,13 +261,15 @@ async def get_studio_inputs_with_path(
...
+            if e.status == Status.NOT_FOUND and workspace_id != "":
                 # Ignore this error, since it simply means we have to check if inputs got deleted in this workspace or fetch from mainline as last resort.
                 pass
             else:
                 raise
+        except Exception as e:
+            msg = f"Unable to get Inputs '{input_path}' of Studio '{studio_id}' in workspace '{workspace_id}'."
```
ClausHolbechArista: This should be handled by the decorator

python-avd/pyavd/_cv/client/studio.py
```diff
@@ -321,17 +309,20 @@ async def get_studio_inputs_with_path(
...
-        except Exception as e:  # pylint: disable=broad-exception-caught
-            e = get_cv_client_exception(e, f"Studio ID '{studio_id}, Workspace ID '{workspace_id}', Path '{input_path}'") or e
-            if isinstance(e, CVResourceNotFound):
+        except GRPCError as e:
+            if e.status == Status.NOT_FOUND:
                 # Ignore this error, since it simply means we no inputs are in the studio so we will return the default value.
                 return default_value
             raise
+        except Exception as e:
+            msg = f"Unable to get Inputs '{input_path}' of Studio '{studio_id}' in workspace '{workspace_id}'."
```
ClausHolbechArista: Thius should be handled by the decorator

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: I removed this block completely as it should never be used in the end (this use case should always be prevented by previous checks as long as `min_items_for_splitting_attempt` is at least 2 (which is already enforced in __init__)).
For another case where i tried to use logger.exception as well - i removed logging and converted it to msg which is passed into the custom (CV) exception

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Fixed/removed

python-avd/pyavd/_cv/client/studio.py
alexeygorbunov: Fixed/removed

python-avd/tests/pyavd/cv/client/test_async_decorators.py
alexeygorbunov: no, probably not, removed this.

ClausHolbechArista (approved):
LGTM. Thank you for implementing these improvements.
Only comment is that we may not need a configurable `min_items_for_splitting_attempt`, but now we have it so we can see if we ever need to use it. Thanks again.

ClausHolbechArista:
@joelbreton2 can you test this? I suggest you first run the CI workflows towards cv-staging with the `devel` code, and try to get unavailable errors. Then run the same workflows with this PR code with -v and look for retries in the logs.
We expect the workflow to go through without issues in the latter case, but failing in the first.

gmuloc (approved):
Agreeing with the changes - logging need a more thorought check.

Locally with cv_deploy scenario I can see the warnings printed during the run. I would expect the same in molecule

PR #5333, opened 2025-04-25
Feat(cv_deploy): Handle WorkspaceStreamResponse with ResponseStatus.UNSPECIFIED

alexeygorbunov:
## Change Summary

Properly handle WorkspaceStreamResponses in an intermediate state.
This addresses new CV behavior where response in an intermediate state is now published once request is picked by the backend.

## Related Issue(s)

- N/A

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- Avoid returning unless response matching request_id reached terminal state (SUCCESS/FAIL)

## How to test
`cv_deploy` CI molecule test

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_cv/client/workspace.py
+++ b/python-avd/pyavd/_cv/client/workspace.py
@@ -11,6 +11,7 @@ from pyavd._cv.api.arista.workspace.v1 import (
     Request,
     RequestParams,
     Response,
+    ResponseStatus,
     Workspace,
     WorkspaceConfig,
     WorkspaceConfigDeleteRequest,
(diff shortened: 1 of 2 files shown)


python-avd/pyavd/_cv/client/workspace.py
```diff
@@ -246,12 +247,20 @@ async def wait_for_workspace_response(
...
                 if request_id in response.value.responses.values:
-                    LOGGER.info("wait_for_workspace_response: Got response for request '%s': %s", request_id, response.value.responses.values[request_id])
-                    return response.value.responses.values[request_id], response.value
-                LOGGER.debug(
-                    "wait_for_workspace_response: Got workspace update but not for request_id '%s'. Workspace State: %s",
-                    request_id,
-                    response.value.state,
-                )
+                    LOGGER.info(
+                        "wait_for_workspace_response: Got response for request '%s' at '%s' UTC: %s",
```
ClausHolbechArista: Instead of adding timestamps to the logs, we can just tell the logger to spit out timestamps as well. Why did you need the time here?

python-avd/pyavd/_cv/client/workspace.py
```diff
@@ -246,12 +247,20 @@ async def wait_for_workspace_response(
...
+                        request_id,
+                        response.time,
+                        response.value.responses.values[request_id],
+                    )
+                    if response.value.responses.values[request_id].status != ResponseStatus.UNSPECIFIED:
```
ClausHolbechArista, line 252: Please update line 228 to briefly explain the new behavior

ClausHolbechArista: also missing a "for" in line 231

ClausHolbechArista:
Tested this PR and it works as expected.

python-avd/pyavd/_cv/client/workspace.py
alexeygorbunov: Added mostly for a better visibility but now remember we didn't want these timestamps in the past, so removing them from here

python-avd/pyavd/_cv/client/workspace.py
alexeygorbunov: Done

marcusmcpeek (approved):
This change worked for me against cv-staging.

ClausHolbechArista (approved):

PR #5338, opened 2025-04-28
Feat(eos_designs): Add more feature support toggles for platform_settings

alexeygorbunov:
## Change Summary

Enforce platform-based limitations for the following EOS features:
 - `queue-monitor`. Affected AVD eos_designs knobs: `queue_monitor_length`, `queue_monitor_streaming`
 - `hardware speed-groups`. Affected AVD eos_designs knobs: `platform_speed_groups`
 - `hardware counter`. Affected AVD eos_designs knobs: `hardware_counters.features.[]`
 - `l2 mtu`. Affected AVD eos_designs knobs: `<connected_endpoints_keys.key>.[].adapters.[].l2_mtu`, `network_ports.[].l2_mtu`
 - `l2 mru`. Affected AVD eos_designs knobs: `<connected_endpoints_keys.key>.[].adapters.[].l2_mru`, `network_ports.[].l2_mru`

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## How to test
- pytest/molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,9 +84,15 @@ keys:
         feature_support:
           type: dict
           keys:
+            queue_monitor:
+              type: bool
+              default: true
```
gmuloc, line 92: maybe we should add a description to explain the relation to queue_monitor_length_notify? (seems like queue_monitor is a superset of the other - what happens if I set queue_monitor to false and queue_monitor_length_notify to true?)

same for bgp / bgp_wait_install

alexeygorbunov: I removed this and other per-feature "parent" keys and defined instead "hardware_features" abstracted/aggregated feature which controls supportability of all hardware-dependent features. Setting it to `false` disables all of them together.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
...
           keys:
+            hardware_features:
+              type: bool
+              default: true
+              description: |
+                An abstracted feature indicating aggregated support for the following common EOS capabilities supported on hardware platforms only:
+                  - `queue-monitor`. Included structured configuration paths: `queue_monitor_length`, `queue_monitor_streaming`.
+                  - `storm-control`. Included structured configuration paths: `ethernet_interfaces.[].storm_control`, `port_channel_interfaces.[].storm_control`.
+                  - `PoE`. Included structured configuration paths: `ethernet_interfaces.[].poe`.
+                  - `BGP wait-install`. Included structured configuration paths: `router_bgp.updates.wait_install`, `router_bgp.vrfs.[].updates.wait_install`.
```
ClausHolbechArista: Let's not include this here, since it directly overlaps with another setting. This should only cover things that are not already covered.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
...
           type: dict
           keys:
+            hardware_features:
+              type: bool
+              default: true
+              description: |
+                An abstracted feature indicating aggregated support for the following common EOS capabilities supported on hardware platforms only:
+                  - `queue-monitor`. Included structured configuration paths: `queue_monitor_length`, `queue_monitor_streaming`.
+                  - `storm-control`. Included structured configuration paths: `ethernet_interfaces.[].storm_control`, `port_channel_interfaces.[].storm_control`.
+                  - `PoE`. Included structured configuration paths: `ethernet_interfaces.[].poe`.
```
ClausHolbechArista: Let's not include this here, since it directly overlaps with another setting. This should only cover things that are not already covered.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
         feature_support:
           type: dict
           keys:
+            hardware_features:
+              type: bool
+              default: true
+              description: |
+                An abstracted feature indicating aggregated support for the following common EOS capabilities supported on hardware platforms only:
+                  - `queue-monitor`. Included structured configuration paths: `queue_monitor_length`, `queue_monitor_streaming`.
+                  - `storm-control`. Included structured configuration paths: `ethernet_interfaces.[].storm_control`, `port_channel_interfaces.[].storm_control`.
```
ClausHolbechArista: I think we can cover this by the existing `interface_storm_control`. We don't support any global storm-control stuff in eos_designs yet, so there is no breaking change. You can add a new feature called `storm_control` and deprecate the old if you like. That can be done in a separate PR to keep things simple here.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
         feature_support:
           type: dict
           keys:
+            hardware_features:
+              type: bool
+              default: true
+              description: |
+                An abstracted feature indicating aggregated support for the following common EOS capabilities supported on hardware platforms only:
+                  - `queue-monitor`. Included structured configuration paths: `queue_monitor_length`, `queue_monitor_streaming`.
```
ClausHolbechArista: We should rather say which config it includes. Regular users don't know structured config by heart.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
         feature_support:
           type: dict
           keys:
+            hardware_features:
```
ClausHolbechArista: Call this `hardware_dependent`.

After addressing the other comments, to avoid overlap, we can reconsider if we even need this group, since it may only have 3 or 4 left here. In that case adding them as separate entries may be simpler and more intuitive for the user.

The advantage of having things separate is that users can flip these individually as EOS evolves. We can only change the defaults for platforms in major versions.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
...
+                  - `storm-control`. Included structured configuration paths: `ethernet_interfaces.[].storm_control`, `port_channel_interfaces.[].storm_control`.
+                  - `PoE`. Included structured configuration paths: `ethernet_interfaces.[].poe`.
+                  - `BGP wait-install`. Included structured configuration paths: `router_bgp.updates.wait_install`, `router_bgp.vrfs.[].updates.wait_install`.
+                  - `hardware speed-group`. Included structured configuration paths: `hardware.speed_groups`.
+                  - `hardware counter feature` for all features except `acl`, `mpls tunnel`, `nexthop`, `pbr`, `vni (decap|encap)`, `vtep (decap|encap)`. Included structured configuration paths: `hardware_counters.features.[]`.
+                  - `L2 MRU`. Included structured configuration paths: `ethernet_interfaces.[].l2_mru`, `port_channel_interfaces.[].l2_mru`.
+                Feature should typically be set to "False" for all virtual EOS platforms (e.g., vEOS-lab, cEOSLab) and cloudEOS.
+            hardware_and_cloudeos_features:
```
ClausHolbechArista: If this ends up only having L2 MTU, we should just have a feature flag for `l2_mtu` to keep things simple.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -299,21 +321,21 @@ keys:
...
           bgp_update_wait_for_convergence: false
-          bgp_update_wait_install: false
-          interface_storm_control: false
-          queue_monitor_length_notify: false
           evpn_gateway_all_active_multihoming: true
         management_interface: Management0
         reload_delay:
           mlag: 300
           non_mlag: 330
       - platforms:
```
ClausHolbechArista: FYI we intend to make this a case-insensitive regex match - tracked in some other issues.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -131,6 +151,9 @@ keys:
               type: bool
               default: false
               description: Support for all-active EVPN gateway redundancy.
+            hardware_counter_feature:
```
ClausHolbechArista: Avoid adding `_feature` since we are already under `feature_support`.
```suggestion
            hardware_counters:
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -84,6 +84,26 @@ keys:
         feature_support:
```
ClausHolbechArista, line 84: Please add something in the description for each feature whether it will ignore the feature or it will raise an error.

ClausHolbechArista (changes requested):
In general, I would prefer split these into small feature_support flags, and implement them in separate PRs, so you don't have as much back and forth. But if you prefer to keep them here, that is also fine for me.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: this `feature` in this case is the actual CLI key

```
veos-4-33-0f(config-s-sac)#hardware counter ?
  feature  Counter feature
veos-4-33-0f(config-s-sac)#
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Added. Example: `Feature will be ignored on unsupported platforms.`

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Added comment regarding dependency between `queue_monitor` and `queue_monitor_length_notify`.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: thanks!

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Completely removed this aggregated feature and now use separate specific features instead.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Completely removed this aggregated feature and now use separate specific features instead.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Fixed. Using existing `interface_storm_control `

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Removed this section completely. Every specific feature relies on it's own description.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Removed this aggregated feature.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Removed this aggregated feature

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ClausHolbechArista: we don't need to follow CLI here. Also this may cover more CLI in the future.

ClausHolbechArista: What we are tracking here are platform capabilities, and the capability is hardware counters.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,45 @@ def router_multicast(self) -> None:
...
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.hardware_counter.supported:
+            self.custom_structured_configs.root.append(EosCliConfigGen._from_dict({"hardware_counters": None}))
+            return
+        hardware_counters = self.inputs.hardware_counters._cast_as(EosCliConfigGen.HardwareCounters)
+
+        # Filter different hardware counter features basedon the platform supportability
+        if self.shared_utils.platform_settings.feature_support.hardware_counter.feature.supported:
+            hardware_counters.features = hardware_counters.features._filtered(
+                lambda x: (
```
ClausHolbechArista: please don't use single-letter variables. Use something meaningful instead, so the code is easier to read.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,45 @@ def router_multicast(self) -> None:
...
+                    x.name != "mpls lfib"
+                    and get_v2(
```
ClausHolbechArista: All this is super complex. It would have been much simpler with just a list of strings like `unsupported_features: ["mpls lfib"]` under the platform setting. Also the schema becomes much simpler with less maintenance.

ClausHolbechArista: For the unsupported_features items you may be able to use a $ref to pick up the valid_values from the hardware_counters schema.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -413,7 +463,13 @@ def mac_address_table(self) -> None:
 
     @structured_config_contributor
     def queue_monitor_streaming(self) -> None:
-        """queue_monitor_streaming set based on queue_monitor_streaming data-model."""
+        """queue_monitor_streaming set based on queue_monitor_streaming data-model and platform_settings.feature_support.queue_monitor fact."""
+        if not self.inputs.queue_monitor_streaming:
+            return
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
+            self.custom_structured_configs.root.append(EosCliConfigGen._from_dict({"queue_monitor_streaming": None}))
```
ClausHolbechArista: Take a look at _from_null instead. Would be great if we could avoid creating a full EosCliConfigGen instance, but instead just override the combined one.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -302,18 +541,53 @@ keys:
           bgp_update_wait_for_convergence: false
           bgp_update_wait_install: false
           interface_storm_control: false
+          queue_monitor: false
           queue_monitor_length_notify: false
```
ClausHolbechArista: no longer needed when you disable queue monitor.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -288,8 +499,36 @@ keys:
           bgp_update_wait_for_convergence: false
```
ClausHolbechArista: Changing the default values could be considered a breaking change. Can we move the change of defaults to it's own PR, so we can discuss that separately? Anyone wanting it sooner can just add their own `custom_platform_settings` once this one has merged.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: sure, renamed to `hardware_counters`

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Like this?
```diff
- self.custom_structured_configs.root.append(EosCliConfigGen._from_dict({"queue_monitor_streaming": None}))
+ self.custom_structured_configs.nested.queue_monitor_streaming = EosCliConfigGen.QueueMonitorStreaming._from_null()
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -131,6 +160,188 @@ keys:
...
+                    General supportability of the `hardware counter` feature.
+                    Setting this key to `false` for the specific platform will disable support of any specific hardware counter functionality (`feature`, `drop`, etc.) for this platform.
+                  default: true
+                feature:
+                  type: dict
+                  keys:
+                    supported:
+                      type: bool
+                      description: |-
+                        General supportability of the `hardware counter feature` feature.
```
gmuloc: I am not a native English speeaker but supportability feels a bit clunky not sure if my suggestion is better though :)
```suggestion
                        General support of the `hardware counter feature` feature.
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: yeah I think that will be more efficient. Thanks

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
ClausHolbechArista:
```suggestion
                        General support for the `hardware counter` CLI.
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: fixed, replaced with `feature`

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Removed this complexity related to `mpls lfib` with `units` flag set for cEOS/vEOS.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: right, returning default instance if whole `queue_monitor` is disabled:

```
    def queue_monitor_length(self) -> None:
        ...
        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
            self.custom_structured_configs.nested.queue_monitor_length = EosCliConfigGen.QueueMonitorLength._from_null()
            return
```

But both are needed in case only "nofity" sub-feature of `queue-monitor` is not supported

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Rolled back all changes (related to the feature support) under existing platform definitions. Use `custom_platform_settings` for tests

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Revised wording but suggested `scope` may not accurately reflect real impact.
Please check if updated description makes sense

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -103,22 +179,45 @@
...
+
+          # Support for LANZ notifying mode. Requires the parent `queue_monitor` feature to be enabled.
+          # Feature will be ignored on unsupported platforms.
           queue_monitor_length_notify: <bool; default=True>
+
+          # Support for configuration of per interface storm-control settings on Ethernet or Port-channel interfaces.
+          # Feature will be ignored on unsupported platforms.
           interface_storm_control: <bool; default=True>
+
+          # Support for configuration of per interface PoE settings on Ethernet or Port-channel interfaces.
```
ClausHolbechArista:
```suggestion
          # Support for PoE.
```

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -103,22 +179,45 @@
...
+
+          # Support for LANZ.
+          # Feature will be ignored on unsupported platforms.
+          queue_monitor: <bool; default=True>
+
+          # Support for LANZ notifying mode. Requires the parent `queue_monitor` feature to be enabled.
+          # Feature will be ignored on unsupported platforms.
           queue_monitor_length_notify: <bool; default=True>
+
+          # Support for configuration of per interface storm-control settings on Ethernet or Port-channel interfaces.
```
ClausHolbechArista:
```suggestion
          # Support for storm-control.
```

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -132,6 +231,113 @@
 
           # Support for all-active EVPN gateway redundancy.
           evpn_gateway_all_active_multihoming: <bool; default=False>
+
+          # Support for enabling counters for features using programmable hardware counter resources.
+          # Feature will be ignored on unsupported platforms.
+          hardware_counters:
+
+            # General support for the `hardware counter` CLI.
+            # Setting this key to `false` for the specific platform will disable support of any specific hardware counter functionality (`feature`, `drop`, etc.) for this platform.
```
ClausHolbechArista:
```suggestion
            # Setting this key to `false` for a platform will disable support for all hardware counter functionality (`feature`, `drop`, etc.) for this platform.
```

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -210,6 +439,113 @@
...
+            supported: <bool; default=True>
+            feature:
+
+              # General support for the `hardware counter feature` CLI.
+              # Setting this key to `false` for the specific platform will disable support of any specific hardware counter feature for this platform.
+              supported: <bool; default=True>
```
ClausHolbechArista: duplicate of the parent level.

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -210,6 +439,113 @@
...
+
+              # Support for the hardware counter for the `acl` feature.
```
ClausHolbechArista: Since all these descriptions just repeats the key name I think we can move the description to a generic one on `feature` and remove them from each key here.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,35 @@ def router_multicast(self) -> None:
...
+          - hardware_counters.features variable.
+          - platform_settings.feature_support.hardware_counters fact.
+        """
+        if not self.inputs.hardware_counters:
+            return
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.hardware_counters.supported:
+            self.custom_structured_configs.nested.hardware_counters = EosCliConfigGen.HardwareCounters._from_null()
+            return
+        hardware_counters = self.inputs.hardware_counters._cast_as(EosCliConfigGen.HardwareCounters)
```
ClausHolbechArista: I don't think you need the cast as here, since the hardware_counters is reusing the eos cli config gen class. (Check if pyright complains when you remove the cast_as)

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,35 @@ def router_multicast(self) -> None:
...
-        self.structured_config.hardware_counters = self.inputs.hardware_counters
+        """
+        Set hardware_counters.
+
+        Contributing data sources:
+          - hardware_counters.features variable.
+          - platform_settings.feature_support.hardware_counters fact.
+        """
+        if not self.inputs.hardware_counters:
+            return
```
ClausHolbechArista, line 187: You probably need to adjust this a bit, in case you want to block out eos_cli_config_gen inputs.

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
ClausHolbechArista: If this is a different support item than overall hardware_counters we should consider moving it out so we have 
```yaml
hardware_counters: <bool>
hardware_counters_features:
  supported: <bool>
  features:
    ...
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,35 @@ def router_multicast(self) -> None:
...
+        if self.shared_utils.platform_settings.feature_support.hardware_counters.feature.supported:
+            hardware_counters.features = hardware_counters.features._filtered(
+                lambda feature: get_v2(
+                    self.shared_utils.platform_settings.feature_support.hardware_counters.feature,
+                    feature.name.replace(" ", "_").replace("-", "_"),
+                    # Assume all uncovered/new features are supported
+                    default=True,
+                )
+            )
+        elif hasattr(hardware_counters, "features"):
```
ClausHolbechArista: I don't think this check will work as you expect since it is a class. It will always delete it. (This is because we override getattr in our classes)
```suggestion
        elif hardware_counters.features:
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -173,6 +200,8 @@ def hardware(self) -> None:
 
         Converting nested dict to list of dict to support avd_v4.0.
         """
+        if not self.shared_utils.platform_settings.feature_support.hardware_speed_group:
```
ClausHolbechArista, line 218: Here you are not overriding cli gen. We need to find a consistent aproach.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -288,9 +317,20 @@ def load_interval(self) -> None:
...
+        Set queue_monitor_length.
+
+        Contributing data sources:
+          - queue_monitor_length data-model
+          - platform_settings.feature_support.queue_monitor fact
+          - platform_settings.feature_support.queue_monitor_length_notify fact.
+        """
         if not self.inputs.queue_monitor_length:
             return
+        # Avoid collision with eos_cli_config_gen
```
ClausHolbechArista: This has no effect since you already skip out above.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
```diff
@@ -289,3 +289,23 @@ def _get_adapter_phone(
             raise AristaAvdError(msg)
 
         return output_type(trunk=adapter.phone_trunk_mode, vlan=adapter.phone_vlan)
+
+    def _get_adapter_l2_mtu(
+        self: AvdStructuredConfigConnectedEndpointsProtocol,
+        adapter: EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem.ConnectedEndpointsItem.AdaptersItem,
+    ) -> int | None | UndefinedType:
```
ClausHolbechArista: You can just use None here instead of Undefined.

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
alexeygorbunov: All occurrences fixed

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
alexeygorbunov: I wanted `hardware_counters.supported` flag to control overall support for hardware counters (features + drop). `hardware_counters.feature.supported` to control overall support for hardware counters features. `hardware_counters.feature.X` - to control support for a specific feature.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Would section right below this not do this?

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: True, fixed

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
alexeygorbunov: Changed code based on the suggested data model

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
alexeygorbunov: Done

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: eos_designs (`platform_speed_groups`) and eos_cli_config_gen (`hardware.speed_groups.[].(speed_group|serdes)`) keys are not overlapping for this data model

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: On the line above we are returning None in case there is no input for `queue_monitor_length`. Line below covers the case where there is a valid input from the user but platform does not support this feature and we need to return None at the same time patching structured config (as eos_designs and eos_cli_config_gen overlaps in this case)

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/utils.py
alexeygorbunov: Changed in both funcs

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: ah yes, that is why this is different. Maybe elaborate a bit in the comment.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: Ack. I forgot the background of overriding in the other one. Thanks.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,39 @@ def router_multicast(self) -> None:
...
+        Contributing data sources:
+          - hardware_counters.features variable.
+          - platform_settings.feature_support.hardware_counters fact.
+          - platform_settings.feature_support.hardware_counters_features fact.
+        """
+        if not self.inputs.hardware_counters:
+            return
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.hardware_counters:
+            self.custom_structured_configs.nested.hardware_counters = EosCliConfigGen.HardwareCounters._from_null()
```
```suggestion
        if not self.shared_utils.platform_settings.feature_support.hardware_counters:
            # Since we use the same data model in eos_cli_config_gen, it would pick up the input vars unless we explicitly set it to null.
            self.custom_structured_configs.nested.hardware_counters = EosCliConfigGen.HardwareCounters._from_null()
```

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
ClausHolbechArista: Thank you.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -137,6 +166,121 @@ keys:
               type: bool
               default: false
               description: Support for all-active EVPN gateway redundancy.
```
ClausHolbechArista:
```suggestion
              description: |-
                Support for all-active EVPN gateway redundancy.
                An error will be raised if the feature is enabled for unsupported platforms.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -137,6 +166,121 @@ keys:
...
+                Feature will be ignored on unsupported platforms.
+              default: true
+            hardware_counters_features:
+              type: dict
+              keys:
+                supported:
+                  type: bool
+                  description: |-
+                    General support for the `hardware counter feature` CLI.
+                    Setting this key to `false` for the specific platform will disable support of any specific hardware counter feature for this platform.
```
ClausHolbechArista:
```suggestion
                    Setting this key to `false` for the specific platform will ignore all hardware counter features for this platform.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -137,6 +166,121 @@ keys:
...
+                  default: true
+                features:
+                  type: dict
+                  description:  Per-feature support for the hardware counters.
```
ClausHolbechArista:
```suggestion
                  description: |-
                    Per-feature support for the hardware counters.
                    Feature will be ignored on unsupported platforms.
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,39 @@ def router_multicast(self) -> None:
...
+            return
+        hardware_counters = self.inputs.hardware_counters
```
ClausHolbechArista: Since we may be in-place updating.
```suggestion
        hardware_counters = self.inputs.hardware_counters._deepcopy()
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,39 @@ def router_multicast(self) -> None:
...
+                    feature.name.replace(" ", "_").replace("-", "_"),
+                    # Assume all uncovered/new features are supported
+                    default=True,
+                )
+            )
+        else:
+            if hardware_counters.features:
+                del hardware_counters.features
+            # Avoid collision with eos_cli_config_gen
+            self.custom_structured_configs.nested.hardware_counters.features = EosCliConfigGen.HardwareCounters.Features._from_null()
```
ClausHolbechArista: I don't think we need this here, since this is not a root-level key, so our generated key already takes precendence.
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -288,9 +321,20 @@ def load_interval(self) -> None:
...
+        Contributing data sources:
+          - queue_monitor_length data-model
+          - platform_settings.feature_support.queue_monitor fact
+          - platform_settings.feature_support.queue_monitor_length_notify fact.
+        """
         if not self.inputs.queue_monitor_length:
             return
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
+            self.custom_structured_configs.nested.queue_monitor_length = EosCliConfigGen.QueueMonitorLength._from_null()
```
```suggestion
        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
            # Since we use the same data model in eos_cli_config_gen, it would pick up the input vars unless we explicitly set it to null.
            self.custom_structured_configs.nested.queue_monitor_length = EosCliConfigGen.QueueMonitorLength._from_null()
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -487,7 +531,13 @@ def mac_address_table(self) -> None:
 
     @structured_config_contributor
     def queue_monitor_streaming(self) -> None:
-        """queue_monitor_streaming set based on queue_monitor_streaming data-model."""
+        """queue_monitor_streaming set based on queue_monitor_streaming data-model and platform_settings.feature_support.queue_monitor fact."""
+        if not self.inputs.queue_monitor_streaming:
+            return
+        # Avoid collision with eos_cli_config_gen
+        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
+            self.custom_structured_configs.nested.queue_monitor_streaming = EosCliConfigGen.QueueMonitorStreaming._from_null()
```
```suggestion
        if not self.shared_utils.platform_settings.feature_support.queue_monitor:
            # Since we use the same data model in eos_cli_config_gen, it would pick up the input vars unless we explicitly set it to null.
            self.custom_structured_configs.nested.queue_monitor_streaming = EosCliConfigGen.QueueMonitorStreaming._from_null()
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
ClausHolbechArista: Maybe there is a corner case if you _only_ give features, where we could end up with an empty dict.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Suggested fix applied

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Changed

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: - Following our discussion - converted schema to the following view:

```yaml
            hardware_counters:
              type: bool
              description: |-
                Support for enabling counters using programmable hardware counter resources.
                Setting this key to `false` for the specific platform will ignore all hardware counter features for this platform.
              default: true
            hardware_counter_features:
              type: dict
              description:  |-
                Per-feature support for the hardware counters.
                Feature will be ignored on unsupported platforms.
              keys:
                acl:
                  type: bool
...
```

- Modified `def hardware_counters` to reflect schema changes and to cover use case where features are supported in general but each feature is explicitly disabled for the platform

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,38 @@ def router_multicast(self) -> None:
...
+        # Filter different hardware counter features based on the platform supportability
+        hardware_counters.features = hardware_counters.features._filtered(
+            lambda feature: get_v2(
+                self.shared_utils.platform_settings.feature_support.hardware_counter_features,
+                feature.name.replace(" ", "_").replace("-", "_"),
+                # Assume all uncovered/new features are supported
+                default=True,
+            )
+        )
+        # Use case where all specific features are filtered out leaving an empty list
```
ClausHolbechArista, line 203: nice catch! but what if features are empty but the log is set? You should probably only override the features to the `null` value here.

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -163,8 +163,38 @@ def router_multicast(self) -> None:
...
+        if not hardware_counters.features:
+            # Since we use the same data model in eos_cli_config_gen, it would pick up the input vars unless we explicitly set it to null.
+            self.custom_structured_configs.nested.hardware_counters = EosCliConfigGen.HardwareCounters._from_null()
```
```suggestion
            self.custom_structured_configs.nested.hardware_counters.features = EosCliConfigGen.HardwareCounters.Features._from_null()
```

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/HARDWARE_FEATURES_ON_VIRTUAL_PLATFORMS.yml
```diff
@@ -0,0 +1,409 @@
...
+    # Expected to be filtered out for virtual platforms
+    - name: flow-spec
+    # Expected to be filtered out for virtual platforms
+    - name: gre tunnel interface
+      direction: out
+    # Expected to be filtered out for virtual platforms
+    - name: ip
+    # Expected to be filtered out for virtual platforms
+    - name: mpls interface
+    # Is not accepted by CLI of the [cv]EOS devices but is permitted by AVD to simplify schema
```
ClausHolbechArista: Please make this more clear. I think we are just matching on mpls lfib overall, and not evaluating mpls lfib + units_packets as an individual case right?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+no enable password
+no aaa root
+!
+hardware counter feature acl out mac
+hardware counter feature decap-group
+hardware counter feature directflow
+hardware counter feature ecn
+hardware counter feature flow-spec
+hardware counter feature gre tunnel interface out
+hardware counter feature ip
```
ClausHolbechArista: % Incomplete command when testing on 720XP

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature mpls interface
+hardware counter feature mpls lfib out units packets
+hardware counter feature mpls lfib in
+hardware counter feature mpls tunnel
+hardware counter feature multicast
```
ClausHolbechArista: % Incomplete command when testing on 720XP

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature nexthop
+hardware counter feature pbr in
+hardware counter feature pdp
```
ClausHolbechArista: % Incomplete command when testing on 720XP

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature policing interface
+hardware counter feature qos
+hardware counter feature qos dual-rate-policer
+hardware counter feature route
```
ClausHolbechArista: % Incomplete command when testing on 720XP

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
+!
+no enable password
+no aaa root
+!
+hardware counter feature acl out mac
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature decap-group
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature directflow
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature ecn
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature flow-spec
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature gre tunnel interface out
+hardware counter feature ip
+hardware counter feature mpls interface
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature mpls lfib out units packets
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature mpls lfib in
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature mpls tunnel
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature multicast
+hardware counter feature nexthop
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature pbr in
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature pdp
+hardware counter feature policing interface
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature qos
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature qos dual-rate-policer
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature route
+hardware counter feature routed-port
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature segment-security in
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature subinterface
```
ClausHolbechArista: % Incomplete command

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature tapagg
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature traffic-class
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature traffic-policy
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature vlan
```
ClausHolbechArista: % Unavailable command (not supported on this hardware platform)

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
```diff
@@ -0,0 +1,160 @@
...
+hardware counter feature vlan-interface
```
ClausHolbechArista: % Incomplete command

ClausHolbechArista (commented):
test result on 720XP-48ZC2

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Sure, changed back to `EosCliConfigGen.HardwareCounters.Features._from_null()`

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
alexeygorbunov: Fixed

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/HARDWARE_FEATURES_ON_VIRTUAL_PLATFORMS.yml
alexeygorbunov: Right, we are matching on `mpls lfib` but CLI of [cv]EOS will effectively reject command `hardware counter feature mpls lfib units .*` (struct config with `units_packets: true`). I simply wanted to keep this comment and both use cases (with and without units) to track/remember this.
But if this is excessive - removing it and keeping one use case only.
Fixed.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/HARDWARE_FEATURES_720XP.cfg
alexeygorbunov: I'm not sure i can make all generated commands compliant with all involved platforms as i have not done any feature supportability tests against physical platforms. 720 was added to the inventory to confirm that PoE is skipped for all virtual platforms but still added correctly on at least 1 physical platform that supports it (720XP).

alexeygorbunov: Adjusted inputs and feature support for involved hardware platforms as well. All produced CLI code is now compliant with all related platforms

ClausHolbechArista (approved):

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -175,8 +175,38 @@ def router_multicast(self) -> None:
...
+            return
+        hardware_counters = self.inputs.hardware_counters._deepcopy()
+
+        # Filter different hardware counter features based on the platform supportability
+        hardware_counters.features = hardware_counters.features._filtered(
+            lambda feature: get_v2(
+                self.shared_utils.platform_settings.feature_support.hardware_counter_features,
+                feature.name.replace(" ", "_").replace("-", "_"),
+                # Assume all uncovered/new features are supported
+                default=True,
```
Vibhu-gslab, line 200: In the description we have mentioned `Feature will be ignored on unsupported platforms.` but here we have set `default` as `true`, looks like contradict statements.
May be we can update the description like `Feature will be ignored on unsupported platforms when set to false`.

ClausHolbechArista: Maybe the description is not clear enough. The feature will be ignored on unsupported platforms, but what makes a platform unsupported is to set this to false. So by default most platforms are supported.

ClausHolbechArista: The line you are commenting on is specifically for hardware_counters where the config input model is just a list of strings, so it could easily contain things we don't cover in the feature toggles. So this default is just saying "yes something is supported because we don't know the option"

Vibhu-gslab: > The line you are commenting on is specifically for hardware_counters where the config input model is just a list of strings, so it could easily contain things we don't cover in the feature toggles. So this default is just saying "yes something is supported because we don't know the option"

Oh yes got that, comment is not just for this line it is for every key where `default: true` and description says `Feature will be ignored on unsupported platforms.`
It will all make sense if we can just make the description little better, may be like `Feature will be ignored on unsupported platforms when set to false.`

ClausHolbechArista: It is really `Feature will be ignored on platforms where this is 'false'`

ClausHolbechArista: @Vibhu-gslab updated in https://github.com/aristanetworks/avd/pull/5338/commits/39e07f8b634c6eaa24e59626c13bda76338e1052. Please review.

Vibhu-gslab: Looks good!

Vibhu-gslab (approved):
LGTM!

PR #5343, opened 2025-04-29
Fix(eos_designs): queue monitor length with missing notifying

### LINKED ISSUE #5331: queue_monitor_length_notify: false fails if notifying is not set
### Issue Summary

`del queue_monitor_length.notifying` fails when `queue_monitor_length_notify: false` and `notifying` is not set in the inventory. This was introduced by #5127 

eos_designs

### AVD version

5.3.0 and later

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

None

### Steps to reproduce

```shell

```

### Relevant log output

```shell

```

### Contributing Guide

ankudinov:
## Change Summary

Fix the case of missing `notifying` key introduced in #5127

## Related Issue(s)

Fixes #5331 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Check for missing class attribute before `del`. Returning `None` looks even more straightforward, but changes the config when compared to current behaviour.

## How to test
Use you inventory and check if:
- config wasn't changed compared to the previous AVD version
- there is no error when `notifying: true` is not set under `queue_monitor_length`

This was already tested on a customer inventory and with updated Molecule test (with and without the fix).
It could be feasible to move Molecule test somewhere else. Open for the feedback.


ClausHolbechArista (approved):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -295,7 +295,8 @@ def queue_monitor_length(self) -> None:
         # Remove notifying key if not supported by the platform settings.
         queue_monitor_length = self.inputs.queue_monitor_length._cast_as(EosCliConfigGen.QueueMonitorLength)
         if not self.shared_utils.platform_settings.feature_support.queue_monitor_length_notify:
```
```suggestion
        if not self.shared_utils.platform_settings.feature_support.queue_monitor_length_notify and hasattr(queue_monitor_length, "notifying"):
```

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -295,7 +295,8 @@ def queue_monitor_length(self) -> None:
...
-            del queue_monitor_length.notifying
+            if hasattr(queue_monitor_length, "notifying"):
+                del queue_monitor_length.notifying
```
```suggestion
            del queue_monitor_length.notifying
```

ClausHolbechArista (changes requested):

ClausHolbechArista (approved):

Shivani-gslab (approved):

PR #5359, opened 2025-05-02
Refactor(plugins): Improve generation of schema classes

ClausHolbechArista:
- Improve generation of schema classes to avoid very long lines, which then requires an E501 exception for ruff.


ClausHolbechArista:
We still need E501 rule for eos_cli_config_gen and eos_designs, but with this we no longer need it for eos_designs_facts.

gmuloc (approved):

PR #5362, opened 2025-05-02
Refactor(eos_designs): Modify data format for node id pool manager

ClausHolbechArista:
## Change Summary

Modify data format for node id pool manager

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Allow customization of the node id pool name using `fabric_numbering_node_id_pool`.
- Changing the pool key from dict to string.
- Changing the assignment key from dict to string.
- Changing the file layout to be nested dicts instead of nested lists of dicts.
- requires deletion of data file for existing deployments.

## How to test

Existing molecule and pytest updated.

### User Checklist

- N/A

### Repository Checklist


carlbuchmann (changes requested):
Please add a test with `fabric_numbering_node_id_pool: fabric_name={fabric_name}{dc_name?</dc_name=}{pod_name?</pod_name=}{type?</type=}{rack?</rack=}` to test overriding the default and also testing all available template fields.

carlbuchmann (commented):
Tested:

- [x] custom `fabric_numbering_node_id_pool: fabric_name={fabric_name}{dc_name?</dc_name=}{pod_name?</pod_name=}{type?</type=}{rack?</rack=}` and got desired result for unique id for each switch in same node type key.
- [x] Manually updated FABRIC-ids.yml file to get custom id on a node. id was retained
- [x] Removed devices from inventory, and devices were removes from FABRIC-ids.yml
- [x] error raised when incorrect data in FABRIC-ids.yml

carlbuchmann:
> Please add a test with `fabric_numbering_node_id_pool: fabric_name={fabric_name}{dc_name?</dc_name=}{pod_name?</pod_name=}{type?</type=}{rack?</rack=}` to test overriding the default and also testing all available template fields.

Resolved - thank you!

carlbuchmann (approved):
LGTM

python-avd/pyavd/_eos_designs/schema/schema_fragments/fabric_numbering_node_id_pool_key.schema.yml
```diff
@@ -0,0 +1,26 @@
...
+keys:
+  fabric_numbering_node_id_pool:
+    type: str
+    documentation_options:
+      table: fabric-numbering
+    description: |-
+      Name of Node ID pool or template used to render the name of each Node ID pool.
+      For each device the Node ID is assigned from a pool shared by all devices rendering the same pool name.
+      This can be modified to include fewer or more fields to keep separate pools or to use the same pool across areas.
+      This can be a template using the AVD string formatter syntax: https://avd.arista.com/devel/roles/eos_designs/docs/how-to/custom-descriptions-names.html#avd-string-formatter-syntax.
```
alexeygorbunov: Is this a valid link? Not working for me

```suggestion
      This can be a template using the AVD string formatter syntax: https://avd.arista.com/devel/ansible_collections/arista/avd/roles/eos_designs/docs/how-to/custom-descriptions-names.html#avd-string-formatter-syntax.
```

python-avd/pyavd/api/pool_manager/node_id_pools.py
```diff
@@ -7,43 +7,33 @@
...
 @dataclass()
-class NodeIdPoolCollection(PoolCollection[NodeIdPoolKey, NodeIdAssignmentKey, int]):
+class NodeIdPoolCollection(PoolCollection[int]):
     pools_key: str = "node_id_pools"
-    pools: dict[PoolKey, Pool[NodeIdPoolKey, NodeIdAssignmentKey, int]] = field(default_factory=dict)
-    pool_cls: type[Pool[NodeIdPoolKey, NodeIdAssignmentKey, int]] = Pool[NodeIdPoolKey, NodeIdAssignmentKey, int]
-    pool_key_cls: type[NodeIdPoolKey] = NodeIdPoolKey
-    assignment_cls: type[PoolAssignment[NodeIdAssignmentKey, int]] = PoolAssignment[NodeIdAssignmentKey, int]
-    assignment_key_cls: type[NodeIdAssignmentKey] = NodeIdAssignmentKey
+    pools: dict[str, Pool[int]] = field(default_factory=dict)
```
alexeygorbunov: Is `pools` defined here for future use? It shouldn't be `_pools`?

python-avd/pyavd/api/pool_manager/base_classes.py
alexeygorbunov, line 1: Generic question - is this expected that allocation of the node_ids for 2 (or any x > 1) new devices raises an exception if 2nd device is trying to explicitly use ID which is the "next" available at the time when node_id for the 1st new device is allocated?

Example:
- have "SPINE1" and "SPINE2" with pool-allocated IDs of 1 and 2.
- trying to add
```
    - name: SPINE3
    - name: SPINE4
      id: 3
```
- allocation for SPINE4 fails because ID 3 gets allocated to SPINE3 before logic processes SPINE4. 
`Unfortunately the 'id: 3' is not available in the Node ID pool at this time. The 'id' setting must either be removed or changed. If you prefer to keep the 'id' setting, the next available value is 4 for host 'SPINE4'.`

It's not very logical to assign such non-consecutive values but if user does -  user may get confused that AVD reports ID of 3 as occupied although it's not tracked in the pool file.

python-avd/pyavd/api/pool_manager/node_id_pools.py
```diff
@@ -7,43 +7,33 @@
...
         """Returns the pool key to use for this device."""
-        return NodeIdPoolKey(
-            fabric_name=shared_utils.fabric_name, dc_name=shared_utils.inputs.dc_name, pod_name=shared_utils.inputs.pod_name, type=shared_utils.type
+        return AvdStringFormatter().format(
+            shared_utils.inputs.fabric_numbering_node_id_pool,
+            fabric_name=shared_utils.fabric_name,
+            dc_name=shared_utils.inputs.dc_name,
+            pod_name=shared_utils.inputs.pod_name,
+            type=shared_utils.type,
+            rack=shared_utils.node_config.rack,
```
alexeygorbunov, line 35: Once Hybrid Campus PR is ready - would it make sense to extend this list with campus-related attributes of the nodes (campus/campus_pod/campus_access_pod)?

python-avd/pyavd/api/pool_manager/base_classes.py
ClausHolbechArista: I am ok with this behavior for now. Since the logic runs seperately for each device, we have no way of knowing ahead of time, which one to try out. Maybe we could add some information about who is using it in the error message, so it helps the user figure out what is going on.

python-avd/pyavd/api/pool_manager/node_id_pools.py
ClausHolbechArista: Most likely yes.

python-avd/pyavd/api/pool_manager/node_id_pools.py
ClausHolbechArista: It should have been _pools, but we don't really need it since the type var will work fine.
```suggestion
```

ClausHolbechArista: fixed

alexeygorbunov (approved):
LGTM

python-avd/pyavd/api/pool_manager/base_classes.py
```diff
@@ -14,56 +14,32 @@
...
 
-    def as_dict(self) -> dict:
-        return {"key": asdict(self.key), "value": self.value}
-
 
 @dataclass
-class Pool(Generic[T_PoolKeyType, T_AssignmentKeyType, T_ValueType]):
-    """One Pool of resources indexed by T_PoolKeyType. Currently only supporting a pool of Integers."""
+class Pool(Generic[T_ValueType]):
+    """One Pool of resources indexed by T_AssignmentKeyType. Currently only supporting a pool of Integers."""
```
gmuloc: `T_AssignmentKeyType` does not exist anymore

```suggestion
    """One Pool of resources indexed by str. Currently only supporting a pool of Integers."""
```

python-avd/pyavd/api/pool_manager/base_classes.py
gmuloc: as discussed on our call this morning we are going to implement this so blocking the PR for now. Thanks @alexeygorbunov

gmuloc (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/converge.yml
```diff
@@ -447,11 +463,8 @@
           ansible.builtin.import_role:
             name: arista.avd.eos_designs
       rescue:
-        - name: Error message
```
gmuloc, line 433: let's do it on all in a different PR

gmuloc (approved):

PR #5376, opened 2025-05-07
Test: Improve python bindings for rust validation

ClausHolbechArista:
Improve python bindings for rust validation so they now return native Python types instead of JSON. Also updating the pyi file so type hints and doc strings reflect the correct types.

Rust enums are not mapped to Enum in Python because of limitations in pyo3, so the class represenation added in pyi file is what comes closest to the actual implementation. Can be seen in action in the tests.


ClausHolbechArista:
@konall please review and evaluate any impact to other repos using this.

konall (approved):
I'm not well-versed in the Python bindings, but the other changes don't impact the dependent Rust crates at all really

gmuloc (approved):

PR #5395, opened 2025-05-13
Refactor(eos_designs): Separate hostname from device unique id

ClausHolbechArista:
TODO:

Go through the changes and various TODO messages, and decide on each how we should proceed.

## Change Summary

Refactor(eos_designs): Separate hostname from device unique id

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

When running AVD from Ansible, the unique identifier for each device is also the hostname.
This comes from Ansible's inventory and is intuitive when using Ansible.

Since AVD can also run in other contexts, where the unique inventory ID may not correspond to the hostname of the device,
it is important to keep the logic separate between `device_uid` and `hostname`.

- All input models pointing to a specific peer device will find the device based on it's `device_uid`.
- The `name` field under node settings is actually the `device_uid`.
- The input models do not offer a way to set the hostname. It can only be set in the various PyAVD function calls.
- Error messages will mention the `device_uid` when relating to a device.
- Descriptions will use the `hostname`.
- Models that support regex matching to relate to devices will match on either `hostname` or `device_uid`. TODO: Decide if this is what we want.
- Currently `device_uid` is not available for custom description logic. Only the `hostname`.

## How to test

Added a fancy pytest that patches the eos_designs_unit_tests molecule scenario to provide separate UID vs hostnames. Then run and compare the produced configs. This should catch it if we miss something later. Unfortunately the test is very slow because it walks all the inputs many times to look for hostnames and replace.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -430,8 +430,3 @@ platform_speed_groups:
         speed_groups: [ 3, 2, 5/1 ]
       - speed: 25G
         speed_groups: [ 1, 2, 4 ]
-
-# Set Aboot password with sha512 hash
```
gmuloc, line 434: why?

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/group_vars/CV_TOPOLOGY.yml
```diff
@@ -12,21 +12,21 @@ type: l3leaf
 mgmt_interface: null # To get the coverage for get(self.cv_topology_config, "mgmt_interface")
 use_cv_topology: true
 cv_topology:
-  - hostname: CV-MLAG-OSPF-L3LEAF1A
+  - hostname: uid_CV-MLAG-OSPF-L3LEAF1A
```
gmuloc, line 15: I am not sure I follow the purpose of this change

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschematools.py
```diff
@@ -177,16 +177,6 @@ def handle_validation_exceptions(self, exceptions: Generator, mode: str | None)
                 self.ansible_display.error(message, wrap_text=False)
         return counter
 
-    def validate_schema(self) -> int:
```
gmuloc, line 180: TODO - check why this is removed in this PR

docs/contribution/eos_designs_internal_notes.md
```diff
@@ -22,6 +22,26 @@ flowchart TD
...
+
+### Separate Device UID from Device Hostname
+
+When running AVD from Ansible, the unique identifier for each device is also the hostname.
+This comes from Ansible's inventory and is intuitive when using Ansible.
+
+Since AVD can also run in other contexts, where the unique inventory ID may not correspond to the hostname of the device,
+it is important to keep the logic separate between `device_uid` and `hostname`.
+
+- All input models pointing to a specific peer device will find the device based on it's `device_uid`.
```
gmuloc:
```suggestion
- All input models pointing to a specific peer device will find the device based on its `device_uid`.
```

docs/contribution/eos_designs_internal_notes.md
```diff
@@ -22,6 +22,26 @@ flowchart TD
...
+- The `name` field under node settings is actually the `device_uid`.
+- The input models do not offer a way to set the hostname. It can only be set in the various PyAVD function calls.
+- Error messages will mention the `device_uid` when relating to a device.
+- Descriptions will use the `hostname`.
+- Models that support regex matching to relate to devices will match on either `hostname` or `device_uid`. TODO: Decide if this is what we want.
```
gmuloc, line 42: this seems potentially annoying if they are not related but end up overlapping

ansible_collections/arista/avd/plugins/plugin_utils/schema/avdschematools.py
ClausHolbechArista: It was just removed because it is unused and showed up in my editor.

docs/contribution/eos_designs_internal_notes.md
ClausHolbechArista: agreed, hence we need to discuss further and decide how to proceed. I have talked with others about maybe adding a special schema type (or maybe just str format) of "device name" or similar, which could accept different syntaxes:
- `"foo"` or `{"hostname": "foo"}` for referencing the device by hostname.
- `{"device_uid": "uid_foo"}` for referencing the device by device_uid.

So all schema fields where we need a device name, we would modify it to use this format. Same for regex, we could also allow this dual form, where the regex will match one or the other.

ClausHolbechArista: Or maybe just `"foo"` or `"$hostname=foo"` or `"$device_id=uid_foo"` to not have to juggle types.

PR #5410, opened 2025-05-16
Test: Rust schema validation: Use configuration object instead of feature

ClausHolbechArista:
Rust schema validation: Use configuration object instead of feature

This change modifies the signature of the validation functions to also take an optional config. Config only contains a single knob to relax the check for required variables in the root dict. This is used by the language server since it validates separate docs only containing a partial model.


rust/validation/src/validation/dict.rs
```diff
@@ -81,9 +81,8 @@ fn validate_keys(schema: &Dict, input: &Map<String, Value>, ctx: &mut Context) {
                 Some(Value::Null) | None => {
                     // nullish values don't need to be validated beyond a requiredness check
 
-                    #[cfg(feature = "relaxed_validation_on_root_dicts")]
                     // Don't validate required keys if we are at the rool level.
```
gmuloc:
```suggestion
                    // Don't validate required keys if we are at the root level.
```

gmuloc: should this comment be changed?

gmuloc (approved):

PR #5443, opened 2025-05-26
Refactor(eos_designs): Add pytest coverage for network_services/port_channel_interfaces.py

laxmikantchintakindi:
## Change Summary

Add pytest coverage for network_services/port_channe_interfaces.py (99%)

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add pytest coverage for network_services/port_channe_interfaces.py.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/port_channel_interfaces.py
```diff
@@ -90,9 +90,6 @@ def _set_l3_port_channels(
         """
         for vrf in tenant.vrfs:
             for l3_port_channel in vrf.l3_port_channels:
-                if l3_port_channel.node != self.shared_utils.hostname:
-                    continue
```
laxmikantchintakindi, line 94: This is already filtered in shared_utils/filtered_tenants.py::filtered_l3_port_channels.

python-avd/pyavd/_eos_designs/structured_config/network_services/port_channel_interfaces.py
```diff
@@ -168,7 +165,7 @@ def _set_l3_port_channels(
                     ospf_authentication = l3_port_channel.ospf.authentication
                     if ospf_authentication == "simple" and (ospf_simple_auth_key := l3_port_channel.ospf.simple_auth_key) is not None:
                         port_channel_interface._update(ospf_authentication=ospf_authentication, ospf_authentication_key=ospf_simple_auth_key)
-                    elif ospf_authentication == "message-digest" and (ospf_message_digest_keys := l3_port_channel.ospf.message_digest_keys) is not None:
```
laxmikantchintakindi: Emtpy l3_port_channel.ospf.message_digest_keys is not being considered as None.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/network-services-l3-port-channels.yml
```diff
@@ -174,9 +174,35 @@ tenants:
                   key: testpass1
                 - id: 2
                   key: testpass2
+                  # 'id' and 'key' are not set
+                  # TODO: AVD 6.0.0 Make 'id' a primary key and set 'hash_algorithm' and 'key' required
+                - hash_algorithm: "sha256"
+            # Test 'ip ospf authentication message-digest' is not rendered when message-digest keys are incomplete
```
Shivani-gslab:
```suggestion
                  # Test when 'id' and 'key' are not set under message-digest keys.
                  # TODO: AVD 6.0.0 Make 'id' a primary key and 'key' required
                - hash_algorithm: "sha256"
```

Shivani-gslab (commented):
Tried to write test-case for the uncovered line 64, 
```
if potential_parent_interface.name not in configured_physical_po:
                  
      self.structured_config.port_channel_interfaces.append(potential_parent_interface)
```
it can be covered only when the interface number is same eg. `7 in Ethernet7` for the endpoint with and without sub-interfaces. If we give same interface name `Ethernet7` it throws duplication error.

it works with this, since the interface number is same but not the name -
```
endpoints:
    - id: 60
       nodes: [network_services_point_to_point]
       interfaces: [Vlan7]
       port_channel:
          mode: "on"
```
but this is not a valid input for port-channel.
need help if there is any other case for this.
@ClausHolbechArista @gmuloc

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/network-services-l3-port-channels.yml
laxmikantchintakindi: It will still render `ip ospf authentication message-digest`, as other two keys are present.

Shivani-gslab: it will render for first 2 elements but not this one, we can see the same in config

alexeygorbunov:
> Tried to write test-case for the uncovered line 64,
> 
> ```
> if potential_parent_interface.name not in configured_physical_po:
>                   
>       self.structured_config.port_channel_interfaces.append(potential_parent_interface)
> ```
> 
> it can be covered only when the interface number is same eg. `7 in Ethernet7` for the endpoint with and without sub-interfaces. If we give same interface name `Ethernet7` it throws duplication error.
> 
> it works with this, since the interface number is same but not the name -
> 
> ```
> endpoints:
>     - id: 60
>        nodes: [network_services_point_to_point]
>        interfaces: [Vlan7]
>        port_channel:
>           mode: "on"
> ```
> 
> but this is not a valid input for port-channel. need help if there is any other case for this. @ClausHolbechArista @gmuloc

I think this can be hit by using ethernet interfaces with IDs which produce same results for `"".join(re.findall(r"\d",))`.
Examples can be: `interfaces: [Ethernet5/1]` and `interfaces: [Ethernet51]`. Both will produce targeted PO with ID of 51.

```
      - name: TESTPO-SUB
        type: vpws-pseudowire
        subinterfaces:
          - number: 1051
        endpoints:
          - id: 62
            nodes: [network_services_point_to_point]
            interfaces: [Ethernet5/1]
            port_channel:
              short_esi: "0102:BEEF:F10D"
              mode: "active"
          - id: 63
            nodes: [fake_node]
            interfaces: [Ethernet5/1]
            port_channel:
              mode: "active"


      - name: TESTPO
        type: vpws-pseudowire
        endpoints:
          - id: 64
            nodes: [network_services_point_to_point]
            interfaces: [Ethernet51]
            port_channel:
              short_esi: "0102:BEEF:F20D"
              mode: "active"
          - id: 65
            nodes: [fake_node]
            interfaces: [Ethernet51]
            port_channel:
              mode: "active"
```

`potential_parent_interface` in this case is `<PortChannelInterfacesItem(name=Port-Channel51, peer_type=system, shutdown=False, switchport=<Switchport(enabled=False)>, evpn_ethernet_segment=<EvpnEthernetSegment(identifier=0000:0000:0102:BEEF:F10D, route_target=01:02:BE:EF:F1:0D)>, lacp_id=0102.BEEF.F10D)>` with `configured_physical_po` already containing Po51 (`{'Port-Channel11', 'Port-Channel10', 'Port-Channel4', 'Port-Channel51'}`)


Resulting config:
```
interface Port-Channel51
   no shutdown
   no switchport
   !
   evpn ethernet-segment
      identifier 0000:0000:0102:BEEF:F20D
      route-target import 01:02:BE:EF:F2:0D
   lacp system-id 0102.BEEF.F20D
!
interface Port-Channel51.1051
   no shutdown
   !
   encapsulation vlan
      client dot1q 1051 network client
!

interface Ethernet5/1
   no shutdown
   channel-group 51 mode active
!

interface Ethernet51
   no shutdown
   channel-group 51 mode active
!

patch panel
   patch TESTPO
      connector 1 interface Port-Channel51
      connector 2 pseudowire bgp vpws TENANT_A pseudowire TESTPO
   !
   patch TESTPO-SUB_1051
      connector 1 interface Port-Channel51.1051
      connector 2 pseudowire bgp vpws TENANT_A pseudowire TESTPO-SUB_1051
   !
```

BUt produced config is not what "user" really asks for. So maybe there is another valid use case for this if statement

python-avd/pyavd/_eos_designs/structured_config/network_services/port_channel_interfaces.py
ClausHolbechArista: but the logic below will not run if the list is empty. The for will skip and the if below will be false. We already have a TODO in the schema which will simplify this logic.

ClausHolbechArista:
From what I can test, the line 
```python
    if potential_parent_interface.name not in configured_physical_po:                  
        self.structured_config.port_channel_interfaces.append(potential_parent_interface)
```

is already being tested for `network_services_point_to_point` so we don't have an issue here any longer right?

ClausHolbechArista:
> is already being tested for network_services_point_to_point so we don't have an issue here any longer right?

I misunderstood the case not being tested. After fiddling and trying various input combinations, I decided to remove the condition, since we cannot produce valid config where this becomes false. Now we will properly report duplicate interfaces for the case @alexeygorbunov tried out (overlapping port-channel numbers).

python-avd/pyavd/_eos_designs/structured_config/network_services/port_channel_interfaces.py
laxmikantchintakindi: Reverted this condition.

gmuloc (approved):

alexeygorbunov (approved):
LGTM

Shivani-gslab (approved):

PR #5447, opened 2025-05-26
Bump(pyavd): Add PyYAML to PyAVD dependencies

ClausHolbechArista:
The optional PoolManager depends on PyYAML so we need to add that as a general dependency in PyAVD.


carl-baillargeon (approved):
LGTM

gmuloc (approved):

PR #5452, opened 2025-05-27
Feat(eos_cli_config_gen): Add cli_config_gen support for E-Tree options

### LINKED ISSUE #5448: Feat(eos_cli_config_gen): Add support for E-Tree leaf config under VLAN
### Enhancement summary

Add support for the E-Tree leaf configuration for both VXLAN and MPLS: [https://www.arista.com/en/support/toi/eos-4-24-0f/14481-evpn-e-tree-for-mpls](https://www.arista.com/en/support/toi/eos-4-24-0f/14481-evpn-e-tree-for-mpls)

eos_cli_config_gen

### Use case example

The CLI config is as follows:
```
switch#config
 switch(config)#vlan 1000
switch(config-vlan-1000)#e-tree role leaf
switch(config-vlan-etree-1000)#remote leaf host drop
...
vlan 1000
   e-tree role leaf
      remote leaf host drop
```

This would extend the existing `vlans` schema. I'd suggest the following additions:
```yaml
vlans:
  - id: <int; required; unique>
    etree:
       # All VLANs by default have the E-Tree root role
       leaf_role: <bool; default: False>
       # enables remote leaf hosts to instead be installed as explicit drop routes in the local FDB
       remote_leaf_host_drop: <bool; default: false>
```

I don't see any real benefits to going for a deeper structure, closer to the CLI, but I'm open either way

### Additional context

_No response_

### Contributing Guide

colinmacgiolla:
## Change Summary

This PR allows the E-Tree Leaf Role to be set (all VLANs are root role by default). It also allows for the Remote Leaf Host Drop option to be enabled. Both are on a per-VLAN basis.

It also fixes some header inconsistencies in the documentation headers while I was there.

## Related Issue(s)

Fixes #5448 

## Component(s) name

## Proposed changes
This modifies the existing VLAN schema with the new keys:
```yaml
vlans:
  - id: <int; required; unique>
    e_tree:
       # All VLANs by default have the E-Tree root role
       leaf_role: <bool; default: False>
       # enables remote leaf hosts to instead be installed as explicit drop routes in the local FDB
       remote_leaf_host_drop: <bool; default: false>
```

## How to test
Tested on 7280 running 4.33.1F. Note that the option `remote leaf host drop` requires 4.33.0F+.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host1/vlans.yml
+++ b/ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host1/vlans.yml
@@ -22,6 +22,12 @@ vlans:
     private_vlan:
       type: isolated
       primary_vlan: 110
+  - id: 201
+    tenant: Tenant_B
+    name: ETREE_LEAF
+    e_tree:
+      leaf_role: true
+      remote_leaf_host_drop: true
   - id: 3010
     tenant: Tenant_A
     name: MLAG_iBGP_TENANT_A_PROJECT01
--- a/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
+++ b/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
@@ -32,7 +32,7 @@
 
 #### Private VLANs
 
-| Primary Vlan ID | Secondary VLAN ID | Private Vlan Type |
+| Primary VLAN ID | Secondary VLAN ID | Private VLAN Type |
 | --------------- | ----------------- | ----------------- |
 {%         for vlan in vlans | arista.avd.natural_sort('id') %}
 {%             if vlan.private_vlan.type is arista.avd.defined and
@@ -43,7 +43,7 @@
 {%     endif %}
 {%     if ns.show_e_tree_table %}
 
-#### E-Tree role
+#### E-Tree Role
 
 | VLAN ID | VLAN Name | Role | Remote Leaf Host Drop |
 | ------- | --------- | ---- | --------------------- |
--- a/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
+++ b/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
@@ -42,6 +42,7 @@
 {%         endfor %}
 {%     endif %}
 {%     if ns.show_e_tree_table %}
+
 #### E-Tree role
 
 | VLAN ID | VLAN Name | Role | Remote Leaf Host Drop |
@@ -53,10 +54,11 @@
 {%             else %}
 {%                 set row_role = 'Root' %}
 {%             endif %}
-{%             set row_host_drop = vlan.e_tree.remote_host_leaf_drop | arista.avd.default('-' ) %}
+{%             set row_host_drop = vlan.e_tree.remote_leaf_host_drop | arista.avd.default('-' ) %}
 | {{ vlan.id }} | {{ row_name }} | {{ row_role }} | {{ row_host_drop }} |
 {%         endfor %}
 {%     endif %}
+
 ### VLANs Device Configuration
 
 ```eos
--- a/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
+++ b/python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
@@ -5,7 +5,7 @@
 #}
 {# doc - vlans #}
 {% if vlans is arista.avd.defined %}
-{%     set ns = namespace(show_private_vlan_table = false) %}
+{%     set ns = namespace(show_private_vlan_table = false, show_e_tree_table = false) %}
 
 ## VLANs
 
@@ -16,6 +16,9 @@
 {%     for vlan in vlans | arista.avd.natural_sort('id') %}
 {%         if vlan.private_vlan is arista.avd.defined %}
 {%             set ns.show_private_vlan_table = true %}
+{%         endif %}
+{%         if vlan.e_tree.leaf_role is arista.avd.defined %}
+{%             set ns.show_e_tree_table = true %}
 {%         endif %}
 {%         set row_name = vlan.name | arista.avd.default('-') %}
 {%         if vlan.trunk_groups is arista.avd.defined %}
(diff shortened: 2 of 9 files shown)


python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/vlans.schema.yml
```diff
@@ -43,6 +43,15 @@ keys:
           items:
             type: str
             description: Trunk Group Name.
+        e_tree:
+          type: dict
+          keys:
+            leaf_role:
+              type: bool
+              description: Set the VLAN into the E-Tree leaf role. By default all VLANs are in root role
```
MaheshGSLAB:
```suggestion
              description: Set the VLAN into the E-Tree leaf role. By default all VLANs are in root role.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/vlans.schema.yml
```diff
@@ -43,6 +43,15 @@ keys:
...
+            remote_leaf_host_drop:
+              type: bool
+              description: Enables remote leaf hosts to instead be installed as explicit drop routes in the local FDB
```
MaheshGSLAB:
```suggestion
              description: Enables remote leaf hosts to instead be installed as explicit drop routes in the local FDB.
```
It also requires to define the leaf_role as true. So update the description.

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/vlans.j2
```diff
@@ -16,6 +16,12 @@ vlan {{ vlan.id }}
 {%     for trunk_group in vlan.trunk_groups | arista.avd.natural_sort %}
    trunk group {{ trunk_group }}
 {%     endfor %}
+{%     if vlan.e_tree is arista.avd.defined and vlan.e_tree.leaf_role is arista.avd.defined(true) %}
```
MaheshGSLAB:
```suggestion
{%     if vlan.e_tree.leaf_role is arista.avd.defined(true) %}
```
Because by checking  vlan.e_tree.leaf_role we are already checking the vlan.e_tree so no need to add explicitly.

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
```diff
@@ -17,6 +17,9 @@
 {%         if vlan.private_vlan is arista.avd.defined %}
 {%             set ns.show_private_vlan_table = true %}
 {%         endif %}
+{%         if vlan.e_tree.leaf_role is arista.avd.defined %}
+{%             set ns.show_e_tree_table = true %}
+{%         endif %}
```
MaheshGSLAB, line 22: Right now we are setting the namespace for every vlan id. However we can add the logic like:

```suggestion
{%         if vlan.private_vlan is arista.avd.defined and not ns.show_private_vlan_table %}
{%             set ns.show_private_vlan_table = true %}
{%         endif %}
{%         if vlan.e_tree.leaf_role is arista.avd.defined and not ns.show_e_tree_table %}
{%             set ns.show_e_tree_table = true %}
{%         endif %}
```
Lets confirm with @ClausHolbechArista and @gmuloc  first.

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/vlans.j2
MaheshGSLAB: Also the order is not correct as per EOS. I am not able to add the suggestion because of unchanged code but this config will come after the address locking section plus there is exclamation mark between address locking section and e-tree section. Here is the sample output.

```
s1-leaf1(config-vlan-201)#sh ac
vlan 201
   name ETREE_LEAF
   trunk group mahesh
   private-vlan isolated primary vlan 110
   !
   address locking
      address-family ipv4
   !
   e-tree role leaf
      remote leaf host drop
```

MaheshGSLAB (changes requested):

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/vlans.j2
colinmacgiolla: That's not what I was getting in 4.32
```
vlan 20
   name TEST
   state suspend
   trunk group LALA
   e-tree role leaf
   private-vlan community primary vlan 11
   !
   address locking
      address-family ipv4
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/vlans.schema.yml
colinmacgiolla: Done

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/vlans.j2
colinmacgiolla: Looks like it changed between 4.32 and 4.33...

MaheshGSLAB: @ClausHolbechArista @gmuloc 
command `remote leaf host drop` is not supported on 4.32.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host1/vlans.yml
```diff
@@ -22,6 +22,12 @@ vlans:
     private_vlan:
       type: isolated
       primary_vlan: 110
+  - id: 201
+    tenant: Tenant_B
+    name: ETREE_LEAF
+    e_tree:
+      leaf_role: true
```
ClausHolbechArista, line 29: Do we expect other roles? If so `role: leaf` with valid_values (only one value for now) would be good(?)

python-avd/pyavd/_eos_cli_config_gen/j2templates/documentation/vlans.j2
ClausHolbechArista: I think setting again and again is "cheaper" than doing the extra if condition.
It would be better to have a list of private vlans and etree vlans and add the vlan objects there, and then only iterate through those in the tables below - but that is a bigger change I won't put on Colin. The current booleans will work as well. These are rarely used features, so the case will be false for 99%.

ansible_collections/arista/avd/molecule/eos_cli_config_gen/inventory/host_vars/host1/vlans.yml
colinmacgiolla: There are only two, you are either a root (default) or you are a leaf

MaheshGSLAB (approved):

Vibhu-gslab (approved):

ClausHolbechArista (approved):

PR #5456, opened 2025-05-28
Feat(eos_designs): Add a platform setting to allow/disallow mtu on sub-interfaces

### LINKED ISSUE #5178: vEOS-LAB default platform settings should prevent `mtu` on subinterfaces
### Enhancement summary

This happens on p2p-vrfs uplinks. AVD devel generates

```
site2-leaf1(config-if-Et3.101)#show run int et3.200
interface Ethernet3.200
   description P2P_site2-wan1_Ethernet1.200_VRF_BLUE
   mtu 9214
   encapsulation dot1q vlan 200
   flow tracker sampled FLOW-TRACKER
   vrf BLUE
   ip address 10.0.2.12/31
```

But this gets an error when deployed on vEOS because mtu is not supported on subinterfaces

```
site2-leaf1(config-if-Et3.101)#mt?
  mtu  not supported on this hardware platform
```

eos_designs

### Use case example

WAN cv-pathfinder example deployed to vEOS fails on border...

Add a `platform_settings.feature` knob to indicate if `mtu` is supported or not on subinterfaces and set it to True by default and to False for vEOS in the default platform settings.

1/ setting `per_interface_mtu: false` under `custom_platform_settings` for `VEOS` works but it also removes MTU on the physical interface which is not ideal (and on MLAG...)

2/ custom_structured_configuration can also help but it is tedious to maintain as the subinterfaces are created dynamically based on the VRFs.

### Additional context

_No response_

### Contributing Guide

laxmikantchintakindi:
## Change Summary

Support to allow/disallow MTU on sub-interfaces through platform_settings.

## Related Issue(s)

Fixes #5178

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
This happens on p2p-vrfs uplinks. AVD devel generates

```
site2-leaf1(config-if-Et3.101)#show run int et3.200
interface Ethernet3.200
   description P2P_site2-wan1_Ethernet1.200_VRF_BLUE
   mtu 9214
   encapsulation dot1q vlan 200
   flow tracker sampled FLOW-TRACKER
   vrf BLUE
   ip address 10.0.2.12/31
```
But this gets an error when deployed on vEOS because mtu is not supported on subinterfaces.
```
site2-leaf1(config-if-Et3.101)#mt?
  mtu  not supported on this hardware platform
```
Add a platform_settings.feature_support knob to indicate if mtu is supported or not on subinterfaces and set it to True by default and to False for vEOS in the default platform settings.

## How to test
Use platform_settings as below:
```
platform_settings:
  - platforms: [ default ]
    default_interface_mtu: 1234
    feature_support:
      subinterface_mtu: false
```
Add an l3 sub-interface with mtu set as below:
```
tenants:
  - name: TENANT1
    mac_vrf_vni_base: 10000
    vrfs:
      - name: default
        vrf_id: 1
        l3_interfaces:
          - interfaces: [Ethernet10.10]
            ip_addresses: [10.10.10.10/24]
            nodes: [your-host]
            mtu: 9000
            enabled: true
```
`mtu 9000` should not be configured for Ethernet10.10.

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/platform_settings.yml
```diff
@@ -1,7 +1,13 @@
-type: l2leaf
-l2leaf:
+---
```
MaheshGSLAB, line 1: I think you should add new test instead of updating the existing one. Also it is possible that we are loosing overall coverage by change this l2leaf to l3leaf

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/platform_settings.yml
```diff
@@ -11,10 +17,28 @@ platform_settings:
...
+    mac_vrf_vni_base: 10000
+    vrfs:
+      - name: default
+        vrf_id: 1
+        l3_interfaces:
+          - interfaces: [Ethernet10.10]
+            ip_addresses: [10.10.10.10/24]
+            nodes: [platform_settings]
+            # mtu will not be configured for the sub-interface
+            mtu: 9000
```
MaheshGSLAB, line 42: Can you check is there any default MTU 9214 on these l3_interfaces, If yes then we should give the preference of subinterface_mtu over to default. But if someone has explicitly added the mtu then may be raise an error to get user know that in vEOS platform mtu is not supported on subinterface.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/platform_settings.yml
laxmikantchintakindi: This host is to test only platform settings and we are not losing any configs.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/platform_settings.yml
laxmikantchintakindi: There is no default MTU for l3_interface.

MaheshGSLAB (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -93,6 +93,12 @@ keys:
             poe:
               type: bool
               default: false
+            subinterface_mtu:
+              type: bool
+              default: true
+              description: |
+                Allow/disallow MTU configuration under sub-interfaces.
+                When this key is set to False, MTU is not rendered under the sub-interface even though it is set in the inputs.
```
Shivani-gslab:
```suggestion
                When this key is set to False, MTU is not rendered under the sub-interface even if it is set in the interfaces inputs.
```

Shivani-gslab (changes requested):

Shivani-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/network_services/ethernet_interfaces.py
```diff
@@ -166,6 +166,8 @@ def _set_l3_interfaces(
                     self._set_ipv4_acl(acl)
 
                 if "." in interface_name:
+                    if not self.shared_utils.platform_settings.feature_support.subinterface_mtu:
```
gmuloc: create a `get_interface_mtu` shared_utils helper that receives the interface name and do all these checks to write the logic in one place.

please move the call to this helper to line 135 in this file to avoid setting and then nullifying the MTU.

There is also some occurence of this in the `underlay/utils` `_get_l2_as_subint` method where you should call the helper

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/ethernet_interfaces.py
laxmikantchintakindi: I have created a new shared_utils allow_interface_mtu() to check if mtu is allowed or not. Used this property everywhere.

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -165,6 +163,14 @@ def default_interface_mtu(self: SharedUtilsProtocol) -> int | None:
     def evpn_multicast(self: SharedUtilsProtocol) -> bool:
         return self.switch_facts.evpn_multicast is True
 
+    def interface_mtu(self: SharedUtilsProtocol, interface_name: str | None = None) -> bool:
```
Shivani-gslab:
```suggestion
    def interface_mtu(self: SharedUtilsProtocol, interface_name: str) -> bool:
```

Shivani-gslab (dismissed):

python-avd/pyavd/_eos_designs/shared_utils/misc.py
Shivani-gslab: Also, I think the method should be named - `allow_interface_mtu`

laxmikantchintakindi: This is required as I used this method in python-avd/pyavd/_eos_designs/shared_utils/inband_management.py

laxmikantchintakindi: @gmuloc Looking for Guillaume's suggestion on method name.

ansible_collections/arista/avd/examples/cv-pathfinder/intended/structured_configs/site1-border2.yml
```diff
@@ -89,7 +87,6 @@ ethernet_interfaces:
 - name: Ethernet4.100
   description: P2P_site1-wan2_Ethernet2.100_VRF_BLUE
   shutdown: false
-  mtu: 9214
```
gmuloc: feels like a lot of changes can we not contain this in a smaller test?

gmuloc: Ok ignore this - this is on purpose - I got confused because I haev forced mtu in my lab on subif

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/host_vars/default_interface_mtu_platform.yml
```diff
@@ -1,11 +1,34 @@
...
+      bgp_as: 101
 
 default_interface_mtu: 8765 # This will not be used since platform settings takes precedence.
 
 platform_settings:
   - platforms: [ default ]
     default_interface_mtu: 1234
+    feature_support:
+      # Allow rendering mtu configuration under sub-interface
+      subinterface_mtu: true
```
gmuloc: is it not the default?

gmuloc: a bif confused for this one

laxmikantchintakindi: It is true by default.

gmuloc: so why add it here? is it for covergae?

laxmikantchintakindi: No. Removed.

python-avd/pyavd/_eos_designs/shared_utils/misc.py
gmuloc: let's use `interface_mtu_supported`

laxmikantchintakindi: done.

python-avd/pyavd/_eos_designs/structured_config/mlag/__init__.py
```diff
@@ -64,7 +64,7 @@ def vlan_interfaces(self) -> None:
             ),
             shutdown=False,
             no_autostate=True,
-            mtu=self.shared_utils.p2p_uplinks_mtu,
+            mtu=self.shared_utils.p2p_uplinks_mtu if self.shared_utils.interface_mtu_supported(main_vlan_interface_name) else None,
```
ClausHolbechArista: I think we should do a helper function like `get_interface_mtu(interface: str, mtu: int | None) -> int | None` and then we can check for the platform support in that helper and we can identify if it is a subinterface and check that too.

ClausHolbechArista: so instead of the one you have now where you just return a bool, just give it the mtu var and return that or none.

laxmikantchintakindi: This helper is also supporting `python-avd/pyavd/_eos_designs/shared_utils/inband_management.py::inband_mgmt_mtu`. Can I keep this method and create a new `get_interface_mtu` and use `interface_mtu_supported` in new method?

ClausHolbechArista: I think the helper could be used anywhere.
pseudocode for the new method:
```python
def get_interface_mtu(self: SharedUtilsProtocol, interface_name: str, configured_mtu: int | None) -> int | None:
    if not self.platform_settings.feature_support.per_interface_mtu:
        return None
    if "." in interface_name and not self.platform_settings.feature_support.subinterface_mtu:
        return None
    return configured_mtu
```

laxmikantchintakindi: Done. Yes this worked at all places. But could not get rid of `self.platform_settings.feature_support.per_interface_mtu` check in `python-avd/pyavd/_eos_designs/shared_utils/inband_management.py::inband_mgmt_mtu`.

ClausHolbechArista: all good, that one is fine to keep.

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -290,6 +296,7 @@ keys:
           interface_storm_control: false
           queue_monitor_length_notify: false
           evpn_gateway_all_active_multihoming: true
+          subinterface_mtu: false
```
ClausHolbechArista: I think many other platforms have this issue. I think we should avoid updating the default platform settings in this PR, since it will be a breaking change to do so.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
laxmikantchintakindi: Removed the default value.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-LEAF1A.cfg
```diff
@@ -115,6 +115,18 @@ interface Ethernet9
...
    switchport
 !
+interface Ethernet10
+   no shutdown
+   no switchport
+!
+interface Ethernet10.10
+   description test mtu not rendered for VEOS platform
+   no shutdown
+   mtu 9000
```
ClausHolbechArista: The purpose of the test was to check that the platform did _not_ update this. But since we removed the update of default platform settings, this is no longer relevant. Let's revert the changes to this test, since you cover it in the platform_settings tests.

ansible_collections/arista/avd/roles/eos_designs/docs/tables/platform-settings.md
```diff
@@ -185,6 +191,10 @@
           interface_storm_control: <bool; default=True>
           poe: <bool; default=False>
 
+          # Allow/disallow MTU configuration under sub-interfaces.
```
ClausHolbechArista:
```suggestion
          # Support for MTU configuration under sub-interfaces.
```

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/intended/configs/DC1-LEAF1A.cfg
laxmikantchintakindi: Removed this test. Added new test for underlay sub-interfaces.

gmuloc (approved):

ClausHolbechArista (approved):

PR #5485, opened 2025-06-04
Feat(pyavd): Add possibility to combine nested schema object

gmuloc:
## Change Summary

Combining is different from merging. It will:

* append_unique for AvdLists.
* raise if a "leaf" key is clashing between a and b
* for AvdIndexedLists:
  * append new items from b in a (new in the sense of new primary key)
  * for primary keys present in both a and b, the combining is valid only if the items are strictly equals, otherwise raise

## Related Issue(s)

Should allow to combine lists between VRFs in network_services in a follow-up PR

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

cf summary

## How to test

Unit tests

### Repository Checklist


python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -297,3 +297,34 @@ def _compare(self, other: object) -> bool:
...
+
+        if self._created_from_null or other._created_from_null:
+            # Set the flag to the value of other and overwrite with data from other.
+            self._created_from_null = other._created_from_null
+            # Replace with the "other" list.
+            self._items = other._items.copy()
+            return
+
+        for item in other:
+            # TODO: Do we need to copy the item to prevent some unsavory things later if other is modified?
```
ClausHolbechArista: I think I would like us to move towards creating new objects when we do operations like this. Pro: Will not modify input data. Con: Performance.
Potential long term plan: Move merge/combine/update operations to Rust.
So this method should really be `_combined` which returns a new instance. Let's discuss before you change anything.

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -256,3 +256,38 @@ def _compare(self, other: object) -> bool:
...
+        """
+        cls = type(self)
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        list_merge = "append_unique"
+
+        if self._created_from_null or other._created_from_null:
+            # Set the flag to the value of other and set list_merge to replace so we overwrite with data from other below.
```
ClausHolbechArista: Are we sure we want this behavior for combine? It is basically saying that `other.key: null` overwrites `this.key: [something]` Instead it should raise IMO.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,66 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+
+        Raises:
+            AristaAvdDuplicateDataError: If any item from other is conflicting with an item from self when appending.
+        """
+        cls = type(self)
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        if other._created_from_null:
```
ClausHolbechArista: Same as above. Let's not silently merge null objects.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,66 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+            field_type = self._fields[field]["type"]
+            if issubclass(field_type, AvdBase):
+                # Merge in to the existing object
+                old_value = cast("AvdBase", old_value)
+                new_value = cast("AvdBase", new_value)
+                old_value._combine(new_value)
+                continue
+
+            if field_type is dict:
+                # TODO: In place combining without schema - need to set the keys probably
```
ClausHolbechArista: This only happens for a few places with ambigous structured config schemas. For now I think we can just do a comparison like any other "leaf" key.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,66 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+
+        # TODO: check the two next lines
+        if other._created_from_null:
+            # Inherit the _created_from_null attribute to make sure we output null values instead of empty dicts.
+            self._created_from_null = True
+        elif self._created_from_null:
+            # We merged into a "null" class, but since we now have proper data, we clear the flag.
+            self._created_from_null = False
+
+        # TODO: Handle clashing custom data when combining - for now hardcoding list_merge"
```
ClausHolbechArista: Our merger in _utils have a feature similar to this combine, which can raise in case of collisions.

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: if created_form_null are the same - continue with all the logic afterwards - prsent a conflict

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -297,3 +297,34 @@ def _compare(self, other: object) -> bool:
...
+
+        Raises:
+            AristaAvdDuplicateDataError: If any item from other is conflicting with an item from self when appending.
+        """
+        cls = type(self)
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        if self._created_from_null or other._created_from_null:
```
gmuloc: change the null

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: remove the check

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: there should be a conservative_merge to check for collisions - use this (strategy)

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc: as discussed - for now we keep it like this and we will advise later

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: calling merge with ` same_key_strategy="must_match"`

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+
+        # If the value of _created_from_null are different, consider it as a conflict.
+        if self._created_from_null != other._created_from_null:
+            raise AristaAvdDuplicateDataError(type(self).__name__, str(self._dump()), str(other._dump()))
+
+        for field, new_value in other.items():
+            old_value = self._get_defined_attr(field)
+
+            if old_value is Undefined:
+                # value not set so we can just pick the old_value
```
alexeygorbunov: Shouldn't be `new_value` here?

```suggestion
                # value not set so we can just pick the new_value
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+                setattr(self, field, new_value)
+                continue
+
+            # Combine new value
+            field_type = self._fields[field]["type"]
+            if issubclass(field_type, AvdBase):
+                # Merge in to the existing object
+                old_value = cast("AvdBase", old_value)
+                new_value = cast("AvdBase", new_value)
+                old_value._combine(new_value)
```
alexeygorbunov, line 502: We try to combine without first checking if `old_value` == `new_value`?

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+            AristaAvdDuplicateDataError: If any item from other is conflicting with an item from self when appending.
+        """
+        cls = type(self)
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        # If the value of _created_from_null are different, consider it as a conflict.
+        if self._created_from_null != other._created_from_null:
+            raise AristaAvdDuplicateDataError(type(self).__name__, str(self._dump()), str(other._dump()))
```
alexeygorbunov: Would it make sense to add some clarification/context to the `context` as exception raised due to mismatching `_created_from_null` attribute does not really tell why two instances are not "combinable" as they look like they are:
```
Found duplicate objects with conflicting data while generating configuration for SomeList. [1, 2] conflicts with [2, 3, 4].

OR

Found duplicate objects with conflicting data while generating configuration for SomeIndexedList. [{'name': 'one', 'some_int': 1}, {'name': 'two', 'some_int': 2}] conflicts with [{'name': 'two', 'some_int': 2}, {'name': 'three', 'some_int': 3}, {'name': 'four', 'some_int': 4}]
```

python-avd/tests/pyavd/schema/test_data_combining.py
```diff
@@ -0,0 +1,164 @@
...
+    b = data_merging_schema_class._from_dict(b_data)
+
+    # Setting _created_from_null to True for a
+    getattr(a, key)._created_from_null = True
+    with pytest.raises(AristaAvdDuplicateDataError, match="Found duplicate objects with conflicting data while generating configuration for "):
+        a._combine(b)
+
+    # Setting _created_from_null to True for b
+    getattr(a, key)._created_from_null = False
+    getattr(a, key)._created_from_null = True
```
alexeygorbunov: is this supposed to be set for `b`? Otherwise this test is similar to lines 146-147

python-avd/tests/pyavd/schema/test_data_combining.py
```diff
@@ -0,0 +1,164 @@
...
+
+A_INDEXED_LIST = {"some_indexed_list": [{"name": "one", "some_int": 1}, {"name": "two", "some_int": 2}]}
+B_INDEXED_LIST = {"some_indexed_list": [{"name": "two", "some_int": 2}, {"name": "three", "some_int": 3}, {"name": "four", "some_int": 4}]}
+C_INDEXED_LIST = {"some_indexed_list": [{"name": "two", "some_int": 666}]}
+UNIQUE_A_AND_B_INDEXED_LISTS = {
+    "some_indexed_list": [{"name": "one", "some_int": 1}, {"name": "two", "some_int": 2}, {"name": "three", "some_int": 3}, {"name": "four", "some_int": 4}]
+}
+
+A_DICT = {"some_dict": {"some_nested_key": "blah"}}
+B_DICT = {"some_dict": {"some_nested_key": "blah"}}
```
alexeygorbunov: Is `B_DICT` needed (as it's the same as `A_DICT`)?

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: where do you see this? the one created from null should be empty (probably my tests are cheating)

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: yes - AvdBase type will take care of that

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +458,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+
+            if old_value is Undefined:
+                # value not set so we can just pick the old_value
+                setattr(self, field, new_value)
+                continue
+
+            # Combine new value
+            field_type = self._fields[field]["type"]
+            if issubclass(field_type, AvdBase):
+                # Merge in to the existing object
```
gmuloc:
```suggestion
                # Combine into the existing object
```

python-avd/tests/pyavd/schema/test_data_combining.py
gmuloc: yup!
```suggestion
    getattr(b, key)._created_from_null = True
```

python-avd/pyavd/_schema/models/avd_model.py
alexeygorbunov: These are messages of the actual exceptions raised by tests but i indeed forgot that in reality instances created from null will be empty. So we should be OK then!

python-avd/pyavd/_schema/models/avd_base.py
```diff
@@ -153,3 +157,15 @@ def _deepmerged(
...
+
+    @abstractmethod
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
+
+        Args:
```
ClausHolbechArista, line 167: add raises

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -297,3 +299,31 @@ def _compare(self, other: object) -> bool:
...
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        # If the value of _created_from_null are different, consider it as a conflict.
+        if self._created_from_null != other._created_from_null:
+            raise AristaAvdDuplicateDataError(type(self).__name__, str(self._dump()), str(other._dump()))
+
+        for item in other:
+            # TODO: Do we need to copy the item to prevent some unsavory things later if other is modified?
```
ClausHolbechArista: I don't think we do that for merging, so we should not here either. It was a general performance optimization.

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -257,3 +260,30 @@ def _compare(self, other: object) -> bool:
...
         return all(item == other_items[index] for index, item in enumerate(items))
+
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
+
+        for AvdIndexedList this is simply an append, if any conflict occurs a duplicate error will be raised.
```
ClausHolbechArista:
```suggestion
        for AvdList this is simply an append_unique, if any conflict occurs a duplicate error will be raised.
```

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -257,3 +260,30 @@ def _compare(self, other: object) -> bool:
...
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        # If the value of _created_from_null are different, consider it as a conflict.
+        if self._created_from_null != other._created_from_null:
+            raise AristaAvdDuplicateDataError(type(self).__name__, str(self._dump()), str(other._dump()))
+
+        # Append non-existing items.
+        self._items.extend(new_item for new_item in other._items if new_item not in self._items)
```
ClausHolbechArista: Why not just:
```python
for item in other:
  self.append_unique(item)
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +460,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
         return all(value == other.__dict__[field] for field, value in self.items() if field not in ignore_fields)
+
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
+
+        for AvdIndexedList this is simply an append, if any conflict occurs a duplicate error will be raised.
```
ClausHolbechArista: AvdModel.

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +460,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+            AristaAvdDuplicateDataError: If any item from other is conflicting with an item from self when appending.
+        """
+        cls = type(self)
+        if not isinstance(other, cls):
+            msg = f"Unable to combine type '{type(other)}' into '{cls}'"
+            raise TypeError(msg)
+
+        # If the value of _created_from_null are different, consider it as a conflict.
+        if self._created_from_null != other._created_from_null:
+            raise AristaAvdDuplicateDataError(type(self).__name__, str(self._dump()), str(other._dump()))
```
ClausHolbechArista: Should we translate None to "null" ?

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +460,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+                old_value = cast("AvdBase", old_value)
+                new_value = cast("AvdBase", new_value)
+                old_value._combine(new_value)
+                continue
+
+            if field_type is dict:
+                # TODO: In place combining without schema - need to set the keys probably
+                setattr(self, field, new_value)
+                continue
+
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -457,3 +460,59 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+            if old_value != new_value:
+                # this also applies to fields where field_type is dict
```
ClausHolbechArista, line 506: No until you remove the code above ;)

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: I don't know if it makes sense when reading python error message

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: Hmm I had removed this in a commit something went wrong

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: Main use case is with yaml/json-based input, so I think it will be better with null

gmuloc: I am not sure it is a good plan for when we will need to differentiate a _created_from_null AvdString vs the AvdString "null". But we can revisit then.

ClausHolbechArista:
We don't need this for 5.x so I will push it out.

python-avd/tests/pyavd/schema/combined.schema.yml
```diff
@@ -5,7 +5,8 @@ keys:
     type: list
     primary_key: name
     display_name: IP Extended Access-Lists
-    description: 'AVD currently supports 2 different data models for extended ACLs:
+    description:
+      "AVD currently supports 2 different data models for extended ACLs:
```
ankudinov: I feel this is unrelated to the PR. Looks like a very valid fix though, but ideally I'd submit another one.

ankudinov (approved):

python-avd/tests/pyavd/schema/combined.schema.yml
gmuloc: indeed

gmuloc: I need to fix all of this

python-avd/pyavd/_schema/models/avd_base.py
```diff
@@ -101,8 +101,12 @@ def _strip_empties(self) -> None:
         """In-place update the instance to remove data matching the given strip_values."""
 
     @abstractmethod
-    def _dump(self, include_default_values: bool = False) -> dict | list:
-        """Dump data into native Python types with or without default values."""
+    def _dump(self, include_default_values: bool = False) -> dict | list | str:
+        """
+        Dump data into native Python types with or without default values.
+
+        It honors _created_from_null, returning the string "null".
```
ClausHolbechArista: That is misleading. If it is created from null it should return the type `None`

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -297,3 +299,31 @@ def _compare(self, other: object) -> bool:
             return False
 
         return all(item._compare(other[key]) for key, item in self.items())
+
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
```
ClausHolbechArista, line 309: Should we mention that this is a "shallow" operation, meaning it will not dig into each item and attempt to combine those?

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -463,3 +466,52 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
...
+            field_type = self._fields[field]["type"]
+            if issubclass(field_type, AvdBase):
+                # Combine into the existing object
+                old_value = cast("AvdBase", old_value)
+                new_value = cast("AvdBase", new_value)
+                old_value._combine(new_value)
+                continue
+
+            if old_value != new_value:
+                # this also applies to fields where field_type is dict
```
ClausHolbechArista, line 506: we should not have any dicts left in the schema. So maybe we can start enforcing that keys is always set.

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc:
```suggestion
```

python-avd/pyavd/_schema/models/avd_model.py
gmuloc: probably in another PR?

python-avd/pyavd/_schema/models/avd_base.py
gmuloc: indeed - just a type hint issue here

python-avd/pyavd/_schema/models/avd_base.py
```diff
@@ -101,8 +101,12 @@ def _strip_empties(self) -> None:
         """In-place update the instance to remove data matching the given strip_values."""
 
     @abstractmethod
-    def _dump(self, include_default_values: bool = False) -> dict | list:
-        """Dump data into native Python types with or without default values."""
+    def _dump(self, include_default_values: bool = False) -> dict | list | str:
```
gmuloc:
```suggestion
    def _dump(self, include_default_values: bool = False) -> dict | list | str | None:
```

python-avd/pyavd/_schema/models/avd_indexed_list.py
gmuloc: hmm thats probably an issue then as it is not shallow in model where I am calling combine on the nested object. Maybe I should not call append then but rather I should call combine further if the object is already present

gmuloc (commented):
lets clean up all consideration of `created_from_null` for combine

python-avd/pyavd/_schema/models/avd_base.py
```diff
@@ -101,7 +101,7 @@ def _strip_empties(self) -> None:
         """In-place update the instance to remove data matching the given strip_values."""
 
     @abstractmethod
-    def _dump(self, include_default_values: bool = False) -> dict | list:
+    def _dump(self, include_default_values: bool = False) -> dict | list | str:
```
ClausHolbechArista:
```suggestion
    def _dump(self, include_default_values: bool = False) -> dict | list:
```

python-avd/pyavd/_schema/models/avd_indexed_list.py
```diff
@@ -297,3 +297,26 @@ def _compare(self, other: object) -> bool:
...
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
+
+        for AvdIndexedList this is simply an append, if any conflict occurs a duplicate error will be raised.
```
ClausHolbechArista:
```suggestion
        For AvdIndexedList this is simply an append, if any conflict occurs a duplicate error will be raised.
```

python-avd/pyavd/_schema/models/avd_list.py
```diff
@@ -257,3 +257,26 @@ def _compare(self, other: object) -> bool:
...
         return all(item == other_items[index] for index, item in enumerate(items))
+
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
+
+        for AvdList this is simply an append_unique, if any conflict occurs a duplicate error will be raised.
```
ClausHolbechArista:
```suggestion
        For AvdList this is simply an append_unique, if any conflict occurs a duplicate error will be raised.
```

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -463,3 +464,48 @@ def _compare(self, other: object, ignore_fields: tuple[str, ...] = ()) -> bool:
 
         # Use regular __eq__ check on all nested models, since we do not carry over ignore_fields.
         return all(value == other.__dict__[field] for field, value in self.items() if field not in ignore_fields)
+
+    def _combine(self, other: Self) -> None:
+        """
+        Update instance by combining the other instance in.
+
+        Combining is different from merging in the sense that it will raise if there is a conflict
+        between one of our elements and the other elements.
```
ClausHolbechArista:
```suggestion
        between one of our properties and the same property of the other.
        
        Properties of types AvdModel, AvdList and AvdIndexedList are recursively combined.
```

ClausHolbechArista (approved):

PR #5495, opened 2025-06-10
CI(cv_deploy): Fix molecule runner git issue for cv_deploy

alexeygorbunov:
## Change Summary

Remove dynamically-created directories at the end of the related tests.
This will avoid git changes tracked by molecule action runner

## Related Issue(s)

Fixes #

## Component(s) name

`arista.avd.cv_dploy`

## Proposed changes
Delete dynamically-created directories created during the test execution. This prevents molecule action runner from discovering untracked dirs/files and returning non-zero exit code

## How to test
`molecule test -s cv_deploy`
Example: https://github.com/aristanetworks/avd-internal/actions/runs/15547724549/job/43772405491

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/cv_deploy/duplicated_devices_strict_system_mac.yml
```diff
@@ -248,6 +248,14 @@
           ignore_errors: true
           when: cv_deploy_results.workspace.id is not none and cv_deploy_results.workspace.id | length > 0
 
+        - name: "{{ test_id | upper }} Remove directory with dynamically created structured configs (required for molecule runner)."
+          tags: ["{{ test_id }}"]
+          run_once: true
+          delegate_to: localhost
+          ansible.builtin.file:
```
ClausHolbechArista: An empty directory is not tracked in git. If we have some files in here, maybe we should just track those in git? Alternatively I think it would be better to gitignore the path instead of deleting again.

alexeygorbunov: This is a dir with dynamically-generated structured config files (we fetch live inventory data from CV for CI devices (to not have CI dependent on a static values of serial_number and system_mac_address of the CI devices) so tracking it in git will be problematic as files inside may change if CI devices are redeployed with new serial/mac.
I removed that extra task and added whole dir to the gitignore.

ClausHolbechArista (approved):

gmuloc (approved):

PR #5524, opened 2025-06-16
Feat(anta_runner): Added support for VerifyVxlanConfigSanity anta tests

MaheshGSLAB:
## Change Summary

Added support for VerifyVxlanConfigSanity anta tests

## Related Issue(s)

Fixes # part of https://github.com/aristanetworks/avd-internal/issues/271

## Component(s) name

`arista.avd.anta_runner

## Proposed changes
Added support for VerifyVxlanConfigSanity anta tests into anta_runner role.

## How to test
Run anta_runner molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_anta/input_factories/vxlan.py
```diff
@@ -0,0 +1,30 @@
...
+    def create(self) -> list[VerifyVxlanVniBinding.Input] | None:
+        """Create a list of bindings for the `VerifyVxlanVniBinding` test."""
+        bindings = {}
+        vxlan_config = self.structured_config.vxlan_interface.vxlan1.vxlan
+
+        for vlan in vxlan_config.vlans:
+            bindings[vlan.vni] = vlan.id
+
+        for vrf in vxlan_config.vrfs:
+            bindings[vrf.vni] = vrf.name
```
MaheshGSLAB: If we add VLANs or VRFs with VNIs from custom_structured_configuration.vxlan_interface, and the same VNI already exists in structured_configuration, the VNI-to-VLAN/VRF mapping will be overwritten — since the VNI acts as the key and must always be unique.

This behavior aligns with EOS as well, where a single VNI (e.g., 1111) can only be mapped to one VLAN or VRF. You cannot map multiple VLANs (e.g., VLAN 11 and VLAN 12) to the same VNI (1111).

carl-baillargeon (changes requested):
Hi @MaheshGSLAB, please remove `VerifyVxlanVniBinding` for now per the test matrix. Thanks

python-avd/pyavd/_anta/input_factories/vxlan.py
MaheshGSLAB: removed the test

MaheshGSLAB:
> Hi @MaheshGSLAB, please remove `VerifyVxlanVniBinding` for now per the test matrix. Thanks

removed

ansible_collections/arista/avd/molecule/anta_runner/anta/avd_catalogs/allow_bgp_vrfs_run/dc1-spine1.json
```diff
@@ -572,5 +572,19 @@
...
+        "result_overwrite": {
+          "custom_field": "Generated by AVD"
+        },
+        "filters": {
+          "tags": [
+            "dc1-spine1"
+          ]
+        }
+      }
+    }
```
carl-baillargeon: Why is it generated for spines? We usually don't expect spines to to be VTEPs.

python-avd/pyavd/_anta/index.py
```diff
@@ -163,6 +163,7 @@
     TestSpec(
         test_class=VerifyTransceiversTemperature,
     ),
+    TestSpec(test_class=VerifyVxlanConfigSanity, contitional_keys=[StructuredConfigKey.VXLAN1_INTERFACE]),
```
MaheshGSLAB: i have wrong argument here, instead of conditional_keys i have contitional_keys which cause to generate the catalog for all devices.

MaheshGSLAB: Even after fixing the typo this is what we are getting from device_context.structured_config. 
```
vxlan_interface=<VxlanInterface(vxlan1=<Vxlan1(vxlan=<Vxlan(source_interface=None)>)>)
```
The source interface is None hence we always have vxlan_interface.vxlan1.vxlan in device_context.structured_config and this cause to generate the test catalog for all devices. But there is no vxlan_interface in our structure config.

i checked the code and we never add the source interface as None also no default but still it generate in object.
```
if self._multi_vtep:
            vxlan.source_interface = "Loopback0"
            vxlan.mlag_source_interface = self.shared_utils.vtep_loopback
        else:
            vxlan.source_interface = self.shared_utils.vtep_loopback
```

ansible_collections/arista/avd/molecule/anta_runner/anta/avd_catalogs/allow_bgp_vrfs_run/dc1-spine1.json
MaheshGSLAB: in anta model.py file we have self.structured_config.vxlan_interface.vxlan1.vxlan.source_interface to check the vtep which make the entry in structure config and conditional key match pass. Updated the model.py file to fix the current issue. We may need to take a look on other checks as well.

python-avd/pyavd/_anta/index.py
carl-baillargeon: > i have wrong argument here, instead of conditional_keys i have contitional_keys which cause to generate the catalog for all devices.

In that case we should add `model_config = ConfigDict(extra="forbid")` to `TestSpec` Pydantic model to avoid allowing wrong fields.

ansible_collections/arista/avd/molecule/anta_runner/anta/avd_catalogs/allow_bgp_vrfs_run/dc1-spine1.json
carl-baillargeon: Ah I see. `is_vtep` is triggered before by `VerifyInterfacesStatusInputFactory` that's why. Good catch. Can you check in other places please? Separate PR though.

python-avd/pyavd/_anta/index.py
carl-baillargeon: When we specify `conditional_keys` the test are skipped at a higher level here: https://github.com/aristanetworks/avd/blob/releases/v5.5.x/python-avd/pyavd/_anta/factories.py#L67. Here we are using `get_v2` which uses `_get` under the hood so we are good for this one.

ansible_collections/arista/avd/molecule/anta_runner/anta/avd_catalogs/allow_bgp_vrfs_run/dc1-spine1.json
MaheshGSLAB: sure

python-avd/pyavd/_anta/index.py
MaheshGSLAB: added model_config = ConfigDict(extra="forbid") to TestSpec

python-avd/pyavd/api/_anta/test_spec.py
```diff
@@ -36,6 +36,7 @@ class TestSpec(BaseModel):
     test_class: type[AntaTest]
     conditional_keys: list[StructuredConfigKey] | None = None
     input_factory: type[AntaTestInputFactory] | None = None
+    model_config = ConfigDict(extra="forbid")
```
carl-baillargeon:
```suggestion
    model_config = ConfigDict(extra="forbid")
    test_class: type[AntaTest]
    conditional_keys: list[StructuredConfigKey] | None = None
    input_factory: type[AntaTestInputFactory] | None = None
```

carl-baillargeon (approved):
LGTM!

python-avd/pyavd/_anta/index.py
gmuloc: > vtep_loopback

@MaheshGSLAB so are you saying we have a key always set to None despite not setting it and no default value?

MaheshGSLAB: yes when we check something through self.strucutre_config and it is None in that case it update None in strucutre_config object as well. For this PR here is issue https://github.com/aristanetworks/avd/blob/releases/v5.5.x/python-avd/pyavd/_anta/models.py#L43

gmuloc (approved):

PR #5530, opened 2025-06-17
Feat(plugins): New logging system for AVD action plugins [WIP]

carl-baillargeon:
## Change Summary

WIP

## Related Issue(s)

Fixes https://github.com/aristanetworks/avd-internal/issues/270

## Component(s) name

`arista.avd.<role-name>`

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/__init__.py
```diff
@@ -7,6 +7,13 @@
 from os import environ
 from pathlib import Path
 
+AVD_LOGGING_V2_ENABLE = environ.get("AVD_LOGGING_V2_ENABLE")
+
+if AVD_LOGGING_V2_ENABLE:
+    from .plugin_utils.utils.avd_logging import init_avd_logging
+
+    init_avd_logging()
```
carl-baillargeon, line 15: I am not sure that initializing logging here is the right approach. This will impact all plugins, meaning we cannot really transition one plugin at a time.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin.py
```diff
@@ -0,0 +1,30 @@
+# Copyright (c) 2025 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from abc import abstractmethod
+from typing import Any
+
+from ansible.plugins.action import ActionBase
+
+
+class AvdActionPlugin(ActionBase):
```
carl-baillargeon, line 10: Later PR but we could potentially add common methods in this. For example `load_structured_configs`, `load_one_structured_config`, etc.

Maybe we should move the logging functions/decorator in here as well?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging.py
```diff
@@ -0,0 +1,305 @@
...
+    def filter(self, record: logging.LogRecord) -> bool:
+        """Adds all keys from the context dict as attributes to the log record."""
+        record.__dict__.update(self.context)
+        return True
+
+
+class AnsibleDisplayHandler(logging.Handler):
+    """Ansible display handler."""
+
+    def __init__(self, display: Display | None = None) -> None:
```
carl-baillargeon, line 93: I know we don't have to do this since Display is a singleton but it makes testing easier (we can pass a mocked Display down the line).

I didn't have much success trying to patch Display (I think it's called pretty early in the Ansible chain or maybe because its a singleton OR... probably because I was not patching the right thing :D)

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging.py
```diff
@@ -0,0 +1,305 @@
...
+            if add_hostname_context:
+                context_data["hostname"] = task_vars.get("inventory_hostname")
+                format_parts.append("<%(hostname)s>")
+
+            format_parts.append("%(message)s")
+            log_format = " ".join(format_parts)
+
+            temp_filters = [ContextFilter(context_data)] if context_data else []
+
+            remove_global_handler = not self._task.args.get("live_display", True)
```
carl-baillargeon, line 226: It feels weird to temporarily remove the "original" handler we initially set in `plugins/__init__.py` to start "fresh".

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging.py
```diff
@@ -0,0 +1,305 @@
...
+                    final_result = func(self, *args, **kwargs)
+                except BaseException as exc:
+                    # Recast errors as AnsibleActionFail
+                    msg = f"Error during plugin execution: {exc}"
+                    raise AnsibleActionFail(msg) from exc
+
+                # Process captured Python warnings and update the result object
+                if captured_warnings:
+                    final_result.setdefault("deprecations", [])
+                    final_result.setdefault("warnings", [])
```
carl-baillargeon, line 249: Python warnings would go here to the appropriate Ansible result fields.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging.py
```diff
@@ -0,0 +1,305 @@
...
+                    log_format=log_format,
+                    remove_global_handler=remove_global_handler,
+                ),
+            ):
+                try:
+                    final_result = func(self, *args, **kwargs)
+                except BaseException as exc:
+                    # Recast errors as AnsibleActionFail
+                    msg = f"Error during plugin execution: {exc}"
+                    raise AnsibleActionFail(msg) from exc
```
carl-baillargeon, line 244: That means we have to remove the try-except blocks we have in our plugins and let them crash so it's handled here.

ansible_collections/arista/avd/plugins/__init__.py
carl-baillargeon: I also feel like the logging configuration is a little bit different per plugin anyways so I don't think we will gain that much in having a "centralized" configuration. I think it would be cleaner and easier to transition if we simply configure logging in each plugin temporarily (using common functions/decorator) and clean-up after each run.

carl-baillargeon: Adding the new `AvdActionPlugin` (see below) + the decorator would basically enable the new logging by itself, making it a lot easier to transition per plugin.

carl-baillargeon:
See #5615.

PR #5546, opened 2025-06-23
Refactor(eos_cli_config_gen)!: Added vrf as unique key for tacacs and radius source interfaces

MaheshGSLAB:
## Change Summary

Added vrf as unique key for tacacs and radius source interfaces. Since EOS supports only unique vrf name with source interface hence added vrf key as unique key. If we add different interface for duplicate vrf the EOS overwrite the command with latest source interface.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

## Proposed changes
Made vrf as unique key for tacacs and radius source interface.

## How to test
Run eos_cli_config_gen molecule

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/ip_tacacs_source_interfaces.schema.yml
+++ b/python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/ip_tacacs_source_interfaces.schema.yml
@@ -9,6 +9,8 @@ keys:
   ip_tacacs_source_interfaces:
     type: list
     display_name: IP Tacacs Source Interfaces
+    unique_keys:
+      - vrf
     items:
       type: dict
       keys:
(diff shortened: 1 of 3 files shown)


Shivani-gslab (approved):
LGTM!

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/ip_tacacs_source_interfaces.schema.yml
```diff
@@ -9,6 +9,8 @@ keys:
   ip_tacacs_source_interfaces:
     type: list
     display_name: IP Tacacs Source Interfaces
+    unique_keys:
+      - vrf
```
gmuloc, line 13: Discussed with @ClausHolbechArista - given that it does not make sense to have the same source interface in multiple VRFs (even if this is allowed by EOS...) we were discussing even adding this:

```suggestion
    primary_key: name
    unique_keys:
      - vrf
```

gmuloc: ok so change of plan looking at the CLI further. We keep it as you have but add a comment for 6.0 to make VRF a primary_key (and the templates will need to be adjusted)

gmuloc:
as discussed closing for now

PR #5569, opened 2025-06-30
Feat(eos_designs): Preview: Digital Twin for ACT. internet_access flag for cloudeos and veos node type

alexeygorbunov:
## Change Summary

Add new boolean flag to control direct Internet access for `cloudeos` (not in ACT documentation but is supported) and `veos` node types

## Related Issue(s)

Fixes #5568

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add new global `digital_twin.fabric.act_internet_access` and node config `digital_twin.act_internet_access` booleans to manage ACT's direct Internet access for cloudeos and veos node type devices

## How to test
- molecule (scenario `eos_designs-twodc-5stage-clos`)
- pytest (for molecule scenarios)

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #5579, opened 2025-07-03
Feat(eos_designs): Preview: Digital Twin for ACT. List of Ethernet ports for veos node types

alexeygorbunov:
## Change Summary

Expose all Ethernet ports of the ACT `veos` node types under "ports" attribute.
This is required to force ACT to attach not only fabric ports (covered by `links` attribute) but as well all ports towards `connected_endpoints` and `network_ports`.
ACT only supports `ports` node-level attribute for `veos` node types. It is not supported by `cloudeos` or any other type.

~Note: Update to the `strip_empties` code is duplicating PR #5569~

## Related Issue(s)

Fixes #aristanetworks/avd-internal#286

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

1. Render "ports" attribute for ACT's `veos` nodes (inside ACT topology file) to include all ethernet_ports.

## How to test
- molecule (scenarios `eos_designs-twodc-5stage-clos` and `digital_twin`)
- pytest (for molecule scenarios)

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_utils/strip_empties.py
```diff
@@ -38,7 +40,7 @@ def strip_empties_from_list(data: list, strip_values_tuple: tuple = (None, "", [
     data : Any
         data to filter
     strip_values_tuple : tuple, optional
-        Value to remove from data, by default (None, "", [], {},)
+        Value to remove from data, by default (None, "", [], {}, (),)
```
gmuloc: so changing this default could be breaking if any of our plugins following semver is using strip_empties with empty tuples that where passing through.
(though .. probably not to be honest)

Just need to make sure we are not changing some existing output or to apply the previous default wherever it could impact.

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -188,6 +189,22 @@ def _get_digital_twin_act(fabric_documentation_facts: FabricDocumentationFacts)
                     node_type=digital_twin_node_type,
                     ip_addr=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..ip_addr", separator=".."),
                     version=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..version", separator=".."),
+                    # Render Ethernet ports for veos node type devices
+                    ports=tuple(
+                        sorted(
+                            (
+                                ethernet_interface["name"]
+                                for ethernet_interface in get(
+                                    fabric_documentation_facts.structured_configs, f"{device}..ethernet_interfaces", [], separator=".."
```
gmuloc, line 243: are you sure you are not getting issues with subinterfaces here?

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -188,6 +189,22 @@ def _get_digital_twin_act(fabric_documentation_facts: FabricDocumentationFacts)
                     node_type=digital_twin_node_type,
                     ip_addr=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..ip_addr", separator=".."),
                     version=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..version", separator=".."),
+                    # Render Ethernet ports for veos node type devices
```
gmuloc: so it says vEOS but I am not seeing where we filter on vEOS devices - is this somewhere before?

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -188,6 +189,22 @@ def _get_digital_twin_act(fabric_documentation_facts: FabricDocumentationFacts)
...
+                    ports=tuple(
+                        sorted(
+                            (
+                                ethernet_interface["name"]
+                                for ethernet_interface in get(
+                                    fabric_documentation_facts.structured_configs, f"{device}..ethernet_interfaces", [], separator=".."
+                                )
+                            ),
+                            # Extract digits from the interface names and use them to sort interfaces using the natural order
+                            # Can not use natural_sort utility here directly due to the triggered CI deps import failure
```
gmuloc, line 248: lets add a TODO on this :)

ansible_collections/arista/avd/extensions/molecule/eos_designs-twodc-5stage-clos/digital_twin/documentation/fabric/TWODC_5STAGE_CLOS-topology.yml
```diff
@@ -1,105 +1,248 @@
...
+    - Ethernet1
+    - Ethernet2
+    - Ethernet3
+    - Ethernet4
 - DC1-POD1-LEAF2B:
     node_type: veos
     ip_addr: 192.168.1.9/16
     version: 4.33.1.1F
+    ports:
+    - Ethernet1
```
gmuloc, line 42: I feel like there will be a subif picked up with the code. Looking at the intended configs there is not a single subinterface in the intended configs :)

gmuloc (changes requested):

python-avd/pyavd/get_fabric_documentation.py
alexeygorbunov: Indeed, sub-interfaces would be added to the topology file. Thanks for catching!
Fixed it now by adding check for a "." present in interface name. Added network services l3_interfaces to confirm

python-avd/pyavd/get_fabric_documentation.py
alexeygorbunov: it's on line 207:

```
185          digital_twin_devices.append(
   186              {
   187                  device: ActNodeSettings(
   188                      # All three values are enforced as non-empty strings during the generation of the metadata part of the structured_config
   189                      node_type=digital_twin_node_type,
   190                      ip_addr=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..ip_addr", separator=".."),
   191                      version=get(fabric_documentation_facts.structured_configs, f"{device}..metadata..digital_twin..version", separator=".."),
   192                      # Render Ethernet ports for veos node type devices
   193                      ports=tuple(
   194                          sorted(
   195                              (
   196                                  ethernet_interface["name"]
   197                                  for ethernet_interface in get(
   198                                      fabric_documentation_facts.structured_configs, f"{device}..ethernet_interfaces", [], separator=".."
   199                                  )
   200                                  if "." not in ethernet_interface["name"]
   201                              ),
   202                              # Extract digits from the interface names and use them to sort interfaces using the natural order
   203                              # Can not use natural_sort utility here directly due to the triggered CI deps import failure
   204                              key=lambda interface_name: list(map(int, re_findall(r"\d+", interface_name))),
   205                          )
   206                      )
   207                      if digital_twin_node_type == "veos"
   208                      else (),
   209                  )
   210              }
   211          )
```

python-avd/pyavd/get_fabric_documentation.py
alexeygorbunov: done

ansible_collections/arista/avd/extensions/molecule/eos_designs-twodc-5stage-clos/digital_twin/documentation/fabric/TWODC_5STAGE_CLOS-topology.yml
alexeygorbunov: Added l3_interfaces in network services and a check for the presence of "." in interface name

python-avd/pyavd/get_fabric_documentation.py
gmuloc: thanks!

python-avd/pyavd/_utils/strip_empties.py
alexeygorbunov: Rolled back all proposed changes to strip_empties. Make `ports` optional to never have an empty tuple (and need to strip it)

ankudinov (approved):
LGTM

ankudinov (approved):
LGTM

ansible_collections/arista/avd/extensions/molecule/digital_twin/inventory/hosts.yml
```diff
@@ -12,6 +12,17 @@ all:
               digital_twin:
                 fabric: {}
                 use_default_interfaces_of_digital_twin_platform: true
+        # Group to test rendering of the Ethernet ports (for veos and clouseos nodes) in the ACT topology file.
```
ClausHolbechArista:
```suggestion
        # Group to test rendering of the Ethernet ports (for veos and cloudeos nodes) in the ACT topology file.
```

ClausHolbechArista (dismissed):
Blocking this. Not sure this is really what we want. We talked about adding support for connected endpoints, and that would solve this automatically. Always outputting all these ports feels wrong to me. Remember it can be thousands of ports for a large config, so we would really want to have more granular control, which is what we talked about adding for connected endpoints.

gmuloc:
> Blocking this. Not sure this is really what we want. We talked about adding support for connected endpoints, and that would solve this automatically. Always outputting all these ports feels wrong to me. Remember it can be thousands of ports for a large config, so we would really want to have more granular control, which is what we talked about adding for connected endpoints.

discussed during maintainer call and we will target outputing everything but disabling them for validation if we know the remote host is not there in the twin

alexeygorbunov:
Added the following adjustments:

```
Render "ports" attribute for ACT's veos nodes (inside ACT topology file) which includes:
All standalone ports towards connected endpoints
All PO member ports towards connected endpoints
All standalone ports defined under network_ports
All POs member ports defined under network_ports
All l3_interfaces defined in network_services and node config
All l3_port_channels defined in network_services and node config
All standalone ports defined under l3_edge/core_interfaces and not connected to another fabric switch
All PO member ports defined under l3_edge/core_interfaces and not connected to another fabric switch
All ports standalone and PO member ports rendered under ports (explained above) will as well get validate_state: false set in their Digital Twin version of the structured config. Parent POs of such member port will get validate_state: false` set as well
```

@gmuloc @ClausHolbechArista

ClausHolbechArista:
> Added the following adjustments:
> 
> ```
> 1. Render "ports" attribute for ACT's veos nodes (inside ACT topology file) which includes:
>   - All standalone ports towards connected endpoints
>   - All POs (and their member ports) towards connected endpoints
>   - All standalone ports defined under network_ports
>   - All POs (and their member ports) defined under network_ports
>   - All l3_interfaces defined in network_services and node config
>   - All l3_port_channels defined in network_services and node config
>   - All standalone ports defined under l3_edge/core_interfaces and not connected to another fabric switch
>   - All POs (and their member ports) defined under l3_edge/core_interfaces and not connected to another fabric switch
> 2. All ports (both POs and their member ports) rendered under ports (explained above) will as well get validate_state: false set in their Digital Twin version of the structured config
> ```
> 
> @gmuloc @ClausHolbechArista

Would it not be simpler to just include all ports that are in the structured config, instead of having to maintain it everywhere where we produce ports? Port-channels should not be in this list. Only physical interfaces.

python-avd/pyavd/_eos_designs/structured_config/network_services/ethernet_interfaces.py
```diff
@@ -308,7 +319,12 @@ def _set_subif_parent_interfaces(self: AvdStructuredConfigNetworkServicesProtoco
             interface = EosCliConfigGen.EthernetInterfacesItem(
                 name=interface_name,
                 shutdown=False,
+                metadata=EosCliConfigGen.EthernetInterfacesItem.Metadata(
+                    peer_type="l3_interface",
+                    # Set validate_state to `False` in Digital Twin mode
+                    validate_state=False if self.shared_utils.digital_twin else Undefined,
```
ClausHolbechArista: I think we should make a helper function in shared utils like `get_interface_validate_state(self: SharedUtilsProtocol, input: bool | None, peer_in_fabric: bool = False) -> bool | Undefined`

For digital twin it would return false unless the peer is in the fabric. For known peers, if input is false that will win, but not if it is true.
For non-digital-twin it will follow the input if not None, and otherwise Undefined.

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -216,6 +243,25 @@ def _get_digital_twin_act(fabric_documentation_facts: FabricDocumentationFacts)
                         and digital_twin_node_type in ["cloudeos", "veos"]
                     )
                     else None,
+                    # Render Ethernet ports for veos node type devices. Skip ports facing other fabric nodes
```
ClausHolbechArista, line 237: as mentioned elsewhere, let's just include all known interfaces in the ports list. We don't kneed to exclude the ones covered by the topology links.

alexeygorbunov: Done, changed back to original approach to include all present Ethernet ports

python-avd/pyavd/_eos_designs/structured_config/network_services/ethernet_interfaces.py
alexeygorbunov: ~Why would we not respect `validate_state` input besides the use case where we know that it is a Digital Twin and remote end of the link is simply not deployed? In all other cases, in both Digital Twin and regular modes users may have reasons to explicitly enable of disable this flag~

Adjusted based on suggestion

python-avd/pyavd/_eos_designs/shared_utils/utils.py
```diff
@@ -319,3 +319,22 @@ def get_local_interface(self: SharedUtilsProtocol, input_interface: str | None)
             case "use_inband_mgmt_interface":
                 return self.inband_mgmt_interface
         return input_interface
+
+    def get_interface_validate_state(self: SharedUtilsProtocol, user_input: bool | None, peer_in_fabric: bool = False) -> bool | UndefinedType:
```
gmuloc, line 323: this could now be moved to StructuredConfigUtils

alexeygorbunov: Done

ClausHolbechArista (approved):

PR #5607, opened 2025-07-09
Feat(eos_designs): Preview: Enhance formatting of the produced ACT Digital Twin topology file

alexeygorbunov:
## Change Summary

Adjust formatting of the produced ACT topology file to make it fully compliant with `yamllint`'s default rule for `indentation` and `document-start`.

## Related Issue(s)

Fixes #aristanetworks/avd-internal#289

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Adjust `eos_designs_documentation.py` plugin to pre-process serialized ACT topology content (to fix indentation) prior to writing it.

I first tried to override specific methods of `AnsibleDumper` (`represent_dict`/`represent_tuple`/`represent_sequence`/`increase_indent`) but no luck therefore fell back to the trivial `re` processing

## How to test
- Molecule scenario `eos_designs-twodc-5stage-clos`
- `yamllint`
```shell
yamllint -d "{extends: default}" molecule/eos_designs-twodc-5stage-clos/digital_twin/documentation/fabric/TWODC_5STAGE_CLOS-topology.yml
```

### User Checklist

- N/A

### Repository Checklist


Shivani-gslab (approved):
LGTM!

Vibhu-gslab (approved):

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -121,8 +123,19 @@ def main(self, validated_args: dict, task_vars: dict, result: dict) -> dict:
 
         if output.digital_twin:
             content = {str(key).replace("_", "-"): value for key, value in asdict(output.digital_twin).items() if value is not None}
+            # Adjust formatting of the dumped Digital Twin artifact
+            match output.digital_twin:
+                case ACTDigitalTwin():
+                    serialized_content = "---\n" + sub(
```
gmuloc: why only for ACT?

alexeygorbunov: We've only been looking into the ACT right now so basically addressing ACT only for now. Other environments may be covered by the same approach as well but we just haven't explored yet.

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -121,8 +123,19 @@ def main(self, validated_args: dict, task_vars: dict, result: dict) -> dict:
...
+                        r"^(\s*-|\s{4})",
+                        lambda match: (" " * 4 + text) if (text := match.group(0)).startswith("  -") else (" " * 2 + text),
+                        yaml.dump(content, Dumper=AnsibleDumper, sort_keys=False, indent=2, width=130),
+                        flags=MULTILINE,
+                    )
+                case _:
+                    serialized_content = (yaml.dump(content, Dumper=AnsibleDumper, sort_keys=False, indent=2, width=130),)
```
carl-baillargeon: Why is this a tuple?
```suggestion
                    serialized_content = yaml.dump(content, Dumper=AnsibleDumper, sort_keys=False, indent=2, width=130)
```
Also, since you are also using `serialized_content` in the case above, you could move this out of the match block:
```suggestion
                    serialized_content = base_yaml
```

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -121,8 +123,19 @@ def main(self, validated_args: dict, task_vars: dict, result: dict) -> dict:
 
         if output.digital_twin:
             content = {str(key).replace("_", "-"): value for key, value in asdict(output.digital_twin).items() if value is not None}
+            # Adjust formatting of the dumped Digital Twin artifact
```
carl-baillargeon, line 128:
```suggestion
            base_yaml = yaml.dump(content, Dumper=AnsibleDumper, sort_keys=False, indent=2, width=130)
            
            # Adjust formatting of the dumped Digital Twin artifact
```

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -121,8 +123,19 @@ def main(self, validated_args: dict, task_vars: dict, result: dict) -> dict:
...
+            match output.digital_twin:
+                case ACTDigitalTwin():
+                    serialized_content = "---\n" + sub(
+                        r"^(\s*-|\s{4})",
+                        lambda match: (" " * 4 + text) if (text := match.group(0)).startswith("  -") else (" " * 2 + text),
+                        yaml.dump(content, Dumper=AnsibleDumper, sort_keys=False, indent=2, width=130),
+                        flags=MULTILINE,
+                    )
```
carl-baillargeon: Consider adding kwargs for clarity:
```suggestion
                    serialized_content = "---\n" + sub(
                        pattern=r"^(\s*-|\s{4})",
                        repl=lambda match: (" " * 4 + text) if (text := match.group(0)).startswith("  -") else (" " * 2 + text),
                        string=base_yaml,
                        flags=MULTILINE,
                    )
```

carl-baillargeon (changes requested):

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
ClausHolbechArista: please use f-strings for compiling strings instead of +

ClausHolbechArista: Also the sub stuff should preferably be done above instead of inline in the string compilation. (for clarity)

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
alexeygorbunov: For the tuple - i must have wrapped it into something else, then removed but haven't cleaned everything properly.
Moved base_yaml outside of the match/case.
Thanks!

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
alexeygorbunov: Done

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
alexeygorbunov: All items should be fixed now

ClausHolbechArista (changes requested):
I prefer if we could avoid this change. I'd rather just exclude the generated files from yamllint. Similar to structured config.
One day we can change our yaml dump tool to one that supports this format natively.

ClausHolbechArista:
Closing. We will not do this, since the current output format matches structured config etc. We should to use a different yaml library in general to fix this.

PR #5615, opened 2025-07-10
Refactor(plugins): Add AvdActionPlugin class with logging config for action plugins

carl-baillargeon:
## Change Summary

Implementation of `AvdActionPlugin` base class on `verify_requirements` to better manage logging.

## Component(s) name

`arista.avd.verify_requirements`

## Proposed changes
- Added `AvdActionPlugin` base class for AVD action plugins
- Added `AvdLoggingConfig` class to control logging settings within a plugin run
- Refactored `verify_requirements` to use this new base class

## How to test
1. Build a playbook for a full deployment using `eos_designs`, `eos_cli_config_gen`, `cv_deploy` and `anta_runner`.
2. Run the playbook with `-vvv` and save the screen output to a file: `ansible-playbook playbooks/fabric-deploy.yml -vvv > output.txt`.
3. Run the same playbook with this PR, also using `-vvv` and save the screen output to a different file `output2.txt`.
4. Open both text files in a file editor, compare the logs and make sure no log is missing.

### Repository Checklist


ansible_collections/arista/avd/plugins/action/verify_requirements.py
```diff
@@ -96,7 +84,7 @@ def _parse_requirements(req_str: str) -> tuple[Requirement, list[str]]:
         req = Requirement(req_str)
     except InvalidRequirement as exc:
         msg = f"Wrong format for requirement {req_str}"
-        raise AnsibleActionFail(msg) from exc
+        raise ValueError(msg) from exc
```
gmuloc, line 88: Note to self: see if this is caught later

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def run(self, tmp: Any = None, task_vars: dict[str, Any] | None = None) -> dict[str, Any]:
+        """Ansible Action entry point."""
+        if task_vars is None:
+            task_vars = {}
+
+        self.result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        return self.run_plugin(task_vars)
```
gmuloc: should you not try except and raise `AnsibleActionFail` from caught exceptions?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
```diff
@@ -0,0 +1,80 @@
...
+        "pyavd": logging.DEBUG,
+        "schema_tools": logging.DEBUG,
+        "anta": logging.DEBUG,
+        "external_libs": logging.DEBUG,
+    },
+}
+"""Map Ansible verbosity levels (0-6) to a dictionary of logger names and their desired logging level."""
+
+EXTERNAL_LIB_LOGGERS = ["asyncio", "httpcore", "httpx", "requests", "urllib3"]
+"""List of common third-party libraries whose log levels are managed collectively. Ignoring `hpack` logger since it is too noisy."""
```
gmuloc: mention in what hpack is used for our sanity :)

gmuloc: maybe others as well

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
```diff
@@ -0,0 +1,80 @@
...
+    Calculate the logging level for a given logger based on Ansible verbosity.
+
+    Args:
+        logger_name: The name of the logger for which to find the level.
+        verbosity: The verbosity level from Ansible Display object (0 for none, 1 for -v, etc.).
+
+    Returns:
+        The calculated logging level.
+    """
+    # If the level is higher than the max defined key (7), fall back to the highest defined level (6)
```
gmuloc: max defined key is 6 right?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_config.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        The calculated logging level.
+    """
+    # If the level is higher than the max defined key (7), fall back to the highest defined level (6)
+    max_defined_verbosity = max(ANSIBLE_VERBOSITY_MAPPING.keys())
+    effective_verbosity = min(verbosity, max_defined_verbosity)
+
+    level_map = ANSIBLE_VERBOSITY_MAPPING[effective_verbosity]
+
+    return level_map.get(logger_name, level_map["external_libs"])
```
gmuloc, line 95: explain that if you don't find the logger name you consider as external lib somewhere :) (I vote comment above 80)

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+
+from .config import get_avd_log_level
+from .handlers import AnsibleDisplayHandler, ContextFilter, SaveToResultHandler
+
+
+def avd_logging(
+    add_hostname_context: bool = False,
+    add_role_context: bool = False,
+    target_loggers: list[str] | None = None,
+    display: Display | None = None,
```
gmuloc: if this is needed only for simplified testing I would vote to not have it at all - maybe we can take a look?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+                context_data["role_name"] = task_vars.get("ansible_role_name")
+                format_parts.append("[%(role_name)s] -")
+            if add_hostname_context:
+                context_data["hostname"] = task_vars.get("inventory_hostname")
+                format_parts.append("<%(hostname)s>")
+
+            format_parts.append("%(message)s")
+            log_format = " ".join(format_parts)
+
+            temp_filters = [ContextFilter(context_data)] if context_data else []
```
gmuloc: make a helper for this

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+
+            # Use the context manager to apply changes and ensure cleanup
+            with (
+                warnings.catch_warnings(record=True) as captured_warnings,
+                avd_logging_context(
+                    logger_names=loggers_to_target, temp_handlers=temp_handlers, temp_filters=temp_filters, log_format=log_format, verbosity=display.verbosity
+                ),
+            ):
+                warnings.simplefilter("always", DeprecationWarning)
```
gmuloc: add a comment above this line to state what you are doing:)

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+
+                try:
+                    final_result = func(self, *args, **kwargs)
+                except BaseException as exc:
+                    # Recast errors as AnsibleActionFail
+                    msg = f"Error during plugin execution: {exc}"
+                    raise AnsibleActionFail(msg) from exc
```
gmuloc: ah so you are doing it in the decorator .. I would do it in your plugin class around to make sure if someone does not use the decorator but your class it is acting properly

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+                    final_result.setdefault("warnings", [])
+                    for w in captured_warnings:
+                        msg = str(w.message)
+                        if issubclass(w.category, DeprecationWarning):
+                            # AvdDeprecationWarning's are added from AvdSchemaTools with more context
+                            # This is a catch-all for other deprecations
+                            final_result["deprecations"].append({"msg": msg})
+                        else:
+                            # Catch-all for standard Python warnings from any library
+                            final_result["warnings"].append(msg)
```
gmuloc: same make a helper

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+            temp_handler.addFilter(temp_filter)
+
+    original_states = {}
+    target_loggers = [logging.getLogger(name) for name in logger_names]
+
+    for logger in target_loggers:
+        # Save original state (level, handlers, propagation)
+        original_states[logger.name] = {"level": logger.level, "handlers": list(logger.handlers), "propagate": logger.propagate}
+        # Defend against lingering handlers from other plugins if not cleaned up
+        logger.handlers.clear()
```
gmuloc: I know why we are doing this but are we being good citizens in general? (we can reevaluate if anyone complains :p )

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        return True
+
+
+class AnsibleDisplayHandler(logging.Handler):
+    """A handler to bridge Python logging to the Ansible Display object for screen output."""
+
+    def __init__(self, display: Display | None = None) -> None:
+        """Initialize the handler."""
+        super().__init__()
+        self.display = display if display is not None else Display()
```
gmuloc: same I feel we can avoid the display parametere as it is a singleton

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_handlers.py
```diff
@@ -0,0 +1,72 @@
...
+
+        Check for a 'color' attribute in the record to allow for explicit color overrides.
+        """
+        message = self.format(record)
+
+        color = getattr(record, "color", None)
+        if color:
+            # If a color is explicitly provided, use it
+            self.display.display(message, color=color)
+            return
```
gmuloc, line 44: should this not respect the level as well?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/handlers.py
```diff
@@ -0,0 +1,72 @@
...
+        elif record.levelno == logging.INFO:
+            self.display.v(message)
+        elif record.levelno == logging.DEBUG:
+            self.display.vvv(message)
+
+
+class SaveToResultHandler(logging.Handler):
+    """A handler that saves warning and error logs to the Ansible result dictionary."""
+
+    def __init__(self, result_dict: dict) -> None:
```
gmuloc: Can we type this better for pyright? :)

ansible_collections/arista/avd/tests/unit/action/test_verify_requirements.py
```diff
@@ -3,6 +3,7 @@
 # that can be found in the LICENSE file.
 
 import os
+import logging
```
gmuloc, line 5: hmm so here we are missing some coverage I suppose and thats why you are getting the failure?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin.py
carl-baillargeon: Agreed. I will move the try except from the decorator to here as discussed.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
carl-baillargeon: I will reword this better.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Maybe in later PR when we do `anta_runner`.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Agreed. Will move it to `AvdActionPlugin`.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: We need it for now until all our plugins are migrated (to avoid having "legacy" handlers from non-migrated plugins).

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,170 @@
...
+
+        # Disabling propagation to avoid duplicate logs in Ansible 'log_path' file
+        logger.propagate = False
```
carl-baillargeon, line 151: Re-evaluate when all plugins are migrated and are no longer using the root logger (cv_workflow, etc.).

ansible_collections/arista/avd/tests/unit/action/test_verify_requirements.py
carl-baillargeon: Correct.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
carl-baillargeon:
```suggestion
    # Any level set above 6 will be treated as 6
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_handlers.py
carl-baillargeon: Display methods like `display.error` do not support the color parameter. They are also doing stuff (like dedup) before calling `display.display`.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: I didn't have much success patching Display... I think it's instantiated pretty early in the Ansible chain or maybe because of the singleton. Maybe I was not patching the right object :) 

Nonetheless I think it's fine to use dependency injection here to simplify testing and avoid weird interactions.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/action/verify_requirements.py
carl-baillargeon: It is caught in `AvdActionPlugin`.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/config.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_config.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Will reorganize when we do the complex ones - anta_runner, cv_workflow, etc.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_handlers.py
carl-baillargeon: See my other comment.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/handlers.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
```diff
@@ -0,0 +1,166 @@
...
+
+from ansible.utils.display import Display
+
+from ansible_collections.arista.avd.plugins.plugin_utils.utils import AvdActionPlugin
+
+from .config import get_avd_log_level
+from .handlers import AnsibleDisplayHandler, ContextFilter, SaveToResultHandler
+
+
+def avd_logging(
```
carl-baillargeon: From Claus: check if we need the decorator at all. We could simply use the base class with class vars and override per plugin. We could use `main` instead of `run_plugin` for clarity.

ClausHolbechArista:
moving to draft while you go through the refactoring we talked about.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: Done in latest commits.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/log_handlers.py
carl-baillargeon: Removed in latest commits.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_logging/decorator.py
carl-baillargeon: I was patching the wrong object. Fixed in latest commits.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+from .log_handlers import AnsibleDisplayHandler, ContextFilter, SaveToResultHandler
+
+
+class AvdActionPlugin(ActionBase):
+    """Base class for AVD Ansible action plugins to provide common functionality."""
+
+    _primary_logger_name: ClassVar[str] = "ansible_collections.arista.avd"
+    _logging_config: ClassVar[AvdLoggingConfig] = AvdLoggingConfig()
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
gmuloc, line 24: use ParamSpec! :p. it looks much better :)

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+                        msg = str(w.message)
+                        if issubclass(w.category, DeprecationWarning):
+                            # AvdDeprecationWarning's are added from AvdSchemaTools with more context
+                            # This is a catch-all for other deprecations
+                            self.result["deprecations"].append({"msg": msg})
+                        else:
+                            # Catch-all for standard Python warnings from any library
+                            self.result["warnings"].append(msg)
+
+        except BaseException as exc:
```
gmuloc: any reason to go for BaseException instead of just the regular Exception?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+            # Recast errors as AnsibleActionFail
+            msg = f"Error during plugin execution: {exc}"
```
gmuloc:
```suggestion
            msg = f"Error during plugin '{self.ansible_name}' execution: '{exc}'"
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+class AvdActionPlugin(ActionBase):
+    """Base class for AVD Ansible action plugins to provide common functionality."""
+
+    _primary_logger_name: ClassVar[str] = "ansible_collections.arista.avd"
+    _logging_config: ClassVar[AvdLoggingConfig] = AvdLoggingConfig()
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        """Initialize the action plugin."""
+        super().__init__(*args, **kwargs)
+        self.result: dict[str, Any] = {}
```
gmuloc, line 27: so as we just discussed you could instead enforce main to receive the result object from run in the signature and it would also work. I don't have a personal favorite so I guess we can reevaluate as we go along?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+        self.logger = logging.getLogger(self._primary_logger_name)
+
+    @abstractmethod
+    def main(self, task_vars: dict[str, Any]) -> None:
+        """
+        This method must be implemented by child plugins with their core logic.
```
gmuloc, line 41: make it clear this will be called from `run` in the docstring

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+
+        self.result = super().run(tmp, task_vars)
+        del tmp  # tmp no longer has any effect
+
+        # Prepare handlers, filters, and format based on logging config and task arguments
+        temp_handlers = []
+        if self._task.args.get("save_logs", False):
+            temp_handlers.append(SaveToResultHandler(result_dict=self.result))
+        if self._task.args.get("live_display", True):
+            temp_handlers.append(AnsibleDisplayHandler(display=Display()))
```
gmuloc: do you need to pass the Display here? I think we don't as it is a singleton. If this is only for tests I would vote not to do it.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+        temp_filters = [ContextFilter(context_data)] if context_data else []
+
+        try:
+            # Use the context manager to apply changes and ensure cleanup
+            with (
+                warnings.catch_warnings(record=True) as captured_warnings,
+                self._logging_context(temp_handlers=temp_handlers, temp_filters=temp_filters, log_format=log_format),
+            ):
+                # DeprecationWarning is ignored by default
+                warnings.simplefilter("always", DeprecationWarning)
```
gmuloc, line 87: we may need to revisit this as this would prevent the user to ignore DeprecationWarnings?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+
+                # Run the plugin
+                self.main(task_vars)
+
+                # Process captured Python warnings and update the result object
+                if captured_warnings:
```
gmuloc: handle this outside of the with block

gmuloc:
```suggestion
            # Process captured Python warnings and update the result object
            if captured_warnings:
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+            msg = f"Error during plugin execution: {exc}"
+            raise AnsibleActionFail(msg) from exc
+        else:
+            return self.result
+
+    @contextmanager
+    def _logging_context(
+        self,
+        temp_handlers: list[logging.Handler],
+        temp_filters: list[logging.Filter],
```
gmuloc, line 118: pyright says you should do one of these


```suggestion
        temp_filters: list[ContextFilter],
```

or

```suggestion
        temp_filters: Sequence[Filter],
```

or type the list line 66 as

```
temp_filter: list[logging.Filter] = ...
```

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
gmuloc: so this looks nicer with Verify Requirement but this does not makwe pyright happy ..

gmuloc: I think we can ignore it it comes from ansible being bad citizen

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+        finally:
+            for logger in target_loggers:
+                # The temporary handlers are the only one present, so clear them
+                logger.handlers.clear()
+
+                # Restore the original state from before we started
+                original_state = original_states[logger.name]
+                logger.setLevel(original_state["level"])
+                logger.handlers.extend(original_state["handlers"])
+                logger.propagate = original_state["propagate"]
```
gmuloc: would look nice with a dataclass :p

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,161 @@
...
+from ansible.utils.display import Display
+
+from .log_config import AvdLoggingConfig, get_avd_log_level
+from .log_handlers import AnsibleDisplayHandler, ContextFilter, SaveToResultHandler
+
+
+class AvdActionPlugin(ActionBase):
+    """Base class for AVD Ansible action plugins to provide common functionality."""
+
+    _primary_logger_name: ClassVar[str] = "ansible_collections.arista.avd"
```
gmuloc, line 21: What if the _primary_logger_name is not part of the LoggingConfig? Then potentially self.logger in the plugin is leading to nowhere during the run / main execution?

gmuloc (commented):
some stuff -

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Ah yes good catch. Added a check in init.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Yes maybe later when we do anta_runner.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Fixed.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Fixed.

carl-baillargeon: Fixed.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: It was indeed to simplify tests but I managed to properly patch the singleton so all good now.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Done.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Right. I also don't have a strong opinion on this. @ClausHolbechArista maybe?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Added your suggestion + ignore for Pyright.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: No. Changed to Exception.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: I think ParamSpec works with Callable only: https://docs.python.org/3/library/typing.html#typing.ParamSpec

_They are used to forward the parameter types of one callable to another callable – a pattern commonly found in higher order functions and decorators. They are only valid when used in Concatenate, or as the first argument to Callable, or as parameters for user-defined Generics._

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: What do you mean? If a user wants to ignore them, he can put `deprecation_warnings = False` in ansible.cfg to not see them via Ansible Display.

We also do this in a `with warnings.catch_warnings` block so the warnings filters reset when we exit the context.

gmuloc: after tests - we acknowledge that we would not respect PYTHONWARNINGS="error::DeprecationWarning" because it would be overwritten but this can be tackled later

gmuloc (approved):
LGTM as it stands (with  next commit ;) ) lets evolve this as we need

gmuloc (approved):
still LGTM

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+
+        Update the `self.result` dictionary attribute in-place to return data to Ansible.
+        """
+
+    def run(self, tmp: Any = None, task_vars: dict[str, Any] | None = None) -> dict[str, Any]:
+        """Ansible Action entry point."""
+        if task_vars is None:
+            task_vars = {}
+
+        self.result = super().run(tmp, task_vars)
```
ClausHolbechArista: you are creating result here as well, so also creating it in __init__ seems weird to me.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
ClausHolbechArista: See comment under run()

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Enforce that the primary logger is always a target for configuration
+        if self._primary_logger_name not in self._logging_config.target_loggers:
+            msg = (
+                f"The _primary_logger_name '{self._primary_logger_name}' must be included "
+                f"in the _logging_config.target_loggers tuple for the plugin to work correctly."
+            )
+            raise ValueError(msg)
+
+    @abstractmethod
```
ClausHolbechArista, line 38: So we also want to prevent subclasses from overriding run? Maybe we should see if abstract something can block that.

ClausHolbechArista (approved):
Minor comments, but overall LGTM

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
gmuloc: so there isnothing native

given that this is only for our usage I feel like the first solution in this thread that should be caught by pyright is good:

https://stackoverflow.com/questions/321024/making-functions-non-override-able 

otherwise we can add the meta class or a check at runtime in __init_subclass__ but I feel we can live with `@final` if it works (which it should based on PEP 591 https://peps.python.org/pep-0591/

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Used update for clarity.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py
carl-baillargeon: Added final.

PR #5638, opened 2025-07-15
Feat(eos_designs): Preview: Digital Twin for ACT. Use standard default creds for vEOS and CloudEOS node types

alexeygorbunov:
## Change Summary

Use standard `cvpadmin/cvp123!` default creds for vEOS and CloudEOS node types

## Related Issue(s)

Fixes #aristanetworks/avd-internal#291

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Use standard `cvpadmin/cv123!` instead of originally-proposed `admin/admin` to provide default compatibility with CVP auto-onboarding workflow.

## How to test
molecule `eos_designs-twodc-5stage-clos` scenario

### User Checklist

- N/A

### Repository Checklist


carl-baillargeon (commented):
Are we ok to expose this?

Also, this username/password combination is required only with the `onboard_device` cvp knob right?

If so, shouldn't we check if the topology has this knob before using this combination? It feels weird to have these defaults if we are not using CVP at all in ACT.

We should also probably add a note somewhere about this.

alexeygorbunov:
> Are we ok to expose this?
> 
> Also, this username/password combination is required only with the `onboard_device` cvp knob right?
> 
> If so, shouldn't we check if the topology has this knob before using this combination? It feels weird to have these defaults if we are not using CVP at all in ACT.
> 
> We should also probably add a note somewhere about this.

As we tested with Petr, these creds will only be used for initial device provisioning and automated CV onboarding. The minute designed AVD config is pushed, those creds will be gone and won't be pushed again.

I have a request from other reviewers to have default creds in place so that digital twin can be generated by the user in "effortless" mode with zero inputs (except a key which enable digital twin). As ACT uses this creds and their lifetime is limited until the first designed config push, i'm not sure if there is any benefit of complicating logic to use one set of default creds if CVP is included and another set of default creds if it's not.

ClausHolbechArista (approved):

ankudinov (approved):

carl-baillargeon (approved):
LGTM!

PR #5640, opened 2025-07-16
Fix(eos_cli_config_gen): Counters for default matches in traffic policies must also be explicitly created

### LINKED ISSUE #5628: Traffic-Policies - counters not created when named counters are defined in the default actions
### Issue Summary

When creating named counters for the default actions e.g.
```yaml
traffic_policies:
  policies:
    - name: TAP
      matches:
        - name: TEST
          source:
             prefix_lists:
                - TEST
           action:
              count: counter_A
      default_actions:
        ipv4:
          count: ipv4_default_ctr
        ipv6:
          count: ipv6_default_ctr

```

They are not created within the traffic-policy
```
   !
   traffic-policy TAP
      counter counter_A
```

eos_cli_config_gen

### AVD version

5.4.0

### Ansible version

N/A

### Python version

N/A

### How do you run AVD ?

### Steps to reproduce

```shell
See description
```

### Relevant log output

```shell

```

### Contributing Guide

colinmacgiolla:
## Change Summary

Named counters for `ipv4-all-default` and `ipv6-all-default` can be created, but must also be explicitly defined for them to work.  

## Related Issue(s)

Fixes #5628

## Component(s) name

## Proposed changes
I reworked some of the logic so that we also check if any counters for the defaults have been created, and add them to the list. Some of the if checks needed to be reworked, as they assumed that you would only have named counters created within match statements, rather than if you just have default_actions

## How to test
The existing tests covered this usecase, just that the outputs were wrong.

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/traffic-policies.j2
+++ b/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/traffic-policies.j2
@@ -46,20 +46,28 @@ traffic-policies
 {# TRAFFIC POLICIES #}
 {%     for policy in traffic_policies.policies | arista.avd.natural_sort('name') %}
    traffic-policy {{ policy.name }}
+{%         set transient_values = namespace() %}
+{%         set transient_values.counters = [] %}
+{%         if policy.default_actions.ipv4.count is arista.avd.defined %}
+{%             do transient_values.counters.append(policy.default_actions.ipv4.count | string) %}
+{%         endif %}
+{%         if policy.default_actions.ipv6.count is arista.avd.defined %}
+{%             do transient_values.counters.append(policy.default_actions.ipv6.count | string) %}
+{%         endif %}
 {%         if policy.matches is arista.avd.defined %}
 {# COUNTER CREATION SECTION #}
-{%             set transient_values = namespace() %}
-{%             set transient_values.counters = [] %}
 {%             for match in policy.matches %}
 {%                 if match.actions.count is arista.avd.defined %}
 {%                     do transient_values.counters.append(match.actions.count | string) %}
 {%                 endif %}
 {%             endfor %}
-{%             if transient_values.counters | length > 0 %}
+{%         endif %}
+{%         if transient_values.counters | length > 0 %}
       counter {{ transient_values.counters | arista.avd.natural_sort | join(' ') }}
       !
-{%             endif %}
+{%         endif %}
 {# MATCH SECTION #}
+{%         if policy.matches is arista.avd.defined %}
 {%             for match in policy.matches %}
       match {{ match.name }} {{ match.type | lower }}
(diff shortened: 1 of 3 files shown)


python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/traffic-policies.j2
```diff
@@ -46,20 +46,28 @@ traffic-policies
 {# TRAFFIC POLICIES #}
 {%     for policy in traffic_policies.policies | arista.avd.natural_sort('name') %}
    traffic-policy {{ policy.name }}
+{%         set transient_values = namespace() %}
+{%         set transient_values.counters = [] %}
+{%         if policy.default_actions.ipv4.count is arista.avd.defined %}
+{%             do transient_values.counters.append(policy.default_actions.ipv4.count | string) %}
```
Shivani-gslab, line 52: for this we can add `convert_types: ["int"]` to the schema

Shivani-gslab: same for the`count` key at other places as well.

Shivani-gslab: Discussed it with @ClausHolbechArista, we don't want to add the logic for generating `counter` command.
we can have a new key `counters`, so that we have the input key 1:1 with the configuration.
Also, we will remove the existing logic for counter in AVD 6.0 and use the new `counters` key.
Let me do a new PR to add this.

ClausHolbechArista: @colinmacgiolla we will create a new PR based on yours. We need to clean this logic up, since we don't want "clever" logic creating config in `eos_cli_config_gen`. Instead we should just have a `counters` key taking a list of strings, so it is the callers responsibility to compile this list. @Shivani-gslab will take care of it based on your work. Thanks.

colinmacgiolla: Just keep in mind that the EOS CLI will accept the usage of named counters, even if they aren't defined. 

But the traffic policy will flag that named counters are used that are not defined if you look at `show traffic-policy interfaces`, and won't actually count, even though there are no config warnings or errors.

ClausHolbechArista: yeah we will help with a description, but eos_cli_config_gen should just translate - not ensure a working config.

Shivani-gslab: PR #5674

Shivani-gslab:
Implemented in PR #5674

PR #5646, opened 2025-07-17
CI: Enable type checking for `pyavd._cv`

gmuloc:
## Change Summary

WIP - need to discuss some points with @ClausHolbechArista 

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_utils/guaranteed_not_none.py
```diff
@@ -0,0 +1,31 @@
...
+if TYPE_CHECKING:
+    from typing import TypeVar
+
+    # TODO: Once we drop 3.13 support this can be imported from typing.
+    from typing_extensions import TypeIs
+
+    T = TypeVar("T")
+
+
+def guaranteed_not_none(value: T | None) -> TypeIs[T]:
```
ClausHolbechArista: this is not what I had in mind.

I wanted:

```python
guaranteed_not_none(value: T | None) -> T:
    if value is None:
        raise ValueError("Value should never be None here")
    return value
```

ClausHolbechArista (changes requested):

python-avd/pyavd/_cv/client/configlet.py
```diff
@@ -50,6 +55,16 @@
 LOGGER = getLogger(__name__)
 
 
+class ConfigletApiTuple(NamedTuple):
```
carl-baillargeon, line 53: We are using dataclasses from models.py for this use case now.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -64,6 +66,10 @@ async def __aexit__(self, _exc_type: type[BaseException] | None, _exc_val: BaseE
             self._channel.close()
             self._channel = None
 
+    @property
+    def channel(self) -> Channel:
+        return guaranteed_not_none(self._channel)
```
carl-baillargeon, line 71: I don't think we should use `guaranteed_not_none` on code we have control on. It should be used when working with API classes created from proto files. @ClausHolbechArista please confirm.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -130,10 +130,13 @@ class GRPCRequestHandler:
...
-    func_signature: Signature
-    bound_arguments: BoundArguments
-    current_arguments_dict: dict
+
+    # These instance variables are set when the decorator is called
+    # Mark them as Optional and initialize to None for Pyright's sake.
+    func: Callable[..., Coroutine[Any, Any, Any]] | None = None
+    func_signature: Signature | None = None
+    bound_arguments: BoundArguments | None = None
+    current_arguments_dict: dict[str, Any] | None = None
```
carl-baillargeon, line 139: These would not be needed after refactoring the decorator in #6169.

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -501,7 +508,7 @@ def generate_internet_exit_metadata(metadata: dict, device: CVDevice, studio_sch
     return services_dict, warnings
 
 
-def generate_applications_metadata(metadata: dict) -> dict:
+def generate_applications_metadata(metadata: dict[Any, Any]) -> dict[str, list[dict[Any, Any]] | dict[str, list[dict[Any, Any]]]]:
```
carl-baillargeon, line 511: I wonder if we can/should spend the energy using `TypedDict` in this module. Carrying raw data around is pretty hard to follow.

carl-baillargeon (commented):
Partial review.

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -154,7 +163,7 @@ async def _sync_studio_roots(
         manual_ids = [container_id for container_id in existing_root_ids if not container_id.startswith(AVD_ENTITY_PREFIX)]
         new_ordered_ids = desired_root_ids + manual_ids
 
-        await cv_client.set_studio_inputs(
+        _ = await cv_client.set_studio_inputs(
```
gmuloc:
```suggestion
        await cv_client.set_studio_inputs(
```

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -168,13 +177,13 @@ async def _sync_studio_roots(
...
-                cast("str", existing_container.display_name)
+                existing_container.display_name or ""
                 for container_id in stale_avd_ids
                 if (existing_container := existing_containers_by_id.get(container_id)) is not None
             ]
         )
 
         # TODO: Build a 'delete_configlet_containers' gRPC API
-        await gather(*[cv_client.delete_configlet_container(workspace_id=workspace_id, assignment_id=container_id) for container_id in stale_avd_ids])
+        _ = await gather(*[cv_client.delete_configlet_container(workspace_id=workspace_id, assignment_id=container_id) for container_id in stale_avd_ids])
```
gmuloc:
```suggestion
        await gather(*[cv_client.delete_configlet_container(workspace_id=workspace_id, assignment_id=container_id) for container_id in stale_avd_ids])
```

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -96,23 +98,23 @@ async def _sync_configlets(cv_manifest: CVManifest, deployment_result: DeployToC
...
         for configlet in existing_configlets
-        if (configlet_id := cast("str", configlet.key.configlet_id)).startswith(AVD_ENTITY_PREFIX) and configlet_id not in desired_configlet_ids
+        if (configlet_id := guaranteed_not_none(configlet.key.configlet_id)).startswith(AVD_ENTITY_PREFIX) and configlet_id not in desired_configlet_ids
     }
 
     if configlets_to_delete:
         LOGGER.info("deploy_static_config_studio_manifest_to_cv: Removing %d AVD-managed configlets which are no longer used.", len(configlets_to_delete))
         deployment_result.removed_static_config_configlets.extend(configlets_to_delete.values())
-        await cv_client.delete_configlets(workspace_id=workspace_id, configlet_ids=list(configlets_to_delete.keys()))
+        _ = await cv_client.delete_configlets(workspace_id=workspace_id, configlet_ids=list(configlets_to_delete.keys()))
```
gmuloc:
```suggestion
        await cv_client.delete_configlets(workspace_id=workspace_id, configlet_ids=list(configlets_to_delete.keys()))
```

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -96,23 +98,23 @@ async def _sync_configlets(cv_manifest: CVManifest, deployment_result: DeployToC
         LOGGER.info("deploy_static_config_studio_manifest_to_cv: Applying changes for %d configlets (create/update)...", len(cv_manifest.configlets))
         deployment_result.deployed_static_config_configlets.extend(configlet.avd_configlet for configlet in cv_manifest.configlets)
         configlet_tuples = [configlet.api_tuple for configlet in cv_manifest.configlets]
-        await cv_client.set_configlets_from_files(workspace_id=workspace_id, configlets=configlet_tuples)
+        _ = await cv_client.set_configlets_from_files(workspace_id=workspace_id, configlets=configlet_tuples)
```
gmuloc:
```suggestion
        await cv_client.set_configlets_from_files(workspace_id=workspace_id, configlets=configlet_tuples)
```

python-avd/pyavd/_cv/client/__init__.py
gmuloc: the problem is that we are typing _channel as `Channel | None`

and when we call `self._channel` in function it can fail

we can use something else than `guaranteed_not_none` but then we should basically raise a RunTimeError stating you should not call the other function without a channel. so why not use this code?

gmuloc: the only thing we can argue against guaranteed_not_none is that it returns a ValueError

python-avd/pyavd/_cv/client/async_decorators.py
gmuloc: Even in #6169 it would still report `reportUninitializedInstanceVariable` which gets flagged for me.

probably we should add the following in our tool.pyright config:

```
typeCheckingMode = "standard"
```

so that we end up with the same warnings across all our env

python-avd/pyavd/_cv/client/__init__.py
carl-baillargeon: The thing is `self._channel` _can_ be None when we initialize the class and/or we use these methods outside of an async context manager. You need a running event loop to initialize it. A RuntimeError would at least be better but I think a better approach would be to avoid having None in the first place. Maybe we should get inspirations from other libraries like aiohttp which require an event loop at the init time directly to create their client to avoid this situation I think.

python-avd/pyavd/_cv/client/async_decorators.py
carl-baillargeon: Alexey and I discussed offline to refactor the decorator which would remove these anyways.

gmuloc: when will these be implemented?

python-avd/pyavd/_cv/client/__init__.py
ClausHolbechArista: we can have a get_channel() helper that guarantees the return type and raises otherwise.

ClausHolbechArista: hm I guess that is actually what you built here as a property. nvm then :)

gmuloc:
we will revisit later

PR #5679, opened 2025-07-30
Refactor(cv_deploy): Revise deploy_to_cv() to separate AVD inputs

ashenoy-arista:
## Change Summary
Revise deploy_to_cv() to maintain AVD inputs as read-only while we fetch and update relevant data from CV Inventory

## Related Issue(s)
https://github.com/aristanetworks/avd/issues/5673

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

When triggering "cv_deploy" role using inputs from structured config, we would like to treat these inputs as read-only to avoid any accidental modification of the same while reviewing or fetching information from CV Inventory/Studios during intermediate steps of deploy.
We would also target generating result for tags/configs/metadata that were added/removed/skipped as applicable across all input devices in overall result.

## How to test

Updated tests under "cv_deploy" to refer to input tags for devices.
Ran "cv_deploy" playbook for tenant `aawg-demo` in cv-staging cluster and manually tested scenarios for device/interface tag addition/removal and validated updated tags, workspace and overall summary of changes in result.

### User Checklist

- N/A

### Repository Checklist

- [ x] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.


ashenoy-arista:
- Initial set of changes with revised data model to hint Avd inputs as read-only while we gather/update relevant information from CV Inventory as part of deploy_to_cv() workflow.
- This approach adds some debug validation to help compare intermediate results such as tags for add/remove with existing approach and raises exception upon finding any deviation. This came in handy in preliminary testing on test tenant using given set of structured configs as inputs (used examples/cv-pathfinder).

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -337,6 +360,131 @@ async def build_object_for_device(
...
+        structured_config_dir: str | None,
+        structured_config_suffix: str,
+        configuration_dir: str,
+        configlet_name_template: str,
+    ) -> AvdDevice:
+        """
+        Build AvdDevice object for one device.
+
+        Parameters:
+            device_list: List of device hostnames.
```
ClausHolbechArista: This looks wrong

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -337,6 +360,131 @@ async def build_object_for_device(
...
+            structured_config_dir: Path to structured config files.
+            structured_config_suffix: Suffix for structured config files.
+            configuration_dir: Path to EOS config files.
+            configlet_name_template: Python string template used for naming configlets. Ex. "AVD-${hostname}"
+        Return:
+            AvdDevice object representing device info such as host name, serial number, system MAC address and other info, along with device and interface tags,
+            CVPathfinderMetaData and EOS config info.
+
+        Workflow:
+            Per device:
```
ClausHolbechArista:
```suggestion
```

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -337,6 +360,131 @@ async def build_object_for_device(
...
+            system_mac_address=system_mac_address,
+        )
+
+        del structured_config
+        return avd_device_object
+
+
+def is_valid_object(avd_device: AvdDevice | None) -> bool:
+    """Checks if the AvdDevice instance is valid (not None in this case)."""
+    return avd_device is not None
```
ClausHolbechArista: Seems a bit needless TBH. Just to a `if avd_device` in the calling code, since None would be falsy and an instance of AvdDevice would be truthly.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -28,30 +30,58 @@ async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvRes
...
 
     if not configs:
         return
+    if not internal_devices:
+        return
 
     # Build todo with CVEosConfig objects that exist on CloudVision. Add the rest to skipped.
     result.skipped_configs.extend(config for config in configs if not config.device._exists_on_cv)
+
+    # maintain list of candidate devices present in CV inventory for reviewing config
```
ClausHolbechArista: Instead of looping and creating helper lists multiple times, we could just parse of the data once and skip the devices that do not exist or do not have a config. When we skip, we can then update a list to be used for the log message.

In general please see if you can avoid looping many times over the same data. Imagine 10000 devices when you layout the code.

The function calls to cloudvision to actually upload the configs and create containers _do_ have to be done for the whole list and not one-by-one.

ClausHolbechArista: alternatively log the skip once per device we skip and update the result. Skipped because of no config should be debug level log. skip for missing device could be info.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -58,6 +67,16 @@ class CVInterfaceTag:
...
 
 
+@dataclass(frozen=True)
+class AvdInterfaceTag:
+    """Input interface tag generated by AVD for a particular device."""
+
+    label: str
+    value: str
+    interface: str | None = None
+    """Must be set for interface tag to be valid"""
```
ClausHolbechArista: Let's just enforce that it is set. Otherwise it is not an interface tag.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -116,17 +142,27 @@ class DeployToCvResult:
     workspace: CVWorkspace | None = field(default_factory=CVWorkspace)
     change_control: CVChangeControl | None = None
     deployed_configs: list[CVEosConfig] = field(default_factory=list)
+    deployed_configs2: dict[str, AvdEosConfig] = field(default_factory=dict)
```
ClausHolbechArista: Maybe list a list of devices for which we have deployed the config?

ClausHolbechArista: This shoudl use your DeviceResult class instead. Don't try to replicate the old results, if we can build something better.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
ashenoy-arista: Have revised the logic, please check

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: a) For now, have preserved the members of class DeployToCvResult
Made changes to underlying logic in deploy related wrappers to populate members as earlier. That way no change needed to existing ansible-based tests that check for result upon deploy operation.
b) It is not clear to me if we are adding value by populating individual members of DeviceResult class (part of InternalDevice class) to hint further on tags/config/pf_metadata being added/removed/verified/skipped.
This is already addressed by class DeployToCvResult (final summary result of deploy).
Please check

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon, line 1: We can introduce breaking changes in AVD 6.0 so let's take the opportunity to update these models in this PR.

I'd formalize it into three distinct layers for clarity:

1- Input layer (AvdXXX) - Frozen dataclasses representing the desired state provided by the user: `AvdDevice`, `AvdStudioInputs`, etc.

2- Workflow layer (WorkflowXXX) - Mutable, internal-only classes that carry state through the deployment process.
  - Rename `InternalDevice` to `WorkflowDevice` containing the original input like you did: `avd_device: AvdDevice`.
  - Will be updated with information from CV (serial_number, etc.).
  - It can contain helper methods to generate payloads for `CVClient`, similar to `api_tuple` on the existing `CVConfiglet` model.
  - It will also carry per-device results throughout the workflow.

3- `DeployToCvResult`: 
  - The final `DeployToCvResult` could report on the `AvdXXX` objects the user provided if possible. This makes the output directly relatable to the input.
  - The `CVXXX` models could be eliminated entirely at this point.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -177,6 +234,96 @@ class CVTimeOuts:
...
+    def get_one_cv_device_tag(self, avd_device_tag: AvdDeviceTag) -> CVDeviceTag:
+        # caller to ensure avd_device_tag is valid
+        return CVDeviceTag(label=avd_device_tag.label, value=avd_device_tag.value, device=self.get_cv_device())
+
+    def get_one_cv_interface_tag(self, avd_interface_tag: AvdInterfaceTag) -> CVInterfaceTag:
+        # caller to ensure avd_interface_tag is valid
+        return CVInterfaceTag(label=avd_interface_tag.label, value=avd_interface_tag.value, device=self.get_cv_device(), interface=avd_interface_tag.interface)
+
+    def get_cv_pathfinder_metadata(self) -> CVPathfinderMetadata:
+        return CVPathfinderMetadata(metadata=self.avd_device.pathfinder_metadata.metadata, device=self.get_cv_device())
```
carl-baillargeon: We wouldn't need these anymore if we decide to return per-workflow result objects to be added to the final `DeployToCvResult` instance.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -171,7 +155,8 @@ async def deploy_to_cv(
 
                 # Deploy device tags
                 await deploy_tags_to_cv(
-                    tags=device_tags,
+                    tag_type="device",
+                    internal_devices=internal_devices,
```
carl-baillargeon: Since we provide all `internal_devices` with their tags (device + interface), can we await the coroutine only once and remove the `tag_type` parameter?

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: Great suggestions! Thank you
- Added WorkflowDevice class with DeviceResult member to accumulate per-device changes such as deployed/skipped/removed for device/interface tags, config and cv_pathfinder metadata
- We then make use of this per-device result to populate overall deployment result
- class DeployToCvResult - for now making no changes here

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: Retaining these under WorkflowDevice class
These are used to populate members within class DeployToCvResult

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
ashenoy-arista: Added a helper wrapper that takes in tag_type parameter. 
deploy_tags_to_cv() now calls this helper with each tag_type - "device" and "interface"
This allows caller to await on the coroutine once.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: Agreed on 3- `DeployToCvResult`. Let's keep the same interface in this PR to comply with molecule tests.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: As discussed, keeping `DeployToCvResult` as is for this PR to comply with molecule tests.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
carl-baillargeon: See my other comment about calling `get_tags` once for both tag types.

carl-baillargeon (commented):
Few comments.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -255,31 +247,23 @@ async def build_objects(
...
-            Tuple containing (<EOS Configs to deploy>, <Device Tags to deploy>, <Interface Tags to deploy>, <CV Pathfinder Metadata to deploy>).
+            List of <AvdDevice>.
 
         Workflow:
             Per device:
               - Read and load structured config
               - If is_deployed is false, skip the device.
               - Read serial_number & system_mac from structured config.
-              - Create CVDevice object and add to list of device_objects.
+              - Create list of AvdDevice objects.
```
carl-baillargeon: This is per device so this is inaccurate.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -370,26 +352,37 @@ async def build_object_for_device(
...
             for one_interface_tags in all_interface_tags
             if "interface" in one_interface_tags and "tags" in one_interface_tags
             for interface_tag in one_interface_tags["tags"]
             if "name" in interface_tag and "value" in interface_tag
-        ]
+        )
 
         # Build WAN metadata object for this device.
-        cv_pathfinder_metadata_objects = []
+        avd_cv_pathfinder_metadata_object = AvdPathfinderMetadata(metadata={})
```
carl-baillargeon:
```suggestion
        avd_cv_pathfinder_metadata_object = AvdPathfinderMetadata(metadata= get(structured_config, "metadata.cv_pathfinder", default={}))
```
Then you can remove the if statement below.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -28,46 +28,58 @@ async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvRes
...
-    LOGGER.info("deploy_configs_to_cv: %s skipped configs because the devices are missing from CloudVision.", len(result.skipped_configs))
-    todo_configs = [config for config in configs if config.device._exists_on_cv]
-    LOGGER.info("deploy_configs_to_cv: %s todo configs.", len(todo_configs))
+    # Build todo with WorkflowDevice objects that exist on CloudVision. Add the rest to skipped.
+    todo_devices_with_configs: list[WorkflowDevice] = []
+    skipped_devices = 0
+    for device in workflow_devices:
+        if not device.in_cv_inventory:
+            skipped_devices += 1
+            device.result.config.skipped.extend([device.avd_device.config])
```
carl-baillargeon: I see you are using the `.extend([<single_item>])` pattern in a few places in the code. Any specific reason why?
```suggestion
            device.result.config.skipped.append(device.avd_device.config)
```

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -28,46 +28,58 @@ async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvRes
...
+            LOGGER.info("deploy_configs_to_cv: Skipping config for device %s missing in CloudVision Inventory", device.avd_device.hostname)
+        else:
+            todo_devices_with_configs.append(device)
+
+    LOGGER.info("deploy_configs_to_cv: %s skipped configs because the devices are missing from CloudVision.", skipped_devices)
```
carl-baillargeon, line 50: Can we just use `len()` here and avoid the counter?
```suggestion
    LOGGER.info("deploy_configs_to_cv: %s skipped configs because the devices are missing from CloudVision.", len(device.result.config.skipped))
```

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
```diff
@@ -15,19 +17,31 @@
...
     cv_client: CVClient,
 ) -> None:
     """
-    Deploy Tags updating result with success, skipped, errors, warnings.
+    Deploy tags including device and interface tags.
+
+    Please refer to helper_deploy_tags_to_cv() below for detailed implementation.
+    """
+    for tag_type in ["device", "interface"]:
+        await helper_deploy_tags_to_cv(tag_type, workflow_devices, deployment_result.workspace, strict, cv_client)
```
carl-baillargeon: I feel like treating the two tag types separately complicates the logic a lot. Please check if we can use `cv_client.get_tags()` with `element_type=None` to retrieve **all** user tags at once and go from there. Same for the assignments.

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
```diff
@@ -15,19 +17,31 @@
...
 
-    Tags can be either Device Tags or Interface Tags but *not* a combination.
 
+async def helper_deploy_tags_to_cv(
```
carl-baillargeon: Thinking about the complexity here with the multiple loops over devices and the temporary placeholders for results, I wonder if we could simplify this by introducing `WorkflowTag`?

For example:
```
class DeploymentStatus(Enum):
    PENDING = 1
    DEPLOYED = 2
    SKIPPED = 3
    REMOVED = 4

@dataclass
class WorkflowTag:
    input: AvdDeviceTag | AvdInterfaceTag
    parent_device: WorkflowDevice
    status: DeploymentStatus = DeploymentStatus.PENDING
```
We could create one flat list of all `WorkflowTag` objects from all devices and update the `.status` on each as we go.

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
```diff
@@ -37,42 +51,66 @@ async def deploy_tags_to_cv(
...
+    # Build list of devices with tag objects if device exists in CloudVision. Add the rest to skipped.
+    for device in workflow_devices:
+        tags_matching_tag_type = device.avd_device.device_tags if tag_type == "device" else device.avd_device.interface_tags
+        result_placeholder_skipped = device.result.device_tags.skipped if tag_type == "device" else device.result.interface_tags.skipped
+        if not device.in_cv_inventory:
+            if len(tags_matching_tag_type) > 0:
+                result_placeholder_skipped.extend(list(tags_matching_tag_type))
+            continue
+        if not device.serial_number:
+            # TODO: Check if logic below to raise exception is fine
```
carl-baillargeon: This check should be done earlier, probably in `verify_devices_on_cv` when we update the serial numbers.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -56,11 +42,11 @@ async def deploy_to_cv(
     """
     Deploy various objects to CloudVision.
 
-    For any device referred under `configs`, `device_tags` and `interface_tags` the device:
+    For any device referred under `avd_devices`:
     - The device must be present in the CloudVision Inventory and onboarded to the "Inventory & Topology Studio".
         - TODO: See if we can onboard ZTP devices and/or preprovision.
     - The hostname will we updated in the I&T Studio.
-    - The `serial_number` and `system_mac_address` properties will be inplace updated in the given CVDevice objects.
+    - The `serial_number` and `system_mac_address` properties will be updated within per-device result member.
```
carl-baillargeon: "within per-device result member"? This looks wrong.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -72,12 +58,10 @@ async def deploy_to_cv(
...
             It is not supported to reuse an existing Change Control, so the `id` field should not be set in the given CVChangeControl object. \
             The `id` and `state` properties will be inplace updated in the given CVChangeControl object.
-        configs: Configs to be deployed using the "Static Configlet Studio".
-        device_tags: Device Tags to be deployed and assigned.
-        interface_tags: Interface Tags to be deployed and assigned.
+        avd_devices: AvdDevice objects representing device info, config, device/interface tags and CVPathfinderMetaData. \
+            This signifies read-only input from AVD intended to be deployed to CV.
         studio_inputs: Studio Inputs to be deployed. \
             It is not supported to update overlapping input paths for the same studio in the same deployment.
-        cv_pathfinder_metadata: Special metadata for CV Pathfinder solution. Metadata will be combined and deployed to the hidden metadata studio.
```
carl-baillargeon, line 80: Please validate if the metadata Studio is still hidden on CV. Otherwise we should keep this description.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -212,17 +174,20 @@ async def deploy_to_cv(
...
                 await deploy_cv_pathfinder_metadata_to_cv(
-                    cv_pathfinder_metadata=cv_pathfinder_metadata,
-                    result=result,
+                    workflow_devices=workflow_devices,
+                    deployment_result=result,
                     cv_client=cv_client,
                 )
 
+                # Gather per-device result from each workflow device and set within overall result
+                gather_per_device_results(workflow_devices=workflow_devices, deployment_result=result)
```
carl-baillargeon: Let's have a method `update_from_workflow_devices` on `DeployToCvResult` instead.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -61,6 +61,14 @@ class CVDeviceTag:
     device: CVDevice | None = None
 
 
+@dataclass(frozen=True)
+class AvdDeviceTag:
```
carl-baillargeon: Let's move the `AvdXXX` classes to another module called `input_models.py` for clarity.

carl-baillargeon: Let's also add TODOs in the new module for the other models not migrated yet - `AvdWorkspace`, `AvdStudioInputs`, etc.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -85,6 +102,13 @@ class CVPathfinderMetadata:
     device: CVDevice | None = None
 
 
+@dataclass(frozen=True)
+class AvdPathfinderMetadata:
+    """Input Pathfinder metadata generated by AVD."""
+
+    metadata: dict
```
carl-baillargeon: Even with `frozen=True`, metadata is mutable. Something to keep in mind. Not sure if we can guard against that.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -160,6 +184,28 @@ class CVDevice:
...
 
 
+@dataclass(frozen=True)
+class AvdDevice:
+    """Input generated by AVD for a particular device to be deployed."""
+
+    config: AvdEosConfig
+    device_tags: tuple[AvdDeviceTag]
+    interface_tags: tuple[AvdInterfaceTag]
+    pathfinder_metadata: AvdPathfinderMetadata
```
carl-baillargeon: Let's put these fields as optionals. A user might want to push a single device tag only: `my_device = AvdDevice(hostname="DC1-LEAF1A", device_tags=(AvdDeviceTag,))`.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -177,6 +233,66 @@ class CVTimeOuts:
...
+
+
+@dataclass
+class DeviceResult:
+    """Device specific results from deployment to CV."""
+
+    device_tags: DeploymentResult[AvdDeviceTag]
+    interface_tags: DeploymentResult[AvdInterfaceTag]
+    config: DeploymentResult[AvdEosConfig]
+    pathfinder_metadata: DeploymentResult[AvdPathfinderMetadata]
```
carl-baillargeon: Related to my comment here: https://github.com/aristanetworks/avd/pull/5679/files#r2383804793

Let's replace this by an Enum:

```suggestion
class DeploymentStatus(Enum):
    PENDING = 1
    DEPLOYED = 2
    SKIPPED = 3
    REMOVED = 4
```
Then we can create the appropriate `WorkflowXXX` stateful models for all deployment objects. See the other comments in this module.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -177,6 +233,66 @@ class CVTimeOuts:
...
+class WorkflowDevice:
+    """Internal mutable representation of a device for deployment workflow with inputs from AVD."""
+
+    avd_device: AvdDevice
+    result: DeviceResult
+    in_cv_inventory: bool = False
+    serial_number: str | None = None
+    """Device serial number retrieved from CV"""
+    system_mac_address: str | None = None
+    """Device system mac address retrieved from CV"""
```
```suggestion
    input: AvdDevice
    
    # State discovered from CloudVision
    in_cv_inventory: bool = False
    serial_number: str | None = None
    system_mac_address: str | None = None
    
    # Containers for stateful workflow objects
    config: WorkflowConfig | None = None
    device_tags: list[WorkflowTag] = field(default_factory=list)
    interface_tags: list[WorkflowTag] = field(default_factory=list)
    pathfinder_metadata: WorkflowPathfinderMetadata | None = None
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -169,6 +215,16 @@ class CVEosConfig:
     """By default "AVD_<hostname>"""
 
 
+@dataclass(frozen=True)
+class AvdEosConfig:
```
carl-baillargeon: Let's also add the corresponding `WorkflowEosConfig`:
```
@dataclass
class WorkflowConfig:
    input: AvdEosConfig
    parent_device: WorkflowDevice
    status: DeploymentStatus = DeploymentStatus.PENDING
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -85,6 +102,13 @@ class CVPathfinderMetadata:
     device: CVDevice | None = None
 
 
+@dataclass(frozen=True)
+class AvdPathfinderMetadata:
```
carl-baillargeon: Let's also add `WorkflowPathfinderMetadata`:
```
@dataclass
class WorkflowPathfinderMetadata:
    input: AvdPathfinderMetadata
    parent_device: WorkflowDevice
    status: DeploymentStatus = DeploymentStatus.PENDING
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -70,6 +78,15 @@ class CVInterfaceTag:
     """Must be set if device is set"""
 
 
+@dataclass(frozen=True)
+class AvdInterfaceTag:
```
carl-baillargeon: For tags, let's add `WorkflowTag`:
```
@dataclass
class WorkflowTag:
    input: AvdDeviceTag | AvdInterfaceTag
    parent_device: WorkflowDevice
    status: DeploymentStatus = DeploymentStatus.PENDING
```

carl-baillargeon: Consider Generics.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: Please consider introducing other `WorkflowXXX` objects as suggested in the comments of this module. I think this design will simplify the logic quite a lot in the sub-workflows. You can also add methods/properties to them as needed.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -17,47 +17,53 @@
...
 
-    Returns a list of CVDevice objects found to exist on CloudVision.
+    Returns a filtered list of WorkflowDevice objects found to exist in CloudVision.
     """
     # Using set to only include a device once.
+    # Each device will have in_cv_inventory member set to False initially until it is reviewed.
     device_tuples = {
-        (device.serial_number, device.system_mac_address, device.hostname if not any([device.serial_number, device.system_mac_address]) else None)
-        for device in devices
-        if device._exists_on_cv is None
```
carl-baillargeon, line 62: Why are we removing this check?

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -67,60 +73,67 @@ async def verify_devices_in_cloudvision_inventory(
...
         len(existing_device_tuples),
     )
 
-    if missing_devices := [device for device in devices if not device._exists_on_cv]:
+    if missing_devices:
         warnings.append(
             missing_devices_handler(missing_devices=missing_devices, skip_missing_devices=skip_missing_devices, context="CloudVision Device Inventory")
         )
 
+    # TODO: Should we add a validation (possibly raise an exception)
```
carl-baillargeon: @ClausHolbechArista Please confirm here.

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -9,15 +9,24 @@
...
 
 if TYPE_CHECKING:
-    from .models import CVDevice
+    from .models import AvdDevice
 
 LOGGER = getLogger(__name__)
 
 
-def verify_device_inputs(devices: list[CVDevice], warnings: list[Exception], *, strict_system_mac_address: bool) -> None:
+def verify_device_inputs(avd_devices: list[AvdDevice] | None, warnings: list[Exception], *, strict_system_mac_address: bool) -> None:
```
carl-baillargeon: Why are we accepting None here? If we don't have any devices we shouldn't be calling this function in the first place no?

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -29,27 +38,31 @@ def verify_device_inputs(devices: list[CVDevice], warnings: list[Exception], *,
...
     """
     duplicated_devices = DuplicatedDevices()
 
-    # Deduplicate CVDevice objects as original `devices` list may contain duplicated items
-    unique_devices = list({id(device): device for device in devices}.values())
+    if not devices:
+        return duplicated_devices
+
+    # build a list of CVDevice instance for each AvdDevice instance
+    unique_devices = list({id(device): device.get_cv_device() for device in devices}.values())
```
carl-baillargeon: In the old workflow, it was possible to have the same `CVDevice` object twice because of the `deploy_to_cv` inputs but now I don't think it's possible with AvdDevice right?

carl-baillargeon: Umm I guess it is technically possible to provide the same AvdDevice object to `deploy_to_cv` twice but that would be a misusage and we should probably let the logic below catch it I think.

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -118,3 +140,30 @@ def duplicated_devices_handler(
         duplicated_devices.system_mac_address.set_serial_number,
     )
     warnings.append(CVDuplicatedDevices("Duplicated devices found in inventory", duplicated_devices.system_mac_address.set_serial_number))
+
+
+def build_workflow_device_inputs(avd_devices: list[AvdDevice] | None) -> list[WorkflowDevice]:
```
carl-baillargeon: Same comment as above, we shouldn't be accepting None here.

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -118,3 +140,30 @@ def duplicated_devices_handler(
...
+        return workflow_devices
+    for avd_device in avd_devices:
+        # Device result members are initialized with respective members with empty values.
+        # This will make it easier for subsequent wrappers to access and update these members while working on them.
+        device_result = DeviceResult(
+            device_tags=DeploymentResult[AvdDeviceTag](),
+            interface_tags=DeploymentResult[AvdInterfaceTag](),
+            config=DeploymentResult[AvdEosConfig](),
+            pathfinder_metadata=DeploymentResult[AvdPathfinderMetadata](),
+        )
```
carl-baillargeon: No longer needed if we use `DeploymentStatus`.
```suggestion
```

carl-baillargeon: After all changes I think this function could be replaced by a simple list comprehension in `deploy_to_cv`. We could also have a class method like: `WorkflowDevice.from_avd_device(avd_device)`.

python-avd/pyavd/_eos_designs/structured_config/network_services/utils_zscaler.py
```diff
@@ -64,9 +73,25 @@ async def _generate_zscaler_endpoints(self: AvdStructuredConfigNetworkServicesPr
...
 
         async with CVClient(servers=[cv_server], token=cv_token) as cv_client:
-            cv_device = CVDevice(self.shared_utils.hostname, self.shared_utils.serial_number, self.shared_utils.system_mac_address)
-            cv_inventory_devices: list[CVDevice] = await verify_devices_in_cloudvision_inventory(
-                devices=[cv_device],
+            avd_device = AvdDevice(
+                config=AvdEosConfig(file=""),
+                device_tags=tuple(),  # noqa: C408
+                interface_tags=tuple(),  # noqa: C408
+                pathfinder_metadata=AvdPathfinderMetadata(metadata={}),
```
carl-baillargeon: These fields will no longer be needed if you update them to optional fields.
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils_zscaler.py
```diff
@@ -64,9 +73,25 @@ async def _generate_zscaler_endpoints(self: AvdStructuredConfigNetworkServicesPr
...
+                hostname=self.shared_utils.hostname,
+                serial_number=self.shared_utils.serial_number,
+                system_mac_address=self.shared_utils.system_mac_address,
+            )
+            device_result = DeviceResult(
+                device_tags=DeploymentResult[AvdDeviceTag](),
+                interface_tags=DeploymentResult[AvdInterfaceTag](),
+                config=DeploymentResult[AvdEosConfig](),
+                pathfinder_metadata=DeploymentResult[AvdPathfinderMetadata](),
+            )
```
carl-baillargeon: Same here. Not needed if we use `DeploymentStatus`.
```suggestion
```

carl-baillargeon (changes requested):

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
ashenoy-arista: Moving validation to verify_devices_in_cloudvision_inventory() under verify_devices_on_cv.py

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
ashenoy-arista: Restoring previous verbage

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
ashenoy-arista: Rephrased this and included under `avd_devices` input. We no longer have explicit input arg for `cv_pathfinder_metadata` as it is contained within each `avd_device`

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: Implemented WorkflowXXX objects under class WorkflowDevice as suggested.
Thank you for this suggestion.

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
ashenoy-arista: We now call get_tags() and get_tag_assignments() once for fetching ALL user-created tags (device and interface tags). Later, after determining specific tags to be added and deleted, we make separate cv_client calls for set_tag_assignments() and delete_tag_assignments() for "device" and "interface" tag_type based on current cv_client API usage.

python-avd/pyavd/_cv/workflows/verify_inputs.py
ashenoy-arista: Update function signature to not accept None for `avd_devices`.

python-avd/pyavd/_cv/workflows/verify_inputs.py
ashenoy-arista: We expect device name (input) to be provided only once in device_list.
No validation added yet, if needed we could add check in build_objects() within plugins/action/cv_workflow.py and raise appropriate exception if duplicate device names are found.

python-avd/pyavd/_eos_designs/structured_config/network_services/utils_zscaler.py
ashenoy-arista: Under AvdDevice class, config/device_tags/interface_tags/pathfinder_metadata members are not optional - defaulting to None or empty list. This will help support workflows that may perform selected deployment changes for a given device.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -19,7 +21,7 @@
 STATIC_CONFIGLET_STUDIO_ID = "studio-static-configlet"
 
 
-async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvResult, cv_client: CVClient) -> None:
+async def deploy_configs_to_cv(workflow_devices: list[WorkflowDevice], deployment_result: DeployToCvResult, cv_client: CVClient) -> None:
```
carl-baillargeon: With the new structure we don't really need to provide the full `DeployToCvResult` instance anymore. We could simply provide the workspace ID.

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -340,94 +345,119 @@ async def deploy_cv_pathfinder_metadata_to_cv(cv_pathfinder_metadata: list[CVPat
...
-    # Get existing studio inputs
-    existing_studio_inputs = await cv_client.get_studio_inputs(
-        studio_id=CV_PATHFINDER_METADATA_STUDIO_ID,
-        workspace_id=result.workspace.id,
-        default_value=CV_PATHFINDER_DEFAULT_STUDIO_INPUTS,
-    )
-    studio_inputs = deepcopy(existing_studio_inputs)
-    if not isinstance(studio_inputs, dict):
-        # Resetting studio inputs to an empty dict in case we received a wrong type (like None)
-        studio_inputs = {}
```
carl-baillargeon: Why are we moving the logic after parsing the workflow devices?

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
```diff
@@ -39,97 +35,125 @@ async def deploy_tags_to_cv(
...
-
-    # Build a mapping of device serial number to CVDevice.
-    devices_by_serial_number = {
-        tag.device.serial_number: tag.device for tag in deployed_tags if tag.device is not None and tag.device.serial_number is not None
-    }
+    # Build separate dict of serial number mapping to WorkflowDevice object - one for device tags and another for interface tags.
+    # This is needed to determine tags to be unassigned to support workflows that may be deploying
+    # *only* device vs *only* interface tags based on AVD input provided for a given device.
+    devices_by_serial_number_for_device_tags: dict[str, WorkflowDevice] = {}
+    devices_by_serial_number_for_interface_tags: dict[str, WorkflowDevice] = {}
```
carl-baillargeon, line 108: I don't understand why we need to keep two mappings here. `CVTagAssignment` objects already contain all the information needed to reconstruct the appropriate `AvdDeviceTag` or `AvdInterfaceTag` later into the `WorkflowTag`.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -221,6 +273,48 @@ class CVEosConfig:
     """By default "AVD_<hostname>"""
 
 
+@dataclass(frozen=True)
+class AvdDevice:
```
carl-baillargeon, line 283: This should be in `input_models.py`.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -151,59 +128,40 @@ async def deploy_to_cv(
...
-                }.values()
-            )
             # Check structured config of the targeted devices for overlapping `serial_number`s or `system_mac_address`es.
-            verify_device_inputs(devices, result.warnings, strict_system_mac_address=strict_system_mac_address)
+            verify_device_inputs(avd_devices, result.warnings, strict_system_mac_address=strict_system_mac_address)
+
+            # Form list of WorkflowDevice instance from given list of input AvdDevice instance.
+            # We will be then passing list of WorkflowDevice instance to rest of the wrappers
+            # to help identify changes to be done on CV side.
+            workflow_devices = build_workflow_device_inputs(avd_devices)
```
carl-baillargeon: No need to have a separate function in a different module. The comprehension can go here:
```suggestion
            workflow_devices = [WorkflowDevice.from_avd_input(avd_device) for avd_device in avd_devices]
```

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
carl-baillargeon: Same applies for other sub-workflows.

ashenoy-arista: Revising signatures for all deploy_xxx() functions except deploy_static_config_studio_manifest_to_cv() and deploy_studio_inputs_to_cv(). These functions need to update members of `deployment_result` such as `deployed_static_config_containers`, `skipped_static_config_containers`, `removed_static_config_configlets` and `deployed_studio_inputs`. Hence not modifying these for now for allowing ease of access to these members.

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ashenoy-arista: Being consistent with other sub-workflows, if there is no relevant input provided as part of deploy workflow related to pathfinder metadata, we can return right away and avoid making any api calls for fetching input schema for the cv pathfinder metadata studio via get_metadata_studio_schema(...) and cv_client.get_studio_inputs(...).
Also this change helps tests related to mocked api recording from Alexey to pass since there is no such recorded api request/response for pathfinder metadata studios given that test inputs don't include this.

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
ashenoy-arista: Here is my reasoning (sorry for being verbose)
Ideally single `devices_by_serial_number` dictionary built from `deployed_tags` could work.
But if you check existing logic (without changes in this PR), it checks for "device" tags and "interface" separately via making call for deploy_tags_to_cv() each time. Thus maintaining clear separation.

Lets consider existing logic for handling `strict` v/s `non-strict` tags:
```
devices_by_serial_number = {
        tag.device.serial_number: tag.device for tag in deployed_tags if tag.device is not None and tag.device.serial_number is not None
    }
    ...
if strict:
        assignments_to_unassign = {
            assignment for assignment in existing_assignments if assignment.device_id in devices_by_serial_number and assignment not in desired_assignments
        } 
else:
        # Build set of tag labels we have assigned so we know which ones to remove.
        desired_tags_labels = {assignment.label for assignment in desired_assignments}
        assignments_to_unassign = {
            assignment
            for assignment in existing_assignments
            if assignment.device_id in devices_by_serial_number and assignment.label in desired_tags_labels and assignment not in desired_assignments
        }       
```
For `strict` case (when say handling `device` tags) - 
My understanding is that our intention is to identify those assignments which are present in `existing_assignments in CV` but not in `desired_assignments in input` AND 
we have atleast one `device tag` being provided in input for that device.
i.e. if user input does not specify any device_tags for device A, we should NOT be including any device tag assignment for this device within `assignments_to_unassign` even if user has provided `interface` tags input for same device A. Same applies to `non-strict` case as well.

For scenario, where user has provided only interface tags and no device tags as input for a given device, 
if we don't maintain separate dictionaries but instead maintain common dict, we will end up finding device serial_num in common dict due to presence interface tag in input for that device.
This may give us incorrect results when computing `assignments_to_unassign`.

In PR, we have a consolidated deploy_tags_to_cv() for device and interface tags combined.
Having separate dict mapping would help to handle user-provided input combinations that may either include 
both device and interface tags vs only device tags vs only interface tags for a given device.

Similarly in the case of `non-strict` scenario:
```
desired_tags_labels = {assignment.label for assignment in desired_assignments}
```
Now there could be unlikely scenario where the tag label (say "foo") could be identical for device and interface tag. Logic should avoid accidentally matching against tag label coming from device when working on interface tags and vice-versa. 
Hence my proposal on maintaining separate dict mapping for device serial number for deployed device tags vs interface tags to WorkflowTag, when computing `assignments_to_unassign`.

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: class AvdDevice includes method `get_cv_device()` that returns CVDevice and classmethod `from_device_input()` that refers to CVDevice. CVDevice is part of models.py
Hence we may need to keep AvdDevice class in models.py to allow it to reference CVDevice.
Let me know.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
ashenoy-arista: Sounds good.

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
carl-baillargeon: Understood. Thanks.

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
carl-baillargeon: Thanks for the detailed explanation. All clear now.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: Ok we can leave it there for now. We'll move things later in a separate PR if needed.

ashenoy-arista:
Updated logic in verify_devices_on_cv.py to ensure we honor serial_num and system_mac_address for input devices

python-avd/pyavd/_cv/workflows/input_models.py
```diff
@@ -0,0 +1,47 @@
+# Copyright (c) 2023-2025 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from dataclasses import dataclass
+from uuid import NAMESPACE_DNS, uuid5
+
+AVD_NAMESPACE = uuid5(NAMESPACE_DNS, "avd.arista.com")
+AVD_ENTITY_PREFIX = "avd_"
```
alexeygorbunov: Where are `AVD_NAMESPACE` and `AVD_ENTITY_PREFIX` used?

python-avd/pyavd/_cv/workflows/input_models.py
```diff
@@ -0,0 +1,47 @@
...
+    label: str
+    value: str
+    interface: str
+
+
+@dataclass(frozen=True)
+class AvdPathfinderMetadata:
+    """Input Pathfinder metadata generated by AVD."""
+
+    metadata: dict
```
alexeygorbunov: do we need to clarify type annotation here?

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -221,6 +273,48 @@ class CVEosConfig:
...
+    def from_device_input(
+        cls,
+        cv_device: CVDevice,
+        config: AvdEosConfig | None,
+        device_tags: tuple[AvdDeviceTag, ...],
+        interface_tags: tuple[AvdInterfaceTag, ...],
+        pathfinder_metadata: AvdPathfinderMetadata | None,
+    ) -> AvdDevice:
+        # used primarily for testing
+        return AvdDevice(
```
alexeygorbunov:
```suggestion
        return cls(
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -192,6 +195,55 @@ class DeployToCvResult:
...
+    def update_from_workflow_devices(self, workflow_devices: list[WorkflowDevice]) -> None:
+        # iterate thru each workflow device and update deployed/skipped/removed placeholder
+        # for device tags/interface tags/config/pathfinder metadata in overall deployment result
+        for device in workflow_devices:
+            # iterate thru all device_tags checking for deployment status and accumulate under correct bucket
+            for device_tag in device.device_tags:
+                match device_tag.status:
+                    case DeploymentStatus.SKIPPED:
+                        self.skipped_device_tags.append(device.get_one_cv_tag(device_tag.input))
+                    case DeploymentStatus.DEPLOYED:
```
alexeygorbunov, line 208: There is a type checking issue here (and below). Should we explicitly check type of a return to make checker happy?

```suggestion
                    case DeploymentStatus.SKIPPED:
                        if isinstance(cv_tag := device.get_one_cv_tag(device_tag.input), CVDeviceTag):
                            self.skipped_device_tags.append(cv_tag)
                    case DeploymentStatus.DEPLOYED:
```

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -22,14 +22,14 @@
 try:
     from pyavd._cv.workflows.deploy_to_cv import deploy_to_cv
     from pyavd._cv.workflows.models import (
+        AvdDevice,
+        AvdDeviceTag,
```
alexeygorbunov: OK to import from the module where it's already imported (instead of importing from the original module)?

python-avd/pyavd/_cv/workflows/verify_inputs.py
```diff
@@ -29,47 +31,63 @@ def verify_device_inputs(devices: list[CVDevice], warnings: list[Exception], *,
...
 
-    # Group devices based on <CVDevice>.system_mac_address as long as it's not None
+    # Group devices based on <device>.system_mac_address as long as it's not None
     devices_grouped_by_system_mac_address = groupby_obj(
-        list_of_objects=[device for device in devices if device.system_mac_address is not None], attr="system_mac_address", skip_singles=True
+        list_of_objects=[device for device in cv_devices if device.system_mac_address is not None], attr="system_mac_address", skip_singles=True
     )
 
-    # Populate list of CVDevice with duplicated serial_number values
+    # Populate list of AvdDevice with duplicated serial_number values
```
alexeygorbunov: This should be list of CVDevices, not AvdDevices

```suggestion
    # Populate list of CVDevice with duplicated serial_number values
```

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
```diff
@@ -18,113 +18,142 @@
...
+    # Validate all devices present in CV Inventory are populated with serial_number.
+    for device in workflow_devices:
+        if not device.in_cv_inventory:
+            continue
+        if not device.serial_number:
+            LOGGER.info(
+                "verify_devices_in_cloudvision_inventory:: device %s in CV inventory, but no serial number found to be assigned in CV!", device.input.hostname
+            )
+            verify_devices_err_msg = "verify_devices_in_cloudvision_inventory:(): Detected device in CV Inventory, but unable to fetch serial number!"
+            raise CVClientException(verify_devices_err_msg)
```
alexeygorbunov: Is this a realistic scenario (feasible to be tested)? serial_number is a primary key for devices on CV

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
```diff
@@ -39,97 +35,124 @@ async def deploy_tags_to_cv(
...
+            devices_by_serial_number_for_interface_tags[tag.parent_device.serial_number] = tag.parent_device
 
     # If strict, we remove any assignments not specified in the inputs.
     # If not strict, we remove any assignments with the same labels but not specified in the inputs.
+    assignments_to_unassign = set()
     if strict:
-        assignments_to_unassign = {
-            assignment for assignment in existing_assignments if assignment.device_id in devices_by_serial_number and assignment not in desired_assignments
-        }
+        LOGGER.debug("deploy_tags_to_cv: STRICT tags validation enabled for removing unassigned tags")
```
alexeygorbunov: This log message may not be clear as we are not removing tags which are unassigned on CV side but instead we are removing tags which are assigned to devices on CV side  but are not defined in out inputs

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -28,46 +30,62 @@ async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvRes
...
-    LOGGER.info("deploy_configs_to_cv: %s todo configs.", len(todo_configs))
+    # Build todo with WorkflowDevice objects that exist on CloudVision. Add the rest to skipped.
+    todo_workflow_configs: list[WorkflowConfig] = []
+    skipped_devices = 0
+    for device in workflow_devices:
+        if not device.in_cv_inventory:
+            skipped_devices += 1
+            if device.config:
+                device.config.status = DeploymentStatus.SKIPPED
+            LOGGER.info("deploy_configs_to_cv: Skipping config for device %s missing in CloudVision Inventory", device.input.hostname)
```
alexeygorbunov: Should this be inside `if`? Otherwise this may be logged even if device has no `config` attribute set

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
```diff
@@ -28,46 +30,62 @@ async def deploy_configs_to_cv(configs: list[CVEosConfig], result: DeployToCvRes
...
+        # account for workflow where config may not be pushed to CV
```
alexeygorbunov: Does this comment describe following `elif` section?

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -254,23 +259,23 @@ def upsert_edge(metadata: dict, device: CVDevice, studio_inputs: dict, studio_sc
...
             continue
 
-        if get(router, "tags.query") == f"device:{device.serial_number}":
+        if get(router, "tags.query") == f"device:{workflow_device.serial_number}":
             found_index = index
             break
 
     if found_index is None:
-        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: New edge/transit device, adding %s", device.hostname)
+        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: New edge/transit device, adding %s", workflow_device.avd_device.hostname)
```
alexeygorbunov:
```suggestion
        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: New edge/transit device, adding %s", workflow_device.input.hostname)
```

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -254,23 +259,23 @@ def upsert_edge(metadata: dict, device: CVDevice, studio_inputs: dict, studio_sc
...
         studio_inputs_routers.append(edge_metadata)
     else:
-        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Existing edge/transit device, updating %s", device.hostname)
+        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Existing edge/transit device, updating %s", workflow_device.avd_device.hostname)
```
alexeygorbunov:
```suggestion
        LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Existing edge/transit device, updating %s", workflow_device.input.hostname)
```

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
```diff
@@ -340,94 +345,119 @@ async def deploy_cv_pathfinder_metadata_to_cv(cv_pathfinder_metadata: list[CVPat
         vni: 100
     ```
     """
-    LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Got cv_pathfinder_metadata for %s devices", len(cv_pathfinder_metadata))
+    LOGGER.info("deploy_cv_pathfinder_metadata_to_cv: Got %s devices for deploying Pathfinder metadata", len(workflow_devices))
```
alexeygorbunov: This may be confusing as we are counting all devices here, even those that may have nothing to do with pathfinder metadata

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ashenoy-arista: As you indicated, if device is found to be present in CV Inventory, it needs to have a serial_number (device_id) associated. This is primarily a safe-guard check so that rest of deployXXX() routines can be assured that associated serial_number for a device is never None. Hence included this validation here to avoid checking for device serial_number to be non-None within deployXXX().
Let me know if you think is not needed.

python-avd/pyavd/_cv/workflows/input_models.py
ashenoy-arista: Added annotation hint for `metadata`

python-avd/pyavd/_cv/workflows/models.py
ashenoy-arista: Thank you for your suggestion, made changes.

python-avd/pyavd/_cv/workflows/verify_inputs.py
ashenoy-arista: Done

python-avd/pyavd/_cv/workflows/deploy_tags_to_cv.py
ashenoy-arista: Updated log message to just say "deploy_tags_to_cv: STRICT tags validation enabled"
Let me know if that is fine. Just wanted to convey to end-user whether we have strict tags enabled or not based on input provided.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
ashenoy-arista: Good point. 
Also moved `skipped_devices += 1` to be within this if conditional so that we only count those devices for which config is mentioned in input, but device is NOT present in CV inventory, hence we are skipping them.

python-avd/pyavd/_cv/workflows/deploy_configs_to_cv.py
ashenoy-arista: Removed this comment as it not needed.

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ashenoy-arista: Thank you for catching this! Addressed

python-avd/pyavd/_cv/workflows/deploy_cv_pathfinder_metadata_to_cv.py
ashenoy-arista: Revised the logic to print this log message only after we have found a subset of devices for which cv_pathfinder_metadata was provided in input and those devices are present in CV Inventory.

python-avd/pyavd/_cv/workflows/verify_devices_on_cv.py
ashenoy-arista: Removed this validation check for serial number given that if device is in CV Inventory, it should always have serial number associated.

python-avd/tests/pyavd/workflows/test_verify_inputs.py
```diff
@@ -298,10 +336,102 @@
     },
 }
 
+# Variety of deploy input combo for AvdDevice
+AVD_DEVICE_MOCK_DEPLOY_INPUTS = [
```
alexeygorbunov: where is this used?

python-avd/tests/pyavd/workflows/test_verify_inputs.py
```diff
@@ -298,10 +336,102 @@
...
+        config=AvdEosConfig(file="myFilePath5"),
+    ),
+    # AvdDevice input with only pathfinder metadata
+    generate_mock_avd_device_with_deploy_inputs(hostname="switch6", pathfinder_metadata=AvdPathfinderMetadata(metadata={"role": "transit"})),
+]
+
+# Expected members embedded within WorkflowDevice object that would be returned
+# for corresponding AvdDevice input in AVD_DEVICE_MOCK_DEPLOY_INPUTS
+# This is represented as dictionary with hostname from AvdDevice input as key.
+WORKFLOW_DEVICE_EXPECTED_EMBEDDED_RETURN_MEMBERS = {
```
alexeygorbunov: Where is this used?

ansible_collections/arista/avd/extensions/molecule/cv_deploy/duplicated_devices_strict_system_mac.yml
```diff
@@ -200,6 +200,36 @@
           tags: ["{{ test_id }}"]
           ansible.builtin.meta: clear_host_errors
 
+        # Initialize `result_device_configs` var to empty list before appending values to it.
+        # Avoid any effect of value set to this fact by tasks in earlier playbook (duplicated_devices.yml)
+        - name: Initialize result_device_configs fact to empty list
```
alexeygorbunov, line 205:
```suggestion
        - name: "(RESCUE) {{ test_id | upper }} Initialize result_device_configs fact to empty list"
```

python-avd/tests/pyavd/workflows/test_verify_inputs.py
ashenoy-arista: We removed helper function used earlier to build WorkflowDevice instance.
This code is not being used any more. Removing the same

python-avd/tests/pyavd/workflows/test_verify_inputs.py
ashenoy-arista: We removed helper function used earlier to build WorkflowDevice instance.
This code is not being used any more. Removing the same

carl-baillargeon:
Moving to draft until we finalize other cv_deploy features.

carl-baillargeon:
We started implementing this in smaller PRs.

PR #5689, opened 2025-08-04
Refactor(eos_designs): Deprecate support for Jinja templating using `switch.*` facts

### LINKED ISSUE #5231: eos_designs: Emit deprecation warnings for templating with switch.*.
Deprecate templating with switch.*.

Could be done by wrapping the task_vars["switch"] in a custom dict class which emits the deprecation warning. Also handle it in ip / description templating.

_Originally posted by @ClausHolbechArista in https://github.com/aristanetworks/avd/pull/5185#discussion_r2018214279_

ClausHolbechArista:
## Change Summary

Deprecate support for Jinja templating using `switch.*` facts

## Related Issue(s)

Fixes #5231

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

We no longer wish to expose the switch.* facts to end-users.

These are not stable APIs, which has been made clear in documentation for a long time.

Unfortunately we still see a lot of use of these, primarily because we had limited support for management protocols until recently.

For AVD 6.0.0 we will no longer support jinja templates referring to `switch.*`.
Examples:
- `{{ switch.id }}`
- `{{ switch.mgmt_ip }}`
- `{{ switch.mgmt_interface }}`

Users that will be hit by this, have to work out an alternative way of setting what they need. They should open a GitHub issue, so we can work with them to identify alternatives. 
Users with a support agreement can contact Arista support to get assistance.

The templating based on `switch.*` is also unsupportable based on Ansible's changes introduced in version 2.19.

## How to test

- Existing `ansible_only` molecule scenario will emit a deprecation warning when any template contains a reference to `switch.*`.
- Tested removing the `switch.*` use, and the deprecation warning went away.

### Repository Checklist


ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -71,11 +71,28 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
...
+            except Exception as e:
+                msg = "Exception during templating of task_var 'switch'"
+                raise AnsibleActionFail(msg) from e
+
+        task_vars["switch"] = DeprecatedDict(
+            task_vars["switch"],
+            _display=self._display,
+            _message=(
+                "Inline jinja templating based on 'switch.*' facts generated by 'eos_designs' is deprecated. "
+                "Open a GitHub discussion so we can find an alternative."
```
ankudinov: I'd prefer to remove following statement from the message: "Open a GitHub discussion so we can find an alternative."
If customers will be concerned - they will open issue or call support anyway. The message creates false impression of us foreseeing that change will trigger problems and not planning anything in advance.

ClausHolbechArista: done

ankudinov (approved):
LGTM

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -71,11 +71,25 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
         all_facts = AvdSwitchFactsDefaultDict(avd_switch_facts)
 
         # TODO: AVD 6.0.0 remove 'switch'
+        # Fist check if switch itself contains a template - before adding the deprecation notice.
```
```suggestion
        # First check if switch itself contains a template - before adding the deprecation notice.
```

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -71,11 +71,25 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
...
+        task_vars["switch"] = DeprecatedDict(
+            task_vars["switch"],
+            _display=self._display,
+            _message="Inline jinja templating based on 'switch.*' facts generated by 'eos_designs' is deprecated.",
+        )
 
         # Read ansible variables and perform templating to support inline jinja2
         for var in task_vars:
-            if str(var).startswith(("ansible", "molecule", "hostvars", "vars", "avd_switch_facts")):
+            if str(var).startswith(("ansible", "molecule", "hostvars", "vars", "avd_switch_facts", "switch")):
```
gmuloc, line 92: by doing this are we not removing support instead of just deprecating?

ClausHolbechArista: no this is doing templating of _other_ variables pointing to `switch.*`. The check here is just excempting certain variables from the templating, and since we moved `switch.*` above we don't need it here.

gmuloc (approved):

PR #5698, opened 2025-08-06
Refactor: Include ansible tests in Ruff

ClausHolbechArista:
Including ansible tests in Ruff and fixing issues.


ansible_collections/arista/avd/tests/unit/action/test_verify_requirements.py
```diff
@@ -193,11 +197,11 @@ def test__validate_python_requirements_pyavd(extras: bool, running_from_source:
         #     "2.12.6",
         #     True,
         #     True,
-        #     id="deprecated ansible version",
+        #     id="deprecated ansible version",  # noqa: ERA001
```
Vibhu-gslab: These are commented, do we need these ?

ClausHolbechArista: I was not sure why it was commented out instead of removed, so I decided to just ignore. @gmuloc should be back Tomorrow so he can take a look and decide.

ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
```diff
@@ -3,13 +3,13 @@
...
 
-import os
+from pathlib import Path
 
 import pytest
 
 from ansible_collections.arista.avd.plugins.modules.configlet_build_config import get_configlet
 
-CONFIGLETS_DIR = f"{os.path.dirname(os.path.realpath(__file__))}../../inventory/intended/configs"
+CONFIGLETS_DIR = str(Path(__file__).parents[3] / "molecule/eos_designs_unit_tests/intended/configs")
```
gmuloc: why are we targeting different files? are the files not useful anymore?

ansible_collections/arista/avd/tests/unit/action/test_verify_requirements.py
gmuloc: this is commented test because it is testing some commented code in the module. We can remove it and add it back.

ansible_collections/arista/avd/tests/unit/modules/test_configlet_build_config.py
ClausHolbechArista: mistake. I will move this to draft. Thank you for catching.

ansible_collections/arista/avd/tests/unit/modules/test_inventory_to_container.py
```diff
@@ -131,80 +134,80 @@ def test_is_leaf_none_leaf(self) -> None:
...
+    def test_get_devices_preexisting_devices_with_device_filter(self, inventory: dict) -> None:
         output = get_devices(inventory, search_container=SEARCH_CONTAINER, devices=[], device_filter=[GET_DEVICE_FILTER])
         assert [GET_DEVICE_FILTER in item for item in output]
 
-    @pytest.mark.parametrize("DATA", [None])
-    def test_serialize_empty_inventory(self, DATA) -> None:
-        output = serialize(DATA)
+    @pytest.mark.parametrize("data", [None])
+    def test_serialize_empty_inventory(self, data: dict | None) -> None:
+        output = serialize(data)  # pyright: ignore[reportArgumentType]
```
gmuloc, line 180: ugh - I get this but are we mistyping the serialize input as it seems to accept None and you are getting a pyright issue?
Also I would vote for removing the parametrize givent the test

gmuloc: probably same wrong type hints for other places where you have added the pyright ignore?

gmuloc (commented):
LGTM overall but one question

ansible_collections/arista/avd/tests/unit/modules/test_inventory_to_container.py
ClausHolbechArista: The type hint in the function signature matches the parametrize inputs, but the test target may not support this. What we are testing is invalid inputs, so it makes sense that we need to have the ignore.

gmuloc: actually the type hint in `serialize` does _not_ match the input as it states it must be a `dict` but here the parametrized input is `dict|None`, and this test is only testing `None`

we are testing that if we give `None`, then the serialize function gives back `None`.

The return type hint of `serialize` is currently `Tree` and not `Tree | None` so a lot of things, from a type hint point of view are not really great here. The best thing would actually be to raise when we don't receive an iterable.

gmuloc: that being said. we can live with it and not waste more time

gmuloc (approved):

Vibhu-gslab (approved):

ansible_collections/arista/avd/tests/unit/plugins/plugin_utils/eos_validate_state_utils/test_catalog.py
```diff
@@ -2,11 +2,7 @@
 # Use of this source code is governed by the Apache License 2.0
 # that can be found in the LICENSE file.
 
-
-# import pytest
-
-
-# TODO
+# TODO: implement something here
```
ClausHolbechArista:
```suggestion
# TODO: Add tests for Catalog
```

PR #5705, opened 2025-08-07
Doc: Add toggle button to Input Variables [WIP]

### LINKED ISSUE #5653: Doc: Add a toggle to switch between in input variabled to switch between the table and YAML view
Add a toggle to switch between in input variable to switch between the table and YAML

_Originally posted by @carlbuchmann in https://github.com/aristanetworks/avd/issues/5507#issuecomment-3089610610_

JulioPDX:
## Change Summary

- Add a dedicated toggle button to switch between Table and YAML content tabs
- Tries to stay within the same header and view the user is currently working on
- Removes `header.autohide` feature as the header is required to have the toggle always present
- Current implementation requires the entire `header.html` to be overriden from the material mkdocs template
- 

## Related Issue(s)

Fixes #5653 

## Component(s) name

Docs

## Proposed changes

Custom JS has been added to handle the scroll for the user between switching from YAML and Table views. Written to try and maintain user work section/view. Previous iterations would place user in random area of the page depending on the anchor that was selected. CSS is there mainly for styling and is not final. The use of the entire material mkdocs header template was required due to issues/race conditions when objects are loaded to the page. This leads to different elements being missed in the final site render.

## How to test

Serve locally or view generated docs

### User Checklist

- [ ] Finalize styling (colors)
- [ ] Test to make sure it does not interfere with version selector

### Repository Checklist


ClausHolbechArista:
It takes seconds to swap when I click the button. We may have to have some spinning wheel or grey it out while it works.
Also when I click the toggle while the page is loading it flips the toggle button but does not change the layout of the page. To get it to change, I then need to toggle it back and forth again.

JulioPDX:
> It takes seconds to swap when I click the button. We may have to have some spinning wheel or grey it out while it works. Also when I click the toggle while the page is loading it flips the toggle button but does not change the layout of the page. To get it to change, I then need to toggle it back and forth again.

Yes it doesnt track if you switch between using the content tab links and the toggle. I'll have to think about that one but I agree, need a loading wheel or spinner. Thank you

JulioPDX:
@ClausHolbechArista This is working but it tries to do its best to keep the user in the same area of the screen, not perfect though. Should we keep going down this route or abandon?

ClausHolbechArista:
I think we should abandon and focus the efforts on a data model browser component

PR #5725, opened 2025-08-14
Feat(eos_designs): Add support for regex pattern for platform_settings[].platforms

### LINKED ISSUE #4815: Feat(eos_designs): Match using regex on platforms in platform_settings and default_interfaces
### Enhancement summary

as a follow up to #4798  implementation - we should move to using regex fullmatch in the `platform_settings`, `custom_platform_settings` and `default_interfaces` to unify the behavior across `eos_designs` and gives more powerful expression to these areas.

This should be backward compatible to introduce so pretty straightforward regex

eos_designs

### Use case example

easier to express.

cf summary

_No response_

### Additional context

_No response_

### Contributing Guide

laxmikantchintakindi:
## Change Summary

Add support for regex pattern for platform_settings[].platforms.

## Related Issue(s)

Fixes #4815

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add support for regex pattern for platform_settings[].platforms.
### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -44,10 +43,10 @@ def get_platform_settings(self: SharedUtilsProtocol, platform: str | None) -> Eo
         # First look for a matching platform setting specifying our platform
         if platform is not None:
             for platform_setting in self.inputs.custom_platform_settings:
-                if platform in platform_setting.platforms:
+                if self.match_regexes(platform_setting.platforms, platform):
                     return platform_setting
             for platform_setting in self.inputs.platform_settings:
-                if platform in platform_setting.platforms:
+                if self.match_regexes(platform_setting.platforms, platform):
```
Vibhu-gslab, line 49: Update the description in schema for all the changes

laxmikantchintakindi: Updated.

Vibhu-gslab (approved):
LGTM!

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -63,21 +62,18 @@ def get_platform_settings(self: SharedUtilsProtocol, platform: str | None) -> Eo
     @cached_property
     def default_interfaces(self: SharedUtilsProtocol) -> EosDesigns.DefaultInterfacesItem:
         """default_interfaces set based on default_interfaces."""
-        if self.digital_twin and self.inputs.digital_twin.use_default_interfaces_of_digital_twin_platform:
-            device_platform = self.platform or "default"
-        else:
+        device_platform = self.platform or "default"
```
MaheshGSLAB: why we are changing the default behaviour . Like earlier the default is `device_platform = self.original_platform or "default"` but not its `device_platform = self.platform or "default"`

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/hosts.yml
```diff
@@ -707,6 +706,10 @@ all:
           hosts:
             7010TX-LEAF1:
             7010TX-LEAF2:
+        PLATFORM_SETTINGS:
```
MaheshGSLAB: There is platform setting tests group vars above. So either move your host inside that with like regex_pattern_platform_setttings_1/2 or update the current group var and host name with respect to test

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
gmuloc: I was wondering about this change as well.

It seems this was done to save a line but given that the `if FOO` statement was changed to `if not(FOO)`, I ended up concluding (maybe wrongly) that this was not changing the default behavior.

Though I think to avoid confusion we may want to revert this as this seems to be bringing more questions from reviews :)

laxmikantchintakindi: Changed the code.

ansible_collections/arista/avd/molecule/eos_designs_unit_tests/inventory/hosts.yml
laxmikantchintakindi: Done.

MaheshGSLAB (approved):

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -70,14 +69,12 @@ def default_interfaces(self: SharedUtilsProtocol) -> EosDesigns.DefaultInterface
 
         # First look for a matching default interface set that matches our platform and type
         for default_interface in self.inputs.default_interfaces:
-            for platform in default_interface.platforms:
-                if search(f"^{platform}$", device_platform) and self.type in default_interface.types:
-                    return default_interface
+            if self.match_regexes(default_interface.platforms, device_platform) and self.type in default_interface.types:
```
alexeygorbunov, line 72: Should we not use `self.match_regexes` for matching node's `type` here as well for consistency?

gmuloc: we dont' support regexes for types today so not really. Maybe later but thats not what this PR is about so lets not add anything here

gmuloc (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -14,7 +14,7 @@ keys:
     description:
       Custom Platform settings to override the default `platform_settings`.
       This list will be prepended to the list of `platform_settings`.
-      The first entry containing `platforms` matching the `platform` node setting will be chosen.
+      The first entry containing `platforms` matching the full regex of `platform` node setting will be chosen.
```
gmuloc: Hmm I think this needs to be clarified actually.

Technically this is not the `platforms` list mathcing the regex.

It is the countrary.


```suggestion
      The first entry found where the `platform` node setting is fully matched by any regex in the `platforms` list will be chosen.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -23,7 +23,7 @@ keys:
       table: platform-settings
     description:
       Platform settings.
-      The first entry containing `platforms` matching the `platform` node setting will be chosen.
+      The first entry containing `platforms` matching the full regex of `platform` node setting will be chosen.
```
gmuloc:
```suggestion
      The first entry found where the `platform` node setting is fully matched by any regex in the `platforms` list will be chosen.
```

gmuloc (changes requested):

gmuloc (approved):
once CI completes and does its magic of recompiling the schema

PR #5727, opened 2025-08-15
Feat(eos_designs): Preview: Digital Twin for ACT. default_interfaces of original platform.

### LINKED ISSUE #5697: Digital twin: enhance support for default_interfaces
### Enhancement summary

default interfaces need to be enhanced to include the digital twin's platform settings.  

eos_designs

### Use case example

a repo using default_interfaces will build successfully, but deployment to digital twin fails with:

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: pyavd._errors.AristaAvdError: 'uplink_switch_interfaces' is not set on 'DC1-CL1' and 'uplink_switch' 'DC1-SP1' does not have 'downlink_interfaces[0]' set under 'default_interfaces'
fatal: [DC1-SP1]: FAILED! => {"changed": false, "msg": "'uplink_switch_interfaces' is not set on 'DC1-CL1' and 'uplink_switch' 'DC1-SP1' does not have 'downlink_interfaces[0]' set under 'default_interfaces'"}
```

An engineer would configure for their production environment:
```
default_interfaces:
  - types: [l3spine]
    platforms: [7280R]
    downlink_interfaces: [Ethernet2-32]
    mlag_interfaces: [Ethernet1]
  - types: [l2leaf]
    platforms: [7050X3]
    uplink_interfaces: [Ethernet1, Ethernet2]
    downlink_interfaces: [Ethernet4-48]
    mlag_interfaces: [Ethernet3]
```

and wouldn't know to add platforms: [7280R, vEOS]

digital twin logic should add the appropriate digital twin platform type, so that valid configs are generated.

Manually add the digital twins platform info to the repo, manually mangle the yaml to replace the platform keys before executing the playbook.

### Additional context

_No response_

### Contributing Guide

alexeygorbunov:
## Change Summary

Introduce `digital_twin.default_interfaces_of_original_platform` bool key to force AVD to use default_interfaces of the original platform
 
## Related Issue(s)

Fixes #5697

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
- Introduce new key `digital_twin.default_interfaces_of_original_platform` (`False` by default)
- When above key set to `True` - fetch `default_interfaces` using original platform instead of alternate (`platform_settings.[].digital_twin.platform`)

## How to test
Molecule scenario `eos_designs-twodc-5stage-clos` (host `DC1-POD1-L2LEAF1A`)

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/eos_designs-twodc-5stage-clos/verify.yml
```diff
@@ -24,3 +24,24 @@
       delegate_to: 127.0.0.1
       ansible.builtin.import_role:
         name: arista.avd.eos_cli_config_gen
+
+- name: Converge
```
ClausHolbechArista: Please don't modify the playbooks here for tests. It is better to have two hosts under eos_designs_unit_tests testing each behavior specifically, and with minimal other config.

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
```diff
@@ -48,6 +48,12 @@ keys:
               Specifies if the ACT Digital Twin device is deployed with direct access to the Internet.
               This option applies only to the 'cloudeos' and 'veos' node types and will be ignored for all other ACT node types.
               ACT does not provide direct Internet access to cloudeos or veos devices by default.
+      default_interfaces_of_original_platform:
+        type: bool
+        default: false
```
ClausHolbechArista, line 53: Let's do it by default as it seems like the more useful case.
```suggestion
        default: true
```

ClausHolbechArista: Or maybe reverse the behavior and key name so the default is false `use_default_interfaces_of_digital_twin_platform: <bool; default=false>`

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
```diff
@@ -48,6 +48,12 @@ keys:
...
+        description: |-
+           In Digital Twin mode, AVD by default uses the `default_interfaces` settings associated with the alternate platform (identified via `platform_settings.[].digital_twin.platform`).
+           Set key `digital_twin.default_interfaces_of_original_platform` to `True` to use `default_interfaces` settings of the original platform.
```
ClausHolbechArista: Let's try to make the key name as intuitive as possible so the description can be shorter and the brain load lower :)
```suggestion
           In Digital Twin mode, AVD can either use the default interfaces of the original or the digital twin platform (as set in `platform_settings.[].digital_twin.platform`).
```

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -26,8 +26,18 @@ class PlatformMixin(Protocol):
     def platform(self: SharedUtilsProtocol) -> str | None:
         if self.digital_twin:
             return self.digital_twin_platform
+        return self._original_platform
+
+    @cached_property
+    def _original_platform(self: SharedUtilsProtocol) -> str | None:
```
ClausHolbechArista:
```suggestion
    def original_platform(self: SharedUtilsProtocol) -> str | None:
```

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -26,8 +26,18 @@ class PlatformMixin(Protocol):
...
         return default(self.node_config.platform, self.cv_topology_platform)
 
+    @cached_property
+    def _default_interfaces_platform(self: SharedUtilsProtocol) -> str:
+        if self.digital_twin and self.inputs.digital_twin.default_interfaces_of_original_platform:
+            return self._original_platform or "default"
+        return self.platform or "default"
+
     @cached_property
     def digital_twin_platform(self: SharedUtilsProtocol) -> str | None:
```
ClausHolbechArista, line 36: I think this should use self.original_platform as well.

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
```diff
@@ -26,8 +26,18 @@ class PlatformMixin(Protocol):
...
         if self.digital_twin:
             return self.digital_twin_platform
+        return self._original_platform
+
+    @cached_property
+    def _original_platform(self: SharedUtilsProtocol) -> str | None:
         return default(self.node_config.platform, self.cv_topology_platform)
 
+    @cached_property
+    def _default_interfaces_platform(self: SharedUtilsProtocol) -> str:
```
ClausHolbechArista: I don't think we need another property for this. Just put it inside default interfaces, since it is only used there. cached_properties are not free :)

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
alexeygorbunov: Changed according to the later recommendation

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
alexeygorbunov: Changed to suggested description

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
alexeygorbunov: True, fixed!

python-avd/pyavd/_eos_designs/shared_utils/platform_mixin.py
alexeygorbunov: Done, moved it inside `def default_interfaces`

ansible_collections/arista/avd/molecule/eos_designs-twodc-5stage-clos/verify.yml
alexeygorbunov: Moved this to the dedicated `digital_twin` scenario as mixing hosts that require digital twin with normal eos_design_unit_tests hosts creates complexity with pytest

ansible_collections/arista/avd/molecule/digital_twin/inventory/hosts.yml
```diff
@@ -0,0 +1,14 @@
+---
+all:
+  children:
+    DIGITAL_TWIN:
+      children:
+        DIGITAL_TWIN_DEFAULT_INTERFACES:
+          hosts:
+            DIGITAL_TWIN_DEFAULT_INTERFACES_ORIGINAL_PLATFORM:
```
ClausHolbechArista: please use lower case and dash for hostnames and uppercase and underscore for groups.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
carl-baillargeon: Reopening this. I think `use_default_interfaces_of_digital_twin_platform` is a bit misleading. It implies we're choosing a different data model, when the code is actually just overriding the platform lookup key for the _same_ `default_interfaces` model.

I think a more accurate name would be `use_physical_platform_for_default_interfaces`. This also makes more sense with the default set to `true`, which enables the magic. Setting it to `false` would then enforce the strict platform validation, which feels more intuive.

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
carl-baillargeon: We might need to update this if we change back the key per my comment above.

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
alexeygorbunov: Not sure. Term "physical" may as well be confusing as we are not picking here between physical and virtual but instead between original platform (which can be already virtual) and "digital twin" platform

ansible_collections/arista/avd/molecule/digital_twin/inventory/hosts.yml
alexeygorbunov: Fixed

python-avd/pyavd/_eos_designs/schema/schema_fragments/digital_twin.schema.yml
ClausHolbechArista: Since it is harder to identify what we talk about when we refer to the "real" platform, it is simpler to refer to the "digital_twin platform", since that is pretty clear.

carl-baillargeon (approved):
LGTM. Tested locally. New molecule scenario works fine as well.

ClausHolbechArista (approved):

PR #5743, opened 2025-08-19
Refactor(eos_designs,eos_cli_config_gen): Move is_deployed to metadata.is_deployed

MaheshGSLAB:
## Change Summary

Move is_deployed to metadata.is_deployed

## Related Issue(s)

Fixes https://github.com/aristanetworks/avd-internal/issues/296

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
  - [ ] eos_cli_config_gen: Move `is_deployed` to `metadata.is_deployed`.
    - Don't do deprecation warning here, since the raw inputs of `is_deployed` would trigger an error.
    - Update eos_designs to output the new key.
    - Update cv_deploy and anta_runner to read the new key falling back to the old.
    - Update eos_config_deploy_cvp to read the new key falling back to the old.

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/molecule/anta_runner/anta/reports/allow_bgp_vrfs_run/anta_report.csv
```diff
@@ -1,739 +1,79 @@
...
-dc1-leaf1a,VerifyLoggingErrors,unset,,Verifies there are no syslog messages with a severity of ERRORS or higher.,Logging
-dc1-leaf1a,VerifyMlagConfigSanity,unset,,Verifies there are no MLAG config-sanity inconsistencies.,MLAG
-dc1-leaf1a,VerifyMlagInterfaces,unset,,Verifies there are no inactive or active-partial MLAG ports.,MLAG
-dc1-leaf1a,VerifyMlagStatus,unset,,Verifies the health status of the MLAG configuration.,MLAG
-dc1-leaf1a,VerifyRoutingProtocolModel,unset,,Verifies the configured routing protocol model.,Routing
-dc1-leaf1a,VerifyAPIHttpsSSL,unset,,Verifies if the eAPI has a valid SSL profile.,Security
 dc1-leaf1a,VerifyEOSExtensions,unset,,Verifies that all EOS extensions installed on the device are enabled for boot persistence.,Software
 dc1-leaf1a,VerifyEOSVersion,unset,,Verifies the EOS version of the device.,Software
 dc1-leaf1a,VerifyTerminAttrVersion,unset,,Verifies the TerminAttr version of the device.,Software
-dc1-leaf1a,VerifySTPCounters,unset,,Verifies there is no errors in STP BPDU packets.,STP
```
gmuloc: I know this is still draft but just commenting to make sure you caught this - it looks like some work is need in there to make sure we don't break this

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/structured_configs/host1.yml
```diff
@@ -117,6 +116,7 @@ management_interfaces:
   ip_address: 192.168.0.103/24
   type: oob
 metadata:
+  is_deployed: true
```
gmuloc, line 133: should you not keep the `is_deployed` in the deprecated place in one example?

ansible_collections/arista/avd/molecule/anta_runner/anta/reports/allow_bgp_vrfs_run/anta_report.csv
MaheshGSLAB: yeah i fix this in https://github.com/aristanetworks/avd/pull/5743/commits/58ba6218ad542e21a46d30e8ce7cf9d35fd2cb53

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/structured_configs/host1.yml
MaheshGSLAB: so we have metadata. is_deployed default as true so it will always pick either False or True so not sure how to test the deprecated  is_deployed

ClausHolbechArista: Don't have a default value in the schema for the metadata value.

MaheshGSLAB: Even after removing the default value from metadata, the global key is_deployed still defaults to true. As a result, in the structured config, metadata.is_deployed will always be either True or False.

python-avd/pyavd/api/_anta/minimal_structured_config.py
```diff
@@ -56,7 +56,7 @@ def get_minimal_structured_configs(structured_configs: dict[str, dict]) -> dict[
         # Create the minimal structured configuration
         minimal_structured_configs[device] = MinimalStructuredConfig(
             hostname=structured_config["hostname"],
-            is_deployed=get(structured_config, "is_deployed", default=False),
+            is_deployed=default(get(structured_config, "metadata.is_deployed"), get(structured_config, "is_deployed", default=False)),
```
Shivani-gslab, line 62: since we are not generating `is_deployed` from eos_designs in structured_config now, the second part would never have a value true `get(structured_config, "is_deployed", default=False)`

gmuloc: some users may be generating the structured config by other means than eos_designs and then use anta_runner and they could still have top level `is_deployed`

ClausHolbechArista: you can test that with `custom_structured_configuration_is_deployed: true`

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices-v3.j2
```diff
@@ -15,7 +15,8 @@
 cvp_devices:
 {% for device in groups[container_root] | arista.avd.natural_sort %}
 {%     if device | arista.avd.is_in_filter(local_var.device_filter) %}
-{%         if hostvars[device].is_deployed is arista.avd.defined(true) or hostvars[device].is_deployed is not arista.avd.defined %}
+{%         set is_deployed = hostvars[device].metadata.is_deployed | arista.avd.default(hostvars[device].is_deployed) | arista.avd.default(true) %}
```
Shivani-gslab:
```suggestion
{%         set is_deployed = hostvars[device].metadata.is_deployed | arista.avd.default(hostvars[device].is_deployed, true) %}
```

ansible_collections/arista/avd/roles/eos_config_deploy_cvp/templates/cvp-devices.j2
```diff
@@ -14,7 +14,8 @@
 cvp_devices:
 {% for device in groups[container_root] | arista.avd.natural_sort %}
 {%     if device | arista.avd.is_in_filter(local_var.device_filter) %}
-{%         if (hostvars[device].is_deployed is defined and hostvars[device].is_deployed == true) or (hostvars[device].is_deployed is not defined) %}
+{%         set is_deployed = hostvars[device].metadata.is_deployed | arista.avd.default(hostvars[device].is_deployed) | arista.avd.default(true) %}
```
Shivani-gslab:
```suggestion
{%         set is_deployed = hostvars[device].metadata.is_deployed | arista.avd.default(hostvars[device].is_deployed, true) %}
```

Shivani-gslab (changes requested):

Shivani-gslab (approved):
LGTM

carl-baillargeon (approved):
LGTM. 

Tested locally with DC1-SPINE2 is_deployed true or false:

cv_deploy:
image

anta_runner:
image

eos_validate_state:
image



**NOTE:**
We should give an update to https://avd.arista.com/devel/ansible_collections/arista/avd/roles/eos_designs/docs/input-variables.html#flagging-a-device-as-not-deployed in a separate PR.

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/structured_configs/host1.yml
ClausHolbechArista: Mahesh please beware if you are looking at eos_designs or eos_cli_config_gen.
In eos_cli_config_gen none of them should have a default value.
In eos_designs the global `is_deployed` has a default value of true, so eos_designs will always **set** `metadata.is_deployed` as you mention.

The test Guillaume is asking for is that we at least one place in `eos_cli_config_gen_deprecated_vars` test the global `is_deployed: false`. If it's already there all good.

gmuloc: indeed but my comment was on eos_designs so not great.

@MaheshGSLAB we can move the deprecation test to eos_cli_config_gen_deprecated_vars and make sure it works fine.

The test you have written - while working is not necessary :)

gmuloc: ok so I am going in circle - you cannot do eos_cli_config_gen_deprecated_vars because this is not deprecated in schema or we would get deprecation warning because of eos_designs input which we don't want.
What you have done is ok.

gmuloc: so what we really want, actually, is that `cv_deploy` and `anta_runner`, `documentation` still work properly with a structured_config with `is_deployed` at the top level (which is deprecated) and that we don't break that.

This is the real test we would want until people can migrate their structured_config (which they will have to do for 6.0).
The test in eos_designs_deprecated_vars today does not bring any value for now - we can check that the same tests are generated in anta_runner with the "old structured_config)

gmuloc: Adjusted some tests - happy to merge this given the testing BUT:

1/ we are never testing is_deployed in cv_workflow / cv_deploy molecule
2/ the configs in these scenario have no way to be generated so fixed the metadata with the new look.

PR #5781, opened 2025-08-28
Feat(eos_designs): Extend LACP fallback individual options

### LINKED ISSUE #5774: Feat(eos_designs): Allow vlan and native vlan to be configured under lacp_fallback, within the connected_endpoints model.
### Enhancement summary

The current `port_channel` model allows a profile to be defined, when `lacp_fallback` is set to individual. The settings in the profile are then applied to the individual member interfaces e.g. EthernetX so that, for example, PXE boot can be supported.

The feature request is to allow for both `vlans` and `native_vlan` to be configured within `lacp_fallback`, rather than relying on them being configured within the port profile.

eos_designs

### Use case example

We have a small number of `port_profiles` e.g. trunk, access, port-channel trunk, and port-channel access where all the parameters are common. The main difference is the VLANs (and native VLANs) assigned to the adapter.

We can assign the general profile to the adapter, in combination with the vlan e.g.
```yaml
connected_endpoints:
  - name: Host1
    adapters:
      - switches: [switch1a, switch1b]
        switch_ports: [Ethernet1, Ethernet1]
        profile: PP-PO-TRUNK
        vlans: 10,20,30
        native_vlan: 5
```
But currently the only option under `lacp_fallback` when in `individual` mode is a `profile`. Which would cause the # of profiles to explode.
```yaml     
 # LACP fallback configuration.
      lacp_fallback:

        # Either static or individual mode is supported.
        # If the mode is set to "individual" the "individual.profile" setting must be defined.
        mode: <str; "static" | "individual">

        # Define parameters for port-channel member interfaces. Applies only if LACP fallback is set to "individual".
        individual:

          # Port-profile name to inherit configuration.
          profile: <str>

        # Timeout in seconds.
        timeout: <int; default=90>
```

Extend the data model to allow both `vlans` and `native_vlan` to be configured under `lacp_fallback: individual:` e.g.
```yaml     
 # LACP fallback configuration.
      lacp_fallback:
        mode:  "individual"
        individual:
          profile: <str>
          vlans: <str>
          native_vlan: <int>
```
And apply that to the individual member interfaces

_No response_

### Additional context

_No response_

### Contributing Guide

colinmacgiolla:
## Change Summary

This PR implements the changes mentioned in #5774 

## Related Issue(s)

Fixes #5774

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
The changes allow for new keys to be provisioned within the `lacp_fallback` key. I've set it up so that any of the new keys take precedence over what is inherited from the profile. But if the profile is specifying a trunk with a range of VLANs, and the fallback is setting the mode to access with VLAN 5 (see tests) then the switchport trunk... config lines will render, just be ineffective.

```yaml
 # LACP fallback configuration.
      lacp_fallback:
        mode:  "individual"
        individual:
          profile: <str>
          vlans: <str>
          native_vlan: <int>
          mode: <str>
```
### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
```diff
@@ -221,6 +221,24 @@ def _get_ethernet_interface_cfg(
...
+                                "Adapter 'vlans' value must be a single vlan ID when mode is 'access' or 'dot1q-tunnel'. "
+                                f"Got {adapter.port_channel.lacp_fallback.individual.vlans} for interface {ethernet_interface.name}."
+                            )
+                            raise AristaAvdInvalidInputsError(msg) from e
+                    elif ethernet_interface.switchport.mode in ["trunk", "trunk phone"]:
+                        if adapter.port_channel.lacp_fallback.individual.vlans is not None:
+                            ethernet_interface.switchport.trunk.allowed_vlan = adapter.port_channel.lacp_fallback.individual.vlans
+                        if adapter.port_channel.lacp_fallback.individual.native_vlan is not None:
+                            ethernet_interface.switchport.trunk.native_vlan = adapter.port_channel.lacp_fallback.individual.native_vlan
+
```
ClausHolbechArista: I think you could avoid duplicating the code (more maintenance and may be missed == bugs), by using `individual._cast_as(EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem.ConnectedEndpointsItem.AdaptersItem)` and then inheriting the profile on that similar to what we do when inhertiting profiles for the connected endpoints. That would also allow us to add more keys (still not all) to the individual mode by just adding them to the schema. Ping me if I missed something. Thanks.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_adapter_config.schema.yml
```diff
@@ -325,6 +325,22 @@ $defs:
                   profile:
                     type: str
                     description: Port-profile name to inherit configuration.
+                  vlans:
```
laxmikantchintakindi, line 328: Line 320 should be corrected according to new logic.

ClausHolbechArista: done. Thank you!

python-avd/pyavd/_eos_designs/eos_designs_facts/vlans.py
```diff
@@ -102,6 +101,21 @@ def _local_endpoint_vlans_and_trunk_groups(self: EosDesignsFactsGeneratorProtoco
...
                     vlans.update(adapter_vlans)
                     trunk_groups.update(adapter_trunk_groups)
+                    if (
+                        adapter_settings.port_channel.mode
+                        and adapter_settings.port_channel.lacp_fallback.mode == "individual"
+                        and adapter_settings.port_channel.lacp_fallback.individual
+                    ):
+                        individual_adapter = adapter_settings.port_channel.lacp_fallback.individual._cast_as(
+                            EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem.ConnectedEndpointsItem.AdaptersItem
+                        )
```
laxmikantchintakindi: Would it be possible to move this to _parse_adapter_settings?

ClausHolbechArista: I do follow your point, but since this one also calls _parse_adapter_settings it would be a bit of a mess. We could create another helper function doing this part for more reuse, but I decided to just copy it. Let me think about it while we wait for another review. Thanks.

ClausHolbechArista: @laxmikantchintakindi added a function for this. Made it all much simpler so thank you!

laxmikantchintakindi (approved):
LGTM.

ClausHolbechArista (approved):

PR #5782, opened 2025-08-29
Feat(eos_designs): Added support for IGMP snooping querier source interface as vrf_router_id

### LINKED ISSUE #4351: Enable the querier's source addr with the IP of "overlay loopback" associated to the VRF instance
### Enhancement summary

we need to enable the querier's source addr with the IP of "overlay loopback" associated to the VRF instance. Per TAC case analysis, it's important to have the VRF reachability for the querier's source addr. 

Regardless the defined variable in Ansible, the configuration generated is with the string of the variable instead of the value.

## "NETWORK_SERVICES.yml"
# Enable IGMP Snooping (Enabled by default on EOS).
          igmp_snooping_enabled: True
          igmp_snooping_querier:
            enabled: True
**            source_address: "{{lo1012_addrv4}}" **

GSC-MDF-AV03 | SUCCESS => {
**    "lo1012_addrv4": "10.30.92.6" **
}

The eos_cli_generate role only generates the variable string, instead of the value referred by the variable.

** ip igmp snooping vlan 126 querier address {{lo1012_addrv4}} **

## Structure of the 
    # When enabled, IGMP snooping querier will only be configured on L3 devices, i.e., uplink_type: p2p.
    igmp_snooping_querier:
      # Will be enabled automatically if "evpn_l2_multicast" is enabled.
      enabled: <bool>
      # Default IP address of Loopback0.
      source_address: <str>

eos_cli_config_gen

### Use case example

Need to generate the VRF loopback IP as the querier's source, instead of the default global (underlay) loopback0.
** ip igmp snooping vlan 126 querier address 10.30.92.6 **

Need to generate the VRF loopback IP as the querier's source, instead of the default global (underlay) loopback0.
** ip igmp snooping vlan 126 querier address 10.30.92.6 **

_No response_

### Additional context

_No response_

### Contributing Guide

MaheshGSLAB:
## Change Summary

Added support for IGMP snooping querier source interface as vrf_router_id

## Related Issue(s)

Fixes #4351 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
```
# When enabled, IGMP snooping querier will only be configured on L3 devices, i.e., uplink_type: p2p.
igmp_snooping_querier:
  # Will be enabled automatically if "evpn_l2_multicast" is enabled.
  enabled: <bool>
  # Default IP address of Loopback0.
  source_address: <str; "main_router_id" | "vrf_router_id" | "diagnostic_loopback" | "<ip_address>"; default="main_router_id">
```

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +84,33 @@ def _set_ip_igmp_snooping_vlan(
...
+        """
+        source_address_key = default(vlan.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+        if source_address_key is None:
+            return self.shared_utils.router_id
+
+        if vrf is not None:
+            # SVI is attached to a VRF
+            if source_address_key == "vrf_router_id":
+                return self.get_vrf_router_id(vrf, tenant, vrf.bgp.router_id)
+            return self.get_vrf_router_id(vrf, tenant, source_address_key, context="'vrf_router_id' is set to 'diagnostic_loopback' on the SVI")
```
Vibhu-gslab: We can also set it to `main_router_id`, should we include that in the context?

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +84,33 @@ def _set_ip_igmp_snooping_vlan(
...
+
+        # For L2VLANs, 'vrf_router_id' and 'diagnostic_loopback' are treated as 'main_router_id'.
+        if source_address_key in {"main_router_id", "diagnostic_loopback", "vrf_router_id"}:
+            return self.shared_utils.router_id if not self.inputs.use_router_general_for_router_id else None
```
Vibhu-gslab: I think we can use `get_vrf_router_id` here as well with `router_id = main_router_id`.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
Vibhu-gslab: umm.. but for `main_router_id` `context` is not being used.. so may be not.

Vibhu-gslab: when we set `source_address_key= main_router_id` we should not pass `context`, this is the only reason I am keeping this comment open for now.

MaheshGSLAB: If source_address_key is main_router_id, passing context does not matter. The error occurs only when router_id is diagnostic_loopback. We need to clarify in the error message that diagnostic_loopback is set by router_id on a VRF or by vrf_router_id on an SVI.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: yeah but again we need to manipulate the source_address_key before to call get_vrf_router_id. Since vrf_router_id work as main_router_id and diagnostic_loopback work as Loopback0. So wee need to add condition for the same

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,15 @@ $defs:
               description: Will be enabled automatically if "evpn_l2_multicast" is enabled.
             source_address:
               type: str
-              format: ipv4
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the vrf Loopback address.
+                - 'diagnostic_loopback' will configure the vtep loopback.
```
Shivani-gslab:
```suggestion
                - `diagnostic_loopback` will configure the vtep loopback.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,15 @@ $defs:
...
+                - `main_router_id` will configure the Loopback0 IP address.
+                - Any other IPv4 address will be used directly as the source address.
+                - Default IP address of Loopback0
```
Shivani-gslab: so this will be the `main_router_id` ?

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
laxmikantchintakindi: Could we use _get_vtep_diagnostic_loopback_for_vrf directly?

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
MaheshGSLAB: No, if we dont define anything that mean None and for that case we return self.shared_utils.router_id.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: if the source_address_key is actual ipv4 address then we dont want to raise error, we should proceed with that address so we cant use _get_vtep_diagnostic_loopback_for_vrf directly

Shivani-gslab (approved):
LGTM

Vibhu-gslab (approved):
LGTM!

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
```diff
@@ -115,9 +115,15 @@ $defs:
...
-                If not set, IP address of "Loopback0" will be used.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` -> same as `main_router_id`.
+                - `diagnostic_loopback` -> same as `Loopback0`.
+                - `main_router_id` will configure the Loopback0 IP address.
+                - Any other IPv4 address will be used directly as the source address.
+                - Default IP address of Loopback0
+              convert_types:
+                - int
```
ClausHolbechArista:
```suggestion
              description: |-
                Since there is no VRF tied to an L2VLAN, the values  `vrf_router_id`, `diagnostic_loopback` and `main_router_id` will all configure the Loopback0 IP address.
                If set to an IPv4 address, it will be used directly as the source address.
                Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
```

ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,15 @@ $defs:
...
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the vrf Loopback address.
+                - `diagnostic_loopback` will configure the vtep loopback.
+                - `main_router_id` will configure the Loopback0 IP address.
+                - Any other IPv4 address will be used directly as the source address.
+                - Default IP address of Loopback0
+              convert_types:
+                - int
```
ClausHolbechArista:
```suggestion
              default: main_router_id
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,15 @@ $defs:
...
             source_address:
               type: str
-              format: ipv4
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the vrf Loopback address.
+                - `diagnostic_loopback` will configure the vtep loopback.
+                - `main_router_id` will configure the Loopback0 IP address.
+                - Any other IPv4 address will be used directly as the source address.
```
ClausHolbechArista:
```suggestion
                - An IPv4 address will be used directly as the source address.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,15 @@ $defs:
...
+                - Default IP address of Loopback0
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
```diff
@@ -210,9 +210,15 @@ $defs:
...
-            description: |
-              IPv4_address
-              If not set, IP address of "Loopback0" will be used.
+            description: |-
+              The value of `source_address` will be interpreted according to these rules:
+              - `vrf_router_id` will configure the vrf Loopback address.
+              - 'diagnostic_loopback' will configure the vtep loopback.
+              - `main_router_id` will configure the Loopback0 IP address.
+              - Any other IPv4 address will be used directly as the source address.
+              - Default IP address of Loopback0
```
ClausHolbechArista:
```suggestion
              - An IPv4 address will be used directly as the source address.
              Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
```

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +84,33 @@ def _set_ip_igmp_snooping_vlan(
...
+        self: AvdStructuredConfigNetworkServicesProtocol,
+        vlan: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem.SvisItem
+        | EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.L2vlansItem,
+        tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
+        vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem | None,
+    ) -> str | None:
+        """
+        Return the IGMP snooping querier source address for a given VLAN.
+
+        If the source address is not set, the router ID is returned.
```
ClausHolbechArista: If you update the schema as I suggest, this will never happen, since the tenant level will have a default value.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +84,33 @@ def _set_ip_igmp_snooping_vlan(
...
+        For an SVI attached to a VRF, if the source address is 'vrf_router_id', the VRF router ID is returned.
+        For an L2VLAN, 'vrf_router_id' and 'diagnostic_loopback' are treated as 'main_router_id'.
+        """
+        source_address_key = default(vlan.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+        if source_address_key is None:
+            return self.shared_utils.router_id
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -400,6 +400,7 @@ def get_vrf_router_id(
         vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem,
         tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
         router_id: str,
+        context: str = "'router_id' is set to 'diagnostic_loopback' on the VRF",
```
ClausHolbechArista: maybe it is enough just to give the `router_id` as context, since in the new case that would just be changed to `igmp_snooping_querier.source_address`

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
ClausHolbechArista:
```suggestion
            return self.get_vrf_router_id(vrf, tenant, source_address_key, context="'igmp_snooping_querier.source_address' is set to 'diagnostic_loopback' on the VLAN")
```

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
ClausHolbechArista: we should never return none. If we cannot figure out the correct source ip we need to raise.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
MaheshGSLAB: but we need to add the details like its on VRF or VLAN

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: Ok then we need to update the existing get_vrf_router_id method because there also we are returning None. Secondly self.shared_utils.router_id also return str | None(in case if the device is not underlay_router)

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
MaheshGSLAB: updated

MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
ClausHolbechArista: well you can catch it for igmp querier only and raise for that case. We don't have to change the router id logic right now, if you just check the value you get back.

MaheshGSLAB: Added!

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-igmp-snooping-querier-source-address-resolve.yml
```diff
@@ -0,0 +1,34 @@
...
+        vrf_vni: 41
+        svis:
+          - id: 21
+            name: "VLAN_21"
+            tags: ['test_l3']
+
+expected_error_message: >-
+  Unable to determine the IGMP snooping querier source address for VLAN 'VLAN_21' in VRF 'IGMP_QUERIER_TEST_3' in Tenant 'Tenant_A'.
+  The configured 'igmp_snooping_querier.source_address' could not be resolved.
+  If using 'main_router_id', 'diagnostic_loopback', or 'vrf_router_id' for an L2VLAN or SVI, ensure that 'use_router_general_for_router_id' is not true.
```
ClausHolbechArista: I don't understand this logic. I can have `use_router_general_for_router_id` and it should still be able to pick up that router_id here. If our logic prevents this, we need to break up the logic to not get into this situation.

ClausHolbechArista (dismissed):

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-igmp-snooping-querier-source-address-resolve.yml
MaheshGSLAB: For SVI if source_address is main_router_id and use_router_general_for_router_id is true in that case we are raising this error and for L2VLAN if source_address is 'main_router_id', 'diagnostic_loopback', or 'vrf_router_id'  and use_router_general_for_router_id is true in that case we are raising this error. As for L2VLANs, 'vrf_router_id' and 'diagnostic_loopback' are treated as 'main_router_id'.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
```diff
@@ -210,9 +210,15 @@ $defs:
...
-            description: |
-              IPv4_address
-              If not set, IP address of "Loopback0" will be used.
+            description: |-
+              The value of `source_address` will be interpreted according to these rules:
+              - `vrf_router_id` will configure the vrf Loopback address.
+              - 'diagnostic_loopback' will configure the vtep loopback.
+              - `main_router_id` will configure the Loopback0 IP address.
+              - An IPv4 address will be used directly as the source address.
+              Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
```
alexeygorbunov, line 240:
```suggestion
              The value of `source_address` will be interpreted according to these rules:
              - `vrf_router_id` will configure the VRF router ID address.
              - 'diagnostic_loopback' will configure the VRF Diagnostic Loopback address.
              - `main_router_id` will configure the Loopback0 IP address.
              - An IPv4 address will be used directly as the source address.
              Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -331,8 +331,13 @@ $defs:
...
             source_address:
               type: str
-              format: ipv4
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the vrf Loopback address.
+                - `diagnostic_loopback` will configure the vtep loopback.
+                - `main_router_id` will configure the Loopback0 IP address.
+                - An IPv4 address will be used directly as the source address.
```
alexeygorbunov, line 340:
```suggestion
                The value of `source_address` will be interpreted according to these rules:
                - `vrf_router_id` will configure the VRF router ID address.
                - `diagnostic_loopback` will configure the VRF Diagnostic Loopback.
                - `main_router_id` will configure the Loopback0 IP address.
                - An IPv4 address will be used directly as the source address.
```

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +85,44 @@ def _set_ip_igmp_snooping_vlan(
...
+        """
+        Return the IGMP snooping querier source address for a given VLAN.
+
+        For an SVI attached to a VRF, if the source address is 'vrf_router_id', the VRF router ID is returned.
+        For an L2VLAN, 'vrf_router_id' and 'diagnostic_loopback' are treated as 'main_router_id'.
+        """
+        source_address_key = default(vlan.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+
+        if vrf is not None:
+            # SVI is attached to a VRF
```
alexeygorbunov: Does this comment really reflect the content/logic of the following `if` block?

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-igmp-snooping-querier-source-address-resolve.yml
alexeygorbunov: > I don't understand this logic. I can have `use_router_general_for_router_id` and it should still be able to pick up that router_id here. If our logic prevents this, we need to break up the logic to not get into this situation.

I assume this is just due to the following existing logic?

https://github.com/aristanetworks/avd/blob/8b0da657f1fcdf2d54155a2ce1b8e8c6c240735e/python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py#L430-L431

ClausHolbechArista: I know it is because of how we have chosen to implement it, but it makes no sense from a user perspective. We should not add restriction on the user just because of our code reuse. Then we need to break the code up in to better functions that works for more cases.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: my bad the comment was for if vrf is not None:. removed

ansible_collections/arista/avd/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-igmp-snooping-querier-source-address-resolve.yml
MaheshGSLAB: updated the code logic with one more optional argument.

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -400,6 +400,8 @@ def get_vrf_router_id(
         vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem,
         tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
         router_id: str,
+        error_context: str = "'router_id' is set to 'diagnostic_loopback' on the VRF",
+        none_source_interface: bool = False,
```
gmuloc: This functions become too complexe - can we change the code in the calling function instead?

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
```diff
@@ -115,9 +115,10 @@ $defs:
               description: Will be enabled automatically if evpn_l2_multicast is enabled.
             source_address:
               type: str
-              description: |
-                IPv4_address
-                If not set, IP address of "Loopback0" will be used.
+              description: |-
+                Since there is no VRF tied to an L2VLAN, the values  `vrf_router_id`, `diagnostic_loopback` and `main_router_id` will all configure the Loopback0 IP address.
```
gmuloc:
```suggestion
                Since there is no VRF tied to an L2VLAN, the values `vrf_router_id`, `diagnostic_loopback` and `main_router_id` will all configure the Loopback0 IP address.
```

gmuloc: Need to explain the values in here and explain why they all point to Loopback0:


```suggestion
                The value of `source_address` will be interpreted according to these rules:
                - `vrf_router_id`, `diagnostic_loopback` and `main_router_id` will configure the Loopback0 IP since there is no VRF tied to an L2VLAN.
                - An IPv4 address will be used directly as the source address.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
```diff
@@ -115,9 +115,10 @@ $defs:
...
+                If set to an IPv4 address, it will be used directly as the source address.
```
gmuloc:
```suggestion
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
```diff
@@ -226,9 +226,15 @@ $defs:
...
-              If not set, IP address of "Loopback0" will be used.
+            description: |-
+              The value of `source_address` will be interpreted according to these rules:
+              - `vrf_router_id` will configure the VRF router ID address.
+              - 'diagnostic_loopback' will configure the VRF Diagnostic Loopback address.
+              - `main_router_id` will configure the Loopback0 IP address.
+              - An IPv4 address will be used directly as the source address.
+              Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
+            convert_types:
+              - int
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/IGMP_QUERIER_TESTS.yml
```diff
@@ -44,6 +49,7 @@ tenants:
             ip_address_virtual: 10.0.2.1/24
             igmp_snooping_querier:
               version: 3
+              source_address: vrf_router_id
```
gmuloc, line 52: there is never a test for `diagnostic_loopback`. Can we ensure we have testing for all variants on both SVIs and L2VLANS

gmuloc: keep the test where you have `vrf_router_id` -> `diagnostic_loopback` like you have here but make sure we have also in another VRF something wehre `vrf_router_id` would yield some different IPv4/

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +85,46 @@ def _set_ip_igmp_snooping_vlan(
...
+        else:
+            source_address = source_address_key
+
+        if source_address is not None:
+            return source_address
+
+        vrf_info = f" in VRF '{vrf.name}'" if vrf else ""
+        msg = (
+            f"Unable to determine the IGMP snooping querier source address for VLAN '{vlan.name}'{vrf_info} in Tenant '{tenant.name}'. "
+            "The configured 'igmp_snooping_querier.source_address' resolved to None."
```
gmuloc:
```suggestion
            f"The configured 'igmp_snooping_querier.source_address: {source_address_key}' resolved to None."
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -408,6 +410,8 @@ def get_vrf_router_id(
             vrf: The VRF object containing OSPF/BGP and vtep_diagnostic details.
             tenant: The Tenant to which the VRF belongs.
             router_id: The router ID type specified for the VRF (e.g., "vtep_diagnostic", "main_router_id", "none", or an IPv4 address).
```
gmuloc:
```suggestion
            router_id: The router ID type specified for the VRF (e.g., "diagnostic_loopback", "main_router_id", "none", or an IPv4 address).
```

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
gmuloc:
```suggestion
        # Handle "diagnostic_loopback" router ID case
```

gmuloc (changes requested):
let's write all the tests first and see how we can adjust the logic to avoid overcomplicating `utils.py` function

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -422,13 +426,13 @@ def get_vrf_router_id(
...
                     f"Invalid configuration on VRF '{vrf.name}' in Tenant '{tenant.name}'. "
                     "'vtep_diagnostic.loopback' along with either 'vtep_diagnostic.loopback_ip_pools' or 'vtep_diagnostic.loopback_ip_range' must be defined "
-                    "when 'router_id' is set to 'diagnostic_loopback' on the VRF."
+                    f"when {error_context}."
                 )
                 raise AristaAvdInvalidInputsError(msg)
             # Resolve router ID from loopback interface
             return get_ip_from_ip_prefix(interface_data.ip_address)
         if router_id == "main_router_id":
-            return self.shared_utils.router_id if not self.inputs.use_router_general_for_router_id else None
```
gmuloc, line 431: this logic probably needs to be moved to the place where `get_vrf_router_id` is called

MaheshGSLAB: i have updated the code for igmp by which the main_router_id will handle there only so we do not need to update the code where get_vrf_router_id is called

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
MaheshGSLAB: updated

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/IGMP_QUERIER_TESTS.yml
MaheshGSLAB: added few more tests

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -400,29 +400,31 @@ def get_vrf_router_id(
...
         """
         Determine the router ID for a given VRF based on its configuration.
 
         Args:
             vrf: The VRF object containing OSPF/BGP and vtep_diagnostic details.
             tenant: The Tenant to which the VRF belongs.
-            router_id: The router ID type specified for the VRF (e.g., "vtep_diagnostic", "main_router_id", "none", or an IPv4 address).
+            router_id: The router ID type specified for the VRF (e.g., "diagnostic_loopback", "main_router_id", "none", or an IPv4 address).
+            error_context: A string indicating the context of the router ID (e.g., "router_id" for BGP or OSPF).
+            none_source_interface: A boolean indicating if the source-interface is 'none' then raise the error.
```
MaheshGSLAB:
```suggestion
```

ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-svi-igmp-snooping-querier-source-address-resolve.yml
```diff
@@ -0,0 +1,31 @@
...
+      bgp_as: 65101
+      id: 1
+      mgmt_ip: 192.168.200.101/24
+
+tenants:
+  - name: Tenant_A
+    mac_vrf_vni_base: 10000
+    igmp_snooping_querier:
+      enabled: true
+      source_address: none
```
laxmikantchintakindi, line 19: We have not mentioned in schema that we can use `none` keyword. As per schema the only options we have are `vrf_router_id`, `diagnostic_loopback`, `main_router_id` or any IPv4 address.

MaheshGSLAB: none is negative test, it will raise the error so we should not mention a value which going to raise the error.

gmuloc: this does not really feel right as you are using a piece of code written for something else where `"none"` is a valid value to trigger the error message here. 
Also it does not test for the error message for L2VLANs

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_l2vlans.schema.yml
```diff
@@ -115,9 +115,10 @@ $defs:
               description: Will be enabled automatically if evpn_l2_multicast is enabled.
             source_address:
               type: str
-              description: |
-                IPv4_address
-                If not set, IP address of "Loopback0" will be used.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id`, `diagnostic_loopback` and `main_router_id` will configure the Loopback0 IP since there is no VRF tied to an L2VLAN.
+                - An IPv4 address will be used directly as the source address.
```
gmuloc, line 121: Given that this is configured at the tenant level we should add this here:

```suggestion
                - An IPv4 address will be used directly as the source address.
                Overrides `<network_services_key>[].igmp_snooping_querier.source_address`.
```

ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/inventory/host_vars/failure-svi-igmp-snooping-querier-source-address-resolve.yml
gmuloc: let's add some check for "valid IPv4 address" if we go through all the cases in the code and test the negative unit test against that

MaheshGSLAB: added

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -36,20 +38,41 @@ def ip_igmp_snooping(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
...
         vlan: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem.SvisItem
         | EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.L2vlansItem,
         tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
+        vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem | None,
     ) -> None:
         """
         Set ip_igmp_snooping structured_config for one vlan.
 
-        Can be used for both svis and l2vlans
+        Can be used for both SVIs and L2vlans
```
alexeygorbunov:
```suggestion
        Can be used for both SVIs and L2VLANs
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -332,8 +332,13 @@ $defs:
               description: Will be enabled automatically if "evpn_l2_multicast" is enabled.
             source_address:
               type: str
-              format: ipv4
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the VRF router ID address according to `vrf.bgp.router_id`.
```
alexeygorbunov: `vrf.bgp.router_id` is an internal path which may be not clear to the reader. Would it be better to reference full `<network_services_keys.name>[].vrfs[].bgp.router_id

gmuloc: yeah - we are technically inside the VRF already I would think but lets do this

```suggestion
                - `vrf_router_id` will configure the VRF router ID address according to ``<network_services_keys.name>[].vrfs[].bgp.router_id`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
```diff
@@ -231,9 +231,13 @@ $defs:
             description: Will be enabled automatically if evpn_l2_multicast is enabled.
           source_address:
             type: str
-            description: |
-              IPv4_address
-              If not set, IP address of "Loopback0" will be used.
+            description: |-
+              The value of `source_address` will be interpreted according to these rules:
+              - `vrf_router_id` will configure the VRF router ID address according to `vrf.bgp.router_id`.
```
gmuloc:
```suggestion
              - `vrf_router_id` will configure the VRF router ID address according to ``<network_services_keys.name>[].vrfs[].bgp.router_id`.
```

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +112,71 @@ def _set_ip_igmp_snooping_vlan(
...
+        tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
+        vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem,
+    ) -> str:
+        """Return the IGMP snooping querier source address for an SVI."""
+        source_address_key = default(svi.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+
+        match source_address_key:
+            case "main_router_id":
+                source_address = self.shared_utils.router_id
+            case "vrf_router_id":
```
alexeygorbunov, line 152: Is there any value to have `vrf_router_id` here as for this option we are simply passing `vrf.bgp.router_id` (which is either `main_router_id` or `diagnostic_loopback`) into `self.get_vrf_router_id` which will still fallback into either `main_router_id` or `diagnostic_loopback`

gmuloc: yes if you enter `bgp.router_id: 42.42.42.42`

alexeygorbunov: oh, right!

alexeygorbunov (approved):
Besides the usage of `cast`-ing (that we were trying to avoid) it LGTM. Generated outputs look correct for all variants of the inputs.

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +112,71 @@ def _set_ip_igmp_snooping_vlan(
...
+    def _get_l2vlan_igmp_querier_source_address(
+        self: AvdStructuredConfigNetworkServicesProtocol,
+        l2vlan: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.L2vlansItem,
+        tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
+    ) -> str:
+        """Return the IGMP snooping querier source address for an L2VLAN."""
+        source_address_key = default(l2vlan.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+
+        source_address = self.shared_utils.router_id if source_address_key in {"main_router_id", "diagnostic_loopback", "vrf_router_id"} else source_address_key
+        source_address = cast("str", source_address)
```
MaheshGSLAB:
```suggestion
        source_address = cast("str", source_address) if source_address is None else source_address
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_svi_settings.schema.yml
```diff
@@ -231,9 +231,13 @@ $defs:
             description: Will be enabled automatically if evpn_l2_multicast is enabled.
           source_address:
             type: str
-            description: |
-              IPv4_address
-              If not set, IP address of "Loopback0" will be used.
+            description: |-
+              The value of `source_address` will be interpreted according to these rules:
+              - `vrf_router_id` will configure the VRF router ID address according to ``<network_services_keys.name>[].vrfs[].bgp.router_id`.
```
Shivani-gslab:
```suggestion
              - `vrf_router_id` will configure the VRF router ID address according to `<network_services_keys.name>[].vrfs[].bgp.router_id`.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -332,8 +332,13 @@ $defs:
               description: Will be enabled automatically if "evpn_l2_multicast" is enabled.
             source_address:
               type: str
-              format: ipv4
-              description: Default IP address of Loopback0.
+              description: |-
+                The value of `source_address` will be interpreted according to these rules:
+                - `vrf_router_id` will configure the VRF router ID address according to ``<network_services_keys.name>[].vrfs[].bgp.router_id`.
```
Shivani-gslab:
```suggestion
                - `vrf_router_id` will configure the VRF router ID address according to `<network_services_keys.name>[].vrfs[].bgp.router_id`.
```

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
gmuloc: this looks not correct - was it applied?

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +112,71 @@ def _set_ip_igmp_snooping_vlan(
...
+        except AddressValueError:
+            msg = (
+                f"Invalid IGMP snooping querier source address for VLAN '{l2vlan.name}' "
+                f"in Tenant '{tenant.name}'. The value '{source_address}' resolved from "
+                f"'igmp_snooping_querier.source_address: {source_address_key}' "
+                "is not a valid IPv4 address."
+            )
+            raise AristaAvdInvalidInputsError(msg) from None
+
+        return source_address
```
gmuloc:
```suggestion
        if source_address is not None:
            try:
                IPv4Address(source_address)
                return source_address
            except AddressValueError:
                # The next block will raise
                pass
        msg = (
            f"Invalid IGMP snooping querier source address for VLAN '{l2vlan.name}' "
            f"in Tenant '{tenant.name}'. The value '{source_address}' resolved from "
            f"'igmp_snooping_querier.source_address: {source_address_key}' "
            "is not a valid IPv4 address."
        )
        raise AristaAvdInvalidInputsError(msg)
```

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +112,71 @@ def _set_ip_igmp_snooping_vlan(
...
+        svi: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem.SvisItem,
+        tenant: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem,
+        vrf: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem,
+    ) -> str:
+        """Return the IGMP snooping querier source address for an SVI."""
+        source_address_key = default(svi.igmp_snooping_querier.source_address, tenant.igmp_snooping_querier.source_address)
+
+        match source_address_key:
+            case "main_router_id":
+                source_address = self.shared_utils.router_id
```
laxmikantchintakindi, line 151: If use_router_general_for_router_id is enabled, then this will choose wrong router_id.

gmuloc: I think it will return None and raise?

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
MaheshGSLAB: there were two different msg if source address is None and source address is not valid ipv4 address hence i tried to make one message only that even in case of None the source address is not valid ipv4 address. So for a case where source address return None the IPv4Address(source_address) is not able to handle it as it accept str or int. So try to cast the None into a string "None". But your solution is also works with little modification as ruff is not happy with return source address in try block hence use else as suggested by ruff.
```
ruff (legacy alias).....................................................................................................Failed
- hook id: ruff
- exit code: 1

TRY300 Consider moving this statement to an `else` block
   --> python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py:128:17
    |
126 |             try:
127 |                 IPv4Address(source_address)
128 |                 return source_address
    |                 ^^^^^^^^^^^^^^^^^^^^^
129 |             except AddressValueError:
130 |                 # The next block will raise
    |

Found 1 error.
```

laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
```diff
@@ -85,3 +112,73 @@ def _set_ip_igmp_snooping_vlan(
...
+            "is not a valid IPv4 address."
+        )
+        if source_address is not None:
+            try:
+                IPv4Address(source_address)
+            except AddressValueError:
+                raise AristaAvdInvalidInputsError(msg) from None
+            else:
+                return source_address
+        raise AristaAvdInvalidInputsError(msg)
```
MaheshGSLAB, line 138: Tried to write the testcase for uncovered code where source_address is None by router_id as None but when router_id is None then unable to reach the code for network services ip igmp snooping. Hence leaving with uncovered for now.

gmuloc (approved):

python-avd/pyavd/_eos_designs/structured_config/network_services/ip_igmp_snooping.py
gmuloc: addde them in my last commit

PR #5786, opened 2025-09-01
Refactor(eos_designs)!: Set underlay bgp peer groups only when neighbors are using the peer group

### LINKED ISSUE #5521: Remove Underlay BGP group when no neighbors are generated
### Enhancement summary

When devices within AVD fall out of EVPN fabric being able to remove or maybe not even generate the group IPv4-UNDERLAY-PEERS. 

eos_designs

### Use case example

For devices outside the EVPN fabric BGP group IPv4-UNDERLAY-PEERS gets generated no matter the AVD role.  

Below is the internal repository.

https://gitlab.aristanetworks.com/professional-services/bnym-avd

Either a new role for devices falling outside the fabric or some knob maybe within the fabric signifying these are not EVPN devices and underlay protocol BGP group isn't necessary? 

created group_vars file for specific nodes overriding BGP group thus renaming IPv4-UNDERLAY-PEERS to existing BGP group resulting in removing IPv4-UNDERLAY-PEERS.

### Additional context

_No response_

### Contributing Guide

laxmikantchintakindi:
## Change Summary

Set underlay bgp peer groups only when neighbors are using the peer group.

## Related Issue(s)

Fixes #5521

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
When devices within AVD fall out of EVPN fabric being able to remove or maybe not even generate the group IPv4-UNDERLAY-PEERS.

## How to test
When there are no neighbors using the peer group, underlay peer group should not be configured.
### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -138,3 +120,8 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
                             description=f"{f'{link.peer}_{subinterface.peer_interface}'}_vrf_{subinterface_vrf}",
                             bfd=link.bfd,
                         )
+
+    @cached_property
+    def underlay_p2p_links(self: AvdStructuredConfigUnderlayProtocol) -> list[EosDesignsFacts.UplinksItem]:
+        """Return a list of P2P underlay links."""
+        return [link for link in self._underlay_links if link.type == "underlay_p2p"]
```
Shivani-gslab: you can move this method to `underlay/utils.py` and use it at other places as well.
also, as it is an internal method, change the name to `_underlay_p2p_links`

Shivani-gslab (changes requested):

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/shared_utils/underlay.py
```diff
@@ -133,3 +133,33 @@ def underlay_isis_authentication_key(self: SharedUtilsProtocol) -> str | None:
...
+    def address_family_ipv4_peer_group(self: SharedUtilsProtocol) -> EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroupsItem | None:
+        if not self.underlay_ipv6_numbered:
+            address_family_ipv4_peer_group = EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroupsItem(
+                name=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name, activate=True
+            )
+            if self.inputs.underlay_rfc5549 is True:
+                address_family_ipv4_peer_group.next_hop.address_family_ipv6._update(enabled=True, originate=True)
+
+            return address_family_ipv4_peer_group
+        return None
```
MaheshGSLAB: we should not return the None, lets check the self.underlay_ipv6_numbered before calling the method

python-avd/pyavd/_eos_designs/shared_utils/underlay.py
```diff
@@ -133,3 +133,33 @@ def underlay_isis_authentication_key(self: SharedUtilsProtocol) -> str | None:
...
+
+    @cached_property
+    def address_family_ipv6_peer_group(self: SharedUtilsProtocol) -> EosCliConfigGen.RouterBgp.AddressFamilyIpv6.PeerGroupsItem | None:
```
MaheshGSLAB: same as above

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/underlay-filter-redistribute-connected-to-bgp-without-route-map.yml
```diff
@@ -60,18 +60,16 @@ router_bgp:
...
-    maximum_routes: 12000
-    route_map_out: RM-BGP-UNDERLAY-PEERS-OUT
   - name: EVPN-OVERLAY-PEERS
     type: evpn
     update_source: Loopback0
     bfd: true
     ebgp_multihop: 3
     send_community: all
     maximum_routes: 0
+  - name: IPv4-UNDERLAY-PEERS
```
MaheshGSLAB: is this expected change

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/tenant-default-vrf-settings.yml
```diff
@@ -86,18 +86,16 @@ router_bgp:
...
-    maximum_routes: 12000
-    route_map_out: RM-BGP-UNDERLAY-PEERS-OUT
   - name: EVPN-OVERLAY-PEERS
     type: evpn
     update_source: Loopback0
     bfd: true
     ebgp_multihop: 3
     send_community: all
     maximum_routes: 0
+  - name: IPv4-UNDERLAY-PEERS
```
MaheshGSLAB: same here

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -25,53 +25,35 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
...
-                address_family_ipv4_peer_group.next_hop.address_family_ipv6._update(enabled=True, originate=True)
-
-            self.structured_config.router_bgp.address_family_ipv4.peer_groups.append(address_family_ipv4_peer_group)
-
-        if self.shared_utils.underlay_ipv6:
-            self.structured_config.router_bgp.address_family_ipv6.peer_groups.append_new(
-                name=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name, activate=True
-            )
+        # Set BGP peer group only when underlay link is present.
+        if self._underlay_p2p_links:
```
MaheshGSLAB: store the output into a variable otherwise you are keep doing the [link for link in self._underlay_links if link.type == "underlay_p2p"] again and again

python-avd/pyavd/_eos_designs/shared_utils/underlay.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/shared_utils/underlay.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
laxmikantchintakindi: self._underlay_p2p_links is cached property, so it won't be processed repeatedly.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/underlay-filter-redistribute-connected-to-bgp-without-route-map.yml
laxmikantchintakindi: This peer group was added using default vrf in network services. Fixed this.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/tenant-default-vrf-settings.yml
laxmikantchintakindi: This peer group was added using default vrf in network services. Fixed this.

MaheshGSLAB (commented):
improve the coverage

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_eos_designs/shared_utils/underlay.py
laxmikantchintakindi: This is used in core_interfaces_and_l3_edge/router_bgp and underlay/router_bgp. So to not repeat the check, I added it here.

alexeygorbunov (dismissed):
LGTM

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
gmuloc: It seems like you could rewrite this like this and exit early and deindent the rest of the function given that both line 55 and 79 then loop over the `underlay_p2p_links` so would do nothing if it is empty.

```suggestion
        if not self._underlay_p2p_links:
            return
        # <...> the rest of the function
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -119,6 +119,8 @@ def _router_bgp_peer_groups(self: AvdStructuredConfigNetworkServicesProtocol) ->
             self.structured_config.router_bgp.peer_groups.append_new(
                 name=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name,
                 type="ipv4",
+                send_community="all",
+                maximum_routes=12000,
```
gmuloc: so this was needed because when you changed the code  in underlay the lines were missing?

gmuloc: so followinga rabbit hole discussion we need to put this PR on hold.

1/ this will explode once we force using the same structured config for all modules (you are adding a route-map, not configuring password,...) so we will need to fix this after a yet-to-be-commited PR is merged
2/ afterwards, this code should be changed to check if the peer-group is configured and add the route-map in that case

alternatively, we need to move the shared_utils this bit and it would be a lot and we would rather avoid.

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
laxmikantchintakindi: Done.

gmuloc:
@laxmikantchintakindi can you please update coverage on this - I think at least core & l3edge and underlay router BGP are fairly easy to address as it just requires to add some underlay_structured_config to the right molecule scenarii

The one in shared_utils should be sraightforward as well (enabling bfd for underlay peer group in EOS designs)

Shivani-gslab:
> @laxmikantchintakindi can you please update coverage on this - I think at least core & l3edge and underlay router BGP are fairly easy to address as it just requires to add some underlay_structured_config to the right molecule scenarii
> 
> The one in shared_utils should be sraightforward as well (enabling bfd for underlay peer group in EOS designs)

Improved the coverage

MaheshGSLAB (approved):

Vibhu-gslab (approved):

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-cleartext-passwords.cfg
```diff
@@ -144,10 +144,6 @@ router bgp 666
    neighbor PG-TENANT-LEVEL-TRUE password 7 txBXbmEsAhrSxZnkOpWEkw==
    neighbor PG-VRF-LEVEL-TRUE peer group
    neighbor PG-VRF-LEVEL-TRUE password 7 n2siNKbxnE7KcrfEAsHShQ==
-   neighbor UNDERLAY-PEERS peer group
-   neighbor UNDERLAY-PEERS password 7 TNj+rhJ4XPgAD3zuTCxmMw==
```
gmuloc, line 155: are we losing some coverage here?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-preencrypted-passwords.cfg
```diff
@@ -144,10 +144,6 @@ router bgp 666
    neighbor PG-TENANT-LEVEL-FALSE password 7 IQ1xyukPoIcBKZmUUQWyGA==
    neighbor PG-VRF-LEVEL-FALSE peer group
    neighbor PG-VRF-LEVEL-FALSE password 7 +8KX3tMSiFtObNqjE03nkw==
-   neighbor UNDERLAY-PEERS peer group
-   neighbor UNDERLAY-PEERS password 7 0nsCUm70mvSTxVO0ldytrg==
```
gmuloc, line 148: same question

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/core-4-multicast.cfg
```diff
@@ -43,30 +43,23 @@ interface Ethernet3
 interface Loopback0
    description ROUTER_ID
    no shutdown
-   ip address 10.0.0.4/32
```
gmuloc, line 46: investigate

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge4B.cfg
```diff
@@ -358,11 +358,6 @@ router bgp 65000
    update wait-install
    no bgp default ipv4-unicast
    maximum-paths 16
-   neighbor IPv4-UNDERLAY-PEERS peer group
-   neighbor IPv4-UNDERLAY-PEERS route-map RM-BGP-UNDERLAY-PEERS-IN in
-   neighbor IPv4-UNDERLAY-PEERS route-map RM-BGP-UNDERLAY-PEERS-OUT out
```
gmuloc, line 363: hmm if we remove this we should remove teh route-maps as well.
need to check if this should be removed in that case also

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
```diff
@@ -370,10 +370,6 @@ router bgp 65000
    maximum-paths 16
    bgp listen range 192.168.142.0/24 peer-group WAN-OVERLAY-PEERS remote-as 65000
    bgp listen range 192.168.143.0/24 peer-group WAN-OVERLAY-PEERS remote-as 65000
-   neighbor IPv4-UNDERLAY-PEERS peer group
-   neighbor IPv4-UNDERLAY-PEERS route-map RM-BGP-UNDERLAY-PEERS-IN in
```
gmuloc, line 378: same for the route-map

gmuloc (commented):
Halfay comment 
```
rg IPv4-UNDERLAY  | grep password
```
This comes back empty when looking in inventor, let's make sure we still render a password for the underlay group.

Ideally add an underlay BGP peer for the two tests I flagged


Also for cv-pathfinder some route-maps become useless once we remove the peer group so we need to check if 1. removing the peer-group is correct (some routers should have BGP peers in the underlay for some test) 2. if so the route-maps (and any relevant object) should also be removed.

Maybe this cv-pathfinder one can be taken in a second PR

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/network-services-static-routes.yml
```diff
@@ -230,6 +224,10 @@ router_bgp:
     ebgp_multihop: 3
     send_community: all
     maximum_routes: 0
+  - name: IPv4-UNDERLAY-PEERS
+    metadata:
+      type: ipv4
```
gmuloc: so here we lose send_commynity and maximum routes which does not look great

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/CORE_UNIT_TESTS.yml
```diff
@@ -58,6 +58,7 @@ core_router:
       id: 3
     - name: core-4-multicast
       id: 4
+      router_id_pool: 10.255.1.0/24
```
gmuloc: so I suppose the reason for this change is because we are losing the loopback? Can we put a comment?
the full reasonoing is probably because of ipv6 numbered?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/IPV6_NUMBERED_LEAFS.yml
```diff
@@ -28,6 +28,9 @@ bgp_peer_groups:
   ipv4_underlay_peers:
     name: IPv6-UNDERLAY-PEERS
     password: F0bdF6JLcJtagc6PCN8W0Q==
+    bfd: true
+    structured_config:
+      description: Test_for_ipv4_underlay_peers
```
gmuloc: can we use `ipv6` in description ? :)

python-avd/pyavd/_eos_designs/shared_utils/utils.py
```diff
@@ -194,6 +194,23 @@ def underlay_vlan_trunk_groups(self: SharedUtilsProtocol) -> dict[int, set[str]]
                 break
         return vlans
 
+    def set_underlay_bgp_peer_group(self: SharedUtilsProtocol, peer_group: EosCliConfigGen.RouterBgp.PeerGroupsItem) -> None:
+        """Set the structured config for Underlay BGP peer group."""
+        af_type = "ipv4" if not self.underlay_ipv6_numbered else "ipv6"
```
gmuloc: here the name of the peer group could be confusing for users (IPv4) but thats life

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-cleartext-passwords.cfg
laxmikantchintakindi: No. UNDERLAY-PEERS is not used for any peer,

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-preencrypted-passwords.cfg
laxmikantchintakindi: No.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/core-4-multicast.cfg
laxmikantchintakindi: This was removed because underlay_ipv6_numbered: true was set for the host.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-cleartext-passwords.cfg
gmuloc: we are losing the password coverage - please make sure there is one host testing passwords.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-preencrypted-passwords.cfg
gmuloc: same

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/network-services-static-routes.yml
laxmikantchintakindi: Send community and maximum routes should not be there as there are no peers using the peer group. Updated the code in underlay/route_maps.py with this logic.
```
        if (
            self.shared_utils.underlay_routing_protocol == "ebgp"
            and self.shared_utils.wan_role == "client"
            **and self.structured_config.router_bgp.peer_groups.get(self.inputs.bgp_peer_groups.ipv4_underlay_peers.name)**
        ):
            # RM-BGP-UNDERLAY-PEERS-IN
            sequence_numbers = EosCliConfigGen.RouteMapsItem.SequenceNumbers()
            sequence_numbers.append_new(
                sequence=40,
                type="permit",
                set=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Set([f"extcommunity soo {self.shared_utils.evpn_soo} additive"]),
                description="Mark prefixes originated from the LAN",
            )
```

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/IPV6_NUMBERED_LEAFS.yml
laxmikantchintakindi: Done.

laxmikantchintakindi:
Added password coverage.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge4B.cfg
laxmikantchintakindi: Removed.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-pathfinder.cfg
laxmikantchintakindi: REmoved.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/evpn_services_l2_only_false.cfg
gmuloc, line 536: this should also be removed if not used anywhere else

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/network-services-static-routes.cfg
gmuloc, line 167: same is it still used?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-settings.cfg
```diff
@@ -86,10 +86,6 @@ router bgp 65101
    neighbor EVPN-OVERLAY-PEERS ebgp-multihop 3
    neighbor EVPN-OVERLAY-PEERS send-community
    neighbor EVPN-OVERLAY-PEERS maximum-routes 0
-   neighbor IPv4-UNDERLAY-PEERS peer group
-   neighbor IPv4-UNDERLAY-PEERS route-map RM-BGP-UNDERLAY-PEERS-OUT out
```
gmuloc, line 90: same for route-map

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
```diff
@@ -51,6 +51,26 @@ def router_bgp(self: AvdStructuredConfigCoreInterfacesAndL3EdgeProtocol) -> None
                 msg = f"{self.data_model}.p2p_links.[].ip, .subnet or .ip_pool"
                 raise AristaAvdMissingVariableError(msg)
 
+            # Set Underlay BGP peer group first.
+            target_peer_group = self.structured_config.router_bgp.peer_groups.obtain(self.inputs.bgp_peer_groups.ipv4_underlay_peers.name)
```
gmuloc: so I think we need to re-evaluate when I am back as reading from structured config like this force the ordering of the modules (which is a reality today but could be breaking later)

@ClausHolbechArista

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -119,7 +119,11 @@ def _router_bgp_peer_groups(self: AvdStructuredConfigNetworkServicesProtocol) ->
                 self.structured_config.router_bgp.address_family_ipv6.peer_groups.append(af_peer_group)
 
         # router bgp default vrf configuration for evpn
-        if self._vrf_default_evpn and (self._vrf_default_ipv4_subnets or self._vrf_default_ipv4_static_routes["static_routes"]):
+        if (
```
gmuloc, line 122: Add a comment stating it is configured only if it exists already. (which once again induce ordering of modules, which we may not be able to avoid)

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge3A.cfg
```diff
@@ -299,18 +299,6 @@ ip extcommunity-list ECL-EVPN-SOO permit soo 192.168.42.6:404
 ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
    seq 10 permit 192.168.42.0/24 eq 32
 !
-route-map RM-BGP-UNDERLAY-PEERS-IN permit 40
```
gmuloc, line 302: curious what happens to WAN-PATH / PL-WAN-HA-PEER-PREFIXES when they are generated in this route-map and we delete the peer-group.
I think the code currently does not cater for that?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/network-services-static-routes.cfg
laxmikantchintakindi: Fixed.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-settings.cfg
laxmikantchintakindi: Fixed.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/cv-pathfinder-edge3A.cfg
laxmikantchintakindi: Added condition for ASPATH and WAN-PREFIXES.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
laxmikantchintakindi: Added.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/add-subnets-to-redistribution-in-default-vrf-with-sequence30.cfg
```diff
@@ -144,9 +144,6 @@ ip extcommunity-list ECL-EVPN-SOO permit soo 192.168.1.1:1
 ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
    seq 10 permit 192.168.1.0/24 eq 32
 !
-ip prefix-list PL-SVI-VRF-DEFAULT
```
gmuloc, line 147: hmm @laxmikantchintakindi this prefixlist is still used in RM-CONN-2-BGP permit 30 so you may have deleted too much now :)

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/evpn_services_l2_only_false.cfg
```diff
@@ -507,12 +507,6 @@ ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
    seq 10 permit 192.168.255.0/24 eq 32
    seq 20 permit 192.168.254.0/24 eq 32
 !
-ip prefix-list PL-STATIC-VRF-DEFAULT
-   seq 10 permit 10.10.10.0/24
-!
-ip prefix-list PL-SVI-VRF-DEFAULT
-   seq 10 permit 10.0.0.0/24
-!
```
gmuloc, line 515: same you srtill have

```
route-map RM-EVPN-EXPORT-VRF-DEFAULT permit 10
   match ip address prefix-list PL-SVI-VRF-DEFAULT
!
route-map RM-EVPN-EXPORT-VRF-DEFAULT permit 20
   match ip address prefix-list PL-STATIC-VRF-DEFAULT

File: ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/network-services-static-routes.cfg
Comment (gmuloc) [reply ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/network-services-static-routes.cfg:167]: ok so we need to check that RM-EVPN-EXPORT-VRF-DEFAULT will also call these prefix-lists

File: python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py [hunk 23]
```diff
@@ -76,7 +76,6 @@ def _route_maps_vrf_default(self: AvdStructuredConfigNetworkServicesProtocol) ->
             return
 
         self._evpn_export_vrf_default_route_map()
```
gmuloc: so I think you will need to also look at calling this only in the right place and make sure it calls:

```
_set_prefix_lists_vrf_default
```

it may apply to the two other prefix-lists below

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
```diff
@@ -97,7 +97,7 @@ def _vrf_default_ipv4_static_routes(self: AvdStructuredConfigNetworkServicesProt
                 continue
 
             for static_route in static_routes:
-                vrf_default_ipv4_static_routes.add(static_route.prefix or static_route.destination_address_prefix)
+                vrf_default_ipv4_static_routes.add(static_route.prefix)
```
gmuloc: this is a clean up because the key was removed for 6.0?

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -119,7 +119,14 @@ def _router_bgp_peer_groups(self: AvdStructuredConfigNetworkServicesProtocol) ->
...
         # router bgp default vrf configuration for evpn
-        if self._vrf_default_evpn and (self._vrf_default_ipv4_subnets or self._vrf_default_ipv4_static_routes["static_routes"]):
+        # Configured only when ipv4 under Peer-group exists
+        if (
+            self._vrf_default_evpn
+            and (self._vrf_default_ipv4_subnets or self._vrf_default_ipv4_static_routes["static_routes"])
+            and self.structured_config.router_bgp.peer_groups.get(self.inputs.bgp_peer_groups.ipv4_underlay_peers.name)
+        ):
+            self._bgp_underlay_peers_route_map()
+            self._set_prefix_lists_vrf_default()
```
gmuloc: move the call to this inside the self._bgp_underlay_peers_route_map function that will be clearer

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -25,53 +25,42 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
...
-
-        self.structured_config.router_bgp.peer_groups.append(peer_group)
+                    target_peer_group.allowas_in._update(enabled=True, times=1)
 
         # Address Families
         # TODO: - see if it makes sense to extract logic in method
         if not self.shared_utils.underlay_ipv6_numbered:
-            address_family_ipv4_peer_group = EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroupsItem(
-                name=self.inputs.bgp_peer_groups.ipv4_underlay_peers.name, activate=True
+            target_address_family = self.structured_config.router_bgp.address_family_ipv4.peer_groups.obtain(
```
alexeygorbunov: Why do we call variable `target_address_family` if it is really the `PeerGroupsItem`

python-avd/pyavd/_eos_designs/shared_utils/utils.py
```diff
@@ -194,6 +194,23 @@ def underlay_vlan_trunk_groups(self: SharedUtilsProtocol) -> dict[int, set[str]]
...
+        peer_group.metadata.type = af_type
+        if password := self.get_bgp_password(self.inputs.bgp_peer_groups.ipv4_underlay_peers):
+            peer_group.password = password
+        if self.inputs.bgp_peer_groups.ipv4_underlay_peers.bfd:
+            peer_group.bfd = True
+        peer_group.maximum_routes = 12000
+        peer_group.send_community = "all"
+
+    def set_ipv4_address_family(self: SharedUtilsProtocol, address_family: EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroupsItem) -> None:
```
alexeygorbunov: We are setting here attributes of the peer_group (activate or not, etc.). Why do we call func parameter `address_family`? And method should probably called differently (to reflect what is actually sets).

Same present in few other places as well

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/ospf-cleartext-passwords.cfg
```diff
@@ -345,7 +342,6 @@ router bgp 666
    !
    address-family ipv4
       no neighbor EVPN-OVERLAY-PEERS activate
-      neighbor IPv4-UNDERLAY-PEERS activate
```
alexeygorbunov, line 348: As `ipv4 unicast` address family is deactivated globally i'm wondering if we at all need to keep `address-family ipv4` section if there are no actual peer-groups for which we need to activate it (like ipv4 mlag,..). In this case we are rendering it here for no purpose

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-settings.cfg
alexeygorbunov: `PL-STATIC-VRF-DEFAULT` is referenced in `RM-EVPN-EXPORT-VRF-DEFAULT` but is still deleted globally

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/vrfs-rt-import-and-export.cfg
```diff
@@ -74,14 +74,6 @@ ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
    seq 10 permit 10.10.0.0/24 eq 32
    seq 20 permit 10.11.0.0/24 eq 32
 !
-ip prefix-list PL-SVI-VRF-DEFAULT
```
alexeygorbunov, line 77: `PL-SVI-VRF-DEFAULT` is still needed by `RM-CONN-2-BGP` and `RM-EVPN-EXPORT-VRF-DEFAULT`

laxmikantchintakindi: Fixed.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/network_services/utils.py
laxmikantchintakindi: Yes. destination_address_prefix key has been removed in 6.0.

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
laxmikantchintakindi: Done.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/evpn_services_l2_only_false.cfg
laxmikantchintakindi: Reverted.

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -69,13 +93,7 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
 
         # Neighbors and VRF Neighbors
         else:
-            for link in self._underlay_links:
-                if link.type != "underlay_p2p":
-                    continue
-
-                # Adding the peer-group first time we add a neighbor.
-                # TODO: self.set_once_peer_group_ipv4_underlay_peers()
```
gmuloc, line 77: this should stay here

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -34,18 +34,42 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
...
+
+        if self.inputs.bgp_peer_groups.ipv4_underlay_peers.structured_config:
+            self.custom_structured_configs.nested.router_bgp.peer_groups.obtain(target_peer_group.name)._deepmerge(
+                self.inputs.bgp_peer_groups.ipv4_underlay_peers.structured_config, list_merge=self.custom_structured_configs.list_merge_strategy
+            )
+
+        self.shared_utils.set_underlay_bgp_peer_group(target_peer_group)
 
-                # Adding the peer-group first time we add a neighbor interface.
-                # TODO: self.set_once_peer_group_ipv4_underlay_peers()
```
gmuloc, line 47: this should be brought back and used instead of the complex code on the right

python-avd/pyavd/_eos_designs/structured_config/network_services/route_maps.py
```diff
@@ -138,6 +138,7 @@ def set_once_route_map_evpn_export_vrf_default(self: AvdStructuredConfigNetworkS
         This should only be called when any one of self._vrf_default_ipv4_subnets, self._vrf_default_ipv4_static_routes["static_routes"],
         self.shared_utils.is_wan_router is true.
         """
+        # Set prefix-lists from EVPN services in VRF "default" (if any)
```
gmuloc:
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
gmuloc: this needs to be repladced by the call to `set_once_peer_group_ipv4_underlay_peers` @laxmikantchintakindi

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -119,7 +119,11 @@ def _router_bgp_peer_groups(self: AvdStructuredConfigNetworkServicesProtocol) ->
                 self.structured_config.router_bgp.address_family_ipv6.peer_groups.append(af_peer_group)
 
         # router bgp default vrf configuration for evpn
-        if self._vrf_default_evpn and (self._vrf_default_ipv4_subnets or self._vrf_default_ipv4_static_routes["static_routes"]):
+        if (
+            self._vrf_default_evpn
+            and (self._vrf_default_ipv4_subnets or self._vrf_default_ipv4_static_routes["static_routes"])
+            and self.structured_config.router_bgp.peer_groups.get(self.inputs.bgp_peer_groups.ipv4_underlay_peers.name)
```
gmuloc: same here, instead of the check, lets call `set_once_peer_group_ipv4_underlay_peers` to make sure the peer group is there if wee nedd it and then do the manipulation

gmuloc: scratch my previous comment and keep the check for now

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
laxmikantchintakindi: Move this method to shared_utils and called from required modules.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/shared_utils/utils.py
gmuloc: there was an issue opened for this actually so lets tackle it separately

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-settings.cfg
gmuloc: @laxmikantchintakindi is it still the case?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/ospf-cleartext-passwords.cfg
gmuloc: there is another issue tracking this @alexeygorbunov let's not tackleit here. https://github.com/aristanetworks/avd/issues/6301

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/tenant-default-vrf-settings.cfg
laxmikantchintakindi: No.

python-avd/pyavd/_eos_designs/shared_utils/utils.py
gmuloc:
```suggestion
    def set_ipv4_address_family_underlay_bgp_peer_group(self: SharedUtilsProtocol, address_family_peer_group: EosCliConfigGen.RouterBgp.AddressFamilyIpv4.PeerGroupsItem) -> None:
```

python-avd/pyavd/_eos_designs/structured_config/underlay/router_bgp.py
```diff
@@ -34,18 +33,15 @@ def router_bgp(self: AvdStructuredConfigUnderlayProtocol) -> None:
         if not self.shared_utils.underlay_bgp:
             return
 
-        # TODO: Remove from here and do under each neighbor type.
-        self.set_once_peer_group_ipv4_underlay_peers()
+        # Set BGP peer group only when underlay link is present.
+        if not self._underlay_p2p_links:
+            return
+        # Adding the peer-group first time we add a neighbor interface.
```
gmuloc:
```suggestion
        # Adding the peer-group as we know we either have neighbors or neighbor_interfaces to configure
```

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-cleartext-passwords.cfg
gmuloc: @laxmikantchintakindi please add some l3edge for this device that will trigger the generation of the peer group as we need to keep testing for cleartext_password for this peer group. Thanks!

gmuloc (dismissed):

laxmikantchintakindi:
Added l3_edge to bgp-cleartext-passwords.

python-avd/pyavd/_eos_designs/shared_utils/utils.py
```diff
@@ -194,6 +194,23 @@ def underlay_vlan_trunk_groups(self: SharedUtilsProtocol) -> dict[int, set[str]]
                 break
         return vlans
 
+    def set_underlay_bgp_peer_group(self: SharedUtilsProtocol, peer_group: EosCliConfigGen.RouterBgp.PeerGroupsItem) -> None:
```
gmuloc: Actually @laxmikantchintakindi are you still using this? I don't think so right?

https://sonarcloud.io/code?id=aristanetworks_avd&pullRequest=5786&selected=aristanetworks_avd%3Apython-avd%2Fpyavd%2F_eos_designs%2Fshared_utils%2Futils.py&line=199

this is not covered by any test it seems

laxmikantchintakindi: Removed the code block from the file.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-peer-groups-2.cfg
```diff
@@ -87,11 +87,6 @@ router bgp 65001
    neighbor EVPN-OVERLAY-PEERS route-reflector-client
    neighbor EVPN-OVERLAY-PEERS send-community
    neighbor EVPN-OVERLAY-PEERS maximum-routes 0
-   neighbor IPv4-UNDERLAY-PEERS peer group
-   neighbor IPv4-UNDERLAY-PEERS bfd
-   neighbor IPv4-UNDERLAY-PEERS description Description for ipv4_underlay_peers via structured_config
```
ClausHolbechArista, line 99: are we not loosing coverage for this?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/core-4-multicast.cfg
ClausHolbechArista, line 1: all the changes here are very confusing. It makes it very hard to review a vital change like this. Do we really need to do the other changes to tests here?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/IPV6_NUMBERED_LEAF1a.cfg
ClausHolbechArista, line 1: changes that are not related to the purpose of this PR.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/l3_edge_bgp.cfg
ClausHolbechArista, line 1: changes that are not related to the purpose of this PR.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/core-4-multicast.yml
ClausHolbechArista, line 1: many changes here. Are we sure about these?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-peer-groups-2.cfg
laxmikantchintakindi: ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/IPV6_NUMBERED_LEAFS.yml
Added structured_config in this host.

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
ClausHolbechArista, line 62: notice you only configure the peer-group here if include-in-underlay-protocol, but you don't have that condition above where you add the peer-group.

ClausHolbechArista, line 48: since you have continue after this line, we cannot be sure we ever get to the line where you create the peer-group. I would prefer to call it together with the logic that adds the neighbor.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/host_vars/core-4-multicast.yml
```diff
@@ -1,2 +1,5 @@
 ---
 underlay_multicast_pim_sm: true
+
```
ClausHolbechArista: this is a test for core interfaces with multicast. Adding underlay_ipv6 here will mess that up since underlay_ipv6_numbered is not compatible with multicast.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/host_vars/l3_edge_bgp.yml
```diff
@@ -9,6 +9,12 @@ spine:
       loopback_ipv4_pool: 1.2.3.4/24
       bgp_as: 65000
 
+bgp_peer_groups:
+  ipv4_underlay_peers:
+    cleartext_password: arista
+    structured_config:
+      description: Test_for_ipv4_underlay_peers
+
```
ClausHolbechArista: test of bgp_peer_groups should be done under the bgp_peer_groups tests.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-peer-groups-2.cfg
ClausHolbechArista: I don't understand how IPV6_NUMBERED_LEAFS group var would affect the bgp-peer-groups-2 device

python-avd/pyavd/_eos_designs/structured_config/core_interfaces_and_l3_edge/router_bgp.py
laxmikantchintakindi: include_in_underlay_protocol is verified at the start of the loop.

ansible_collections/arista/avd/examples/cv-pathfinder/group_vars/TRANSPORTS.yml
```diff
@@ -80,6 +80,16 @@ spine:
...
+            peer_groups:
+              - name: IPv4-UNDERLAY-PEERS
+                activate: true
+          # Adding peer group.
+          peer_groups:
+            - name: IPv4-UNDERLAY-PEERS
+              metadata:
+                type: ipv4
+              send_community: all
+              maximum_routes: 12000
```
ClausHolbechArista: No longer needed since Guillaume removed the use of this on the peer.
```suggestion
```

ClausHolbechArista (changes requested):
one final cleanup item and we should be ready to merge

ClausHolbechArista (approved):

PR #5806, opened 2025-09-04
Feat(cv_deploy): Proxy support

### LINKED ISSUE #5378: Add web proxy support to pyavd
### Enhancement summary

pyavd currently lacks web proxy support, causing cv_deploy to fail in proxy-restricted environments.

PyAVD

### Use case example

A user runs pyavd in a corporate environment where all outbound traffic must go through an HTTP/HTTPS proxy. When cv_deploy tries to push configuration, the connection fails because pyavd does not respect the proxy settings. Adding proxy support would allow it to function correctly in such environments.

Enable pyavd to detect and use standard proxy settings (e.g. environment variables like HTTP_PROXY and HTTPS_PROXY).

_No response_

### Additional context

_No response_

### Contributing Guide

alexeygorbunov:
## Change Summary
Make `cv_deploy` (for both REST and gRPC) compatible with proxy tunneling using HTTP CONNECT method.

## Related Issue(s)

Fixes #5378

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
Introduce ability to support work through PLAIN HTTP Proxy server using CONNECT mode.

## How to test
See https://github.com/aristanetworks/avd-internal/issues/301

~TODO: Automated CI~

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -182,3 +374,13 @@ def __init__(
...
+        self._proxy_host = proxy_host
+        self._proxy_port = proxy_port
+        self._proxy_username = proxy_username
+        self._proxy_password = proxy_password
+        self._proxy_ssl = proxy_ssl
+        self._proxy_verify_certs = proxy_verify_certs
+        self._proxy_unix_socket_forwarder = proxy_unix_socket_forwarder
+        self._custom_ca_path = custom_ca_path
+        self._proxy_forwarder = None
+        self._proxy_socket_path = None
```
ClausHolbechArista: I think having a separate class to handle the low level connection part would be beneficial.

ClausHolbechArista: I guess most of these are just a parking spot for the config vars until we do _connect. So you could just do a small data class with ProxyForwarderConfig to hold them in one place and pass that to the ProxyForwarder on init.

alexeygorbunov:
I was not able to use the same socket towards Proxy where gRPC can negotiate SSL towards proxy endpoint first and then `unwrap` SSL to then reuse the same socket for SSL negotiatoin with public CV endpoint. Therefore current approach for encrypted traffic between gRPC client and proxy (is required when proxy needs Auth)  uses local unix-socket-based forwarder that handles socket towards proxy and SSL negotiation with it. gRPC then uses this unix-socket as a transport to hit CV and negotiate SSL with it.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -0,0 +1,318 @@
...
+
+if TYPE_CHECKING:
+    from requests import Session
+
+
+class ProxyForwarder:
+    """
+    Unix socket forwarder for proxy tunneling.
+
+    Creates a Unix domain socket server that forwards connections through an HTTP/HTTPS proxy
```
ClausHolbechArista: I am not super happy with this idea. It creates a Python layer for all communication. I will investigate a bit on alternatives.

ClausHolbechArista: This one is used by httpx and others. I think that looks much simpler and seems to work for what we need.
https://github.com/romis2012/python-socks?tab=readme-ov-file#async-asyncio

alexeygorbunov: Used pruxy-socks but was not able to achieve SSL-in-SSL where GPRC negotiates SSL over SSL_based proxy tunnel. Implemented plain HTTP CONNECT proxy (filtering will have to be done using SRC IP)

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Removed most of the attributes as i was not able to make ssl-in-ssl work for grpc

python-avd/pyavd/_cv/client/proxy.py
ClausHolbechArista: Same limitation exists for grpcio which everything else uses, so I think we can live with that. I don't follow the filtering point? Have not looked at the code yet.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -48,21 +52,48 @@ class CVClientProtocol(
...
+    async def _connect(self) -> None:
         # TODO: Verify connection
         # TODO: Handle multinode clusters
 
+        # Load custom CA if provided (for REST and gRPC endpoint verification)
+        if self._custom_ca_path:
+            self._temp_ca_bundle_path = create_ca_bundle_with_custom_ca(self._custom_ca_path)
+
+        # Initialize requests session for REST calls
+        self._session = Session()
```
ClausHolbechArista: no need for this any longer, since we only use if for the verions check, which can be done simpler with just a single request. The set_token one is not used outside of testing, so we can live with a seperate session there.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -72,10 +103,56 @@ def _connect(self) -> None:
...
+                )
+
+            except Exception as e:
+                msg = f"Failed to create proxy connection: {type(e).__name__}: {e}"
+                raise CVClientException(msg) from e
+
+            else:
+                return protocol
+
+        channel._create_connection = proxy_connection
```
ClausHolbechArista, line 123: We should probably hardcode our requirement for grpclib so any changes there won't break this.

```suggestion
        # Override the standard method from grpclib with our proxy variant.
        channel._create_connection = proxy_connection
```

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -84,14 +161,19 @@ def _ssl_context(self) -> ssl.SSLContext | bool:
         The return value (The default ssl context or True) will be passed to grpclib.
         Requests will pick it up from ssl lib itself.
         """
+        # Accepting SonarLint issue: We are purposely implementing no verification of certs.
```
ClausHolbechArista: We should not accept insecure stuff outside of the `if not self._verify_certs`.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @property
+    def proxy_url(self) -> str:
+        """
+        Generate proxy URL for python-socks.
+
+        Returns:
+            HTTP proxy URL for python-socks library.
+        """
+        return f"http://{self.proxy_host}:{self.proxy_port}"
```
ClausHolbechArista: We need to make sure to mention somehwere that the connection to the proxy is HTTP.
We also need to look at authentication here at some point.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -0,0 +1,108 @@
...
+            Raw socket connected to target through proxy.
+        """
+        # Create proxy using python-socks
+        proxy = Proxy.from_url(self.proxy_url)
+
+        # Connect through proxy to target
+        return await proxy.connect(dest_host=self.target_host, dest_port=self.target_port)
+
+
+def create_ca_bundle_with_custom_ca(custom_ca_path: str) -> str:
```
ClausHolbechArista: this is not related to proxy as such, so let's move it to another file

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -14,6 +14,9 @@ class CloudVision:
     servers: str | list[str]
     token: str
     verify_certs: bool = True
+    custom_ca_path: str | None = None
+    proxy_host: str | None = None
+    proxy_port: str | None = None
```
ClausHolbechArista:
```suggestion
    proxy_port: int | None = None
```

python-avd/pyproject.toml
```diff
@@ -26,6 +26,8 @@ dependencies = [
     "Jinja2>=3.0",
     "PyYAML>=6.0.0",
     "requests>=2.27.0",
+    "python-socks>=2.7.2",
```
ClausHolbechArista:
```suggestion
    "python-socks[asyncio]>=2.7.2",
```

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -72,10 +103,56 @@ def _connect(self) -> None:
...
+                    ssl=ssl_context,
+                    server_hostname=self._servers[0] if ssl_context else None,
+                )
+
+            except Exception as e:
+                msg = f"Failed to create proxy connection: {type(e).__name__}: {e}"
+                raise CVClientException(msg) from e
+
+            else:
+                return protocol
```
ClausHolbechArista:
```suggestion
            return protocol
```

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: As discussed - `filtering based on SCR IP` is in this case an approach to limit clients and their "permissions" when going through proxy.

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Ok, mirrored verify/proxies into get/post calls

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: * added comment
* hardcoded "grpclib==0.4.8" as pyavd direct dep

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Moved comment to the proper line and removed exception where it was not needed

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Moved it to __init__ as CVClientProtocol's method

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Changed to int with 8080 being default

python-avd/pyproject.toml
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Added support for basic authentication.
Working on the documentation

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -84,14 +146,19 @@
         The return value (The default ssl context or True) will be passed to grpclib.
         Requests will pick it up from ssl lib itself.
         """
+        context = ssl.create_default_context(purpose=ssl.Purpose.SERVER_AUTH)
```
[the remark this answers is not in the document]
ClausHolbechArista: @alexeygorbunov please address this.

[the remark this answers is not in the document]
alexeygorbunov: Added back original NOSONAR exception as this is done on purpose

alexeygorbunov:
End-to-end CI run: https://github.com/aristanetworks/avd-internal/actions/runs/17594495729/job/49983433581

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -147,6 +212,28 @@ def _set_version(self) -> None:
             msg = f"Unable to get version from CloudVision server. Got {response.text}"
             raise CVClientException(msg) from e
 
+    def create_ca_bundle_with_custom_ca(self) -> str:
```
ClausHolbechArista: There must be a simpler way than having to do all this. Also I don't like that we include certifi and combine with those. If we let a user define the ca path they should give a bundle of everything required.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -157,6 +244,11 @@ def __init__(
         password: str | None = None,
         port: int = 443,
         verify_certs: bool = True,
+        custom_ca_path: str | None = None,
+        proxy_host: str | None = None,
+        proxy_port: int = 8080,
+        proxy_username: str | None = None,
+        proxy_password: str | None = None,
```
ClausHolbechArista, line 222: you are not using these any longer are you? Instead you should add the _proxy_manager

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -0,0 +1,85 @@
...
+
+from typing import TYPE_CHECKING
+
+from python_socks.async_.asyncio import Proxy
+
+if TYPE_CHECKING:
+    import socket
+
+
+class HTTPProxyManager:
```
ClausHolbechArista, line 14: please add the attributes with type hints (no defaults) on the class

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -157,6 +244,11 @@ def __init__(
         password: str | None = None,
         port: int = 443,
         verify_certs: bool = True,
+        custom_ca_path: str | None = None,
```
ClausHolbechArista: you should not set a default value when you are always setting them in __init__.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -25,6 +25,11 @@ class CloudVision:
     username: str | None
     password: str | None
     verify_certs: bool = True
+    custom_ca_path: str | None = None
+    proxy_host: str | None = None
+    proxy_port: int | None = 8080
+    proxy_username: str | None = None
+    proxy_password: str | None = None
```
ClausHolbechArista: please avoid default values here, since we are always setting these anyway.

python-avd/pyavd/_cv/client/__init__.py
ClausHolbechArista: I think we should remove the custom ca bundle from this PR, so we can get the proxy support merged and then work out the CA stuff seperately. The users can still decide to trust their own CA on the OS level instead of asking us to trust it.

ClausHolbechArista (changes requested):

python-avd/pyavd/_cv/client/__init__.py
ClausHolbechArista: Removed it all. For now people can modify their ca bundles on the system.

ClausHolbechArista (approved):

alexeygorbunov:
Latest CI run including live pytest: https://github.com/aristanetworks/avd-internal/actions/runs/17867591636

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -72,10 +76,52 @@ def _connect(self) -> None:
...
+
+            try:
+                # Create socket through proxy using python-socks
+                proxy_sock = await self._proxy_manager.create_socket_for_grpc()
+
+                # Create the gRPC protocol using the proxy socket
+                _, protocol = await loop.create_connection(
+                    lambda: channel._protocol_factory(),
+                    sock=proxy_sock,
+                    ssl=channel._ssl if ssl_context is True else ssl_context,
```
ClausHolbechArista:
```suggestion
                    ssl=channel._ssl,
```

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -134,17 +181,19 @@ def _set_version(self) -> None:
...
                 "https://" + self._servers[0] + "/cvpservice/cvpInfo/getCvpInfo.do",
                 headers={"Authorization": f"Bearer {self._token}"},
                 verify=self._verify_certs,
+                proxies=self._proxy_manager.get_requests_proxies() if self._proxy_manager is not None else None,
                 json={},
             )
-
+            response.raise_for_status()
             self._cv_version = CvVersion(response.json()["version"])
         except (KeyError, JSONDecodeError) as e:
```
ClausHolbechArista, line 205: let us catch the error raised by raise_for_status above, so we can tell the user what happened.

ClausHolbechArista: same elsewhere

ClausHolbechArista: do try/except around response.raise_for_status for that case.
And then limit the other one to around the line 194

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Fixed

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Fixed/changed

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -41,6 +41,22 @@
...
+    type: int
+    default: 8080
+  cv_proxy_username:
+    description: Authentication username for the HTTP CONNECT proxy server.
+    type: str
+    required: false
+  cv_proxy_password:
+    description: Authentication password for the HTTP CONNECT proxy server. It is strongly recommended to use Vault for this.
+    type: str
+    required: false
```
ClausHolbechArista: I know it's late, but the cv_ variables concerns the cloudvision server. So these new ones should be `proxy_host`, `proxy_port`, `proxy_username`, `proxy_password`.

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -348,6 +347,49 @@ Click "Save" to exit the dialogue box.
...
+
+To enable the proxy, set `cv_proxy_host` (port `TCP/8080` will be used by default). If this variable is not defined, a proxy will not be used (default mode).
+
+!!! Warning
+
+    Authentication credentials (when used) are sent to the proxy server using ***HTTP Basic authentication*** over non-encrypted HTTP transport (credentials are only `Base64` encoded, not encrypted). Credentials can be exposed by intercepting and analyzing raw TCP/IP traffic between AVD and Proxy server.
+
+    Please use AVD proxy authentication only when absolutely necessary. Always use other filtering and identification mechanisms (like HTTP filtering based on the client's SRC IP, requested destination domains, etc.) to limit the security risks.
+
+    It is important to note that plain HTTP is used by AVD only for the initial CONNECT request to establish the tunnel with the CloudVision through proxy server. Once the TCP tunnel to CloudVision through proxy server is active, all subsequent AVD communication — including both REST and gRPC calls — is protected within a direct secure TLS session(s) established between AVD and CloudVision ***inside*** the TCP proxy tunnel.
```
```suggestion
    It is important to note that plain HTTP is used by AVD only for the initial CONNECT request to establish the tunnel with the CloudVision through proxy server. Once the TCP tunnel to CloudVision through proxy server is active, all subsequent AVD communication — including both REST and gRPC calls — is protected within a secure TLS session(s) established between AVD and CloudVision ***inside*** the TCP proxy tunnel.
```

python-avd/pyproject.toml
```diff
@@ -75,6 +78,7 @@ requires = [
     "cryptography>=38.0.4",
     "aristaproto>=0.1.1",
     "ruff==0.13.1",
+    "async-timeout>=5.0.1"
```
ClausHolbechArista: did we really need this?

ClausHolbechArista: if so it would be clearer to add `python-socks[asyncio]` as the dependency

ClausHolbechArista (changes requested):

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
alexeygorbunov: Sure, renamed all vars to strip `cv_`

ansible_collections/arista/avd/roles/cv_deploy/README.md

python-avd/pyproject.toml
alexeygorbunov: This was indeed not needed anymore. Fixed.

carl-baillargeon (approved):
LGTM! Tests are very nice 💯

ClausHolbechArista (approved):

PR #5840, opened 2025-09-10
Cut(eos_designs)!: Removed the deprecated keys from underlay_ospf_authentication

MaheshGSLAB:
## Change Summary

Removed the deprecated keys from underlay_ospf_authentication

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Removed the deprecated keys from underlay_ospf_authentication

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist

- [ x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)


Shivani-gslab (approved):
LGTM

Vibhu-gslab (approved):

ansible_collections/arista/avd/molecule/eos_designs_deprecated_vars/intended/configs/host1.cfg
```diff
@@ -95,25 +95,22 @@ interface Ethernet5
    no shutdown
    mtu 9214
    no switchport
-   ip ospf network point-to-point
-   ip ospf area 0.0.0.0
+   ip address 192.168.0.18/31
```
gmuloc, line 98: discussed with @MaheshGSLAB 

this change appeared because when removing OPSF, then the protocol became BGP and in BGP we need IPs to be able to configure the neighbors.

2 things:

1/ we should raise when referring a non existing pool (which was not the case here)
2/ when configuring OSPF we should make sure there is an ip address on the link like we do for BGP.

I will approve this PR but we should have a follow-up issue for validations

docs/porting-guides/6.x.x.md
```diff
@@ -57,9 +57,24 @@ The following data model keys have been removed from `eos_designs` in AVD 6.0.0.
...
+As of AVD 6.0.0, the `cleartext_key` key is required and must be set to generate the configuration and documentation for `underlay_ospf_authentication.message_digest_keys[]`.
+
+```diff
+underlay_ospf_authentication:
+  enabled: true
+  message_digest_keys:
+    - id: 1
++     cleartext_key: arista123arista
+```
+
```
gmuloc: technically we had alreadya requirement of either key or cleartext_key so I think we don't need this entry. As the error on the removed key table is enough 

```suggestion
```

gmuloc (approved):

PR #5842, opened 2025-09-11
Cut(eos_designs)!: Removed the deprecated key overlay_loopback_description

MaheshGSLAB:
## Change Summary

Removed the deprecated key overlay_loopback_description

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Removed the deprecated key overlay_loopback_description

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


Shivani-gslab (commented):
since we are removing this key, we can update the names here as well - 

https://github.com/aristanetworks/avd/blob/devel/ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2

api/interface_descriptions - 
https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/api/interface_descriptions/__init__.py#L478

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
```diff
@@ -3,13 +3,11 @@
...
 #}
-{% if overlay_loopback_description is not arista.avd.defined %}
-{%     if switch.mpls_overlay_role in ["server", "client"] %}
-{%         set overlay_loopback_description = "MPLS_Overlay_peering" %}
-{%     elif switch.mpls_lsr is arista.avd.defined(true) %}
-{%         set overlay_loopback_description = "LSR_Router_ID" %}
-{%     else %}
-{%         set overlay_loopback_description = "EVPN_Overlay_Peering" %}
-{%     endif %}
+{% if switch.mpls_overlay_role in ["server", "client"] %}
```
Shivani-gslab: we should check for `router_id_loopback_description` here, as it is the new key for `overlay_loopback_description`

Shivani-gslab: also we can change the variable name

MaheshGSLAB: I’m not sure if we really need to check this, since it should already be handled and we can simply remove the deprecated keys. However, in this case we only see the deprecated key, so let’s confirm.
@ClausHolbechArista @gmuloc

gmuloc: you should change it as Shivani described @MaheshGSLAB if you look at the previous implementation it was a bit hacked like this:

```
            return self._template(template_path, overlay_loopback_description=data.description, router_id_loopback_description=data.description)
```

we were passing to the template the same value for old and new key. so it was transparent for the users of the new key because it was in the end put inside the `overlay_loopback_description` for the template but now we should be passing only the new one and so the template should be adjusted with this

gmuloc (changes requested):

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
alexeygorbunov: @gmuloc , when (under which conditions) is this template used? Can't find where we engage this default template to render loop0 description..

MaheshGSLAB: updated the template

Shivani-gslab (approved):
LGTM

ansible_collections/arista/avd/roles/eos_designs/templates/interface_descriptions/loopback_interfaces/overlay-loopback.j2
gmuloc: @alexeygorbunov so by default they are not today. Sadly people may be running with old node_Type_keys setting targeting these templates which we have not deprecated.

The templates were the default before 5.x and we could have people still using them for whatever reason.

Let's see if we remove the default templates all together from the role as part of 6.0.0 given that now we have precedence for 
better description formats

gmuloc (approved):
LGTM

PR #5865, opened 2025-09-16
CI: Add required-checks for merge-queue

ClausHolbechArista:
Since github settings requires us to maintain _each_ job as a required status check, this PR instead adds an extra job which depends on all the others and requires either success or skip of all of them. Then github settings can just check for this one job. Easier to maintain and track.


Shivani-gslab (approved):
LGTM

gmuloc (approved):

PR #5887, opened 2025-09-22
Doc: Update schema generated default values to avoid sorting keys.

ClausHolbechArista:
Update schema generated default values to avoid sorting keys.

When the tooling generates default values for dicts or lists it uses a yaml dump function, which by default sorts the keys. This skews the default values in the popups behind the (+) buttons.

This PR removes the sort, so we rely on the order of default values from the schema.


Shivani-gslab (approved):
LGTM

MaheshGSLAB (approved):

PR #5894, opened 2025-09-22
CI: Fix regex for insert-license in pre-commit

ClausHolbechArista:
The regex used for insert-license in pre-commit.yaml was not correct after we moved the schemas to pyavd. This PR updates the regex with a verbose regex (?x) and updates all existing schemas.


carl-baillargeon (approved):

carlbuchmann (approved):
LGTM

PR #5888, opened 2025-09-22
Fix(eos_designs): Avoid restricting valid user tags in generate_cv_tags.device_tags

MaheshGSLAB:
## Change Summary

Fix the invalid customer generate_cv_tags.device_tags

## Related Issue(s)

Fixes #5879 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
I am trying to set tags to devices in cvp with eos_designs and cv_deploy and have defined the following tags.

```
# eos_designs automatically sets topology tag in cvp
generate_cv_tags:
  topology_hints: true
  device_tags:
    - name: DC
      data_path: metadata.dc_name
    - name: DC-Pod
      data_path: metadata.pod_name
    - name: Role
      data_path: type
```

I get the follwing error message, when executing the playbook:

```
fatal: [de-dcp-e-dcn-cs-208-11 -> localhost]: FAILED! => {"changed": false, "msg": "The CloudVision tag name 'generate_cv_tags.device_tags[name=Role] is invalid. System Tags cannot be overridden. Try using a different name for this tag."}
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: pyavd._errors.AristaAvdError: The CloudVision tag name 'generate_cv_tags.device_tags[name=Role] is invalid. System Tags cannot be overridden. Try using a different name for this tag.
```

It claims that the tag `Role`is a fixed system tag in CVP.

But I cannot find a tag named "Role" in CVP under system tags.

So updated the INVALID_CUSTOM_DEVICE_TAGS in code.

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/molecule) testing accordingly. (check the box if not applicable)

PR #5899, opened 2025-09-23
Fix(eos_designs): Avoid restricting valid user tags in generate_cv_tags.device_tags (#5888)

gmuloc:
Cherry-pick of https://github.com/aristanetworks/avd/pull/5888


gmuloc:
wrong manipulation so it seems merged but it is not

PR #5902, opened 2025-09-23
Refactor: Correct AVD meta schema location

carlbuchmann:
## Change Summary

Update schema fragments with the correct meta schema location so we have LSP support when authoring.


gmuloc (approved):
LGTM - were these the only one wrong or missing?

MaheshGSLAB (approved):
LGTM, we should catch this via pre-commit i believe.

ClausHolbechArista (approved):

PR #5933, opened 2025-09-30
CI: Fixing PyRight issue in github action

Vibhu-gslab:
## Change Summary

Fixing PyRight in GitHub action

> [!WARNING]
> ~Pending in #5939 and #5941 to do the right thing :)~
> DONE ✅


.github/workflows/pull-request-management.yml
```diff
@@ -157,6 +157,10 @@ jobs:
     name: Python Linting not covered in pre-commit
     runs-on: ubuntu-latest
     steps:
+      - uses: actions/checkout@v5
+      - name: Install pyavd and Ansible requirements
+        run: |
+          pip install "cryptography" -r ansible_collections/arista/avd/requirements-dev.txt --upgrade
```
gmuloc: cryptography should be installed with dev requirements right?

gmuloc: @ClausHolbechArista for some reason, CI decides to install crypto version 41
and pyright is not able to detect that we catch the ImportError in pyavd._utils.password_utils.password_utils.py and reports it.

for now I told Vibhu to add a min version for crypto in our requirements to make pyright happy

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -19,6 +19,93 @@
 if TYPE_CHECKING:
     from . import SharedUtilsProtocol
 
+# TODO: This should be updated every time the schema is updated.
+# Right now it has to be manually updated.
+SPEED_TYPE = Literal[
```
ClausHolbechArista: I do not want to maintain this here. We need to do this in a better way.

.github/workflows/pull-request-management.yml
gmuloc: @Vibhu-gslab I will create a PR to make cryptography an official requirement of pyavd with min version 43.0.0

python-avd/pyavd/_eos_designs/shared_utils/misc.py
gmuloc: indeed thanks to #5941 we are now in wonderland 🎩 🃏

.github/workflows/pull-request-management.yml
gmuloc: done in #5939

gmuloc (approved):
LGTM if CI passes

ClausHolbechArista (approved):

PR #5941, opened 2025-10-01
Refactor(pyavd): Create type aliases for literal in schema classes

ClausHolbechArista:
Create type aliases for literal in schema classes so we can use them as type hints


python-avd/pyavd/_eos_cli_config_gen/schema/__init__.py
```diff
@@ -127,15 +127,13 @@ class Methods(AvdList[MethodsItem]):
                 Methods._item_type = MethodsItem
 
                 _fields: ClassVar[dict] = {"type": {"type": str}, "methods": {"type": Methods}}
-                type: Literal["none", "start-stop", "stop-only"]
+                type: Type
```
gmuloc, line 130: we may one day have class with something else maybe.
We could try prefixing / suffixing with `T_` / `_T` to be safe (and not go mad :) )

ClausHolbechArista: ? Type is a valid class name. It is not the same as `type`

ClausHolbechArista: Note this is only applicable insite the deeply tested namespace.

gmuloc: what if this shadows `Literal` ? or some other class - changing the meaning inside the class

gmuloc: for instance
```
+  literal:
+    type: str
+    valid_values: ["break-stuff"]

```

generates

```
     WanRouteServers._item_type = WanRouteServersItem

+    Literal: TypeAlias = Literal["break-stuff"]
+

```

gmuloc: Screenshot 2025-10-01 at 14 43 56

gmuloc: Discussed on dev call with @ClausHolbechArista - we acknowledge this is a corner case that could occur but for now we shall live with it - it has never bitten us for the nested class names so far.

#YOLOPS

gmuloc:
Coverage is not reported to Sonar hence the warning - we will accept this for this PR and investigate in a separate one.

gmuloc (approved):

PR #5948, opened 2025-10-01
Doc: fix dual-dc layer3 edge

joelbreton2:
## Change Summary

Fix doc for l3 edge range


carlbuchmann (approved):
LGTM.

This was not fixed in devel as the values have been updated to 6.0.
In the future, I discussed with @joelbreton2 the possibility of splitting the files and avoiding line ranges to prevent these issues in the future. This will be addressed going forward.

JulioPDX (approved):
LGTM

ClausHolbechArista (changes requested):
Should this not be fixed in devel as well? We should not do stuff in the release branch unless it is also done in devel, and if so it is best to cherry-pick for tracking.

ClausHolbechArista (approved):

PR #5965, opened 2025-10-03
Feat(eos_designs): Introduce as-path-based route filtering for evpn_prevent_readvertise_to_server

### LINKED ISSUE #5962: Feat(eos_designs): Introduce as-path-based route filtering for evpn_prevent_readvertise_to_server
### Enhancement summary

Introduce an ability to match and filter out EVPN routes based on the as-path ACL.
Based on SR#655319.

eos_designs

### Use case example

Current implementation uses a route-map which matches routes in BGP table based on the ASN of the peer which advertised this BGP route to the local BGP router. In some cases (when source peer is not in the same ASN that we want to advertise this route to, like iBGP peer or eBGP in another ASN) this may lead to such routes still being advertised to the remote peer although remote peer's ASN is in the AS path list. 

Remote peer will still discard those routes due to the native loop avoidance eBGP logic but this will consume compute resources (impact will depend on the frequency of such advertisements and the number of the advertised routes)

```
route-map RM-EVPN-FILTER-AS65001 deny 10
   match as 65001 <<<< this matches an ASN of the peer we learned the route from
```

Introduce new knob which would use as-path ACL (matching for the presence of the target peer's ASN in the AS PATH path attribute instead of checking source's ASN)

```
ip as-path access-list AS65001 permit _65001_ any
!
route-map RM-EVPN-FILTER-AS65001 deny 10
   match as-path AS65001
```

None

### Additional context

N/A

### Contributing Guide

alexeygorbunov:
## Change Summary

Introduce an ability to match and filter out EVPN routes based on the as-path ACL.

## Related Issue(s)

Fixes #5962 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
- Add new key `evpn_prevent_readvertise_to_server_mode` to control method of identifying "looping" routes:
  - Default mode is `source_peer_asn` (current approach matching on ASN of the source peer).
  - Mode `as_path_acl` created `as-path` ACL to check for the presence of EVPN peer's (route-server) ASN inside route's AS PATH attribute.

## How to test
Molecule test `eos_designs_unit_tests` (new host `evpn-prevent-readvertise-to-server-6`)

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    @structured_config_contributor
+    def as_path(self: AvdStructuredConfigOverlayProtocol) -> None:
+        """Set the structured config for as_path."""
+        if self.shared_utils.overlay_routing_protocol != "ebgp":
+            return
+
+        if self.inputs.evpn_prevent_readvertise_to_server and self.inputs.evpn_prevent_readvertise_to_server_mode == "as_path_acl":
+            remote_asns = natural_sort({rs_dict.get("bgp_as") for rs_dict in self._evpn_route_servers.values()})
```
gmuloc: I don't think this is possible but could this possibly end up with a `None` value in the list?

gmuloc: I guess maybe if the route_server is running OSPF and there is no bgp_as configured for it at all but it should fail somewhere else (maybe)

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
```diff
@@ -26,13 +26,23 @@ def route_maps(self: AvdStructuredConfigOverlayProtocol) -> None:
         if self.shared_utils.overlay_cvx:
             return
 
-        if self.shared_utils.overlay_routing_protocol == "ebgp" and self.inputs.evpn_prevent_readvertise_to_server:
+        if (
+            self.shared_utils.overlay_routing_protocol == "ebgp"
+            and self.inputs.evpn_prevent_readvertise_to_server
+            and self.inputs.evpn_prevent_readvertise_to_server_mode in ["source_peer_asn", "as_path_acl"]
```
gmuloc, line 33: why do you need this line? the schema should guarantee it is in these two values (we usually don't add extra checks for this).


```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
```diff
@@ -26,13 +26,23 @@ def route_maps(self: AvdStructuredConfigOverlayProtocol) -> None:
...
+        ):
             remote_asns = natural_sort({rs_dict.get("bgp_as") for rs_dict in self._evpn_route_servers.values()})
             for remote_asn in remote_asns:
                 route_maps_item = EosCliConfigGen.RouteMapsItem(name=f"RM-EVPN-FILTER-AS{remote_asn}")
-                route_maps_item.sequence_numbers.append_new(
-                    sequence=10, type="deny", match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match([f"as {remote_asn}"])
-                )
+                match self.inputs.evpn_prevent_readvertise_to_server_mode:
```
gmuloc: let's try to move the match out of the for loop here.

Suggestion:


```
match self.inputs.evpn_prevent_readvertise_to_server_mode:
    case "source_peer_asn":
        match_prefix = "as "
    case "as_path_acl":
        match_prefix = "as-path AS"
for remote_asn in remote_asns:
    route_maps_item = EosCliConfigGen.RouteMapsItem(name=f"RM-EVPN-FILTER-AS{remote_asn}")
    route_maps_item.sequence_numbers.append_new(
      sequence=10, type="deny",
      match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match([f"{match_prefix}{remote_asn}"])
    )
```

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
alexeygorbunov: RCF and peer-tag approaches will be added later and those will not need route-maps. So here i'm just liming `route-map`-based solution to 2 options only

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
alexeygorbunov: Not sure. We check first that eBGP is used as an overlay and only then look for the ASNs.
Added the check for both route-maps and as-path ACLs to make sure `None`s are skipped if present

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
alexeygorbunov: Thanks! Used suggested approach.

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
```diff
@@ -0,0 +1,34 @@
...
+
+    @structured_config_contributor
+    def as_path(self: AvdStructuredConfigOverlayProtocol) -> None:
+        """Set the structured config for as_path."""
+        if self.shared_utils.overlay_routing_protocol != "ebgp":
+            return
+
+        if self.inputs.evpn_prevent_readvertise_to_server and self.inputs.evpn_prevent_readvertise_to_server_mode == "as_path_acl":
+            remote_asns = natural_sort({bgp_as for rs_dict in self._evpn_route_servers.values() if (bgp_as := rs_dict.get("bgp_as")) is not None})
+            for remote_asn in remote_asns:
```
gmuloc, line 32: could we end up with duplicate entries if two route servers have the same AS?

gmuloc: hmm they will be overwritten and the same - may remote_asns could be a `set` to avoid useless egneration but that won't impact the output really

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
gmuloc: I see - usually we would rather introduce this in the PR with the new options to make it clear why the line is introduced. I would add a comment stating exactly what you just said that the route-maps are used only for these two options

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -455,7 +455,9 @@ def _set_neighbors(self: AvdStructuredConfigOverlayProtocol) -> None:
                     overlay_peering_interface=data.get("overlay_peering_interface"),
                 )
                 if self.inputs.evpn_prevent_readvertise_to_server:
-                    neighbor.route_map_out = f"RM-EVPN-FILTER-AS{data['bgp_as']}"
+                    match self.inputs.evpn_prevent_readvertise_to_server_mode:
+                        case "source_peer_asn" | "as_path_acl":
```
gmuloc: I think later pyright would complain taht you are not covering all the options in this place for instance (you can auto close this comment :) )

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
alexeygorbunov: We are using set comprehension so we should not end up with any duplicated items here.
I updated the molecule to include 3rd route server so that 2 of the RSs are in the same AS

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
alexeygorbunov: Added comment to clarify

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
alexeygorbunov: Next PR will cover the remaining cases

```python
                    match self.inputs.evpn_prevent_readvertise_to_server_mode:
                        case "source_peer_asn" | "as_path_acl":
                            neighbor.route_map_out = f"RM-EVPN-FILTER-AS{data['bgp_as']}"
                        case "peer_tag":
                            neighbor.peer_tag_in = f"EVPN-PEER-TAG-AS{data['bgp_as']}"
                            neighbor.peer_tag_out_discard = f"EVPN-PEER-TAG-AS{data['bgp_as']}"
                        case "rcf":
                            self.structured_config.router_bgp.address_family_evpn.neighbors.append_new(
                                ip_address=data["ip_address"], rcf_out=f"EVPN-FILTER-AS{data['bgp_as']}()"
                            )
```

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
gmuloc: duh - I missed these! thanks

gmuloc (approved):
LGTM thanks!

gmuloc (approved):
still LGTM ;)

ClausHolbechArista:
This one will have some overlap with the dot-notation PR, since the syntax for the as-path lists would differ and you need to escape the dot.
See:
https://www.arista.com/en/support/toi/eos-4-21-1f/14079-asn-mode-regular-expressions-for-bgp-as-path-attributes

The default style is asn so we don't need to change this config (`ip as-path regex-mode asn`), but maybe we should set it just to be sure it is what we expect.

alexeygorbunov:
> Test: Moleculing the molecule

I did few additional tests to confirm how `as-path acl`s match `AS_PATH` under different combinations of `regex-mode` and `asn notation`:

1. `ip as-path regex-mode: asn` && `bgp asn notation: asdot`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` |
| - | - | - |
| `100.100 ?` | Y | Y |
| `65100 100.100 ?` | Y | Y | 
| `100.100 65100 ?` | Y | Y |
| `65100 100.100 65200 ?` | Y | Y | 

2. `ip as-path regex-mode: asn` && `bgp asn notation: asplain`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` |
| - | - | - |
| `6553700 ?` | Y | Y |
| `65100 6553700 ?` | Y | Y | 
| `6553700 65100 ?` | Y | Y |
| `65100 6553700 65200 ?` | Y | Y | 

3. `ip as-path regex-mode: string` && `bgp asn notation: asdot`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` |
| - | - | - |
| `100.100 ?` | N | Y |
| `65100 100.100 ?` | N | Y | 
| `100.100 65100 ?` | N | Y |
| `65100 100.100 65200 ?` | N | Y |

4. `ip as-path regex-mode: string` && `bgp asn notation: asplain`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` |
| - | - | - |
| `6553700 ?` | Y | N |
| `65100 6553700 ?` | Y | N | 
| `6553700 65100 ?` | Y | N |
| `65100 6553700 65200 ?` | Y | N | 


Results:
1. `permit _100\.100_ any` works for `bgp asn notation: asdot` with both `ip as-path regex-mode: asn` and `ip as-path regex-mode: string`
2. `permit _6553700_ any` works for `bgp asn notation: asplain` with both `ip as-path regex-mode: asn` and `ip as-path regex-mode: string`


Therefore if we use `permit _6553700_ any` format when notation is PLAIN and `permit _100\.100_ any` format when notation is DOTTED - proper routes will be matched no matter what `ip as-path regex-mode` is used.

I added an extra line to escape `.` in AS_PATH if present:
```diff
            for remote_asn in remote_asns:
                entries = EosCliConfigGen.AsPath.AccessListsItem.Entries()
+                escaped_remote_asn = remote_asn.replace(".", r"\.")
~                entries.append_new(type="permit", match=rf"_{escaped_remote_asn}_")
                self.structured_config.as_path.access_lists.append_new(name=f"AS{remote_asn}", entries=entries)
```

If after merging PR 5867 (ASDOTTED) returned `remote_asn` is not already adjusted to the required format - i'll need to add an extra step to adjust it's format. Otherwise nothing extra will be needed.

@ClausHolbechArista @gmuloc

gmuloc:
> > Test: Moleculing the molecule
> 
> I did few additional tests to confirm how `as-path acl`s match `AS_PATH` under different combinations of `regex-mode` and `asn notation`:
> 
> 1. `ip as-path regex-mode: asn` && `bgp asn notation: asdot`
> 
> AS_PATH	`permit _6553700_ any`	`permit _100\.100_ any`
> `100.100 ?`	Y	Y
> `65100 100.100 ?`	Y	Y
> `100.100 65100 ?`	Y	Y
> `65100 100.100 65200 ?`	Y	Y
> 2. `ip as-path regex-mode: asn` && `bgp asn notation: asplain`
> 
> AS_PATH	`permit _6553700_ any`	`permit _100\.100_ any`
> `6553700 ?`	Y	Y
> `65100 6553700 ?`	Y	Y
> `6553700 65100 ?`	Y	Y
> `65100 6553700 65200 ?`	Y	Y
> 3. `ip as-path regex-mode: string` && `bgp asn notation: asdot`
> 
> AS_PATH	`permit _6553700_ any`	`permit _100\.100_ any`
> `100.100 ?`	N	Y
> `65100 100.100 ?`	N	Y
> `100.100 65100 ?`	N	Y
> `65100 100.100 65200 ?`	N	Y
> 4. `ip as-path regex-mode: string` && `bgp asn notation: asplain`
> 
> AS_PATH	`permit _6553700_ any`	`permit _100\.100_ any`
> `6553700 ?`	Y	N
> `65100 6553700 ?`	Y	N
> `6553700 65100 ?`	Y	N
> `65100 6553700 65200 ?`	Y	N
> Results:
> 
> 1. `permit _100\.100_ any` works for `bgp asn notation: asdot` with both `ip as-path regex-mode: asn` and `ip as-path regex-mode: string`
> 2. `permit _6553700_ any` works for `bgp asn notation: asplain` with both `ip as-path regex-mode: asn` and `ip as-path regex-mode: string`
> 
> Therefore if we use `permit _6553700_ any` format when notation is PLAIN and `permit _100\.100_ any` format when notation is DOTTED - proper routes will be matched no matter what `ip as-path regex-mode` is used.
> 
> I added an extra line to escape `.` in AS_PATH if present:
> 
> ```diff
>             for remote_asn in remote_asns:
>                 entries = EosCliConfigGen.AsPath.AccessListsItem.Entries()
> +                escaped_remote_asn = remote_asn.replace(".", r"\.")
> ~                entries.append_new(type="permit", match=rf"_{escaped_remote_asn}_")
>                 self.structured_config.as_path.access_lists.append_new(name=f"AS{remote_asn}", entries=entries)
> ```
> 
> If after merging PR 5867 (ASDOTTED) returned `remote_asn` is not already adjusted to the required format - i'll need to add an extra step to adjust it's format. Otherwise nothing extra will be needed.
> 
> @ClausHolbechArista @gmuloc

I think you just need to use the function from @ClausHolbechArista introduced in #6016 which was already merged right?

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2023-2025 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Protocol
+
+from pyavd._eos_cli_config_gen.schema import EosCliConfigGen
+from pyavd._eos_designs.structured_config.structured_config_generator import structured_config_contributor
+from pyavd.j2filters import natural_sort
```
gmuloc, line 11:
```suggestion
from pyavd.j2filters import natural_sort
from pyavd._utils import as_path_list_match_from_bgp_asns
```

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
```diff
@@ -0,0 +1,35 @@
...
+        """Set the structured config for as_path."""
+        if self.shared_utils.overlay_routing_protocol != "ebgp":
+            return
+
+        if self.inputs.evpn_prevent_readvertise_to_server and self.inputs.evpn_prevent_readvertise_to_server_mode == "as_path_acl":
+            remote_asns = natural_sort({bgp_as for rs_dict in self._evpn_route_servers.values() if (bgp_as := rs_dict.get("bgp_as")) is not None})
+            for remote_asn in remote_asns:
+                entries = EosCliConfigGen.AsPath.AccessListsItem.Entries()
+                escaped_remote_asn = remote_asn.replace(".", r"\.")
+                entries.append_new(type="permit", match=rf"_{escaped_remote_asn}_")
```
gmuloc:
```suggestion
                entries.append_new(type="permit", match=rf"_{as_path_list_match_from_bgp_asns([remote_asn])}_")
```

gmuloc (commented):
added the changes to use #6016 method

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
ClausHolbechArista: this is wrong. I am already adding the _ in the function. And it should not be needed to have the _ at the start and end should it?

ClausHolbechArista: @alexeygorbunov did you test without the leading and trailing underscores? According to the TOI it should work fine without them

gmuloc: you are not adding `_` if there is one element. so it is not wrong to maintain what Alexey was generating

ClausHolbechArista:
```suggestion
                entries.append_new(type="permit", match=as_path_list_match_from_bgp_asns([remote_asn]))
```

alexeygorbunov: Extended and re-ran tests (used 4.33.0F).
Summary:
 - wrapping with `_`s or not will have no impact if `ip as-path regex-mode asn`
 - wrapping with `_`s is required when using `ip as-path regex-mode string`. Otherwise such pattern will match additional/incorrect AS_PATHs
 
 Details:
 
## `ip as-path regex-mode: asn` && `bgp asn notation: asdot`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` | `permit 6553700 any` | `permit 100\.100 any` |
| - | - | - | -| - |
| `100.100 ?` | Y | Y | Y | Y |
| `65100 100.100 ?` | Y | Y | Y | Y |
| `100.100 65100 ?` | Y | Y | Y | Y |
| `65100 100.100 65200 ?` | Y | Y | Y | Y |
| `2525.58601 ?` |  |  |  |  |
| `65100 2525.58601 ?` |  |  |  |  |
| `2525.58601 65100 ?` |  |  |  |  |
| `65100 2525.58601 65200 ?` |  |  |  |  |
| `1100.1001 ?` |  |  |  |  |
| `65100 1100.1001 ?` |  |  |  |  |
| `1100.1001 65100 ?` |  |  |  |  |
| `65100 1100.1001 65200 ?` |  |  |  |  |


## `ip as-path regex-mode: asn` && `bgp asn notation: asplain`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` | `permit 6553700 any` | `permit 100\.100 any` |
| - | - | - | -| - |
| `6553700 ?` | Y | Y | Y | Y |
| `65100 6553700 ?` | Y | Y | Y | Y |
| `6553700 65100 ?` | Y | Y | Y | Y |
| `65100 6553700 65200 ?` | Y | Y | Y | Y |
| `165537001 ?` |  |  |  |  |
| `65100 165537001 ?` |  |  |  |  |
| `165537001 65100 ?` |  |  |  |  |
| `65100 165537001 65200 ?` |  |  |  |  |
| `72090601 ?` |  |  |  |  |
| `65100 72090601 ?` |  |  |  |  |
| `72090601 65100 ?` |  |  |  |  |
| `65100 72090601 65200 ?` |  |  |  |  |


## `ip as-path regex-mode: string` && `bgp asn notation: asdot`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` | `permit 6553700 any` | `permit 100\.100 any` |
| - | - | - | -| - |
| `100.100 ?` |  | Y |  | Y |
| `65100 100.100 ?` |  | Y |  | Y |
| `100.100 65100 ?` |  | Y |  | Y |
| `65100 100.100 65200 ?` |  | Y |  | Y |
| `2525.58601 ?` |  |  |  |  |
| `65100 2525.58601 ?` |  |  |  |  |
| `2525.58601 65100 ?` |  |  |  |  |
| `65100 2525.58601 65200 ?` |  |  |  |  |
| `1100.1001 ?` |  |  |  | Y |
| `65100 1100.1001 ?` |  |  |  | Y |
| `1100.1001 65100 ?` |  |  |  | Y |
| `65100 1100.1001 65200 ?` |  |  |  | Y |


## `ip as-path regex-mode: string` && `bgp asn notation: asplain`

| AS_PATH | `permit _6553700_ any` | `permit _100\.100_ any` | `permit 6553700 any` | `permit 100\.100 any` |
| - | - | - | -| - |
| `6553700 ?` | Y |  | Y |  |
| `65100 6553700 ?` | Y |  | Y |  |
| `6553700 65100 ?` | Y |  | Y |  |
| `65100 6553700 65200 ?` | Y |  | Y |  |
| `165537001 ?` |  |  | Y |  |
| `65100 165537001 ?` |  |  | Y |  |
| `165537001 65100 ?` |  |  | Y |  |
| `65100 165537001 65200 ?` |  |  | Y |  |
| `72090601 ?` |  |  |  |  |
| `65100 72090601 ?` |  |  |  |  |
| `72090601 65100 ?` |  |  |  |  |
| `65100 72090601 65200 ?` |  |  |  |  |

@ClausHolbechArista @gmuloc

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
alexeygorbunov: Applied suggestion

python-avd/pyavd/_eos_designs/structured_config/overlay/as_path.py
alexeygorbunov: Applied suggested solution (with `_`s)

gmuloc: thanks @alexeygorbunov for the thorough testing. As agreed yesterday lets get this one in with `_` and cherry-pick it to 5.7.x.

Then lets add a follow up one we won't cherry=pick for the as-path variable for WAN HA in underlay so that we do the same thing (add `_`) in both places

gmuloc (approved):
LGTM

gmuloc:
> ## [Quality Gate Passed](https://sonarcloud.io/dashboard?id=aristanetworks_avd&pullRequest=5965) **Quality Gate passed**
> Issues  [0 New issues](https://sonarcloud.io/project/issues?id=aristanetworks_avd&pullRequest=5965&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=aristanetworks_avd&pullRequest=5965&issueStatuses=ACCEPTED)
> 
> Measures  [0 Security Hotspots](https://sonarcloud.io/project/security_hotspots?id=aristanetworks_avd&pullRequest=5965&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [95.1% Coverage on New Code](https://sonarcloud.io/component_measures?id=aristanetworks_avd&pullRequest=5965&metric=new_coverage&view=list)  [0.0% Duplication on New Code](https://sonarcloud.io/component_measures?id=aristanetworks_avd&pullRequest=5965&metric=new_duplicated_lines_density&view=list)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=aristanetworks_avd&pullRequest=5965)

the coverage missed lines are non exhaustivity in case checks. I think this is because coverage does not understand the type hints and so it cannot know we are looking at all the options.

we can either:
1/ use pragam no-cover for branch to tell the tool we know we are done
2/ replace the last case with `case _` (and some comment)

some context 🧟  https://github.com/pytest-dev/pytest-cov/issues/533 

Thoughts?

ClausHolbechArista:
> the coverage missed lines are non exhaustivity in case checks. I think this is because coverage does not understand the type hints and so it cannot know we are looking at all the options.
> 
> we can either: 1/ use pragam no-cover for branch to tell the tool we know we are done 2/ replace the last case with `case _` (and some comment)
> 
> some context 🧟 [pytest-dev/pytest-cov#533](https://github.com/pytest-dev/pytest-cov/issues/533)
> 
> Thoughts?

IMO we should clean up those unneeded/incomplete match/case. Remember this PR will be cherry picked and live in 5.7 forever.

gmuloc (changes requested):
lets fix coverage

python-avd/pyavd/_eos_designs/structured_config/overlay/route_maps.py
```diff
@@ -26,12 +26,19 @@ def route_maps(self: AvdStructuredConfigOverlayProtocol) -> None:
...
+            and self.inputs.evpn_prevent_readvertise_to_server
+            # Limit the usage of the route-maps for route filtering towards RSs to source_peer_asn and as_path_acl modes.
+            and self.inputs.evpn_prevent_readvertise_to_server_mode in ["source_peer_asn", "as_path_acl"]
+        ):
+            match_prefix = "as " if self.inputs.evpn_prevent_readvertise_to_server_mode == "source_peer_asn" else "as-path AS"
+
+            remote_asns = natural_sort({bgp_as for rs_dict in self._evpn_route_servers.values() if (bgp_as := rs_dict.get("bgp_as")) is not None})
             for remote_asn in remote_asns:
-                route_maps_item = EosCliConfigGen.RouteMapsItem(name=f"RM-EVPN-FILTER-AS{self.shared_utils.get_asn(remote_asn)}")
+                route_maps_item = EosCliConfigGen.RouteMapsItem(name=f"RM-EVPN-FILTER-AS{remote_asn}")
```
alexeygorbunov, line 39: `remote_asn` already follows desired asdot notation format. Extra `get_asn` should not be necessary.

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -454,11 +454,11 @@ def _set_neighbors(self: AvdStructuredConfigOverlayProtocol) -> None:
                     data["ip_address"],
                     route_server,
                     self.inputs.bgp_peer_groups.evpn_overlay_peers.name,
-                    remote_as=self.shared_utils.get_asn(data["bgp_as"]),
+                    remote_as=data["bgp_as"],
```
alexeygorbunov, line 457: `data["bgp_as"]` is already in proper format

gmuloc (dismissed):

gmuloc (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
```diff
@@ -10,7 +10,7 @@ keys:
     documentation_options:
       table: evpn-settings
     description: |
-      Configure route-map on eBGP sessions towards route-servers, where prefixes with the peer's ASN in the AS Path are filtered away.
+      When set to `true`, EVPN BGP routes will not be advertised to an EVPN route-server if they were originated in, or have already traversed, the route-server's AS.
```
ClausHolbechArista: Please don't lead with "when set to true". The knob is to decide on some configuration, so it should be clear that when set we get the behavior, and otherwise we don't. I am not sure I like the new description more than the old. The old could get the word "prefixes" replaced with "updates", but otherwise I think it is more precise.

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server_mode.schema.yml
```diff
@@ -0,0 +1,21 @@
...
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  evpn_prevent_readvertise_to_server_mode:
+    documentation_options:
+      table: evpn-settings
+    description: |
+      `evpn_prevent_readvertise_to_server_mode` controls the method of identifying EVPN routes that should not be advertised to the EVPN route-servers.
+      Only used when `evpn_prevent_readvertise_to_server` is set to `true`.
+      `source_peer_asn` method uses the outbound `route-map` towards EVPN route-server to filter out BGP routes with the source peer's ASN matching ASN of the route-server.
```
ClausHolbechArista:
```suggestion
      `source_peer_asn` mode configures an outbound route-map towards EVPN route-servers which filter out BGP updates learned directly from the ASN of the route-server. This mode will still allow routes learned via any other peer, even if they have the route-server's ASN in the AS-path.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server_mode.schema.yml
```diff
@@ -0,0 +1,21 @@
...
+      `as_path_acl` method uses the outbound `route-map` towards EVPN route-server to filter out BGP routes containing the route-server's ASN in the `AS PATH` attribute.
```
ClausHolbechArista:
```suggestion
      `as_path_acl` mode configures an outbound route-map and as-path access-list which filters out BGP updates with the  uses the outbound `route-map` towards EVPN route-server to filter out BGP routes containing the route-server's ASN anywhere in the AS-path.
```

ClausHolbechArista (changes requested):

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
```diff
@@ -10,7 +10,7 @@ keys:
...
       This is very useful in large-scale networks, where convergence will be quicker by not returning all updates received
       from Route-server-1 to Router-server-2 just for Route-server-2 to throw them away because of AS Path loop detection.
```
gmuloc, line 16:
```suggestion
      from Route-server-1 to Router-server-2 just for Route-server-2 to throw them away because of AS Path loop detection.
      Refer to `evpn_prevent_readvertise_to_server_mode` to control which configuration is generated.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
ClausHolbechArista:
```suggestion
      Prevent sending EVPN BGP updates to the route-server if they came from or passed through the route-server already.
      Refer to `evpn_prevent_readvertise_to_server_mode` to control which configuration style to use.
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
```diff
@@ -10,8 +10,9 @@ keys:
...
+      Refer to `evpn_prevent_readvertise_to_server_mode` to control which configuration is generated.
```
ClausHolbechArista:
```suggestion
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server_mode.schema.yml
```diff
@@ -12,8 +12,8 @@ keys:
     description: |
       `evpn_prevent_readvertise_to_server_mode` controls the method of identifying EVPN routes that should not be advertised to the EVPN route-servers.
       Only used when `evpn_prevent_readvertise_to_server` is set to `true`.
-      `source_peer_asn` method uses the outbound `route-map` towards EVPN route-server to filter out BGP routes with the source peer's ASN matching ASN of the route-server.
-      `as_path_acl` method uses the outbound `route-map` towards EVPN route-server to filter out BGP routes containing the route-server's ASN in the `AS PATH` attribute.
+      `source_peer_asn` mode configures an outbound route-map towards EVPN route-servers which filter out BGP updates learned directly from the ASN of the route-server. This mode will still allow routes learned via any other peer, even if they have the route-server's ASN in the AS-path.
+      `as_path_acl` mode configures an outbound route-map and as-path access-list which filters out BGP updates with the  uses the outbound `route-map` towards EVPN route-server to filter out BGP routes containing the route-server's ASN anywhere in the AS-path.
```
alexeygorbunov: @gmuloc i feel like this final sentence may not be what you actually wanted to write :)

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
alexeygorbunov: Committed as suggested

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server.schema.yml
alexeygorbunov: Committed as suggested

python-avd/pyavd/_eos_designs/schema/schema_fragments/evpn_prevent_readvertise_to_server_mode.schema.yml
gmuloc: indeed

gmuloc: fixed

ClausHolbechArista (approved):

PR #5968, opened 2025-10-06
Refactor(eos_designs): New data model layout

ClausHolbechArista:
experimenting with new ideas for data models.


carlbuchmann:
I noticed this impacted the pool manager as we no longer need to set `type`. Can you add so we also include the role? 
image

ansible_collections/arista/avd/examples/dual-dc-l3ls/inventory.yml
```diff
@@ -5,27 +5,26 @@ all:
     FABRIC:
       children:
         DC1:
+          hosts:
```
gmuloc, line 8: changing this here will impact the example doc so it would be better to do this in a separate scenario. especially if we make this a preview

ansible_collections/arista/avd/examples/dual-dc-l3ls/group_vars/DC1_RACK2.yml
```diff
@@ -0,0 +1,52 @@
...
+    id: 2
+    mgmt_ip: 172.16.1.152/24
+    uplink_switch_interfaces: [Ethernet8, Ethernet8]
+    uplink_switches: [dc1-leaf2a, dc1-leaf2b]
+
+servers:
+  - name: dc1-leaf2-server1
+    adapters:
+      - endpoint_ports: [PCI1, PCI2]
+        switch_ports: [Ethernet5, Ethernet5]
```
gmuloc, line 38: would be good to see how this model behaves for an endpoint that spans multiple racks

python-avd/pyavd/_eos_designs/shared_utils/mlag.py
```diff
@@ -33,13 +33,23 @@ class MlagMixin(Protocol):
...
-        return self.node_type_key_data.mlag_support and self.node_config.mlag and self.node_group_is_primary_and_peer_hostname is not None
+        if not self.node_type_key_data.mlag_support or not self.node_config.mlag:
+            return False
+
+        # Node groups used for mlag peer.
+        if self.node_group_is_primary_and_peer_hostname:
+            return True
+
+        # devices[].mlag_group used for mlag peer.
+        return bool(self.device_config and self.device_config.mlag_group)
```
gmuloc, line 44: I understand this but if we start having this pattern over and over we should probably find some different way to express this (could we use the presence of node_config / device_config?)

python-avd/pyavd/_eos_designs/shared_utils/mlag.py
```diff
@@ -77,15 +87,25 @@ def mlag_peer_l3_ipv6_pool(self: SharedUtilsProtocol) -> str:
...
     def mlag_role(self: SharedUtilsProtocol) -> Literal["primary", "secondary"] | None:
         # Note: self.node_group_is_primary_and_peer_hostname is always set when self.mlag is true, so this is just to make type-checker happy.
-        if self.mlag and self.node_group_is_primary_and_peer_hostname is not None:
+        if not self.mlag:
+            return None
+        if self.node_group_is_primary_and_peer_hostname is not None:
             return "primary" if self.node_group_is_primary_and_peer_hostname[0] else "secondary"
 
+        if self.switch_facts.mlag_peer:
+            return "primary" if natural_sort([self.hostname, self.switch_facts.mlag_peer])[0] == self.hostname else "secondary"
```
gmuloc, line 96: would this not work for both cases?

python-avd/pyavd/_eos_designs/shared_utils/node_config.py
```diff
@@ -63,17 +67,25 @@ def node_config(self: SharedUtilsProtocol) -> EosDesigns._DynamicKeys.DynamicNod
                 <node_type_key>.node_groups.[<node_group>].nodes.[<node>] ->
                     <node_type_key>.nodes.[<node>]
         """
-        node_config = self.node_type_config.nodes.get(self.hostname, default=EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem())
+        if self.device_config:
```
gmuloc, line 70: we should mention in docstrings the chamge of behavior

python-avd/pyavd/_eos_designs/shared_utils/mlag.py
gmuloc: ah no because we return device_config as node_config :) hehe ok

python-avd/pyavd/_eos_designs/shared_utils/node_config.py
```diff
@@ -63,17 +67,25 @@ def node_config(self: SharedUtilsProtocol) -> EosDesigns._DynamicKeys.DynamicNod
...
         if self.node_group_config is not None:
             node_config._deepinherit(
                 self.node_group_config.nodes[self.hostname]._cast_as(EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem, ignore_extra_keys=True)
             )
             node_config._deepinherit(self.node_group_config._cast_as(EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem, ignore_extra_keys=True))
 
-        node_config._deepinherit(
-            self.node_type_config.defaults._cast_as(EosDesigns._DynamicKeys.DynamicNodeTypesItem.NodeTypes.NodesItem, ignore_extra_keys=True)
-        )
+        if self.node_type_config:
```
gmuloc, line 85:
```suggestion
        if self.node_type_config is not None:
```

python-avd/pyavd/_eos_designs/shared_utils/node_config.py
```diff
@@ -86,10 +98,13 @@ def node_group_is_primary_and_peer_hostname(self: SharedUtilsProtocol) -> tuple[
         Returns True, <peer> if this device is the first one in the node_group.
         Returns False, <peer> if this device is the second one in the node_group.
         """
-        if self.node_group_config is None or len(self.node_group_config.nodes) != 2:
-            return None
+        if self.node_group_config:
```
gmuloc, line 101: any reason for this refactoring? this is augmenting the complexity as far as Sonar is concerned

python-avd/pyavd/_eos_designs/shared_utils/node_type.py
```diff
@@ -23,15 +22,12 @@ class NodeTypeMixin(Protocol):
     """
 
     @cached_property
-    def type(self: SharedUtilsProtocol) -> str:
-        """Type fact set based on type variable."""
+    def type(self: SharedUtilsProtocol) -> str | None:
```
gmuloc, line 25: this one is potentially a big one.
Also if I have a device_config with a default_node_type matched I am not sure if all the code I have read until now will work :)

gmuloc: ah no ok it will raise in node_type_keys because both woul dbe set

python-avd/pyavd/_schema/models/avd_model.py
```diff
@@ -228,9 +228,12 @@ def _as_dict(self, include_default_values: bool = False) -> dict:
                 if field in self.__dict__:
                     continue
 
-                default_value = self._get_field_default_value(field)
+                if (default_value := self._get_field_default_value(field)) is None:
```
gmuloc, line 231: this change looks unrelated to the PR and it is now preventing setting empty default values. Given how our classes are built I think this is ok (given that this is what we would automatically generate) but what is the purpose / need for this change?

gmuloc (commented):
overall this looks ok.
there are some difficulty to know for sure in the code if we are using the type or role (device_config / node_config ...) a bit of sometimes we reuse what is already there or shoehorn the data into it and sometimes not.

Probably would be good to add some tests in particular negative unit ones.

That being said the rest is ok.

We should also make a separate example / molecule scenario maybe for this new one.

python-avd/pyavd/_eos_designs/schema/schema_fragments/device_roles.schema.yml
```diff
@@ -0,0 +1,186 @@
...
+        vtep:
+          type: bool
+          default: false
+        custom_ip_addressing:
+          $ref: "eos_designs#/keys/node_type_keys/items/keys/ip_addressing"
+          type: dict
+        custom_interface_descriptions:
+          $ref: "eos_designs#/keys/node_type_keys/items/keys/interface_descriptions"
+          type: dict
+    default:
```
carlbuchmann, line 77: We should probably use a $ref here so we avoid maintenance.

carlbuchmann (commented):
When role is undefinned in default devices roles I get the following error:

```shell
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: RecursionError: maximum recursion depth exceeded
fatal: [overlay-rs1]: FAILED! => 
  msg: 'Unexpected failure during module execution: maximum recursion depth exceeded'
  stdout: ''
```

running with -vvv exposes the issue:

```shell
  File "/home/carlbuchmann/venvs/misc-projects/lib/python3.13/site-packages/pyavd/_eos_designs/shared_utils/device_config.py", line 69, in device_config
    msg = f"Device Role '{self.role}' applied under 'devices' does not exist in `device_roles`."
```

I was currently referencing role: `overlay-controller`, which you currently have commented out.

python-avd/pyavd/_eos_designs/schema/schema_fragments/device_roles.schema.yml
carlbuchmann: nevermind, this has a slightly different model then node_type_keys.default, so we have have to maintain both!
Need to complete the work you started to include all other current default values.

python-avd/pyavd/_eos_designs/shared_utils/mlag.py
ClausHolbechArista: the historic behavior is to take the first device in the node_group. Now we sort them and take the first from the sorted list. This will not give a difference in most cases, but someone might be using asset numbers as hostnames, and then this could give a diff.

python-avd/pyavd/_schema/models/avd_model.py
ClausHolbechArista: I discovered that our dump function dumped all the unset values even if they have no default value. this means you end up with a huge output with benefit. But you could argue that None is also a default value.
I don't remember exactly when I hit this, but I did somewhere where I am calling dump with defaults. I think it is in the manual recasting method.

python-avd/pyavd/_eos_designs/schema/schema_fragments/device_roles.schema.yml
ClausHolbechArista: Correct, currently the model is slightly different, and even includes most keys from devices[], so we can set everything as a role setting. This is something new, and is a bit annoying as it is right now, because we have an almost-super-set of device config under the role. Under the role we have ip addressing etc, but actually these could also be moved to devices[] model so we can set them anywhere we'd like. It would mean the role model would be the same as the device_profiles / devices model except the profile and downlink_pools stuff.

PR #5974, opened 2025-10-07
Feat(eos_designs): Add connected_endpoints key and custom type for each endpoint

ClausHolbechArista:
## Change Summary

Add connected_endpoints key and custom type for each endpoint

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

We will not be supporting dynamic keys as fixtures in CV, so we need a static key to hold the same data. It will also be useful in the future for a hierarchical inventory style in Ansible.

Since the endpoint "type" was always taken from the `connected_endpoints_keys[].type` we needed another way of supplying the type, so I added an optional `type` under each connected endpoint. Defaulting to the existing for dynamic keys. Unset for `connected_endpoints` key.

- Rearranged schema files to $ref to the new key instead of a $def.
- Added type with separate description for dynamic keys (because of default value indirection)
- Added peer_key to the ethernet_interfaces/port_channel_interfaces schema to be used in docs.
- Updated fabric docs to show the type per endpoint.

## How to test

Added molecule coverage.

### User Checklist

- N/A

### Repository Checklist


laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/default_connected_endpoints_description.yml
```diff
@@ -14,10 +14,10 @@ keys:
       Default description or description template to be used on all ports to connected endpoints.
       This can be a template using the AVD string formatter syntax: https://avd.arista.com/stable/ansible_collections/arista/avd/roles/eos_designs/docs/how-to/custom-descriptions-names.html#avd-string-formatter-syntax.
       The available template fields are:
-        - `endpoint_type`: The `type` from `connected_endpoints_keys` like `server`, `router` etc.
+        - `endpoint_type` - the `type` of the connected endpoint either set on the endpoint or taken from `connected_endpoints_keys.type` like `server`, `router` etc.
```
alexeygorbunov: Should this item use the same style?
`connected_endpoints_keys[].type` vs `connected_endpoints_keys.type`?

```suggestion
        - `endpoint_type`: The `type` of the connected endpoint either set on the endpoint or taken from `connected_endpoints_keys[].type` like `server`, `router` etc.
```

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -419,17 +419,26 @@ def all_connected_endpoints(
...
+
+        for dyn_connected_endpoints in self.inputs._dynamic_keys.custom_connected_endpoints:
+            dyn_connected_endpoints_item = dyn_connected_endpoints._cast_as(EosDesigns._DynamicKeys.DynamicConnectedEndpointsItem)
+            dyn_connected_endpoints_item._internal_data.description = self.inputs.custom_connected_endpoints_keys[dyn_connected_endpoints.key].description
+            dyn_connected_endpoints_item._internal_data.type = self.inputs.custom_connected_endpoints_keys[dyn_connected_endpoints.key].type
+            all_connected_endpoints.append(dyn_connected_endpoints_item)
+
+        for dyn_connected_endpoints in self.inputs._dynamic_keys.connected_endpoints:
+            if dyn_connected_endpoints.key not in all_connected_endpoints:
+                dyn_connected_endpoints._internal_data.description = self.inputs.connected_endpoints_keys[dyn_connected_endpoints.key].description
```
alexeygorbunov, line 439: We are modifying here an original instance (without recasting) on purpose? Just differs from the previous loop..

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/port_channel_interfaces.schema.yml
```diff
@@ -855,6 +855,9 @@ keys:
         peer_type:
           type: str
           description: Key only used for documentation or validation purposes.
+        peer_key:
```
alexeygorbunov, line 858: Is this new key used anywhere?

alexeygorbunov (approved):
Added few minor questions but overall LGTM!

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/port_channel_interfaces.schema.yml
ClausHolbechArista: yes in fabric docs.

python-avd/pyavd/_eos_designs/shared_utils/misc.py
ClausHolbechArista: yeah for the _internal_data we do modify the originals in various places.

gmuloc (approved):

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/port_channel_interfaces.schema.yml
laxmikantchintakindi: It is used only for ethernet_interfaces.

PR #5976, opened 2025-10-07
Refactor(pyavd): Improve typing for python-avd/pyavd/api

Vibhu-gslab:
## Change Summary

Fixing pyright errors for python-avd/pyavd/api

## Related Issue(s)

Fixes Part of #https://github.com/aristanetworks/avd-internal/issues/186
Fixes Part of #https://github.com/aristanetworks/avd-internal/issues/321

## Component(s) name

`pyavd.api`

## Proposed changes
Fixing pyright errors

## How to test
Run pre-commit run pyright --all

### Repository Checklist


Shivani-gslab (commented):
Need to check if we can improve the coverage. Sonar cloud is showing 48%

python-avd/pyavd/api/interface_descriptions/__init__.py
```diff
@@ -32,8 +32,8 @@ class AvdInterfaceDescriptions(AvdFacts):
     """
 
     def _template(self, template_path: str, **kwargs: Any) -> str:
-        template_vars = ChainMap(kwargs, self._hostvars)
-        return self.shared_utils.template_var(template_path, template_vars)
+        template_vars = ChainMap(kwargs, dict(self._hostvars))
```
ClausHolbechArista: doing dict on a hostvars object can be very very expensive. We should rather see if we could change the type hint on template_var to take a `Mapping` instead.

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -200,15 +201,19 @@ def p2p_uplinks_ip(self, uplink_switch_index: int) -> str:
 
         prefixlen = self.inputs.fabric_ip_addressing.p2p_uplinks.ipv4_prefix_length
         p2p_ipv4_pool, offset = self._get_p2p_ipv4_pool_and_offset(uplink_switch_index)
-
+        if p2p_ipv4_pool is None or offset is None:
+            msg = "Either p2p_ipv4_pool or offset is None"
```
ClausHolbechArista: Maybe it is better to move this to the `_get_p2p_ipv4/6_pool_and_offset` and update the return type. Then raise separate errors for each key as needed. Also the AristaAvdMissingVariableError just takes a single variable name as argument.

python-avd/pyavd/api/ip_addressing/utils.py
```diff
@@ -200,7 +200,7 @@ def _get_downlink_ipv6_pool_and_offset(self: AvdIpAddressingProtocol, uplink_swi
         )
         raise AristaAvdError(msg)
 
-    def _get_p2p_ipv4_pool_and_offset(self: AvdIpAddressingProtocol, uplink_switch_index: int) -> tuple[str, int]:
+    def _get_p2p_ipv4_pool_and_offset(self: AvdIpAddressingProtocol, uplink_switch_index: int) -> tuple[str | None, int | None]:
```
ClausHolbechArista: these type hints are too loose compared to the error handling logic inside. Please handle all errors in here so you can guarantee that the correct things are returned. This change is what causes all the other changes to be needed, and it makes it harder to track to the source.

python-avd/pyavd/api/interface_descriptions/__init__.py
Vibhu-gslab: Fixed it!

python-avd/pyavd/api/ip_addressing/__init__.py
Vibhu-gslab: Fixed it.

python-avd/pyavd/api/ip_addressing/utils.py
Vibhu-gslab: Fixed it.

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -403,7 +403,7 @@ def vrf_loopback_ip(self, pool: str) -> str:
 
         Used for "vtep_diagnostic.loopback".
         """
-        offset = self.shared_utils.id + self.shared_utils.node_config.loopback_ipv4_offset
+        offset = (self.shared_utils.id if self.shared_utils.id is not None else 0) + self.shared_utils.node_config.loopback_ipv4_offset
```
laxmikantchintakindi:
```suggestion
        offset = (self.shared_utils.id or 0) + self.shared_utils.node_config.loopback_ipv4_offset
```

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -414,7 +414,7 @@ def vrf_loopback_ipv6(self, pool: str) -> str:
 
         Used for "vtep_diagnostic.loopback".
         """
-        offset = self.shared_utils.id + self.shared_utils.node_config.loopback_ipv6_offset
+        offset = (self.shared_utils.id if self.shared_utils.id is not None else 0) + self.shared_utils.node_config.loopback_ipv6_offset
```
laxmikantchintakindi:
```suggestion
        offset = (self.shared_utils.id or 0) + self.shared_utils.node_config.loopback_ipv6_offset
```

laxmikantchintakindi: Why cant we return None if self.shared_utils.id is None?

Vibhu-gslab:
> Need to check if we can improve the coverage. Sonar cloud is showing 48%

Improved coverage

Shivani-gslab (dismissed):

python-avd/pyavd/api/ip_addressing/utils.py
```diff
@@ -104,12 +104,15 @@ def _vtep_loopback_ipv4_address(self: AvdIpAddressingProtocol) -> str | None:
...
+    def _vtep_loopback_ipv6_address(self: AvdIpAddressingProtocol) -> str | None:
         return self.shared_utils.node_config.vtep_loopback_ipv6_address
 
     @cached_property
-    def _vtep_loopback_ipv4_pool(self: AvdIpAddressingProtocol) -> str | None:
-        return self.shared_utils.vtep_loopback_ipv4_pool
+    def _vtep_loopback_ipv4_pool(self: AvdIpAddressingProtocol) -> str:
+        if (vtep_loopback_ipv4_pool := self.shared_utils.vtep_loopback_ipv4_pool) is None:
+            msg = f"Can't assign IP from `vtep_loopback_ipv4_pool` since `underlay_ipv6_numbered: {self.inputs.underlay_ipv6_numbered}` is defined in inputs"
+            raise AristaAvdError(msg)
```
ClausHolbechArista: The test you are doing here has no relation with the error your are raising. Can you guarantee that this is the only case where the variable is None? It could be that the user never gave it.

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -403,7 +403,7 @@ def vrf_loopback_ip(self, pool: str) -> str:
 
         Used for "vtep_diagnostic.loopback".
         """
-        offset = self.shared_utils.id + self.shared_utils.node_config.loopback_ipv4_offset
+        offset = (self.shared_utils.id or 0) + self.shared_utils.node_config.loopback_ipv4_offset
```
ClausHolbechArista: I don't think we want this. If we want the default value of id to be zero, we would set that in the schema. Instead we should raise, complaining about the ID not being set and we need it. I think we have such raises elsewhere you can check.

python-avd/pyavd/_eos_designs/avdfacts.py
```diff
@@ -7,14 +7,14 @@
...
-    from collections.abc import Mapping
+    from collections.abc import MutableMapping
 
     from pyavd._eos_designs.schema import EosDesigns
     from pyavd._eos_designs.shared_utils import SharedUtilsProtocol
 
 
 class AvdFactsProtocol(Protocol):
-    _hostvars: Mapping
+    _hostvars: MutableMapping
```
ClausHolbechArista, line 17: why MutableMapping instead of Mapping?

ClausHolbechArista (dismissed):

python-avd/pyavd/api/ip_addressing/utils.py
Vibhu-gslab: I have updated the error message :)
The reason behind writing the previous msg was the checking of `None` value for `vtep_loopback_ipv4_pool` is being done here https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_designs/shared_utils/ip_addressing.py#L62

python-avd/pyavd/api/ip_addressing/__init__.py
Vibhu-gslab: added the condition to check for id.

python-avd/pyavd/_eos_designs/avdfacts.py
Vibhu-gslab: This is because [here](https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/api/interface_descriptions/__init__.py#L35) we use `ChainMap` which requires MutableMapping `class ChainMap(_collections_abc.MutableMapping):`

python-avd/pyavd/api/ip_addressing/utils.py
ClausHolbechArista: hmm the ipv6 numbered stuff is hacky and causes these weird things now.

laxmikantchintakindi (dismissed):

pyproject.toml
```diff
@@ -122,6 +122,7 @@ runtime-evaluated-base-classes = ["pydantic.BaseModel"]
 [tool.pyright]
 exclude = [
   "ansible_collections/arista/avd",                                      # TODO
+  "python-avd/pyavd/_anta",                                              # TODO
```
gmuloc: I don't think you should add this

pyproject.toml
```diff
@@ -122,6 +122,7 @@ runtime-evaluated-base-classes = ["pydantic.BaseModel"]
...
   "python-avd/pyavd/api",                                                # TODO
```
gmuloc: why is this not removed?

gmuloc (changes requested):

pyproject.toml
Vibhu-gslab: Sorry, sorry added by mistake

pyproject.toml
Vibhu-gslab: by mistake, apologies!!

gmuloc:
@Vibhu-gslab can we augment coverage on this PR?

Shivani-gslab (approved):
LGTM

MaheshGSLAB (approved):

python-avd/pyavd/api/ip_addressing/__init__.py
```diff
@@ -403,7 +404,10 @@ def vrf_loopback_ip(self, pool: str) -> str:
 
         Used for "vtep_diagnostic.loopback".
         """
-        offset = self.shared_utils.id + self.shared_utils.node_config.loopback_ipv4_offset
+        if (switch_id := self.shared_utils.id) is None:
+            msg = f"'id' is not set on '{self.shared_utils.hostname}' to get IP address for a vrf loopback interface"
```
ClausHolbechArista: the text seems to be missing something.

python-avd/pyavd/api/_anta/avd_fabric_data.py
```diff
@@ -61,24 +61,29 @@ def from_structured_config(cls, structured_config: dict[str, Any]) -> AvdDeviceD
...
-            vtep_ip = get(get_item(interface_model, "name", vxlan_source_interface, default={}), "ip_address")
-        else:
-            vtep_ip = None
+            vtep_ip_str = get(get_item(interface_model, "name", vxlan_source_interface, default={}), "ip_address")
+
+        vtep_ip = None
+        if vtep_ip_str:
+            ip = ip_interface(vtep_ip_str).ip
+            if ip.version == 4:
+                vtep_ip = ip
```
gmuloc: write a helper function that convert an ip string to the object or return None otherwise. Though technically we would rather raise I believe. @carl-baillargeon

python-avd/pyavd/api/interface_descriptions/__init__.py
```diff
@@ -457,7 +457,7 @@ def connected_endpoints_port_channel_interface(self, data: InterfaceDescriptionD
             ),
         )
 
-    def router_id_loopback_interface(self, data: InterfaceDescriptionData) -> str:
+    def router_id_loopback_interface(self, data: InterfaceDescriptionData) -> str | None:
```
gmuloc, line 460: this is changing the API which we don't want. lets investigate further.

@ClausHolbechArista

python-avd/pyavd/api/_anta/avd_fabric_data.py
carl-baillargeon: I created the helper here: https://github.com/aristanetworks/avd/pull/6043/files#diff-f098741dde67d6fbc5f157d0c720f91d7eb9aae325b026a68a577f9e2f164018R85

Yes we could raise instead since `loopback_interfaces[].ip_address` should never be an IPv6 address. We have `loopback_interfaces[].ipv6_address` for IPv6.

gmuloc: @Vibhu-gslab lets rebase this PR now that #6043 has merged

python-avd/pyavd/api/interface_descriptions/__init__.py
gmuloc: discussed with Claus - we keep this

python-avd/pyavd/api/_anta/avd_fabric_data.py
Vibhu-gslab: Updated the PR!

python-avd/pyavd/api/_anta/avd_fabric_data.py
```diff
@@ -60,7 +60,11 @@ def from_structured_config(cls, structured_config: dict[str, Any]) -> AvdDeviceD
         }
 
         # Get the Loopback0 IP
-        loopback0_ip = get(get_item(get(structured_config, "loopback_interfaces", []), "name", "Loopback0", default={}), "ip_address")
+        loopback0_ip_str = get(get_item(get(structured_config, "loopback_interfaces", []), "name", "Loopback0", default={}), "ip_address")
+        loopback0_ip = ip_interface(loopback0_ip_str).ip if loopback0_ip_str else None
```
gmuloc: Something like this should remove the need for line 65

```suggestion
        loopback0_ip = IPv4Address(loopback0_ip_str) if loopback0_ip_str else None
```

Vibhu-gslab: Fixed

python-avd/pyavd/_eos_designs/avdfacts.py
gmuloc: https://github.com/microsoft/pylance-release/issues/2097

python-avd/pyavd/api/_anta/avd_fabric_data.py
```diff
@@ -60,7 +60,8 @@ def from_structured_config(cls, structured_config: dict[str, Any]) -> AvdDeviceD
         }
 
         # Get the Loopback0 IP
-        loopback0_ip = get(get_item(get(structured_config, "loopback_interfaces", []), "name", "Loopback0", default={}), "ip_address")
+        loopback0_ip_str = get(get_item(get(structured_config, "loopback_interfaces", []), "name", "Loopback0", default={}), "ip_address")
+        loopback0_ip = IPv4Address(ip_interface(loopback0_ip_str).ip) if loopback0_ip_str else None
```
gmuloc:
```suggestion
        loopback0_ip = IPv4Interface(loopback0_ip_str).ip if loopback0_ip_str else None
```

python-avd/pyavd/api/_anta/avd_fabric_data.py
gmuloc, line 7:
```suggestion
from ipaddress import IPv4Address, IPv4Interface, ip_interface
```

python-avd/pyavd/api/_anta/avd_fabric_data.py
```diff
@@ -106,20 +107,19 @@ def _get_vtep_ips(structured_config: dict[str, Any]) -> tuple[IPv4Address | None
...
+            interface_model = dps_interfaces if "Dps" in vxlan_source_interface else loopback_interfaces
 
             vtep_ip_str = get(get_item(interface_model, "name", vxlan_source_interface, default={}), "ip_address")
             vtep_ip = vtep_ip_addr if vtep_ip_str and isinstance((vtep_ip_addr := ip_interface(vtep_ip_str).ip), IPv4Address) else None
 
         # Get the MLAG VTEP IP (for Multi-VTEP MLAG feature)
         vxlan_mlag_source_interface = get(structured_config, "vxlan_interface.vxlan1.vxlan.mlag_source_interface")
-        if vxlan_mlag_source_interface is not None:
-            mlag_vtep_ip_str = get(get_item(interface_model, "name", vxlan_mlag_source_interface, default={}), "ip_address")
+        if vxlan_mlag_source_interface is not None and "Dps" not in vxlan_mlag_source_interface:
```
gmuloc: discussed on call - we should one day enforce valid pattern in schema
```suggestion
        if vxlan_mlag_source_interface is not None:
```

gmuloc:
```suggestion
        if vxlan_mlag_source_interface is not None:
```

python-avd/pyavd/api/_anta/avd_fabric_data.py
```diff
@@ -106,20 +107,19 @@ def _get_vtep_ips(structured_config: dict[str, Any]) -> tuple[IPv4Address | None
...
+            interface_model = dps_interfaces if "Dps" in vxlan_source_interface else loopback_interfaces
 
             vtep_ip_str = get(get_item(interface_model, "name", vxlan_source_interface, default={}), "ip_address")
             vtep_ip = vtep_ip_addr if vtep_ip_str and isinstance((vtep_ip_addr := ip_interface(vtep_ip_str).ip), IPv4Address) else None
 
         # Get the MLAG VTEP IP (for Multi-VTEP MLAG feature)
         vxlan_mlag_source_interface = get(structured_config, "vxlan_interface.vxlan1.vxlan.mlag_source_interface")
         if vxlan_mlag_source_interface is not None:
-            mlag_vtep_ip_str = get(get_item(interface_model, "name", vxlan_mlag_source_interface, default={}), "ip_address")
+            mlag_vtep_ip_str = get(get_item(loopback_interfaces, "name", vxlan_mlag_source_interface, default={}), "ip_address")
```
gmuloc, line 123:
```suggestion
            # Only supported with loopback interfaces
            mlag_vtep_ip_str = get(get_item(loopback_interfaces, "name", vxlan_mlag_source_interface, default={}), "ip_address")
```

carl-baillargeon (approved):
LGTM.

gmuloc (approved):
OK if CI passes

PR #5989, opened 2025-10-08
Doc: Create flat yaml doc file with full data model

ClausHolbechArista:
To assist with PR reviews we will saving a full yaml file showing the full data model of each schema, so when we later generate the tables etc during mkdocs, we can still easily evaluate the descriptions and layout of the data model.


ClausHolbechArista:
closing for now since we are still not removing the generated markdown files.

PR #5994, opened 2025-10-09
Refactor(anta_runner): Fix the pyright errors for pyavd/_anta/

MaheshGSLAB:
## Change Summary

Fix the pyright errors for pyavd/_anta/

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.anta_runner`

## Proposed changes
Fix the pyright errors for pyavd/_anta/

## How to test
Run the pyright hook from pre-commit

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_anta/input_factories/interfaces.py
```diff
@@ -37,7 +37,7 @@ class VerifyInterfacesStatusInputFactory(AntaTestInputFactory):
 
     def create(self) -> list[VerifyInterfacesStatus.Input] | None:
         """Create a list of inputs for the `VerifyInterfacesStatus` test."""
-        interfaces = []
+        interfaces: list[InterfaceState] = []
```
MaheshGSLAB, line 40: speed and lanes are optional so i am not getting why it says that speed and lanes are missing arguments. However even if i give as None then it reflect on catalogs which is again useless.
@gmuloc @carl-baillargeon

carl-baillargeon: It looks like Pyright doesn't understand positional arguments on Pydantic `Field`:

This doesn't work:
`speed: float | None = Field(None, ge=1, le=1000)`

But this does:
`speed: float | None = Field(default=None, ge=1, le=1000)`

carl-baillargeon: https://github.com/microsoft/pyright/issues/7166

carl-baillargeon: Let's ignore Pyright for now and see if we want to update ANTA.

gmuloc: agreed: let's add a type ignore and a TODO here @MaheshGSLAB I opened https://github.com/aristanetworks/anta/issues/1340 in ANTA

MaheshGSLAB: added

python-avd/pyavd/_anta/logs.py
```diff
@@ -33,12 +33,15 @@ class TestLoggerAdapter(LoggerAdapter):
...
 
-    def process(self, msg: LogMessage, kwargs: dict) -> tuple[str, dict]:
+    def process(self, msg: LogMessage, kwargs: MutableMapping[str, object]) -> tuple[str, MutableMapping[str, object]]:
         """Process the message and kwargs before logging."""
         # Keep the extra dict in kwargs to pass it to the formatter if needed (following the standard LoggerAdapter behavior)
         kwargs["extra"] = self.extra
 
         # Extract the device, test, and context from extra
+        if self.extra is None:
+            return str(msg.value), kwargs
```
carl-baillargeon: We should still return the processed_msg here (without the prefix since self.extra is None).

You could probably move the logic of this function to accommodate this.

python-avd/pyavd/_anta/logs.py
```diff
@@ -49,20 +52,27 @@ def process(self, msg: LogMessage, kwargs: dict) -> tuple[str, dict]:
...
             kwargs.pop(field, None)
 
-        return f"{prefix} {msg}", kwargs
+        return f"{prefix} {processed_msg}", kwargs
 
     @contextmanager
     def context(self, context: str) -> Generator[TestLoggerAdapter, None, None]:
         """Temporarily add context to the logger."""
-        original_extra = dict(self.extra)
+        if self.extra is None:
```
carl-baillargeon: Since our custom adapter relies on extra **not** being None, we should probably override `__init__` and raise if extra is not provided with device and test. We would have the proper type everywhere and we wouldn't need all that additional complexity.

python-avd/pyavd/_anta/input_factories/security.py
```diff
@@ -54,7 +54,7 @@ def create(self) -> list[VerifySpecificIPSecConn.Input] | None:
             # Add static peers to the list of IP security connections
             for static_peer in path_group.static_peers:
                 peer_ip = ip_interface(static_peer.router_ip).ip
-                if (static_peer.router_ip, "default") not in added_peers:
+                if (static_peer.router_ip, "default") not in added_peers and isinstance(peer_ip, IPv4Address):
```
carl-baillargeon: I think we had that discussion before but I can't remember... Do we support IPv6 static peers? If so, we should probably put the `isinstance` check on a separate line, add a log and continue since IPv6 is not supported in ANTA `IPSecPeer` input model. Otherwise let's just add a comment above saying that we don't support IPv6 static peers.

carl-baillargeon (changes requested):

python-avd/pyavd/_anta/input_factories/security.py
MaheshGSLAB: updated the code

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_anta/logs.py
```diff
@@ -95,6 +107,7 @@ class LogMessage(Enum):
     PATH_GROUP_NO_LOCAL_INTERFACES = "path group {path_group} skipped - No local interfaces found"
     PATH_GROUP_NO_STATIC_PEERS = "path group {path_group} skipped - No static peers configured"
     NO_STATIC_PEERS = "skipped - No static peers configured in any path groups"
+    IPv6_STATIC_PEER = "static peer {peer} skipped - ANTA not support IPv6 static peer"
```
laxmikantchintakindi:
```suggestion
    IPv6_STATIC_PEER = "static peer {peer} skipped - ANTA does not support IPv6 static peer"
```

python-avd/pyavd/_anta/input_factories/avt.py
```diff
@@ -24,14 +26,17 @@ def create(self) -> list[VerifyAVTSpecificPath.Input] | None:
...
-        static_peers: set[str] = set()
+        static_peers: set[IPv4Address] = set()
 
         for path_group in path_groups:
             if not path_group.static_peers:
                 self.logger_adapter.debug(LogMessage.PATH_GROUP_NO_STATIC_PEERS, path_group=path_group.name)
                 continue
             for static_peer in path_group.static_peers:
-                static_peers.add(static_peer.router_ip)
+                if isinstance(ip_address(static_peer.router_ip), IPv4Address):
```
laxmikantchintakindi: Could we only check if `:` is in the static_peer.router_ip?. If : is present in it, its Ipv6 address.

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: updated

laxmikantchintakindi (approved):

python-avd/pyavd/_anta/logs.py
```diff
@@ -33,36 +34,47 @@ class TestLoggerAdapter(LoggerAdapter):
...
+            formatted_msg = msg_value.format(**kwargs)
+        except Exception:
+            formatted_msg = msg_value
+
+        # Remove used fields to not interfere with logging kwargs
         for field in fields:
             kwargs.pop(field, None)
 
-        return f"{prefix} {msg}", kwargs
+        return f"{prefix} {formatted_msg}", kwargs
```
carl-baillargeon: I don't think we need to change all this. We can format the msg first and return early if self.extra is None. Check if the following code works (very close to the original code):

```
    def process(self, msg: LogMessage, kwargs: MutableMapping[str, Any]) -> tuple[str, MutableMapping[str, Any]]:
        """Process the message and kwargs before logging."""
        # Keep the extra dict in kwargs to pass it to the formatter if needed (following the standard LoggerAdapter behavior)
        kwargs["extra"] = self.extra

        # Format the LogMessage using the provided kwargs and extract the fields name from the message string
        fields = [field_name for _, field_name, _, _ in string.Formatter().parse(msg.value) if field_name is not None]
        formatted_msg = msg.value.format(**kwargs)

        # Removing the fields name from kwargs to preserve standard logging kwargs only that should always be passed through (e.g. exc_info, stack_info, etc.)
        for field in fields:
            kwargs.pop(field, None)

        if self.extra is None:
            return formatted_msg, kwargs

        # Extract the device, test, and context from extra
        device = self.extra["device"]
        test = self.extra["test"]
        context = self.extra.get("context")

        prefix = f"<{device}> {test}"
        if context:
            prefix += f" {context}"

        return f"{prefix} {formatted_msg}", kwargs
```

python-avd/pyavd/_anta/logs.py
```diff
@@ -33,36 +34,47 @@ class TestLoggerAdapter(LoggerAdapter):
...
 
     @contextmanager
     def context(self, context: str) -> Generator[TestLoggerAdapter, None, None]:
         """Temporarily add context to the logger."""
-        original_extra = dict(self.extra)
+        original_extra = dict(getattr(self, "extra", {}) or {})
```
carl-baillargeon: We need to keep the original_extra as None if it was None, not an empty dict.
```suggestion
        original_extra = dict(self.extra) if self.extra is not None else None
```

python-avd/pyavd/_anta/logs.py
```diff
@@ -33,36 +34,47 @@ class TestLoggerAdapter(LoggerAdapter):
...
         try:
-            self.extra["context"] = context
+            mutable_extra = dict(original_extra)
+            mutable_extra["context"] = context
+            self.extra = mutable_extra
```
carl-baillargeon:
```suggestion
            new_extra = dict(self.extra, context=context) if self.extra is not None else {"context": context}
            self.extra = new_extra
```

python-avd/pyavd/_anta/input_factories/avt.py
```diff
@@ -24,14 +26,17 @@ def create(self) -> list[VerifyAVTSpecificPath.Input] | None:
...
+                    static_peers.add(IPv4Address(static_peer.router_ip))
+                else:
+                    self.logger_adapter.debug(LogMessage.IPv6_STATIC_PEER, peer=static_peer.router_ip)
```
carl-baillargeon: Avoid creating twice the IP object.
```suggestion
                static_peer_ip = ip_address(static_peer.router_ip)
                if isinstance(static_peer_ip, IPv4Address):
                    static_peers.add(static_peer_ip )
                else:
                    self.logger_adapter.debug(LogMessage.IPv6_STATIC_PEER, peer=static_peer.router_ip)
```

python-avd/pyavd/_anta/input_factories/router_path_selection.py
```diff
@@ -38,26 +38,32 @@ def create(self) -> list[VerifySpecificPath.Input] | None:
...
-                        dps_path = DpsPath(
-                            peer=static_peer.router_ip, path_group=path_group.name, source_address=source_address, destination_address=destination_address
-                        )
-                        all_dps_paths.append(dps_path)
+                    if isinstance(ip_address(static_peer.router_ip), IPv4Address):
+                        for destination_address in static_peer.ipv4_addresses:
+                            dps_path = DpsPath(
+                                peer=IPv4Address(static_peer.router_ip),
+                                path_group=path_group.name,
+                                source_address=IPv4Address(source_address),
```
carl-baillargeon, line 65: source_address is already an `IPv4Address` instance no?
```suggestion
                                source_address=source_address,
```

python-avd/pyavd/_anta/input_factories/router_path_selection.py
```diff
@@ -38,26 +38,32 @@ def create(self) -> list[VerifySpecificPath.Input] | None:
...
 
-                source_address = ip_interface(ip_address).ip
+                source_address = ip_interface(interface_ip_address).ip
                 for static_peer in path_group.static_peers:
-                    for destination_address in static_peer.ipv4_addresses:
-                        dps_path = DpsPath(
-                            peer=static_peer.router_ip, path_group=path_group.name, source_address=source_address, destination_address=destination_address
-                        )
-                        all_dps_paths.append(dps_path)
+                    if isinstance(ip_address(static_peer.router_ip), IPv4Address):
```
carl-baillargeon: Same as my other comment, avoid creating twice the IP object.

python-avd/pyavd/_anta/logs.py
```diff
@@ -95,6 +107,7 @@ class LogMessage(Enum):
     PATH_GROUP_NO_LOCAL_INTERFACES = "path group {path_group} skipped - No local interfaces found"
     PATH_GROUP_NO_STATIC_PEERS = "path group {path_group} skipped - No static peers configured"
     NO_STATIC_PEERS = "skipped - No static peers configured in any path groups"
+    IPv6_STATIC_PEER = "static peer {peer} skipped - ANTA does not support IPv6 static peer"
```
carl-baillargeon: Let's also add the path_group in the message for more context, similar to the other `PATH_GROUP_XXX` messages.
```suggestion
    PATH_GROUP_IPV6_STATIC_PEER = "static peer {peer} skipped - ANTA does not support IPv6 static peer"
```

carl-baillargeon (changes requested):

python-avd/pyavd/_anta/input_factories/router_path_selection.py
MaheshGSLAB: no it can be IPv6 as well because it came from `source_address = ip_interface(interface_ip_address).ip`.

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/input_factories/avt.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/input_factories/router_path_selection.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: updated

carl-baillargeon: @MaheshGSLAB Please see my comment above the suggestion :) - Let's also add the path_group in the message for more context, similar to the other PATH_GROUP_XXX messages.

python-avd/pyavd/_anta/input_factories/router_path_selection.py
carl-baillargeon: That means for the type checker we could be doing `IPv4Address(<IPv6Address instance>)` which feels wrong.

Maybe we could simply continue if source_address is an instance of IPv6Address? This would never happen because  it is taken from `ip_address` and not `ipv6_address` so it's fine I think.

MaheshGSLAB: yeah right it taken from ip_address so it will be ipv4 address only. So just doing continue and not any logger.

python-avd/pyavd/_anta/logs.py
MaheshGSLAB: sorry i missed the comment, updated now

carl-baillargeon (approved):
LGTM.

PR #5986, opened 2025-10-08
CI: Capping pydantic<2.12 for CI

gmuloc:
## Change Summary

ANTA is failing currently with pydantic 2.12 released last night.
cf https://github.com/pydantic/pydantic/issues/12348

for now unblocking CI for the rest

PR #6021, opened 2025-10-16
CI: Capping pydantic<2.12 for CI (#5986)

gmuloc:
Cherry-pick #5986


gmuloc:
will be done in #6020 together

PR #6065, opened 2025-10-27
Refactor: Rename avdutils to pyavd-utils

ClausHolbechArista:
Rename/move the WIP avdutils python library to pyavd-utils. Also move the bindings of Rust code into the same folder structure.


pyavd-utils/Makefile
```diff
@@ -0,0 +1,36 @@
...
+
+.PHONY: publish
+publish: ## Publish pyavd package to PyPI (build first)
+	pip3 install twine && \
+	twine check dist/* && \
+	twine upload -r testpypi dist/* && \
+	twine upload dist/*
+
+.PHONY: uv-publish
+uv-publish: ## Publish pyavd package to PyPI (build first)
```
gmuloc, line 32: do we want one command that pushes to both testpypi and pypi in a row?

pyavd-utils/tests/validation/test_validate_json.py
```diff
@@ -7,14 +7,14 @@
...
-from avdutils.validation import Issue, Value, Violation, init_store_from_fragments, validate_json
+from pyavd_utils.validation import Issue, Value, Violation, init_store_from_fragments, validate_json
 
 
 @pytest.fixture
 def init_store() -> None:
     org_path = sys.path
     # Insert /python-avd into the python path to be able to import constants from schema_tools.
-    mocked_path = [str(Path(__file__).parents[4] / "python-avd"), *org_path]
+    mocked_path = [str(Path(__file__).parents[3] / "python-avd")]
```
gmuloc, line 17: so here we ues to expand with mocked_path, now we don't and it still works - this feels a bit surprising? (or maybe we could repalce all this with some with block?

gmuloc (commented):
LGTM overall - just two questions

pyavd-utils/tests/validation/test_validate_json.py
ClausHolbechArista: So before we inserted the python-avd directory before the existing path. Now we just replace the path to only have the python-avd path, and that works. I needed this to trigger the problem with the path locally. It is not surprising that everything works, because it is only mocked during the test. All the imports are done outside.

pyavd-utils/Makefile
ClausHolbechArista: this was just taken from python-avd. Someone had a task to move the publishing to a workflow, so once that is ready all this goes away.

gmuloc:
Should we increase coverage here?

rust/validation/Cargo.toml
```diff
@@ -11,7 +11,7 @@ pyo3-build-config = { workspace = true }
 avdschema = { path = "../avdschema" }
 derive_more = { workspace = true, features = ["display", "from"] }
 ordermap = { workspace = true, features = ["serde"] }
-pyo3 = { workspace = true, features = ["abi3", "extension-module"], optional = true }
```
gmuloc: why do we remove `extension-module`? This seems to be required as per the doc for unix targets: https://pyo3.rs/main/building-and-distribution#the-extension-module-feature

.github/workflows/pull-request-management.yml
```diff
@@ -682,6 +682,10 @@ jobs:
         # This is needed for arm machines since they don't ship with 2024 edition.
       - name: Upgrade rust and cargo
         run: rustup update stable
+        # This is needed because we have pyo3 tests.
```
gmuloc: is this because of some already installed version on the runner?

ClausHolbechArista: on windows the default version is 3.9

rust/validation/Cargo.toml
ClausHolbechArista: they have not updated their docs. The new way is an environment variable (ref https://github.com/PyO3/pyo3/issues/5202) which is set by default with the new version of setuptools-rust (ref https://github.com/PyO3/setuptools-rust/pull/540)

ClausHolbechArista: I need to clean this up since cargo tests are not working when you have extension-module turned on.

.github/workflows/pull-request-management.yml
gmuloc: hehehe fair enough - lets put this in the comment so we know why :p

rust/validation/Cargo.toml
gmuloc: confirmed

.github/workflows/pull-request-management.yml
ClausHolbechArista: updated comment

gmuloc (approved):
LGTM - as discussed on the call, a couple of more tests will be added

PR #6070, opened 2025-10-28
Fix(anta_runner): VerifyStormControlDrops should run only when the feature is enabled

### LINKED ISSUE #5769: Fix(anta_runner): VerifyStormControlDrops should run only when the feature is enabled
Storm control is not supported on certain hardware platforms so ANTA emits a warning which could be avoided by not running the test in the first place.

vitthalmagadum:
## Change Summary

VerifyStormControlDrops should run only when the feature is enabled

## Related Issue(s)

Fixes #5769

## Component(s) name

`arista.avd.anta_runner`

## Proposed changes
VerifyStormControlDrops should run only when the feature is enabled

## How to test
Run Molecule

### User Checklist
- N/A

### Repository Checklist


python-avd/pyavd/_anta/input_factories/interfaces.py
```diff
@@ -94,3 +94,20 @@ def create(self) -> list[VerifyPortChannels.Input] | None:
                 ignored_interfaces.append(po_intf.name)
 
         return [VerifyPortChannels.Input(ignored_interfaces=natural_sort(ignored_interfaces))] if ignored_interfaces else [VerifyPortChannels.Input()]
+
+
+class VerifyStormControlDropsInputFactory(AntaTestInputFactory[VerifyStormControlDrops.Input]):
+    """
+    Input factory class for the `VerifyStormControlDrops` test.
+
+    Run the test if find at least one Ethernet or Port-Channel interface configured with storm-control on the device. Otherwise skip the test.
```
carl-baillargeon: @gmuloc Do you think this is fine or should we instead build a list of interfaces (Ethernet + Port-Channel) to target?

We would return something like `VerifyStormControlDrops.Input(interfaces=interfaces_with_storm_control_enabled)`.

This could potentially generate a lot of inputs.

gmuloc: this is fine like it is

gmuloc:
```suggestion
    Generate the test inputs only if any Ethernet or Port-Channel interfaces are configured with storm-control.
```

carl-baillargeon (approved):
LGTM!

gmuloc (approved):
I am thinking the custom platform trick may hbe difficult to maintain in the long run.

PR #6082, opened 2025-10-29
Feat(cv_deploy): Support environment variables as a way to manage proxy settings

alexeygorbunov:
## Change Summary

Add support to respect the following environment variables in cases where proxy-related settings ***are not explicitly passed to `cv_deploy/cv_workflow`***:
 - `https_proxy`
 - `HTTPS_PROXY`
 - `all_proxy`
 - `ALL_PROXY`
 - `no_proxy`
 - `NO_PROXY`

## Related Issue(s)
N/A

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- If proxy-related settings are not explicitly passed into `cv_deploy`/`cv_workflow` -> fallback to checking env variable for the presence of the standard proxy-related variables (which are normally used by `requests` / `urllib` / etc)

## How to test
- `cv_deploy` molecule scenario
- `test_cv_client_proxy.py` & `test_cv_client_proxy_live.py` pytest module

Latest CI run: https://github.com/aristanetworks/avd-internal/actions/runs/18916912183

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -19,64 +25,70 @@ class HTTPProxyManager:
...
+        self.scheme = scheme
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
         self.target_host = target_host
         self.target_port = target_port
 
-    @property
+    @cached_property
```
carl-baillargeon: Using `@cached_property` for properties that depend on mutable state is not a good idea. Let's remove them entirely and make them regular `@property` or simple attributes. All computations are lightweight (string formatting, boolean checks) and don't need caching anyways.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -90,3 +102,81 @@ async def create_socket_for_grpc(self) -> socket.socket:
 
         # Connect through proxy to target
         return await proxy.connect(dest_host=self.target_host, dest_port=self.target_port)
+
+    @cached_property
+    def use_proxy(self) -> bool:
```
carl-baillargeon: This property does a lot. It should be a simple attribute access, no side effects. Please check my other comment in `__init__`.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -19,64 +25,70 @@ class HTTPProxyManager:
...
 
         Returns:
             HTTP proxy URL.
         """
-        if self.proxy_username and self.proxy_password:
+        if self.username and self.password:
             # Excempting the lines below from Sonar since we cannot use HTTPS here.
-            return f"http://{self.proxy_username}:{self.proxy_password}@{self.proxy_host}:{self.proxy_port}"  # NOSONAR
-        return f"http://{self.proxy_host}:{self.proxy_port}"  # NOSONAR
+            return f"http://{self.username}:{self.password}@{self.host}:{self.port}"  # NOSONAR
```
carl-baillargeon: Does that work when credentials have special characters like `@`, `:`, `/`? Maybe we need to URL encoding.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -90,3 +102,81 @@ async def create_socket_for_grpc(self) -> socket.socket:
...
+
+        # Fallback to the proxy-related settings passed via environment variables
+        return self.proxy_set_via_env and not self.bypass_proxy
+
+    @cached_property
+    def bypass_proxy(self) -> bool:
+        """Checks if target CloudVision server is included in environment variables `no_proxy` or `NO_PROXY` forcing proxy bypass."""
+        for env_variable_candidate in ["no_proxy", "NO_PROXY"]:
+            if isinstance(no_proxy_candidate := environ.get(env_variable_candidate), str) and len(no_proxy_candidate) > 0:
+                return bool({f"{self.target_host}", f"{self.target_host + ':' + str(self.target_port)}"} & set(no_proxy_candidate.split(",")))
```
carl-baillargeon: Do we need to handle whitespaces? `"localhost, .example.com"` -> `" .example.com"` won't match.

Also, should we support wildcard/suffix matching (`.example.com` should match `api.example.com`)?

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -90,3 +102,81 @@ async def create_socket_for_grpc(self) -> socket.socket:
...
+        )
+
+    @staticmethod
+    def proxy_candidate_is_usable(proxy_candidate: CVProxyServerCandidate) -> bool:
+        """Verifies that mandatory proxy settings are set in Proxy server candidate."""
+        return all(
+            [
+                proxy_candidate.scheme == "http",
+                isinstance(proxy_candidate.host, str) and len(proxy_candidate.host) > 0,
+                isinstance(proxy_candidate.port, int) and (0 < proxy_candidate.port < 65536),
```
carl-baillargeon: This is correct but the following is more explicit:
```suggestion
                isinstance(proxy_candidate.port, int) and (1 <= proxy_candidate.port <= 65535),
```

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -19,64 +25,70 @@ class HTTPProxyManager:
...
         """
-        self.proxy_host = proxy_host
-        self.proxy_port = proxy_port
-        self.proxy_username = proxy_username
-        self.proxy_password = proxy_password
+        self.scheme = scheme
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
```
carl-baillargeon: Here it feels like the constructor accepts the parameters from CVClient but doesn't actually finalize the configuration. The real logic happens lazily when `use_proxy` property is first accessed so we don't know if the proxy is actually configured until we use the object.

I think ALL configuration logic should be moved here in the `__init__`.

Something like:
```
self._use_proxy = False

# Check bypass rules first
if self._should_bypass_proxy():
    logger.debug("...")
    return
    
# Try explicit parameters
if self._is_valid_proxy_config(scheme, host, port):
    logger.debug("...")
    self._set_proxy_config(scheme, host, port, username, password)
    self._use_proxy = True
    return

# Fall back to environment variables
if self._configure_from_environment():
    logger.debug("...")
    self._use_proxy = True
    return

logger.debug("No proxy configured, connecting directly")  # Or whatever
```

And then we can simplify `use_proxy` a lot:
```
@property
def use_proxy(self) -> bool:
    return self._use_proxy
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -33,6 +33,15 @@ class CloudVision:
     proxy_password: str | None
 
 
+@dataclass
+class CVProxyServerCandidate:
```
carl-baillargeon: I don't think this class would be needed if the logic is moved to `__init__`.

carl-baillargeon (changes requested):

carl-baillargeon:
Also, please add debug logs to understand what is going on. Thanks!

python-avd/pyavd/_cv/client/proxy.py
ClausHolbechArista: why then even have _use_proxy? you could just set self.use_proxy and not have it as a `@property` but just a native attribute.

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Removed all @cached_properties

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Refactored to move logic away from this property

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Added `urllib.parse.quote` to handle special symbols in username & password when passed explicitly

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: String should have comma-separated values. Don't this we need to handle whitespaces.
Added support for wildcard domains, ipv4 and ipv6 CIDRs and IPs

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: FIxed

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Refactor almost everything. Will require a fresh review

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Removed it

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -361,6 +359,18 @@ To enable the proxy, set `proxy_host` (port `TCP/8080` will be used by default).
...
+
+!!! Note
+
+    If proxy-related settings are passed to `cv_deploy` explicitly and execution environment has proxy-related environment variables set (`https_proxy`/`HTTPS_PROXY`/`all_proxy`/`ALL_PROXY`/`no_proxy`/`NO_PROXY`), such environment variables will be ignored for all REST and gRPC calls initiated by `cv_deploy`.
+
+### Configure proxy settings explicitly
+
+To enable proxy server explicitly, set `proxy_host`. Setting `proxy_port` (port `TCP/8080` will be used by default), `proxy_username` and `proxy_password` is optional.
+
+If valid `proxy_host` (must be non-empty string) and `proxy_port` (greater than 0, less than 65536) are set, `cv_deploy` will ignore (for both REST and gRPC calls) any proxy-related environment variables (`https_proxy`/`HTTPS_PROXY`/`all_proxy`/`ALL_PROXY`/`no_proxy`/`NO_PROXY`) and will force all REST and gRPC egress connections through this proxy.
```
```suggestion
If valid `proxy_host` (must be non-empty string) and `proxy_port` (1-65535) are set, `cv_deploy` will ignore (for both REST and gRPC calls) any proxy-related environment variables (`https_proxy`/`HTTPS_PROXY`/`all_proxy`/`ALL_PROXY`/`no_proxy`/`NO_PROXY`) and will force all REST and gRPC egress connections through this proxy.
```

ClausHolbechArista: this repeats the note above. Do we need it twice?

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -390,6 +400,115 @@ proxy_username: "avd_proxy_user"
...
+    proxy_port: 3128
+    proxy_username: "p:r/o$x@yuser"
+    proxy_password: "p:r/o$x@ypassword"
+    ```
+
+    will be equal to setting the following via the environment variable: `http://p%3Ar%2Fo%24x%40yuser:p%3Ar%2Fo%24x%40ypassword@proxy.local.domain:3128`
+
+### Configure proxy settings using environment variables
+
+If proxy-related settings are not passed to `cv_deploy` explicitly, `cv_deploy` will try to discover usable proxy server (scheme is `http`, host is a non-empty string, port is greater than 0 and less than 65536) using environment variables in the following order:
```
```suggestion
If proxy-related settings are not passed to `cv_deploy` explicitly, `cv_deploy` will try to discover usable proxy server (scheme is `http`, host is a non-empty string, port is greater in the range 1-65535) using environment variables in the following order:
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -390,6 +400,115 @@ proxy_username: "avd_proxy_user"
...
+    ```yaml
+    proxy_host: proxy.local.domain
+    proxy_port: 3128
+    proxy_username: "p:r/o$x@yuser"
+    proxy_password: "p:r/o$x@ypassword"
+    ```
+
+    will be equal to setting the following via the environment variable: `http://p%3Ar%2Fo%24x%40yuser:p%3Ar%2Fo%24x%40ypassword@proxy.local.domain:3128`
+
+### Configure proxy settings using environment variables
```
ClausHolbechArista: I think this section is way to detailed for a user-facing doc. Maybe we could just refer to some RFC or similar that explains the behavior of these environment variables.

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -390,6 +400,115 @@ proxy_username: "avd_proxy_user"
...
+
+```code
+export no_proxy='*,www.arista.io,www.arista.io:443,.arista.io,.arista.io:443,34.67.65.165,34.67.65.165/32,34.67.65.0/24,2a06:98c1:58::1f6,2a06:98c1:58::1f6/128,2a06:98c1:58::/64'
+```
+
+Table below explains how each of the items in the environment variable above would impact an AVD decision tree of selecting proxy bypass settings:
+
+!!! Note
+
+    Convention `<fqdn/ip>:<port>` used in `Matching CloudVision destinations` and `Non-matching CloudVision destinations` columns of the table below means that AVD is instructed to run deployment against CloudVision / CVaaS `<fqnd/ip>` over port `<port>`. Non-standard port assumes usage of an intermediate proxy/load-balancer.
```
```suggestion
    Convention `<fqdn/ip>:<port>` used in `Matching CloudVision destinations` and `Non-matching CloudVision destinations` columns of the table below means that AVD is instructed to run deployment against CloudVision / CVaaS `<fqdn/ip>` over port `<port>`. Non-standard port assumes usage of an intermediate proxy/load-balancer.
```

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -206,6 +143,14 @@ def _set_version(self) -> None:
             msg = f"Unable to get version from CloudVision server. Got {response.text if response else 'No response'}"
             raise CVClientException(msg) from e
 
+    @property
+    def use_proxy(self) -> bool:
```
ClausHolbechArista: Should this be _use_proxy? Or do we expect external consumers to read this?

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -52,7 +49,7 @@ class CVClientProtocol(
     _username: str | None
     _password: str | None
     _cv_version: CvVersion | None = None
-    _proxy_manager: HTTPProxyManager | None = None
+    cv_connection_manager: CVConnectionManager
```
ClausHolbechArista: I don't think we want the connection manager to be an external API, so lets prefix with underscore.

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -85,8 +357,287 @@ async def create_socket_for_grpc(self) -> socket.socket:
...
+        """Verifies validity of the requested proxy server scheme."""
+        if not (result := proxy_scheme_candidate == "http") and raise_on_failure:
+            msg = f"Scheme '{proxy_scheme_candidate}' of the proxy server requested via '{candidate_source}' is not supported by AVD."
+            raise AristaAvdInvalidInputsError(msg)
+
+        return result
+
+    @staticmethod
+    def _proxy_host_is_valid(
+        proxy_host_candidate: str | None, candidate_source: T_ProxyConfigurationSource | None, raise_on_failure: bool = False
```
ClausHolbechArista, line 407: Why do we need the raise_on_failure? This is a validation method and it should either always raise or never raise. It would probably be simplest to just return the boolean and pass fewer things here, and then emit the proper error with the known context in the calling function. But maybe I am missing something.

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Used suggested port range format and removed `Note` as it was indeed saying the same thing

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Applied suggestion

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: My understanding is that this is more a convention and is not regulated by any RFC. Therefore packages/libraries are not obliged to use the same logic/etc (like some support uppercase env variable although others don't) although do follow it in most cases. Because of this i just wanted to be very clear about actual proxy-related behavior implemented here for AVD so that users don't need to guess or rely on other tools (like curl or requests) hoping that their behavior would match us 100%

ansible_collections/arista/avd/roles/cv_deploy/README.md

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Fixed, changed to private attribute `_ cv_connection_manager `

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Was just thinking to publicly expose this property so that it's known if CVClient is using proxy or not, but can for sure make it private. Changed it to private

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: I wanted to reuse the same validation methods (for scheme, fqdn, port, etc) for both explicitly passed parameters and those learned via ENV.

My way of thinking was to raise (`raise_on_failure=True`) only when validation is failed for explicitly passed parameters (like user is trying to instantiate CVClient explicitly telling to use proxy port 0) as in this case this data is explicitly passed to cv_deploy and we just need to let user know that what was requested is not correct. 

But at the same time i wanted to be more graceful (`raise_on_failure=False`) when validating URIs learned via ENV because in this case proxy-related ENVs may be configured system-wide and for different purposes and i just didn't want to fail AVD execution if there is an incorrect proxy-related ENV var that was actually set for something else (non AVD)

alexeygorbunov:
Moved most of the proxy-related cv_deploy documentation into the separate How-To guide

mkdocs.yml
```diff
@@ -198,6 +198,7 @@ nav:
       - Custom Structured Configuration: ansible_collections/arista/avd/roles/eos_designs/docs/how-to/custom-structured-configuration.md
       - Generate CloudVision Tags: ansible_collections/arista/avd/roles/eos_designs/docs/how-to/cloudvision-tags.md
       - Migrate to cv_deploy: docs/howto/migrate-to-cv-deploy.md
+      - Proxy server support in cv_deploy: docs/howto/cv_deploy/proxy.md
```
JulioPDX:
```suggestion
      - Proxy Server Support in cv_deploy: docs/howto/cv_deploy/proxy.md
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -326,6 +326,28 @@ ansible_collections/arista/avd/roles/cv_deploy/defaults/main/directories.yml
 --8<--
 ```
 
+#### Proxy server support
+
+The `arista.avd.cv_deploy` role supports connecting to CloudVision through an **HTTP CONNECT** proxy server.
+
+Below settings allow modifying the default proxy-related behavior. The values below are the default values.
```
JulioPDX:
```suggestion
Use the settings below to modify proxy behavior. The fields currently display the default values
```

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -326,6 +326,28 @@ ansible_collections/arista/avd/roles/cv_deploy/defaults/main/directories.yml
...
+proxy_port: 8080
+
+# Set authentication username for the HTTP CONNECT proxy server.
+# proxy_username: <str>
+
+# Set authentication password for the HTTP CONNECT proxy server.
+# proxy_password: <str>
+```
+
+`cv_deploy` role can also discover proxy server settings using environment variables. Please refer to the [Proxy server support in cv_deploy](../../../../../docs/howto/cv_deploy/proxy.md) how-to guide for additional details.
```
JulioPDX:
```suggestion
`cv_deploy` role can also discover proxy server settings using environment variables. Please refer to the [Proxy Server Support in cv_deploy](../../../../../docs/howto/cv_deploy/proxy.md) how-to guide for additional details.
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
+<!--
+  ~ Copyright (c) 2026 Arista Networks, Inc.
+  ~ Use of this source code is governed by the Apache License 2.0
+  ~ that can be found in the LICENSE file.
+  -->
+
+# Proxy server support in cv_deploy
```
JulioPDX:
```suggestion
# Proxy Server Support in cv_deploy
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+The `arista.avd.cv_deploy` role supports connecting to CloudVision through an [HTTP CONNECT](https://en.wikipedia.org/wiki/HTTP_tunnel#HTTP_CONNECT_method) proxy server, with or without basic authentication.
+
+!!! Warning
+
+    Authentication credentials (when used) are sent to the proxy server using ***HTTP Basic authentication*** over non-encrypted HTTP transport (credentials are only `Base64` encoded, not encrypted). Proxy server credentials can be exposed by intercepting and analyzing raw TCP/IP traffic between AVD and Proxy server. Please always use other filtering and identification mechanisms (like HTTP filtering based on the client's SRC IP, requested destination domains, etc.) to limit the security risks.
```
JulioPDX:
```suggestion
    Authentication credentials (when used) are sent to the proxy server via ***HTTP Basic authentication*** over an unencrypted HTTP connection (credentials are only Base64-encoded, not encrypted). Proxy server credentials can be exposed by intercepting and analyzing raw TCP/IP traffic between AVD and the proxy server. Please always use additional filtering and identification mechanisms (such as HTTP filtering based on the client’s SRC IP, requested destination domains, etc.) to mitigate security risks.
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+    It is important to note that plain HTTP is used by AVD only for the initial CONNECT request to establish the tunnel with the CloudVision through proxy server. Once the TCP tunnel to CloudVision through proxy server is active, all subsequent AVD communication — including both REST and gRPC calls — is protected within a secure TLS session(s) established between AVD and CloudVision ***inside*** the TCP proxy tunnel.
```
JulioPDX:
```suggestion
    It is important to note that AVD uses plain HTTP only for the initial CONNECT request to establish a tunnel to CloudVision through the proxy server. After the TCP tunnel to CloudVision through the proxy server is active, all subsequent AVD communication — including both REST and gRPC calls — is protected within a secure TLS session(s) established between AVD and CloudVision ***inside*** the TCP proxy tunnel.
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+There are two ways to enable proxy server for `cv_deploy`: explicit and via environment variables.
+
+## Configure proxy settings explicitly
+
+To enable proxy server explicitly, set `proxy_host`. Setting `proxy_port` (port `TCP/8080` will be used by default), `proxy_username` and `proxy_password` is optional.
+
+If valid `proxy_host` (must be non-empty string) and `proxy_port` (1-65535) are set, `cv_deploy` will ignore (for both REST and gRPC calls) any proxy-related environment variables (`https_proxy`/`HTTPS_PROXY`/`all_proxy`/`ALL_PROXY`/`no_proxy`/`NO_PROXY`) and will force all REST and gRPC egress connections through this proxy.
+
+Example of the configuration to use unauthenticated HTTP proxy using CONNECT method:
```
JulioPDX:
```suggestion
Example configuration to use an unauthenticated HTTP proxy using the CONNECT method:
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+```yaml
+proxy_host: proxy.local.domain
+proxy_port: 3128
+```
+
+Example of the configuration to use authenticated HTTP proxy using CONNECT method:
```
JulioPDX:
```suggestion
Example configuration to use an authenticated HTTP proxy using the CONNECT method:
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+    Example:
+
+    ```yaml
+    proxy_host: proxy.local.domain
+    proxy_port: 3128
+    proxy_username: "p:r/o$x@yuser"
+    proxy_password: "p:r/o$x@ypassword"
+    ```
+
+    will be equal to setting the following via the environment variable: `http://p%3Ar%2Fo%24x%40yuser:p%3Ar%2Fo%24x%40ypassword@proxy.local.domain:3128`
```
JulioPDX:
```suggestion
    Will be equal to setting the following via the environment variable: `http://p%3Ar%2Fo%24x%40yuser:p%3Ar%2Fo%24x%40ypassword@proxy.local.domain:3128`
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Configure proxy settings using environment variables
+
+If proxy-related settings are not passed to `cv_deploy` explicitly, `cv_deploy` will try to discover usable proxy server (scheme is `http`, host is a non-empty string, port is in the range 1-65535) using environment variables in the following order:
```
JulioPDX:
```suggestion
If proxy-related settings are not passed to `cv_deploy` explicitly, `cv_deploy` will try to discover a usable proxy server (scheme is `http`, host is a non-empty string, port is in the range 1-65535) using environment variables in the following order:
```

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+1. (Step 1) Check if proxy bypass is requested for CloudVision
```
JulioPDX: Mentioning Step 1 and Step 2 seems unnecessary in an ordered list

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,164 @@
...
+
+    ```mermaid
```
JulioPDX, line 62: I'm also not sure if a user needs the decision tree when setting proxy settings, this could be represented by which takes precedence but the team may feel different.

docs/howto/cv_deploy/proxy.md
alexeygorbunov: True, removed `(Step x)`

JulioPDX (approved):
LGTM

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Also removed large table with no_proxy comparison examples for different patterns

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Removed `raise_on_failure `. Logic will raise on incorrect proxy settings for both explicit and env-based approaches

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -3,80 +3,352 @@
...
+
+        if not self.proxy_discovered:
+            LOGGER.info("<CVProxyManager>: No proxy environment variables discovered. AVD will not use proxy server.")
+            self.use_proxy = False
+            return
+
+        LOGGER.info(
+            "<CVProxyManager>: Proxy settings were discovered using environment variable '%s'.",
+            self._env_var_proxy_name,
+        )
```
alexeygorbunov, line 269: @carl-baillargeon this is regarding your question if info logs are emitted to inform about working through proxy

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: All these extra details were moved to how-to guide

ansible_collections/arista/avd/roles/cv_deploy/README.md
```diff
@@ -439,6 +439,28 @@ cv_deploy_validate_inputs_batch_size: 10
     cv_strict_system_mac_address: true
     ```
 
+#### Proxy server support
+
+The `arista.avd.cv_deploy` role supports connecting to CloudVision through an **HTTP CONNECT** proxy server.
```
gmuloc, line 577: somehow I feel like the warning in here made sense (though it is important in the other one as well) given that it warns of a potential credential breach and people may stop here.

We can decide to ignore this comment and merge :)

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -3,80 +3,352 @@
...
+    def _parse_env_var_proxy_content(self) -> None:
+        LOGGER.debug("<CVProxyManager>: Parsing content of the environment variable '%s'...", self._env_var_proxy_name)
+        try:
+            self._parsed_env_var_proxy_content = urlparse(self._env_var_proxy_content)
+            # urlparse will success parsing any URL <str> but will raise at access time if port value is invalid
+            # Try accessing proxy port attribute to catch any issues with its value
+            _ = self._parsed_env_var_proxy_content.port
+        except Exception as e:
+            msg = (
+                f"AVD faced an exception trying to extract proxy server settings from string '{self._env_var_proxy_content}' learned using environment"
```
gmuloc: we should not output _env_var_proxy_content as it could leak the password for proxy in logs / exception message

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,436 @@
...
+http://proxy-server.local:8081
+http://user1:pass1@10.10.10.10:8081
+http://user1:pass1@proxy-server.local:8081
+```
+
+!!! Note
+
+    When setting proxy credentials via environment variables, make sure to quote all special symbols (like `:`, `@`, etc. ) otherwise such proxy URI string will be invalid.
+
+Examples below show invalid values of `https_proxy`/`HTTPS_PROXY`/`all_proxy`/`ALL_PROXY` environment variables which will be ignored by `cv_deploy`:
```
gmuloc: this is not accurate with the mermaid chart and the code as we would raise exceptions upon wrong values

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -85,8 +357,296 @@ async def create_socket_for_grpc(self) -> socket.socket:
...
+        LOGGER.debug("<CVProxyBypassManager>: Reading proxy bypass environment variables: '%s'...", NO_PROXY_ENV_VARS)
+        for env_variable_candidate in NO_PROXY_ENV_VARS:
+            if env_no_proxy_candidate := environ.get(env_variable_candidate):
+                LOGGER.debug("<CVProxyBypassManager>: Environment variable '%s' is found.", env_variable_candidate)
+                self._env_var_no_proxy_name = env_variable_candidate
+                self._env_var_no_proxy_content = env_no_proxy_candidate
+                return
+        LOGGER.debug("<CVProxyBypassManager>: No proxy bypass environment variables found.")
+
+    def _process_env_var_no_proxy_content(self) -> None:
```
gmuloc, line 530: should we accept

```
NO_PROXY="localhost, www.arista.io"
```

so handling spaces (we split on "," but we could have spaces

ansible_collections/arista/avd/roles/cv_deploy/README.md
alexeygorbunov: Added same warning section under role's configuration but now this can trigger "duplication" concerns)

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: true, removed env content from the message

docs/howto/cv_deploy/proxy.md
alexeygorbunov: thanks, this is matching original "graceful" logic which was later changed to strict (fail on error). Reworded ths sentense

python-avd/pyavd/_cv/client/proxy.py
alexeygorbunov: Carl already asked:) https://github.com/aristanetworks/avd/pull/6082#discussion_r2479553641

Interesting that `urllib`'s doc clearly states it requires comma-separated list of hostnames: 

https://docs.python.org/3/library/urllib.request.html

```
The no_proxy environment variable can be used to specify hosts which shouldn’t be reached via proxy; if set, it should be a comma-separated list of hostname suffixes, optionally with :port appended, for example cern.ch,ncsa.uiuc.edu,some.host:8080.
```

But in the end they seem to still strip (and allow hostnames with whitespaces): https://github.com/python/cpython/blob/8cad740557f3a0860497bde5b02ad9c2b8ea9cf6/Lib/urllib/request.py#L1933

```bash
>>> from urllib.request import proxy_bypass_environment
>>> proxy_bypass_environment("host1.com", {"no": "localhost"})
False
>>> proxy_bypass_environment("host1.com", {"no": "localhost,host1.com"})
True
>>> proxy_bypass_environment("host1.com", {"no": "localhost, host1.com"})
True
>>> proxy_bypass_environment("host1.com", {"no": "localhost,  host1.com "})
True
>>> proxy_bypass_environment("host1.com", {"no": "localhost,  host2.com "})
False
>>> 
```

alexeygorbunov: Added stripping for no_proxy/NO_PROXY items and updated test cases.
Not mentioning this in the documentation (just like urllib) though (expecting users to enter clear inputs)

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -3,80 +3,349 @@
...
+    _proxy_configuration_source: T_ProxyConfigurationSource
+    """Source of the proxy server information."""
+    _env_var_proxy_name: T_ProxyEnvVars | None
+    """Name of the selected environment variable."""
+    _env_var_proxy_content: str | None
+    """Raw content of the selected environment variables."""
+    _parsed_env_var_proxy_content: ParseResult
+    """Parsed proxy server information."""
+
+    def __init__(
```
gmuloc, line 193: this function is very big - we should probably look at splitting up so it is more manageable in the future

python-avd/pyavd/_cv/client/proxy.py
```diff
@@ -85,8 +354,296 @@ async def create_socket_for_grpc(self) -> socket.socket:
...
+    def proxy_bypass_discovered(self) -> bool:
+        return bool(self._get_env_var_no_proxy_name() and self._get_env_var_no_proxy_content())
+
+    def _get_env_var_no_proxy_name(self) -> T_NoProxyEnvVars | None:
+        return getattr(self, "_env_var_no_proxy_name", None)
+
+    def _get_env_var_no_proxy_content(self) -> str | None:
+        return getattr(self, "_env_var_no_proxy_content", None)
+
+    def bypass_proxy_for_destination(self, target_host: str, target_host_format: T_HostFormat, target_port: int) -> bool:
```
gmuloc, line 560: need to verify we document that we are case sensitive in the doc

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -268,10 +214,11 @@ def __init__(
             password: Password to use for authentication if token is not set.
             port: TCP port to use for the connection.
             verify_certs: Disables SSL certificate verification if set to False. Not recommended for production.
-            proxy_host: HTTP proxy hostname.
-            proxy_port: HTTP proxy port.
-            proxy_username: Proxy authentication username.
-            proxy_password: Proxy authentication password.
+            proxy_scheme: Proxy server scheme (http/https).
```
gmuloc, line 217: is this correct (https will be blocked right?)

gmuloc: probably worth using Literal in this case for typing + fixing the docstring

python-avd/tests/pyavd/cv/client/test_cv_client_proxy_live.py
```diff
@@ -0,0 +1,324 @@
...
+
+
+@pytest.mark.skipif(environ.get("CV_LIVE_TEST") is None, reason="CV_LIVE_TEST env variable is not set. Live cv_deploy proxy tests are skipped.")
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("targeted_cv"),
+    [
+        pytest.param(
+            {
+                "cv_access_token": environ.get("CV_PRD_ACCESS_TOKEN", default=""),
```
gmuloc, line 27: hmm wondering if all of these would not benefit from the cv_environment fixture and if we could not do the same stuff for proxy variables. this can be a TODO

ansible_collections/arista/avd/roles/cv_deploy/README.md
gmuloc, line 1: the feature applies to cv_workflow too, but the detailed how-to is cv_deploy-
only. I would add a short note in cv_workflow docs saying the same env-var discovery applies when proxy settings are not explicitly passed.

docs/howto/cv_deploy/proxy.md
```diff
@@ -0,0 +1,436 @@
...
+    proxy_password: "p:r/o$x@ypassword"
+    ```
+
+    Will be equal to setting the following via the environment variable: `http://p%3Ar%2Fo%24x%40yuser:p%3Ar%2Fo%24x%40ypassword@proxy.local.domain:3128`
+
+## Configure proxy settings using environment variables
+
+If proxy-related settings are not passed to `cv_deploy` explicitly, `cv_deploy` will try to discover a usable proxy server (scheme is `http`, host is a non-empty string, port is in the range 1-65535) using environment variables in the following order:
+
+1. Check if proxy bypass is requested for CloudVision
```
gmuloc, line 60: I feel this table would work better and ask less from the reader:


```
• | Priority | Setting | Behavior |
  | -------- | ------- | -------- |
  | 1 | Explicit `proxy_host`, `proxy_port`, `proxy_username`, `proxy_password` | Explicit settings win and ignore all proxy-related
  environment variables. |
  | 2 | `no_proxy` | If CloudVision matches, `cv_deploy` does not use a proxy. |
  | 3 | `NO_PROXY` | Used only if `no_proxy` is not set. If CloudVision matches, `cv_deploy` does not use a proxy. |
  | 4 | `https_proxy` | First proxy server candidate. |
  | 5 | `HTTPS_PROXY` | Used only if `https_proxy` is not set. |
  | 6 | `all_proxy` | Used only if no HTTPS proxy variable is set. |
  | 7 | `ALL_PROXY` | Last proxy server candidate. |
```

gmuloc: and then remove the very nice but more requiring mermaid graph

PR #6083, opened 2025-10-30
Refactor(anta_runner): Enable LLDP validation for external (non-AVD) peers

### LINKED ISSUE #5971: Feat(anta_runner): Add a knob to check LLDP neighborship even if a device is not in the inventory
### Enhancement summary

Today Anta tests for LLDP Neighborship are only created for devices which are in the inventory. This will never work for non-Arista devices which speak LLDP and are connected to the Arista Fabric. Therefore it would be good to have a knob to check the LLDP neighborship for connected_endpoints even if they are not in the inventory.  

anta_runner

### Use case example

- connected_endpoints defining non-Arista devices
- connected_endpoints defining Arista devices in a different inventory (separate inventories for DC, Campus, OOB, etc.)

If under the connected_endpoint the "validate_lldp" is set and the peer and peer_interface are defined, validate LLDP neighborship even if not in the inventory

_No response_

### Additional context

_No response_

### Contributing Guide

MaheshGSLAB:
## Change Summary

Enable LLDP validation for external (non-AVD) peers

## Related Issue(s)

Fixes #5971 

## Component(s) name

`arista.avd.anta_runner`

## Proposed changes
Updated the connected endpoint eos_designs logic for validate_lldp:
Previously, the value was being set to None when the input was either None or True. Now, if the input is None, it remains None, and if the input is True, it is explicitly set to True. The behavior for False remains unchanged.

In anta_runner, the logic has been updated so that when validate_lldp is True, the peer availability and peer interface shutdown checks are skipped.

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


vitthalmagadum (approved):

carl-baillargeon (changes requested):
Blocking this until https://github.com/aristanetworks/anta/issues/1370 and https://github.com/aristanetworks/anta/issues/1368 merged and ANTA v1.7 is released.

carl-baillargeon:
@MaheshGSLAB ANTA v1.7.0 has been released, we can finalize this one :) Thanks!

python-avd/pyavd/_anta/input_factories/connectivity.py
```diff
@@ -28,6 +28,7 @@ class Metadata(Protocol):
 
             peer: str
             peer_interface: str
+            validate_lldp: bool
```
carl-baillargeon: No longer needed per my suggestion below.
```suggestion
```

python-avd/pyavd/_anta/input_factories/connectivity.py
```diff
@@ -57,9 +58,14 @@ def create(self) -> Iterator[VerifyLLDPNeighbors.Input]:
...
             peer_interface = intf.metadata.peer_interface
+            validate_lldp = intf.metadata.validate_lldp
 
             # LLDP neighbor is the FQDN when dns domain is set in EOS
-            fqdn = f"{peer_name}.{dns_domain}" if (dns_domain := self.fabric_data.devices[peer_name].dns_domain) is not None else peer_name
+            fqdn = (
+                f"{peer_name}.{dns_domain}"
+                if validate_lldp is not True and (dns_domain := self.fabric_data.devices[peer_name].dns_domain) is not None
+                else peer_name
+            )
```
carl-baillargeon: We added a knob in ANTA to ignore the FQDN so this is no longer required. The `create` method can be simplified a lot:
```
    def create(self) -> Iterator[VerifyLLDPNeighbors.Input]:
        """Generate the inputs for the `VerifyLLDPNeighbors` test."""
        neighbors = [
            LLDPNeighbor(port=intf.name, neighbor_device=intf.metadata.peer, neighbor_port=intf.metadata.peer_interface)
            for intf in self.structured_config.ethernet_interfaces
            if self._is_interface_candidate(intf)
        ]

        if not neighbors:
            self.logger_adapter.debug(LogMessage.NO_INPUTS_GENERATED)
            return

        yield VerifyLLDPNeighbors.Input(neighbors=natural_sort(neighbors, sort_key="port"), require_fqdn=False)
```
Notice the `require_fqdn=False`.

python-avd/pyavd/_anta/input_factories/connectivity.py
```diff
@@ -57,9 +58,14 @@ def create(self) -> Iterator[VerifyLLDPNeighbors.Input]:
 
             peer_name = intf.metadata.peer
             peer_interface = intf.metadata.peer_interface
+            validate_lldp = intf.metadata.validate_lldp
```
carl-baillargeon: These are no longer needed per my suggestion below.
```suggestion
```

python-avd/pyavd/_anta/input_factories/connectivity.py
```diff
@@ -87,10 +93,14 @@ def _is_interface_candidate(self, interface: EosCliConfigGen.EthernetInterfacesI
             self.logger_adapter.debug(LogMessage.INPUT_MISSING_FIELDS, identity=interface.name, fields="metadata.peer, metadata.peer_interface")
             return False
 
-        if not self.is_peer_available(interface.metadata.peer, identity=interface.name):
-            return False
+        if not interface.metadata.validate_lldp:
```
carl-baillargeon: Let's be more clear here.
```suggestion
        # Default behavior (None) implies we treat the peer as an AVD-managed device in the inventory
        treat_as_avd_peer = interface.metadata.validate_lldp is None
        if treat_as_avd_peer:
```

python-avd/pyavd/_anta/input_factories/connectivity.py
carl-baillargeon: Please confirm but I think we can also remove the `dns_domain` field from the class since it is no longer used.

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
```diff
@@ -200,7 +200,7 @@ def _get_ethernet_interface_cfg(
             port_profile=adapter.profile,
             peer_key=connected_endpoint._internal_data.context,
             validate_state=False if adapter.validate_state is False else None,
-            validate_lldp=False if adapter.validate_lldp is False else None,
+            validate_lldp=adapter.validate_lldp,
```
carl-baillargeon, line 203: We need to update the schema descriptions of `validate_lldp` to explain the new behaviors (3 scenarios). Valid for both eos_design and eos_cli_config_gen schemas.

python-avd/pyavd/_anta/input_factories/connectivity.py
carl-baillargeon, line 1: Please update the class docstring with the new changes.

carl-baillargeon (changes requested):

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: removed

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: In ANTA we have require_fqdn field and it treat as 
```yaml
    anta.tests.connectivity:
      - VerifyLLDPNeighbors:
          require_fqdn: False
          neighbors:
            - port: Ethernet1
              neighbor_device: DC1-SPINE1
              neighbor_port: Ethernet1
            - port: Ethernet2
              neighbor_device: dc1-leaf2-server1
              neighbor_port: iLO
      - VerifyLLDPNeighbors:
          require_fqdn: True
          neighbors:
            - port: Ethernet1
              neighbor_device: DC1-SPINE1.local.com
              neighbor_port: Ethernet1
    ```
    so when require_fqdn is true we use neighbor device with domain names. I am not getting why we keep require_fqdn: False static in our code. In ANTA default as True.

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: removed

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/connected_endpoints/ethernet_interfaces.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: updated

python-avd/pyavd/_anta/input_factories/connectivity.py
MaheshGSLAB: Ok got it in anta code we are checking info["systemName"].startswith(f"{neighbor.neighbor_device}.") in case of  require_fqdn: False so yes there are catalog changes in AVD side but it handles in anta.

carl-baillargeon (approved):
LGTM. Tested locally with connected endpoints.

PR #6084, opened 2025-10-30
Refactor(pyavd): Add get_validated_data to pyavd-utils

ClausHolbechArista:
> [!WARNING]
> ~let's merge #6112 first and then try to get this one in merge queue~

Adding a new validation function to the rust based schema validation:

```python

def get_validated_data(data_as_json: str, schema_name: Literal["eos_cli_config_gen", "eos_designs"]) -> GetValidatedDataResult:
    """
    Validate data against a schema specified by name and return the data after coercion and validation.

    This function includes the coerced data encoded as JSON in the return value.
    This data also contains default values that got inserted during validation.

    Args:
        data_as_json: Structured data dumped as JSON.
        schema_name: The name of the schema to validate against.

    Returns:
        GetValidatedDataResult holding the coerced data and the ValidationResult with lists of violations and coercions as Feedback objects.
    """
```


rust/validation/src/validation/store.rs
```diff
@@ -115,6 +123,23 @@ impl StoreValidate<&str> for Store {
...
         }
     }
+    fn validate_value(
+        &self,
+        value: &mut Value,
+        schema_name: &str,
+        configuration: Option<&Configuration>,
+    ) -> ValidationResult {
+        if let Ok(schema_type) = Schema::try_from(schema_name) {
+            self.validate_value(value, schema_type, configuration)
```
gmuloc: so here we now have a schema_type that is the schema_name for the other implementation :) I understand whats going on and thats just variable names but :p

gmuloc (approved):

pyavd-utils/tests/validation/test_get_validated_data.py
```diff
@@ -0,0 +1,44 @@
...
+    assert validated_data == (
+        '{"ethernet_interfaces":[{"name":"Ethernet1","unknown":12345,"ospf_authentication_key_type":"7"}],"avd_data_validation_mode":"error",'
+        '"config_end":false,"generate_default_config":false,"generate_device_documentation":true,"transceiver_qsfp_default_mode_4x10":true}'
+    )
+    validation_result = coercion_and_validation_result.validation_result
+    assert len(validation_result.violations) == 1
+    feedback = validation_result.violations[0]
+    assert feedback.path == ["ethernet_interfaces", "0", "unknown"]
+    assert isinstance(feedback.issue, Issue.Validation)
+    assert isinstance(feedback.issue._0, Violation.UnexpectedKey)
```
carl-baillargeon, line 44: While testing I noticed that `Violation.UnexpectedKey` didn't have the key attribute:
```bash
(.dev) ~/git_projects/avd/pyavd-utils (refactor/pyavd-utils/add-coerce-and-validate-json-function ✔) python3 ~/git_projects/stuff/avd/schema_testing.py
ip_address
  File "/home/carl/git_projects/stuff/avd/schema_testing.py", line 20, in <module>
    print(v.issue._0.key)
          ^^^^^^^^^^^^^^
AttributeError: 'validation.Violation_UnexpectedKey' object has no attribute 'key'
```
If we translate this data to upper layers in our code it might be useful to know which key was unexpected maybe?

carl-baillargeon: Ah I see there is the `path` attribute available that would include the unexpected key.

carl-baillargeon (approved):
LGTM! Tested locally with different valid/invalid configs.

PR #6089, opened 2025-10-31
Refactor(eos_designs)!: Use re.fullmatch instead of re.match for schema pattern-based validation

alexeygorbunov:
## Change Summary

Use re.fullmatch instead of re.match for schema pattern-based validation

TODOs:
  - Document new restrictions/patterns in 6.0.0 porting guide?

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.eos_designs`

## How to test
`eos_designs_unit_tests` molecule scenario

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -1027,7 +1027,7 @@ $defs:
                             description: |-
                               Ethernet interface name like 'Ethernet2'.
                               Member interface cannot be subinterface.
-                            pattern: "Ethernet[\\d/]+"
+                            pattern: "Ethernet(\\d+/){0,2}\\d+"
```
ClausHolbechArista: we cannot say how many digits and slashes may be needed. For now just keep the existing patterns if they works. We want to move this to a "format" in the future, so we can maintain a pattern for ethernet interfaces in one place.

alexeygorbunov: Relaxed patterns for Eth/Po interfaces in terms of the possible depth (chassis, module, etc.)
At the same time initial patterns seem to not be very precise as they match incorrect things

```
>>> re.fullmatch("Ethernet[\\d/]+", "Ethernet1/1/")
<re.Match object; span=(0, 12), match='Ethernet1/1/'>
>>>
```

ClausHolbechArista: I understand the current ones are not necessarily exact, but it is not the purpose of this PR to fix that all over. We need to work the proper patterns and implement that consistently across the board.

gmuloc: @alexeygorbunov - not for this PR but the proper pattern would be `pattern: "Ethernet\\d+(/\\d+)*"` to force starting by a number and then an optional pattern of `/x`

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type.schema.yml
```diff
@@ -212,7 +212,7 @@ $defs:
             type: list
             items:
               type: str
-              pattern: "Ethernet[\\d/]+"
+              pattern: "Ethernet\\d+(-\\d+)?(,\\d+(-\\d+)?)*(/\\d+(-\\d+)?(,\\d+(-\\d+)?)*)*"
```
alexeygorbunov: NOTE: This pattern must be changed. Key description claims support for the interface ranges. Current pattern ` "Ethernet[\\d/]+"` will not match any range

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type.schema.yml
alexeygorbunov, line 238: NOTE: This key does not claim support for interface ranges. Therefore not changing the pattern. Any values using interface ranges will fail validation.

alexeygorbunov, line 871: NOTE: No pattern-enforced check right now. Keeping it as is.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
alexeygorbunov: I rolled back all regex patterns changes that were not mandatory (put notes explaining why proposed pattern changes are needed in cases where it's not a simple `^` & `$` removal

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type.schema.yml
ClausHolbechArista: This will not match `"Ethernet1, Ethernet3"` which is also allowed. Let's simplify the pattern for now and then revisit later.

Not sure if this will work:
```suggestion
              pattern: "Ethernet[\\d/,(Ethernet) -]+"
```

ClausHolbechArista: we do support the ranges here. We do in many places. We validate the ranges intensively in the expansion logic, so maybe we could just do `Ethernet.*` for now.

alexeygorbunov: `range_expand` seems to not enforce any strict syntax or patterns

```
>>> from pyavd.j2filters import range_expand
>>> range_expand("Ethernet1-")
['Ethernet1']
>>> range_expand("Ethernet1-2/")
['Ethernet1', 'Ethernet2']
>>> range_expand("Abc1-2/")
['Abc1', 'Abc2']
>>>
```

alexeygorbunov: Relaxed this patterns to `Ethernet.*`

ClausHolbechArista (approved):
Thank you for this. We will improve the patterns to tighten the inbputs later.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/HARDWARE_FEATURES_ON_VIRTUAL_PLATFORMS.yml
```diff
@@ -259,9 +259,9 @@ l3leaf:
           uplink_switches:
             - HARDWARE_FEATURES_VEOS_LAB
           uplink_interfaces:
-            - Ethernet1.1
+            - Ethernet10
```
gmuloc, line 262: given that the test is to check feature_support for sflow subinterfaces disabled, changing this would decrease coverage right?

gmuloc: discussed in a call: still covered in l3_edge and network tenants

gmuloc (approved):
Need a follow up PR to update the patterns

PR #6113, opened 2025-11-04
Refactor(pyavd-utils): simplify invalid schema error

gmuloc:
> [!Warning]
> Need to be rebased on top of #6084

## Change Summary

Move parsing schema name to a helper

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

### User Checklist

- N/A

### Repository Checklist


rust/validation/src/validation/store.rs
```diff
@@ -88,68 +88,93 @@ impl StoreValidate<Schema> for Store {
...
+            Err(_) => Ok({
+                let mut ctx = Context::new(self, None);
+                ctx.add_violation(Violation::InvalidSchema {
+                    schema: schema_name.into(),
+                });
+                ctx.into()
+            }),
+        }
+    }
+    fn process_schema_string_unfaillible<F>(&self, schema_name: &str, func: F) -> ValidationResult
```
ClausHolbechArista: Needing to have two of these makes it clear that this pattern is broken. Let's try to find a better way. I am ok with removing the schema error from the violations and instead failing the result. This means we can change the return type on the validate_value to also be a result. And then this can be dropped and we just use try_from(string)?

ClausHolbechArista (approved):

PR #6130, opened 2025-11-10
CI(pyavd-utils): Fix pyright warnings

ClausHolbechArista:
- Fix pyright warnings caused by missing .so files for python-utils
- Elevate such warnings to errors so it blocks the CI
- ignore .tox in pyright - only relevant for local use where running tox first can trigger lots of errors from other libraries installed inside the tox venvs.


gmuloc (approved):
LGTM

PR #6135, opened 2025-11-11
Feat(eos_designs): Added support for TLS in aaa_settings.radius.servers

### LINKED ISSUE #6124: Feat(eos_designs): Add support for TLS in `aaa_settings.radius.servers`
Add support for TLS in `aaa_settings.radius.servers` items similar to `radius_server.hosts[].tls` in eos_cli_config_gen. If TLS is specified, key/cleartext_key must be ignored to avoid the following error:

```bash
fatal: [DC2-SPINE1 -> localhost]: FAILED! => {"changed": false, "msg": "'`aaa_settings.radius.servers[host=radsec.ag01c01.agni.arista.io].key` or `aaa_settings.radius.servers[host=radsec.ag01c01.agni.arista.io].cleartext_key`' is required but was not found."}
```

MaheshGSLAB:
## Change Summary

Added support for TLS in aaa_settings.radius.servers.

## Related Issue(s)

Fixes #6124 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Add support for TLS in aaa_settings.radius.servers items similar to radius_server.hosts[].tls in eos_cli_config_gen. If TLS is specified, key/cleartext_key must be ignored to avoid the following error:
```
fatal: [DC2-SPINE1 -> localhost]: FAILED! => {"changed": false, "msg": "'`aaa_settings.radius.servers[host=radsec.ag01c01.agni.arista.io].key` or `aaa_settings.radius.servers[host=radsec.ag01c01.agni.arista.io].cleartext_key`' is required but was not found."}
```

## How to test
Run Molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/aaa_settings.yml
```diff
@@ -135,6 +135,10 @@ keys:
                     To protect the password at rest it is strongly recommended to make use of a vault or similar.
                   convert_types:
                     - int
+                tls:
+                  type: dict
+                  description: When TLS is configured, `key` and `cleartext_key` are ignored.
```
laxmikantchintakindi, line 140: I would consider removing this logic in j2 template. IMO we should avoid EOS logics in template.

laxmikantchintakindi: s1-leaf1(config)#sh run | s radius
radius-server host 192.168.0.1 key 7 0207165218120E
radius-server host 11.10.10.5 vrf MGMT key 7 020706580F
radius-server host 11.10.10.5 vrf MGMT tls ssl-profile bacd
aaa group server radius atds
   server 192.168.0.1
ip radius source-interface Management0

laxmikantchintakindi: Both key and tls profile can be configured on EOS.

MaheshGSLAB: yes both keys are supported simultaneous now, @ClausHolbechArista @gmuloc is it ok to update the j2 template?

gmuloc: what do you mean by now @MaheshGSLAB - has it changed in some recent version of EOS? What versions have you tested upon?

gmuloc: also - is it the same for TACACS?

gmuloc: Works both for 4.35 and 4.33

```
localhost(config)#show run | se rad
radius-server host 1.1.1.1 key 7 070D2D4346051D090D01070D443E27376820262E5E17040C075F555D13551D020B
radius-server host 1.1.1.1 tls ssl-profile orororr
```

Lets fix the template in another PR so we can have a proper PR and porting guide entry

this was mentioned in the PR notes that we should not render key at the time: https://github.com/aristanetworks/avd/pull/4194 maybe worth investigate what has changed (maybe it was just a misunderstanding of what should go into eos_cli_config_gen)

gmuloc: discussed on call: let's not touch it on this PR and right now no one is requiring us to change. It is breaking but a nice to have

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
```diff
@@ -648,7 +648,7 @@ def radius_servers(self) -> None:
                 )
 
             server_key = self._get_tacacs_or_radius_server_password(server)
```
gmuloc: so this would raise is both key and cleartext_key are missing according to the function. But the schema is stating that you will ignore it. So we end up requiring something when actually it might not be needed right?

the only test you have configured is adding a key so we will not see this.

Maybe you should have something like `if not server.tls:` to guard sevrer_key but now we end up doing something uniquely because of the ignore in the schema which is meh...

Today we required key or cleartext key because we want traffic to be encrypted. if TLS removes this needs then we should change `_get_tacacs_or_radius_server_password` to not raise if tls is set.

@ClausHolbechArista for comment

Vibhu-gslab (approved):

gmuloc (changes requested):

python-avd/pyavd/_eos_designs/structured_config/base/__init__.py
gmuloc: discussed - let's do an `if not server.tls:` ... 
```suggestion
            if not server.tls:
                server_key = self._get_tacacs_or_radius_server_password(server)
```

(server_key won't be set for 651 so need more adjustment)

gmuloc:
@MaheshGSLAB can you check we have full test coverage on this afte rthis update?

MaheshGSLAB:
> @MaheshGSLAB can you check we have full test coverage on this afte rthis update?

@gmuloc Coverage is full but i updated the condition to check the tls.enabled as if enabled is false in that case we are still generating the tls structure config but j2 template will ignore it. So we need tls.enabled to get actual tls configuration.

gmuloc:
> > @MaheshGSLAB can you check we have full test coverage on this afte rthis update?
> 
> @gmuloc Coverage is full but i updated the condition to check the tls.enabled as if enabled is false in that case we are still generating the tls structure config but j2 template will ignore it. So we need tls.enabled to get actual tls configuration.

thanks!

gmuloc (approved):

PR #6154, opened 2025-11-13
Refactor: Improve Rust validation

ClausHolbechArista:
- Add deprecation warnings and errors (Still missing conflicts)
- Refactor the validation result model to have errors, warnings, infos


pyavd-utils/rust/pyvalidation/src/lib.rs
```diff
@@ -34,44 +34,41 @@ mod validation {
...
-                violations: value
-                    .violations
+                errors: value
+                    .errors
+                    .iter()
+                    .map(|feedback| Feedback {
+                        message: feedback.issue.to_string(),
+                        path: feedback.path.to_owned().into(),
+                    })
+                    .collect(),
```
gmuloc: we could write a helper function taking a Vec and returning the Vec of Feedback to avoid repeaying three times the full stuff here

pyavd-utils/tests/validation/test_get_validated_data.py
```diff
@@ -8,25 +8,20 @@
...
-        '"config_end":false,"generate_default_config":false,"generate_device_documentation":true,"transceiver_qsfp_default_mode_4x10":true}'
+        '"config_end":false,"transceiver_qsfp_default_mode_4x10":true}'
     )
     validation_result = coercion_and_validation_result.validation_result
-    assert len(validation_result.violations) == 0
-
-    assert len(validation_result.coercions) == len(expected_coercions)
-    for coercion in validation_result.coercions:
-        assert (coercion.path, coercion.message) in expected_coercions, f"Coercion not expected: {coercion.path}, {coercion.message}"
+    assert len(validation_result.errors) == 0
```
gmuloc: should we not assert on warning as well for completion?

pyavd-utils/pyavd_utils/validation.pyi
```diff
@@ -6,21 +6,35 @@
...
 class ValidationResult:
     """Result of data validation."""
 
-    violations: list[Feedback]
-    coercions: list[Feedback]
-    insertions: list[Feedback]
+    violations: list[Violation]
+    deprecations: list[Deprecation]
 
 class GetValidatedDataResult:
```
gmuloc: would be better if this was called `ValidatedDataResult`

pyavd-utils/rust/pyvalidation/benches/performance.rs
```diff
@@ -0,0 +1,49 @@
...
+            b.iter(|| get_validated_data(py, TEST_DATA, "eos_designs"))
+        )
+    });
+}
+
+criterion_group!(
+    benches,
+    benchmark_init_store_from_fragments,
+    benchmark_get_validated_data
+);
```
gmuloc, line 48: not sure I know much about criterion but why do we have some group stuff in one benchmark and not in the other? (sample size..)

pyavd-utils/rust/pyvalidation/src/lib.rs
```diff
@@ -19,60 +19,76 @@ fn get_store() -> &'static Store {
...
-                .for_each(|feedback| match feedback.issue {
-                    validation::feedback::Issue::Coercion(_) => coercions.push(feedback.into()),
-                    validation::feedback::Issue::DefaultValueInserted() => {
-                        insertions.push(feedback.into())
+                .for_each(|feedback| match &feedback.issue {
+                    validation::feedback::WarningIssue::Deprecated(deprecated) => {
+                        result.deprecations.push(Deprecation {
+                            message: deprecated.to_string(),
+                            path: feedback.path.to_owned().into(),
+                            removed: false,
```
gmuloc, line 84: Note to self: how do we handle `removed: true`?

rust/avdschema/src/schema/any.rs
```diff
@@ -52,3 +52,33 @@ impl From<&mut AnySchema> for String {
...
+
+impl Shortcuts for AnySchema {
+    fn is_required(&self) -> bool {
+        match self {
+            Self::Bool(schema) => schema.is_required(),
+            Self::Int(schema) => schema.is_required(),
+            Self::Str(schema) => schema.is_required(),
+            Self::List(schema) => schema.is_required(),
+            Self::Dict(schema) => schema.is_required(),
+        }
```
gmuloc: We could use a macro here
```rust
macro_rules! delegate {
    ($self:ident, $method:ident, $($arg:expr),*) => {
        match $self {
            Self::Bool(schema) => schema.$method($($arg)*),
            Self::Int(schema) => schema.$method($($arg)*),
            Self::Str(schema) => schema.$method($($arg)*),
            Self::List(schema) => schema.$method($($arg)*),
            Self::Dict(schema) => schema.$method($($arg)*),
        }
    };
}
```

and then

```
delegate!(self, is_required,)

File: rust/avdschema/src/utils/schema_from_path.rs [hunk 7]
```diff
@@ -74,31 +75,25 @@ impl SchemaKeys {
...
-                        })
-                        .collect::<Vec<_>>()
-                })
-                .unwrap_or_default(),
+                        .iter()
+                        .map(|(key, dynamic_key_info)| {
+                            (key.to_owned(), SchemaKey::from(dynamic_key_info))
+                        }).collect::<Vec<_>>()
+                }).into_iter().flatten()
+
```
gmuloc: here we create an unecessary internediate Vec:


```suggestion
                .get_dynamic_keys(dict)
                .unwrap_or_default()
                .into_iter()
                .map(|key, dynamic_keys_info| {
                     (key.to_owned(), SchemaKey::from(&dynamic_key_info))
                })
```

rust/validation/src/validation/any.rs
gmuloc, line 17: could use delegate macro here as well

rust/validation/src/validation/list.rs
```diff
@@ -96,18 +89,10 @@ fn validate_unique_keys(schema: &List, items: &[Value], ctx: &mut Context) {
         for (current_trail, current_item) in &items {
             for (trail, item) in &items {
                 if current_trail != trail && current_item == item {
-                    ctx.violations.push(Feedback {
-                        path: {
-                            let mut path = ctx.path.clone();
-                            path.extend_from_slice(current_trail);
-                            path
-                        },
+                    ctx.result.errors.push(Feedback {
```
gmuloc: I feel the compelxity of this method could be enhanced as right now we are going to loop over the items for each unique_key and then do a n square comparison - we could instead build a per unique key `Map <vlaue, list<trail>>`where we add the trail for each value (going through items only once) and then iter over the map and add a violation for any Map value

rust/validation/src/validation/store.rs
```diff
@@ -54,37 +54,49 @@ impl StoreValidate<Schema> for Store {
         schema_name: Schema,
         configuration: Option<&Configuration>,
     ) -> Result<ValidationResult, StoreValidateError> {
+        debug!("Validating JSON");
```
gmuloc, line 57: could add a edbug between from_str and validate to indicate when deserialization is done

rust/validation/src/validation/store.rs
```diff
@@ -54,37 +54,49 @@ impl StoreValidate<Schema> for Store {
...
+        debug!("Validating YAML Done");
+        Ok(result)
     }
     fn validate_value(
         &self,
         value: &mut Value,
         schema_name: Schema,
         configuration: Option<&Configuration>,
     ) -> ValidationResult {
+        debug!("Validating serde_yaml::Value");
```
gmuloc: is it not serde_json?

rust/validation/src/validation/store.rs
```diff
@@ -54,37 +54,49 @@ impl StoreValidate<Schema> for Store {
...
         let schema = self.get(schema_name);
         schema.coerce(value, &mut ctx);
+        debug!("Validating serde_yaml::Value Coercion Done");
         schema.validate_value(value, &mut ctx);
-        ctx.into()
+        debug!("Validating serde_yaml::Value Done");
+        ctx.result
     }
     fn coerce_value(&self, value: &mut Value, schema_name: Schema) -> ValidationResult {
+        debug!("Coercing serde_yaml::Value");
```
gmuloc: same here we are importing `use serde_json::Value;`

pyavd-utils/rust/pyvalidation/src/lib.rs
ClausHolbechArista: this was changed more or less to what you said.

pyavd-utils/tests/validation/test_get_validated_data.py
ClausHolbechArista: I was lazy, but things got changed inbetween so I think we cover all of them now.

pyavd-utils/pyavd_utils/validation.pyi
ClausHolbechArista: Ok, I will adjust

pyavd-utils/rust/pyvalidation/benches/performance.rs
ClausHolbechArista: Yes, we only use the group stuff when we need to. Because it takes so long to initialize the store, I had to lower the amount of samples for that benchmark.

pyavd-utils/rust/pyvalidation/src/lib.rs
ClausHolbechArista: This should be handled now, but if you find mistakes, let me know.
With the latest changes, it is handled 1. in coercion with an explicit check and 2. in validation where we try to get the schema for a key found in the data, and for the None case we add the Removed error.

rust/avdschema/src/utils/schema_from_path.rs
ClausHolbechArista: done and also improved the one above.

rust/validation/src/validation/dict.rs
```diff
@@ -2,116 +2,128 @@
...
-        for (key, key_schema) in keys {
-            match input.get(key) {
-                Some(Value::Null) | None => {
-                    // nullish values don't need to be validated beyond a requiredness check
+        let dynamic_keys_schemas = get_dynamic_keys_schemas(schema, input);
+        input.iter().for_each(|(input_key, input_value)| {
+            ctx.state.path.push(input_key.to_owned());
+            if let Some(key_schema) = keys.get(input_key) {
+                check_deprecation(input_key, key_schema, input, ctx);
+                key_schema.validate(input_value, ctx);
```
gmuloc: so here if check_deprecation is adding an error, we still validate the key? Often we don't have type information anymore right?

rust/validation/src/validation/dict.rs
```diff
@@ -2,116 +2,128 @@
...
+    if let Some(deprecation) = key_schema.deprecation()
+        && deprecation.warning
+    {
+        if deprecation.removed.unwrap_or_default() {
+            ctx.add_error(Violation::DeprecatedRemoved(Removed::from_schema(
+                &ctx.state.path,
+                deprecation,
+            )));
+        } else {
+            // TODO: Catch conflict.
```
gmuloc: what kind of conflict are you expecting?


EDIT:
Ah you mean catching conflicts between old and new key?

rust/validation/src/feedback.rs
```diff
@@ -56,28 +59,89 @@ impl From<&str> for Value {
...
+        self.0.is_empty()
+    }
+    pub(crate) fn clone_with_slice(&self, slice: &[String]) -> Self {
+        let mut new = self.clone();
+        new.0.extend_from_slice(slice);
+        new
+    }
+}
+impl std::fmt::Display for Path {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
gmuloc, line 88: maybe a nice docstring for this one :)

rust/validation/src/feedback.rs
```diff
@@ -56,28 +59,89 @@ impl From<&str> for Value {
...
+        let mut string = String::default();
+        for (index, element) in self.0.iter().enumerate() {
```
gmuloc, line 90: so for now it looks like the Path is not able to handle the IndexedList with primary key right?

rust/validation/src/feedback.rs
```diff
@@ -125,8 +189,13 @@ pub enum Violation {
     #[display("The value '{found}' is below the minimum allowed '{minimum}'.")]
     ValueBelowMinimum { minimum: i64, found: i64 },
     /// The value is not unique as required.
-    #[display("The value is not unique among similar items. Conflicting items: {other_path:?}")]
-    ValueNotUnique { other_path: Vec<String> },
+    #[display("The value is not unique among similar items. Conflicting item: {other_path}")]
+    ValueNotUnique { other_path: Path },
+    /// The input data model is deprecated and cannot be used in conjunction with the new data model.
+    #[display("The input data model is deprecated and cannot be used in conjunction with the new data model '{other_path}'.{url}")]
+    DeprecatedConflict { other_path: Path, url: UrlField },
```
gmuloc, line 205: so this is not implemented yet right?

rust/validation/src/feedback.rs
```diff
@@ -162,3 +231,229 @@ pub enum ViolationValidValues {
...
+            remove_in_version: Some("6.0.0".to_string()),
+            url: Some("my.url".to_string()),
+            ..Default::default()
+        }
+    }
+
+    #[test]
+    fn deprecated_from_schema() {
+        let deprecated =
+            Deprecated::from_schema(&Path::from_iter(["foo"]), &get_deprecation_test_schema());
```
gmuloc, line 448: this test is using an `avdschema::base::Deprecation` that is `Some(true)` for removed and still being able to convert. I understand this is how the code is built today where it just assumed that whatever is being passed is correct (that is it is indeed _only_ deprecated and not removed) but I feel like this could lead to some error laters given that it does not check that this is a valid object (in the sense deprecation and not remove) it is being asked to render

rust/validation/src/feedback.rs
```diff
@@ -162,3 +231,229 @@ pub enum ViolationValidValues {
...
+
+#[derive(Clone, Debug, PartialEq, Serialize, derive_more::Display)]
+#[display("The input data model '{path}' is deprecated and will be removed{version}.{replacement}{url}")]
+pub struct Deprecated {
+    pub path: Path,
+    pub replacement: ReplacementField,
+    pub version: VersionField,
+    pub url: UrlField,
+}
+impl Deprecated {
```
gmuloc, line 305: following my remark on the tests below, maybe we just make one struct here that does the job for both and honor the `removed` field from `avdschema::base::Deprecation`? Like `DeprecationInfo` or whatever?

rust/validation/src/feedback.rs
```diff
@@ -125,8 +189,13 @@ pub enum Violation {
...
     /// The value is not unique as required.
-    #[display("The value is not unique among similar items. Conflicting items: {other_path:?}")]
-    ValueNotUnique { other_path: Vec<String> },
+    #[display("The value is not unique among similar items. Conflicting item: {other_path}")]
+    ValueNotUnique { other_path: Path },
+    /// The input data model is deprecated and cannot be used in conjunction with the new data model.
+    #[display("The input data model is deprecated and cannot be used in conjunction with the new data model '{other_path}'.{url}")]
+    DeprecatedConflict { other_path: Path, url: UrlField },
+    /// Removed after deprecation of data model.
+    DeprecatedRemoved(Removed),
```
gmuloc:
```suggestion
    RemovedFromSchema(Removed),
```

rust/validation/src/feedback.rs
ClausHolbechArista: it handles it fine, but just like any list ;)

rust/validation/src/feedback.rs
ClausHolbechArista: made a short one.

rust/avdschema/src/schema/any.rs
ClausHolbechArista: I am not sure it will make the code much more readable, since it is not a standard macro. for now I would prefer to just keep the expanded code.

rust/validation/src/feedback.rs
ClausHolbechArista: not yet

rust/validation/src/feedback.rs
ClausHolbechArista: well it is not removed from the schema. it is still in the schema, but no longer accepted in the data.

rust/validation/src/feedback.rs
ClausHolbechArista: That may make it harder to match on in the LSP where we want different severities for these. I will look into it.

rust/validation/src/feedback.rs
ClausHolbechArista: That is already checked in the function calling this, and that check needs to stay there since it also controls if it goes into error or warning.

ClausHolbechArista: If we collapse to a single struct this doesn't matter. But the display templates becomes much more complex and have to be implemented in code instead of the derive shortcuts.

rust/validation/src/feedback.rs
gmuloc: hmm yeah ok so not the nice format :p

rust/validation/src/feedback.rs
gmuloc: ok so `Removed`

rust/avdschema/src/schema/any.rs
ClausHolbechArista: done

rust/validation/src/feedback.rs
ClausHolbechArista: this is what we do in the schema in general because the LSP also needs this to navigate. I can create a "Step" Enum which can either be a string, int or key+value pair -> later.

rust/validation/src/validation/dict.rs
ClausHolbechArista: we *always* have the type, but you are correct, if it is removed or a conflict we should stop there. I will work on this.

rust/validation/src/validation/any.rs
ClausHolbechArista: done

rust/validation/src/validation/dict.rs
ClausHolbechArista: yes this is the conflict between a deprecated model and the new model

rust/validation/src/feedback.rs
ClausHolbechArista: I will keep it as different items. It is only because they have the same origin in our schema model that we would consider collapsing them, but they are two very different "signals" so I prefer to keep them seperate.

PR #6169, opened 2025-11-20
Fix(cv_deploy): (Archived) Enable gRPC message splitting for set_tags, set_tag_assignments and delete_tag_assignments

### LINKED ISSUE #6163: Fix(cv_deploy): Enable gRPC message splitting to set_tags and set_tag_assignments
### Issue Summary

`GRPCRequestHandler` is currently not splitting messages that exceed max message size for `set_tags` / `set_tag_assignments` calls. Therefore return of the `Status.RESOURCE_EXHAUSTED` leads to the deployment failure instead of graceful retry with message splitting

cv_deploy

### AVD version

devel

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

None

### Steps to reproduce

```shell
[pyavd] - deploy_tags_to_cv: Got AAA tag assignments
[pyavd] - deploy_tags_to_cv: Creating BBB tag assignments
...
fatal: [<HIDDEN> -> localhost]: FAILED! => {"changed": true, "errors": ["(<Status.RESOURCE_EXHAUSTED: 8>, 'grpc: received message larger than max (1442369 vs. 1048576)', None)"], "warnings": []}
```

### Relevant log output

```shell

```

### Contributing Guide

alexeygorbunov:
## Change Summary

~! Needs https://github.com/aristanetworks/avd/pull/6225 merged first !~

Enable gRPC message splitting for `set_tags`, `set_tag_assignments` and `delete_tag_assignments`

## Related Issue(s)

Fixes #6163 

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- Change function signatures to accept list vs set
- Enable splitting for lists of tags and tag assignments

## How to test
Live pytest against PRD,STG and ONPREM (`python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py::test_deploy_tags_to_cv_message_splitting`)

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/client/tag.py
```diff
@@ -119,19 +119,19 @@ async def get_tags(
 
         return tags
 
-    @GRPCRequestHandler()
+    @GRPCRequestHandler(list_field="tags")
     async def set_tags(
         self: CVClientProtocol,
         workspace_id: str,
-        tags: set[CVTag],
+        tags: list[CVTag],
```
carl-baillargeon: For a future PR maybe we should think about having a broader type here and cast as a list in the decorator instead. This would provide more flexibility in the calling code (we wouldn't have to convert to a list).
```suggestion
        tags: Iterable[CVTag],
```

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@pytest.fixture(scope="module")
+def cv_tags() -> list[CVTag]:
+    tags_qty = 20000
+    return [CVTag(element_type="device", label=f"pytest-label-{tag_index}", value=f"pytest-value-{tag_index}") for tag_index in range(1, tags_qty + 1)]
+
+
+@pytest.fixture(scope="module")
+def cv_tag_assignments(cv_tags: set[CVTag]) -> list[CVTagAssignment]:
```
carl-baillargeon:
```suggestion
def cv_tag_assignments(cv_tags: list[CVTag]) -> list[CVTagAssignment]:
```

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,93 @@
...
+    device_id = "ABCDEFGHIGKLMNOP"
+    return [CVTagAssignment(element_type=cv_tag.element_type, label=cv_tag.label, value=cv_tag.value, device_id=device_id) for cv_tag in cv_tags]
+
+
+@pytest.mark.skipif(environ.get("CV_LIVE_TEST") is None, reason="CV_LIVE_TEST env variable is not set. Live cv_deploy tests are skipped.")
```
carl-baillargeon, line 803: Do we have plans to run these tests in CI somehow?

python-avd/pyavd/_cv/client/tag.py
ClausHolbechArista: since we need to know the length to be able to split, I think we need to use `Sequence[CVTag]`. Also an iteratable could be consumed after the first iteration worst case.

carl-baillargeon: Agreed Sequence is a better fit here.

alexeygorbunov: Please check my latest updates, i rolled back this changes (so async functions are using specific typing like list/set). I tried to modify async decorator to handle all functions (with `list`, `set` or `tuple` being a type of the parameter that needs to be splitted). This gives a flexibility of supporting different functions with different types of iterables at the same time. Added/adjusted tests as well to confirm functionality

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Outdated code

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Yes, there is a matching PR in AVD internal to run all pytests (including live ones pushing messages "larger than the limit" to PRD/STG and ON-PREM

Latest run example:
https://github.com/aristanetworks/avd-internal/actions/runs/19715197506/job/56485760524

alexeygorbunov:
```
# coverage report --include=pyavd/_cv/workflows/deploy_tags_to_cv.py --show-missing
Name                                       Stmts   Miss  Cover   Missing
------------------------------------------------------------------------
pyavd/_cv/workflows/deploy_tags_to_cv.py      46      0   100%
------------------------------------------------------------------------
TOTAL                                         46      0   100%
```

ClausHolbechArista:
@carl-baillargeon please take a new look at this one.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -121,59 +121,64 @@ class GRPCRequestHandler:
         max_retries (int): Maximum number of retry attempts for Status.UNAVAILABLE. Total attempts = 1 + max_retries.
         initial_delay (int): Initial delay in seconds before the first retry.
         factor (int): Multiplier for the delay in subsequent retries.
-        list_field (str): Name of the parameter to be split if Status.RESOURCE_EXHAUSTED is received.
+        iter_field (str): Name of the parameter to be split if Status.RESOURCE_EXHAUSTED is received.
```
carl-baillargeon: I think `split_field` is a better name.
```suggestion
        split_field (str): Name of the parameter to be split if Status.RESOURCE_EXHAUSTED is received.
```

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -121,59 +121,64 @@ class GRPCRequestHandler:
...
+        self.iter_field = iter_field
         self.min_items_for_splitting_attempt = max(2, min_items_for_splitting_attempt)
 
     def __call__(self, func: Callable[P, T]) -> Callable[P, T]:
         self.func = func
         self.func_signature = signature(func)
 
-        if self.list_field:
-            if not (return_annotation := self._is_list_annotation(self.func_signature.return_annotation, strict=True))[0]:
+        if self.iter_field:
```
carl-baillargeon: We don't need runtime validation since this decorator is used internally by developers. We can safely rely on Pyright. All these checks and `_is_sized_iterable_annotation` could be removed.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -121,63 +121,68 @@ class GRPCRequestHandler:
         max_retries (int): Maximum number of retry attempts for Status.UNAVAILABLE. Total attempts = 1 + max_retries.
         initial_delay (int): Initial delay in seconds before the first retry.
         factor (int): Multiplier for the delay in subsequent retries.
-        list_field (str): Name of the parameter to be split if Status.RESOURCE_EXHAUSTED is received.
+        iter_field (str): Name of the parameter to be split if Status.RESOURCE_EXHAUSTED is received.
```
alexeygorbunov: One of the suggestions was to use `Sequence` instead of an `Iterable` but `Set` is not a subtype of the Sequence

alexeygorbunov: Maybe Collection is a better option as it includes lists, sets and tuples and is sizable

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -271,72 +284,80 @@ async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs:
...
+        if not (self.collection_field and self.func_signature):
+            # No collection_field configured for splitting, execute the call directly (with retries)
             return await self._execute_single_call_with_retries(original_call_args, original_call_kwargs)
 
         bound_arguments = self.func_signature.bind(*original_call_args, **original_call_kwargs)
         current_arguments_dict = bound_arguments.arguments
 
-        list_value: list = current_arguments_dict.get(self.list_field, [])
-        if not isinstance(list_value, list):
+        collection_value: list | set | tuple = current_arguments_dict.get(self.collection_field, [])
```
alexeygorbunov:
```
        collection_value: list | set | tuple = current_arguments_dict.get(self.collection_field, [])
        if not any(isinstance(collection_value, allowed_type) for allowed_type in self.permitted_collection_field_types) or not collection_value:
            msg = (
                f"{self.__class__.__name__} decorator expected the value of the collection_field '{self.collection_field}' for function '{func_name}' "
                f"to be a non-empty list, set or tuple. Got '{collection_value}' of a type '{type(collection_value)}'."
            )
            raise TypeError(msg)
```

-> relaxing requirement to allow func calls with no actual argument passed for `colelction_filed`.

Reason - calling decorated functions with enabled splitting without passing any actual argument, like `existing_configlets = await cv_client.get_configlets(workspace_id=workspace_id)` ran by manifest to fetch all configlets

```
"msg": "Error during plugin execution: GRPCRequestHandler decorator expected the value of the collection_field 'configlet_ids' for function 'get_configlets' to be a non-empty list, set or tuple. Got '[]' of a type '<class 'list'>'."
```

After relaxing requirements:
```
GRPCRequestHandler: Preparing call for 'get_configlets' for collection_field 'configlet_ids' with 0 item(s).
```

python-avd/pyavd/_cv/client/tag.py
carl-baillargeon: During a call this week we decided on using tuples moving forward. For this PR though, let's just use a list like you suggested in the beginning so we don't have to change the decorator code and quickly merge the fix.

We can focus on changing to tuples and update the decorator later.

PR #6225, opened 2025-12-05
CI(cv_deploy): Add deploy_tags_to_cv pytest coverage

### LINKED ISSUE #6163: Fix(cv_deploy): Enable gRPC message splitting to set_tags and set_tag_assignments

alexeygorbunov:
## Change Summary

This is a pre-req for #6169 
Add coverage for `pyavd._cv.workflows.deploy_tags_to_cv`

## Related Issue(s)

Fixes #6163 (partially)

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
Additional pytest coverage

## How to test
pytest

### User Checklist

- N/A

### Repository Checklist


alexeygorbunov:
```
# coverage report --include=pyavd/_cv/workflows/deploy_tags_to_cv.py --show-missing
Name                                       Stmts   Miss  Cover   Missing
------------------------------------------------------------------------
pyavd/_cv/workflows/deploy_tags_to_cv.py      46      0   100%
------------------------------------------------------------------------
TOTAL                                         46      0   100%
```

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,800 @@
...
+@pytest.mark.filterwarnings("ignore:Unverified HTTPS request is being made to host")
+async def test_deploy_tags_to_cv_message_splitting(
+    caplog: pytest.LogCaptureFixture,
+    targeted_cv: dict[str, str],
+    verify_certs: bool,
+    cv_tags_fixture: set[CVTag],
+    cv_tag_assignments_fixture: set[CVTagAssignment],
+) -> None:
+    """Test ability to gracefully push amount of Tags and Assignments which exceeds the message limit (1837788 vs. 1048576 max)."""
+    with does_not_raise(), caplog.at_level(INFO):
```
gmuloc: I don't think you need does_not_raise (a nullcontext) - I generally use this in parametrize tests to handle some tests that raise and some that don't but here it does not feel useful (and it does not mean that it cannot raise at all :) )

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,800 @@
...
+    assert mock_cv_client.delete_tag_assignments.call_count == delete_tag_assignments_call_count
+    for arg_key, arg_value in delete_tag_assignments_call_args.items():
+        assert mock_cv_client.delete_tag_assignments.call_args[1][arg_key] == arg_value
+
+
+## Live TAGs tests ##
+@pytest.mark.skipif(environ.get("CV_LIVE_TEST") is None, reason="CV_LIVE_TEST env variable is not set. Live cv_deploy tests are skipped.")
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("targeted_cv", "verify_certs"),
```
gmuloc: if we are repeating this pattern in other places we can create this fixture:

```
@dataclass
class CvEnvironment:
    id: str
    cv_server: str
    cv_access_token: str
    verify_certs: bool

@pytest.fixture(scope="session", params=["PRD", "STG", "ONPREM"])
def cv_environment(request) -> CvEnvironment:
    """
    Fixture that loads environment variables and yields a CvEnvironment dataclass instance.
    Skips the test case if either the server URL or the access token is missing.
    """
    env_prefix: str = request.param
    
    cv_server = os.environ.get(f"CV_{env_prefix}_SERVER")
    cv_access_token = os.environ.get(f"CV_{env_prefix}_ACCESS_TOKEN")
    

    missing_vars = []
    if not cv_server or cv_server.strip() == "":
        missing_vars.append(f"CV_{env_prefix}_SERVER")
    if not cv_access_token or cv_access_token.strip() == "":
        missing_vars.append(f"CV_{env_prefix}_ACCESS_TOKEN")
        
    if missing_vars:
        pytest.skip(
            f"Skipping test case for {env_prefix} because mandatory environment variable(s) are missing: {', '.join(missing_vars)}"
        )

    verify_certs: bool = (env_prefix != "ONPREM")

    return CvEnvironment(
        id=f"CVAAS_{env_prefix}",
        cv_server=cv_server.strip(),
        cv_access_token=cv_access_token.strip(),
        verify_certs=verify_certs,
    )
```

gmuloc: example output:

```
CV_LIVE_TEST=1 pytest test_deploy_tags_to_cv.py::test_deploy_tags_to_cv_message_splitting  -vvv
============================================================================================================================= test session starts =============================================================================================================================
platform darwin -- Python 3.11.11, pytest-9.0.1, pluggy-1.6.0 -- /Users/guillaumemulocher/.pyenv/versions/3.11.11/envs/avd-3.11.11/bin/python
cachedir: .pytest_cache
rootdir: /Users/guillaumemulocher/Github/gmuloc/ansible-avd
configfile: pyproject.toml
plugins: mock-3.15.1, asyncio-1.2.0, anyio-4.9.0, xdist-3.8.0
asyncio: mode=Mode.STRICT, debug=False, asyncio_default_fixture_loop_scope=function, asyncio_default_test_loop_scope=function
collected 3 items

test_deploy_tags_to_cv.py::test_deploy_tags_to_cv_message_splitting[PRD] SKIPPED (Skipping test case for PRD because mandatory environment variable(s) are missing: CV_PRD_SERVER, CV_PRD_ACCESS_TOKEN)                                                                 [ 33%]
test_deploy_tags_to_cv.py::test_deploy_tags_to_cv_message_splitting[STG] SKIPPED (Skipping test case for STG because mandatory environment variable(s) are missing: CV_STG_SERVER, CV_STG_ACCESS_TOKEN)                                                                 [ 66%]
test_deploy_tags_to_cv.py::test_deploy_tags_to_cv_message_splitting[ONPREM] SKIPPED (Skipping test case for ONPREM because mandatory environment variable(s) are missing: CV_ONPREM_SERVER, CV_ONPREM_ACCESS_TOKEN)        
```

with the following refactored code:

```
@pytest.mark.skipif(environ.get("CV_LIVE_TEST") is None, reason="CV_LIVE_TEST env variable is not set. Live cv_deploy tests are skipped.")
@pytest.mark.asyncio
@pytest.mark.filterwarnings("ignore:Unverified HTTPS request is being made to host")
async def test_deploy_tags_to_cv_message_splitting(
    caplog: pytest.LogCaptureFixture,
    cv_environment: CvEnvironment,
    cv_tags_fixture: set[CVTag],
    cv_tag_assignments_fixture: set[CVTagAssignment],
) -> None:
    """Test ability to gracefully push amount of Tags and Assignments which exceeds the message limit (1837788 vs. 1048576 max)."""
    with does_not_raise(), caplog.at_level(INFO):
        async with CVClient(
            servers=cv_environment.cv_server,
            token=cv_environment.cv_access_token,
            verify_certs=cv_environment.verify_certs,
        ) as cv_client:
```

python-avd/tests/pyavd/cv/workflows/helpers.py
```diff
@@ -48,3 +59,241 @@ def create_grpc_change_control(
 def generate_id(key: str) -> str:
     """Helper to consistently generate expected IDs for tests."""
     return CVManifest._generate_deterministic_id(key)
+
+
+def get_tags_cv_state(tag_type: Literal["device", "interface", "all"]) -> list[Tag]:
```
gmuloc: looks like we never use "all" - any reason to not just creat two separate functions?

gmuloc (commented):
will make a PR to your branch with suggestion

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: 100%, were copy/pasting from other tests and forgot to remove

python-avd/tests/pyavd/cv/workflows/helpers.py
alexeygorbunov:
```
    if tag_type == "device":
        return device_tags
    if tag_type == "interface":
        return interface_tags
    return device_tags + interface_tags <<<
    ```

Another open refactoring PR is targeting to process all device and interface tags at the same time (in contrast to the current workflow which processes device and interface tags separately). So i'm simply trying to prepare helper funcs that can be easily used/reused for all test cases (once workflow to process device+interface tags at the same time is ready -> testing function can simply request `all` which will lead to a return of the concatenation of all tags (device + interface))

File: python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py (hunk 2 above)
Comment (alexeygorbunov) [reply python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py]: This looks great! I though modified your proposal a bit to make sure that missing env vars are not leading to the simple test skip. If we need to test code against all targeted environments - skipping test against specific environment due to unset (or incorrectly set to empty string) env vars should probably be treated as a test failure

File: python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py [hunk 4]
```diff
@@ -0,0 +1,773 @@
...
+def mock_cv_client() -> MagicMock:
+    """Fixture to provide a mocked CVClient instance with AsyncMocks."""
+    client = MagicMock()
+    # Patch all required async methods with AsyncMock
+    client.get_tags = AsyncMock()
+    client.set_tags = AsyncMock()
+    client.get_tag_assignments = AsyncMock()
+    client.set_tag_assignments = AsyncMock()
+    client.delete_tag_assignments = AsyncMock()
+    return client
```
gmuloc: so my PR was moving this all in one place. It feels a bit more central and it helps to understand what we get when we take the fixture between file.

You don't think it would help?

python-avd/tests/pyavd/cv/workflows/helpers.py
```diff
@@ -48,3 +59,241 @@ def create_grpc_change_control(
...
+    device_tags = [
```
gmuloc: I would still move these tag creation to helpers as this is very hard to follow and understand whats going on

python-avd/tests/pyavd/cv/workflows/helpers.py
```diff
@@ -48,3 +59,241 @@ def create_grpc_change_control(
...
+                element_type=ElementType.INTERFACE,
+                label="mixed_tag_1",
+                value="mixed_tag_1_value_2",
+                device_id="L3_serial",
+                interface_id="Ethernet1",
+                element_sub_type=ElementSubType.DEVICE,
+            ),
+            tag_creator_type=CreatorType.USER,
+        ),
+        TagAssignment(
```
gmuloc: this assignement is the same as the previous one - is it on purpose?

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,761 @@
...
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("strict", "expected_removed_tags", "delete_tag_assignments_call_count", "delete_tag_assignments_call_args"),
+    [
+        pytest.param(True, [], 0, {}, id="STRICT_TRUE"),
+        pytest.param(False, [], 0, {}, id="STRICT_FALSE"),
+    ],
+)
```
gmuloc: Looking at this - I would rather that we have only one pram (strict) and remove the extra ones which are not really needed

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,761 @@
...
+async def test_deploy_tags_to_cv_unassigned_interface_tags(
+    mock_cv_client: MagicMock,
+    strict: bool,
+    expected_removed_tags: list[CVDeviceTag | CVInterfaceTag],
+    delete_tag_assignments_call_count: int,
+    delete_tag_assignments_call_args: dict[str, Any],
+) -> None:
+    """Test assignment of the existing interface tag+value pairs."""
+    mock_cv_client.get_tags.return_value = get_interface_tags_cv_state()
```
gmuloc, line 543: I feel like these longs tests could deserve some comments to understand all the steps

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,761 @@
...
+    for arg_key, arg_value in delete_tag_assignments_call_args.items():
+        assert mock_cv_client.delete_tag_assignments.call_args[1][arg_key] == arg_value
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize(
+    ("strict", "expected_removed_tags", "delete_tag_assignments_call_count", "delete_tag_assignments_call_args"),
+    [
+        pytest.param(True, [], 0, {}, id="STRICT_TRUE"),
+        pytest.param(False, [], 0, {}, id="STRICT_FALSE"),
```
gmuloc: same here - we can just test in that part that there is nbo dleeted tag since we know already thats what we want

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,761 @@
...
+@pytest.mark.asyncio
+@pytest.mark.filterwarnings("ignore:Unverified HTTPS request is being made to host")
+async def test_deploy_tags_to_cv_message_splitting(
+    caplog: pytest.LogCaptureFixture,
+    cv_environment: CvEnvironment,
+    cv_tags_fixture: set[CVTag],
+    cv_tag_assignments_fixture: set[CVTagAssignment],
+) -> None:
+    """Test ability to gracefully push amount of Tags and Assignments which exceeds the message limit (1837788 vs. 1048576 max)."""
+    with does_not_raise(), caplog.at_level(INFO):
```
gmuloc:
```suggestion
    with caplog.at_level(INFO):
```

gmuloc (commented):
some more comment to simplify a bit the inputs

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
```diff
@@ -0,0 +1,761 @@
...
+    assert mock_cv_client.delete_tag_assignments.call_count == delete_tag_assignments_call_count
+    for arg_key, arg_value in delete_tag_assignments_call_args.items():
+        assert mock_cv_client.delete_tag_assignments.call_args[1][arg_key] == arg_value
+
+
+## Live TAGs tests ##
+@pytest.mark.skipif(environ.get("CV_LIVE_TEST") is None, reason="CV_LIVE_TEST env variable is not set. Live cv_deploy tests are skipped.")
+@pytest.mark.asyncio
+@pytest.mark.filterwarnings("ignore:Unverified HTTPS request is being made to host")
+async def test_deploy_tags_to_cv_message_splitting(
```
carl-baillargeon: This is not testing the `deploy_tags_to_cv` workflow itself but rather `CVClient` tag methods. We should probably move this to `../client`.

python-avd/tests/pyavd/cv/workflows/helpers.py
alexeygorbunov: This function has been changed. No "all" option present now

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: By explicitly passing those "empty" items i'm trying to confirm that applying specific combination of tags/assignments does not lead to deleting/un-associating any other tags/assignments, trying to use the same test function structure for simplicity of reading the logic

gmuloc: you could hardcode the values at the end of the test instead of having the parameters all the same, it clutters the test for not much value (add a comment stating what you do in the test)

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Added comments/details to most of the tested steps

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Done it for most of the functions (except those that actually expect different tags/associations removal based on a strict/non-strict logic)

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Removed

python-avd/tests/pyavd/cv/workflows/test_deploy_tags_to_cv.py
alexeygorbunov: Removed live 'test_deploy_tags_to_cv_message_splitting' for now as it can't be tested anyway until decorator is adjusted and input type used by `[gs]et_tags` is covered for splitting.
Keeping Guillaumes `conftest.CvEnvironment` (which will be eventually used by that test function)

carl-baillargeon (approved):
LGTM!

gmuloc (approved):
Thanks!

ClausHolbechArista:
@gmuloc @carl-baillargeon please close the comments after reviewing.

PR #6286, opened 2025-12-23
Feat(eos_designs)!: Add the possibility to configure VRF rt in one direction only

### LINKED ISSUE #6266: Feat(eos_designs): Add the possibility to configure VRF rt in one direction only
### Enhancement summary

In some corner cases it is possible that one want only one direction for the automatically generated RT per VRF like

This feature request is to add some knobs to allow this customization without impacting existing deployments

eos_designs

### Use case example

```
vrf default
   rd 1.1.1.1:500
   route-target import evpn 500:500
   # notice no export
```

```diff
 <network_services_keys.name>:
   - name: <str; required; unique>
     vrfs:
       - name: <str; required; unique>
         rt_override: <str>
+        rt_import: <bool, default: true>
+        rt_export: <bool, default: true>
```

The keys allow for disabling the generation of the `route-target import | export` line for the VRF for _every_ addresse familiy in `address_families`

This should be taken into account in https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py#L268 

This allow to workaround the issue today
```
     - name: default
        vrf_id: 200
        rt_override: 500:500
        bgp:
          structured_config:
            route_targets:
              export: null
```

### Additional context

_No response_

### Contributing Guide

MaheshGSLAB:
## Change Summary

Add the possibility to configure VRF rt in one direction only

## Related Issue(s)

Fixes #6266 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
In some corner cases it is possible that one want only one direction for the automatically generated RT per VRF like

This feature request is to add some knobs to allow this customization without impacting existing deployments
```
<network_services_keys.name>:
     vrfs:
         rt_override: <str>
+        rt_import: <bool, default: true>
+        rt_export: <bool, default: true>
```
## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -276,15 +276,17 @@ def _update_router_bgp_vrf_evpn_or_mpls_cfg(
...
         for af in sorted(vrf_address_families):
-            if (target := get_item(route_targets["import"], "address_family", af)) is None:
-                route_targets["import"].append({"address_family": af, "route_targets": [vrf_rt]})
-            else:
-                target["route_targets"].append(vrf_rt)
+            if vrf.rt_import:
+                if (target := get_item(route_targets["import"], "address_family", af)) is None:
+                    route_targets["import"].append({"address_family": af, "route_targets": [vrf_rt]})
+                else:
+                    target["route_targets"].append(vrf_rt)
```
MaheshGSLAB: We reinitialize `route_targets = {"import": [], "export": []}` on every call to this method. Because vrf_address_families is a set of str, target is always None for each AF, so the else block is never reached—unless there is something I am overlooking.
FYI @gmuloc @ClausHolbechArista

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -297,7 +299,7 @@ def _update_router_bgp_vrf_evpn_or_mpls_cfg(
         if vrf.name == "default" and self._vrf_default_evpn and self._route_maps_vrf_default_check():
             # Special handling of vrf default with evpn.
 
-            if (target := get_item(route_targets["export"], "address_family", "evpn")) is None:
+            if (target := get_item(route_targets["export"], "address_family", "evpn")) is None and vrf.rt_export:
```
laxmikantchintakindi: When rt_import is set to false, this is creating below error for new host.
```
'NoneType' object has no attribute 'setdefault'
```

MaheshGSLAB: updated code logic

laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
```diff
@@ -423,6 +423,14 @@ $defs:
                   rt_override supports two formats:
                     - A single number will be used in the RT assigned number subfield (second part of the RT).
                     - A full RT string with colon separator which will override the full RT.
+              rt_import:
+                type: bool
+                description: Enable/Disable route target import for the VRF of every address family.
```
gmuloc: mention that these keys do not impact `additional_route_targets`.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
gmuloc: I am not sure I follow.
The function should be called only once with all address families (line 161)

so for the first address family it will be set to the values and for the second we will already have one import / export paur and add an extra one?

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -276,15 +276,17 @@ def _update_router_bgp_vrf_evpn_or_mpls_cfg(
         route_targets = {"import": [], "export": []}
```
gmuloc: Actually, looking into this. we can refactor this as we don't really need to do the _from_dict anymore @MaheshGSLAB.

Just append directly to the `bgp_vrf.route_targets.import` / `bgp_vrf.route_targets.export` wherever needed

gmuloc (changes requested):
let's do some cleanup :)

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
MaheshGSLAB: Yes it called once with all address families but its set of str mean no repeat address family so `target := get_item(route_targets["import"], "address_family", af))` will always None for a unique address family so we never touch the else part.

gmuloc: understood - we can remove the else part indeed

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_network_services.schema.yml
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -273,36 +273,27 @@ def _update_router_bgp_vrf_evpn_or_mpls_cfg(
...
+                )
 
         for rt in vrf.additional_route_targets:
-            if rt.type is None:
-                continue
-            if (target := get_item(route_targets[rt.type], "address_family", rt.address_family)) is None:
-                route_targets[rt.type].append({"address_family": rt.address_family, "route_targets": [rt.route_target]})
-            else:
-                target["route_targets"].append(rt.route_target)
+            if rt.type == "import" and rt.address_family and rt.route_target:
```
gmuloc: It appears we should make `type`, `address_family` and `route_target` required in the model - what do you think?

In the current status we would filter thjem anyways

this is from shared_utils so there is no need to do your checks (but maybe you are doing them for pyright and in that case lets just change the schema in your PR and make these fields required)

```
            vrf.additional_route_targets = vrf.additional_route_targets._filtered(
                lambda rt: bool((not rt.nodes or self.hostname in rt.nodes) and rt.address_family and rt.route_target and rt.type in ["import", "export"])
            )
```

gmuloc (changes requested):

gmuloc:
please also check how to increase coverage

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
MaheshGSLAB: Updated the schema and logic

gmuloc: @MaheshGSLAB also please update the porting guide with the schema changes

gmuloc (changes requested):
LGTM just need to update the porting guide to indicate the changes in the schema as they are 'breaking' (previously we would have silently ignored the additional route target with missing fields now we tell the user to give all the fields or to remove the entry)

gmuloc (approved):

PR #6316, opened 2026-01-07
CI: Adding compiled templates back to coverage

ClausHolbechArista:
Adding compiled templates back to coverage. Got removed when fixing package clash in coverage


gmuloc (approved):
approving so we see what happens in devel

PR #6358, opened 2026-01-13
Fix(eos_designs): Improve creation of route-maps, prefix-lists and peer-groups

ClausHolbechArista:
## Change Summary

Improve creation of route-maps, prefix-lists and peer-groups

Precursor for #5786.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Create route-maps, prefix-lists, as-path lists etc as required instead of repeating the logic to deduct twice.
- Use run_once decorator to avoid repeating the work needlesly.

## How to test

- No unexpected changes in molecule
- The few changes seen are actually fixes, since the logic was faulty before. We were referring to objects (route-maps, prefix-lists etc) that were not configured.
- Most of these changes will go away again once we do the second part to only configure peer-groups when needed.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/shared_utils/underlay.py
```diff
@@ -73,9 +73,9 @@ def any_multicast_enabled(self: SharedUtilsProtocol) -> bool:
         return self.underlay_multicast_pim_sm_enabled or self.underlay_multicast_static_enabled
 
     @cached_property
-    def underlay_multicast_rp_interfaces(self: SharedUtilsProtocol) -> list[EosCliConfigGen.LoopbackInterfacesItem] | None:
+    def underlay_multicast_rp_interfaces(self: SharedUtilsProtocol) -> list[EosCliConfigGen.LoopbackInterfacesItem]:
```
gmuloc: Could we even refactor with this?
```suggestion
    def underlay_multicast_rp_interfaces(self: SharedUtilsProtocol) -> EosCliConfigGen.LoopbackInterfaces:
```

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
```diff
@@ -450,23 +456,33 @@ def _set_neighbors(self: AvdStructuredConfigOverlayProtocol) -> None:
...
+                remote_as = data["bgp_as"]
+                if remote_as is None:
+                    # Never happens but here to help the type checker.
+                    continue
                 neighbor = self._create_neighbor(
                     data["ip_address"],
                     route_client,
                     self.inputs.bgp_peer_groups.evpn_overlay_peers.name,
-                    remote_as=self.shared_utils.get_asn(data["bgp_as"]),
+                    remote_as=remote_as,
```
gmuloc: so technically here it could be None for the underlying function and it would not chage a thing

gmuloc: however I am not to sure why you changed the `get_asn` ? would it not cause problems for dot notation?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/add-subnets-to-redistribution-in-default-vrf-with-sequence30.cfg
```diff
@@ -147,6 +147,11 @@ ip prefix-list PL-LOOPBACKS-EVPN-OVERLAY
 ip prefix-list PL-SVI-VRF-DEFAULT
    seq 10 permit 10.10.10.0/24
 !
+route-map RM-BGP-UNDERLAY-PEERS-OUT deny 10
+   match ip address prefix-list PL-SVI-VRF-DEFAULT
```
laxmikantchintakindi, line 151: This route map is not applied anywhere.

python-avd/pyavd/_eos_designs/structured_config/underlay/as_path.py
```diff
@@ -20,13 +20,9 @@ class AsPathMixin(Protocol):
...
-
-        if self.shared_utils.wan_ha and self.shared_utils.use_uplinks_for_wan_ha and self.shared_utils.bgp_as:
-            entries = EosCliConfigGen.AsPath.AccessListsItem.Entries()
-            entries.append_new(type="permit", match=as_path_list_match_from_bgp_asns((self.shared_utils.bgp_as,)))
-            self.structured_config.as_path.access_lists.append_new(name="ASPATH-WAN", entries=entries)
+    @run_once_method
+    def set_once_as_path_acl_aspath_wan(self: AvdStructuredConfigUnderlayProtocol) -> None:
+        """Set as-path access-list ASPATH-WAN."""
+        entries = EosCliConfigGen.AsPath.AccessListsItem.Entries()
+        entries.append_new(type="permit", match=as_path_list_match_from_bgp_asns((self.shared_utils.bgp_as,)))  # pyright: ignore[reportArgumentType]
```
gmuloc, line 27: not sure why the type checking ignore required before here - did we change anything (as bgp_as could always be str | None)

python-avd/pyavd/_eos_designs/structured_config/underlay/loopback_interfaces.py
```diff
@@ -95,9 +95,8 @@ def loopback_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> None:
             self.structured_config.loopback_interfaces.append(vtep_loopback)
 
         # Underlay Multicast RP Loopbacks
-        if self.shared_utils.underlay_multicast_rp_interfaces is not None:
-            for underlay_multicast_rp_interface in self.shared_utils.underlay_multicast_rp_interfaces:
-                self.structured_config.loopback_interfaces.append(underlay_multicast_rp_interface)
+        for underlay_multicast_rp_interface in self.shared_utils.underlay_multicast_rp_interfaces:
+            self.structured_config.loopback_interfaces.append(underlay_multicast_rp_interface)
```
gmuloc: Yeah we could just extend then if we return the object directly here rather than for loop

python-avd/pyavd/_eos_designs/avdfacts.py
```diff
@@ -58,8 +60,9 @@ def clear_cache(self) -> None:
             self.__dict__.pop(key, None)
 
 
-class AvdFacts(AvdFactsProtocol):
+class AvdFacts(RunOnceMethodStateHelper, AvdFactsProtocol):
```
gmuloc: is it needed in this PR?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/add-subnets-to-redistribution-in-default-vrf-with-sequence30.cfg
ClausHolbechArista: it is in for the unused peer-group in router bgp

python-avd/pyavd/_eos_designs/structured_config/underlay/as_path.py
ClausHolbechArista: I think the check was higher up in the function in the old code.

python-avd/pyavd/_eos_designs/avdfacts.py
ClausHolbechArista: nope. Something got applied to the wrong commit in my IDE.

ClausHolbechArista:
```suggestion
class AvdFacts(AvdFactsProtocol):
```

python-avd/pyavd/_eos_designs/structured_config/overlay/router_bgp.py
ClausHolbechArista: get_asn is done inside _create_neighbor

python-avd/pyavd/_eos_designs/avdfacts.py
```diff
@@ -63,3 +63,4 @@ def __init__(self, hostvars: MutableMapping, inputs: EosDesigns, shared_utils: S
         self._hostvars = hostvars
         self.inputs = inputs
         self.shared_utils = shared_utils
+        super().__init__()
```
ClausHolbechArista:
```suggestion
```

gmuloc, line 66: just read about this and this feels magical .. :)

gmuloc (approved):

python-avd/pyavd/_eos_designs/avdfacts.py
ClausHolbechArista: I even forgot and tried to remove and had to bring it back. So indeed.

PR #6428, opened 2026-01-27
Test(eos_designs): Update tests to have BGP peers as needed

ClausHolbechArista:
As preperation for removal of unused BGP overlay peer-groups, this updates the tests to have at least one BGP peer where needed for the tested feature.

This will only add config.


ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/bgp-cleartext-passwords.cfg
```diff
@@ -133,6 +133,7 @@ router bgp 666
    no bgp default ipv4-unicast
    maximum-paths 4
    neighbor EVPN-OVERLAY-PEERS peer group
+   neighbor EVPN-OVERLAY-PEERS next-hop-unchanged
```
ClausHolbechArista, line 136: this is because we are changing this one to be a route-server

gmuloc (approved):

PR #6441, opened 2026-01-28
Feat(pyavd): Add get_device_config_tree helper

mbakalarski:
## Change Summary

Add get_device_config_tree() to pyavd to return rendered device CLI as a nested dictionary (with tests).
Enables structured inspection and downstream processing of generated EOS configurations.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/.github/workflows/pull-request-management.yml
+++ b/.github/workflows/pull-request-management.yml
@@ -603,13 +603,15 @@ jobs:
             id: Main-Tests
             pytest_args: >-
               --ignore=python-avd/tests/pyavd/molecule_scenarios/test_get_device_config.py
+              --ignore=python-avd/tests/pyavd/molecule_scenarios/test_get_device_config_tree.py
               --ignore=python-avd/tests/pyavd/molecule_scenarios/test_negative_eos_designs.py
               --ignore=python-avd/tests/schema_tools/metaschema/test_meta_schema_model.py
             upload_compiled_templates: true
-          - name: "Excluded-Tests (test_get_device_config and test_meta_schema_model)"
+          - name: "Excluded-Tests (test_get_device_config, test_get_device_config_tree and test_meta_schema_model)"
             id: Excluded-test_get_device_config-and-test_meta_schema_model
             pytest_args: >-
               python-avd/tests/pyavd/molecule_scenarios/test_get_device_config.py
+              python-avd/tests/pyavd/molecule_scenarios/test_get_device_config_tree.py
               python-avd/tests/schema_tools/metaschema/test_meta_schema_model.py
           - name: "Excluded-Tests (test_negative_eos_designs)"
             id: Excluded-test_negative_eos_designs
--- a/python-avd/pyavd/__init__.py
+++ b/python-avd/pyavd/__init__.py
@@ -3,6 +3,7 @@
 # that can be found in the LICENSE file.
 from .get_avd_facts import get_avd_facts
 from .get_device_config import get_device_config
+from .get_device_config_tree import get_device_config_tree
 from .get_device_doc import get_device_doc
 from .get_device_structured_config import get_device_structured_config
 from .get_device_test_catalog import get_device_test_catalog
@@ -23,6 +24,7 @@ __version__ = "6.0.0.dev7"
 __all__ = [
     "get_avd_facts",
     "get_device_config",
+    "get_device_config_tree",
     "get_device_doc",
     "get_device_structured_config",
     "get_device_test_catalog",
--- a/python-avd/pyavd/get_device_config_tree.py
+++ b/python-avd/pyavd/get_device_config_tree.py
@@ -1,0 +1,57 @@
+# Copyright (c) 2023-2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .api.schemas import EOSConfig
+
+
+def get_device_config_tree(structured_config: EOSConfig | dict) -> dict[str, dict]:
+    """
+    Render eos_cli_config_gen templates and return a nested dict representing the CLI.
+
+    Args:
+        structured_config:
+            EOSConfig instance or dictionary with the validated structured configuration.
+
+    Returns:
+        Device configuration as a nested dict under a single 'default' key.
+    """
+    from .get_device_config import get_device_config  # noqa: PLC0415
+
+    full_cli = get_device_config(structured_config)
+
+    # Convert the full CLI into a nested dictionary
+    return {"default": _cli_to_tree(full_cli)}
+
+
+def _cli_to_tree(cli: str) -> dict[str, dict]:
+    """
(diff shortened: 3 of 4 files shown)


ClausHolbechArista:
Hi @mbakalarski

Thank you for your pull-request. In AVD we first generate the "structured config" which is a well-defined and schema enforced data model representing the EOS config. From that we generate the EOS CLI which is just text. Going the other way around is very fragile, since you cannot know if something is a config token or a variable field.

Ex.
```eos
router bgp 12345:
   router-id 1.2.3.4
```
would have to be parsed into:
```json
{'router_bgp_12345':['router_id_1.2.3.4']}
```
or
```json
{'router':{'bgp':{'12345':{'router_id':'1.2.3.4'}}}}`
```
with no clear demarkation between tokens and values.

Instead I will suggest to parse the structured config. It is very very close to the rendered config, and with the schema classes we are adding in AVD 6.0 (already in `devel` as `pyavd.api.schemas.EOSConfig`) you get typing, descriptions, structure etc for free.

I don't see a general use of the "blind" coversion from text to structured data. It would be working against our efforts as described above.

Therefore, for now, I don't think we will accept this pull-request.
Please enlighten me if you think I am missing something.

Thank you again.
Claus

mbakalarski:
Hi @ClausHolbechArista 

Thank you for the feedback — I understand and agree that the structured config is the source of truth and CLI is considered lossy.

I wanted to note that Arista’s eAPI already provides a hierarchical JSON view of running-config (essentially a CLI tree), which many tools consume. My PR’s goal was similar: a vendor-aligned, best-effort tree of rendered configs for inspection, diffing, or downstream tooling.

I completely respect that this may be out of scope for AVD core. An alternative could be a small external pyavd helper/extension that provides a tree view without conflicting with the schema-first philosophy.

Thanks again for reviewing and for the guidance!
Michal

ClausHolbechArista:
I think this would fit better in an external EOS utilities library, since it is not depending on any AVD functionalities. I asume it could also work the same with other vendors' configurations files as long as they use a structure based on indentation, so maybe it deserves it's own library.

ClausHolbechArista:
A quick google search revealed some similar existing libraries. Maybe the efforts are better spent on enhancing those? One example: `network-tools/shconfparser`

mbakalarski:
Hi @ClausHolbechArista 

This appears to be native to EOS.
Sorry for this rather long excerpt:

```
ceos01#show running-config | json 
{
    "header": [
        "! device: ceos01 (cEOSLab, EOS-4.35.0F-44178984.4350F (engineering build))"
    ],
    "comments": [],
    "cmds": {
        "no aaa root": null,
        "username arista privilege 15 secret sha512 $6$A3rRo/XbfxQ9PvMJ$jEk0qafOiYjAEyqtMZYCApR0e5u8BWutPMNUtQcUMbaQjqZfsP.EQ/58dKRHJreYV.5MmBU1N9dMbfwHkZTeA0": null,
        "management api http-commands": {
            "comments": [],
            "cmds": {
                "protocol http port 6021": null,
                "no shutdown": null
            }
        },
        "no service interface inactive port-id allocation disabled": null,
        "transceiver qsfp default-mode 4x10G": null,
        "service routing protocols model multi-agent": null,
        "hostname ceos01": null,
        "spanning-tree mode mstp": null,
        "system l1": {
            "comments": [],
            "cmds": {
                "unsupported speed action error": null,
                "unsupported error-correction action error": null
            }
        },
        "management api restconf": {
            "comments": [],
            "cmds": {
                "transport https test": {
                    "comments": [],
                    "cmds": {
                        "ssl profile restconf": null
                    }
                }
            }
        },
        "management security": {
            "comments": [],
            "cmds": {
                "ssl profile restconf": {
                    "comments": [],
                    "cmds": {
                        "certificate restconf.crt key restconf.key": null
                    }
                }
            }
        },
        "interface Ethernet1": {
            "comments": [],
            "cmds": {}
        },
        "ip access-list custom-cp": {
            "comments": [],
            "cmds": {
                "counters per-entry": null,
                "10 permit icmp any any": null,
                "20 permit ip any any tracked": null,
                "30 permit udp any any eq bfd ttl eq 255": null,
                "40 permit udp any any eq bfd-echo ttl eq 254": null,
                "50 permit udp any any eq multihop-bfd micro-bfd sbfd": null,
                "60 permit udp any eq sbfd any eq sbfd-initiator": null,
                "70 permit ospf any any": null,
                "80 permit tcp any any eq ssh telnet www snmp bgp https msdp ldp netconf-ssh gnmi": null,
                "90 permit udp any any eq bootps bootpc ntp snmp ptp-event ptp-general rip ldp": null,
                "100 permit tcp any any eq mlag ttl eq 255": null,
                "110 permit udp any any eq mlag ttl eq 255": null,
                "120 permit vrrp any any": null,
                "130 permit ahp any any": null,
                "140 permit pim any any": null,
                "150 permit igmp any any": null,
                "160 permit tcp any any range 5900 5910": null,
                "170 permit tcp any any range 50000 50100": null,
                "180 permit udp any any range 51000 51100": null,
                "190 permit tcp any any eq 3333": null,
                "200 permit tcp any any eq nat ttl eq 255": null,
                "210 permit tcp any eq bgp any": null,
                "220 permit rsvp any any": null,
                "230 permit tcp any any eq 9340": null,
                "240 permit tcp any any eq 9559": null,
                "250 permit udp any any eq 8503": null,
                "260 permit udp any any eq lsp-ping": null,
                "270 permit udp any eq lsp-ping any range 51900 51999": null,
                "280 permit tcp any any eq 6020": null,
                "290 permit tcp any any eq 6021": null
            }
        },
        "no ip routing": null,
        "system control-plane": {
            "comments": [],
            "cmds": {
                "ip access-group custom-cp in": null
            }
        },
        "router multicast": {
            "comments": [],
            "cmds": {
                "ipv4": {
                    "comments": [],
                    "cmds": {
                        "software-forwarding kernel": null
                    }
                },
                "ipv6": {
                    "comments": [],
                    "cmds": {
                        "software-forwarding kernel": null
                    }
                }
            }
        }
    }
}
```

ClausHolbechArista:
Right, I missed if the purpose was to create the same output locally from a config file.
That means it should stay in an eos-specific library. Or maybe you go the other way and use the generic config parser tool I mentioned from static config files and write a wrapper that can convert the eos `| json` output to that same format, so you get better data to parse on your end. It all depends on the end goal and how you wish to use the data.

mbakalarski:
In my case, an external parsing tool works fine. AVD could potentially benefit from generating formatted CLI output, but since "show running-config | json" is processed on the EOS side, a 100% match maybe be hard to achieve.

ClausHolbechArista:
I will ask the rest of the team if they see wider use of this. Personally I don't support making it part of AVD, since we already have the structured config, which is a much stronger data model.

ClausHolbechArista:
We have decided not to accept this pull request. Still I would like to thank you for your contribution and interest in AVD.

I would strongly suggest using one of the existing tools that can parse network configs and create a structured data set.

Thanks

mbakalarski:
I understand. Thank you for your time and for the suggestion.

PR #6444, opened 2026-01-28
Fix(anta_runner): Remove deduplication between VerifyReachability tests

carl-baillargeon:
## Change Summary

In the current logic, we deduplicated pings between different VerifyReachability tests based on destination and VRF which was very fragile (same destination and VRF but different source would get removed). It was also creating confusion in the report because expected entries were missing since they were deduped.

This PR removes the deduplication checks between different VerifyReachability tests to get the complete entries in the report.
ANTA uses caching anyways so the exact same ping command will never be sent twice. These tests are also turned off by default via the `extra_fabric_validation` knob.

BGP neighbors reachability test is now disabled by default via the `extra_fabric_validation` knob. Molecule files were updated reflecting this.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.anta_runner`

### User Checklist

- N/A

### Repository Checklist


vitthalmagadum (approved):
LGTM

ansible_collections/arista/avd/extensions/molecule/anta_runner/anta/avd_catalogs/default_run/dc1-leaf1b.json
```diff
@@ -118,41 +118,6 @@
...
-          "custom_field": "Generated by AVD"
-        },
-        "filters": {
-          "tags": [
-            "dc1-leaf1b"
-          ]
-        },
-        "hosts": [
-          {
-            "destination": "10.255.0.1",
```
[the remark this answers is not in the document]
MaheshGSLAB: We see the test in default_run_filtered_report but not in default_run.

ansible_collections/arista/avd/extensions/molecule/anta_runner/anta/avd_catalogs/default_run_filtered_report/dc1-leaf1a.json
```diff
@@ -237,11 +237,49 @@
             "vrf": "default",
             "repeat": 1
           },
+          {
+            "destination": "10.255.0.1",
```
MaheshGSLAB, line 241: Ideally loopback0 is `10.255.0.3` and below test already covered in Vtep connectivity test. Can we more smart on detecting such cases? 
{
            "destination": "10.255.0.1",
            "source": "10.255.0.3",
            "description": "dc1-spine1",
            "vrf": "default",
            "repeat": 1
          },

gmuloc: so the problem with being more smart @MaheshGSLAB is that you can run one example with only the BGP tests and get the test for a loopback and when you add VTEP tests then you "lose" your BGP test. Which looks weird and wrong (where did your neighbor go? Have you lost BGP?..)

Yes the test is duplicated but at least you get consistent output when running with different options without too much dark magic.

In ANTA duplicated show command are not sent multiple times to the device (there is caching) so in some cases it is just adding a line

MaheshGSLAB (approved):

PR #6438, opened 2026-01-27
Feat(eos_designs,eos_cli_config_gen): Expose avd_tmp_dir to eos_designs and eos_cli_config_gen roles

carl-baillargeon:
## Change Summary

The new `validate_inputs` action plugin uses a temporary folder to store templated and validated files used internally.

This PR adds a new `avd_tmp_dir` variable to both eos_designs and eos_cli_config_gen roles to store files into this folder instead.

## Related Issue(s)

Fixes https://github.com/aristanetworks/avd-internal/issues/343

## Component(s) name

`arista.avd.eos_designs`
`arista.avd.eos_cli_config_gen`

### User Checklist

- N/A

### Repository Checklist

- [X] My code has been rebased from devel before I start
- [X] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.
- [X] My change requires a change to the documentation and documentation have been updated accordingly.
- [X] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/extensions/molecule) testing accordingly. (check the box if not applicable)

PR #6463, opened 2026-02-02
Feat: Encrypt temporary files when vault is enabled

gmuloc:
## Change Summary

Following the introduction of `validate_inputs`, AVD now leverages tmp files to store templated and validated data. This PR automates vaulting these files when a vault secret is detected when running a playbook.

## Component(s) name

`plugins`

## Proposed changes

* use the DataLoader in the action plugin to decrypt vaulted files
* use the vault encrypt function

## How to test

* tests will be integration tests but would need the avd_tmp_dir PR #6438 to be able to check the tmp files.
* getting an issue in integration tests to use VAULT (nasty ansible-test :) )

### Repository Checklist


ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -324,9 +333,10 @@ def _run_templating_phase(
         self.logger.info("Templating hostvars...")
         start_time = perf_counter()
         successful_hosts = []
+        output_files = []
```
alexeygorbunov: What is the purpose of this variable? Where is it used once populated?

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -431,6 +445,7 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
         hostname: Hostname to process.
         output_path: Directory path where the templated JSON file will be written.
         schema_name: Schema name used for filtering hostvars.
+        loader: Ansible DataLoader for unvaulting files if needed.
```
alexeygorbunov: It is both vaulting and unvaulting (encrypting and decrypting), not just unvolting?

alexeygorbunov:
Would it not make sense to encrypt templated/validated files even if Vault is not used/configured?

alexeygorbunov (approved):
Tested with and without using vaulted input files. With vaulted input - templated and validated files are properly encrypted & decrypted with the initial vault pass

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -453,9 +468,19 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
...
+
+        # Write JSON data to file, vaulted if any vault secret is configured.
         output_file_path = output_path / f"{hostname}.json"
-        with output_file_path.open(mode="w", encoding="utf-8") as f:
-            json.dump(templated_hostvars, f, skipkeys=True, default=lambda _: "<not serializable>", indent=4)
+        with contextlib.suppress(AnsibleVaultError):
+            data = loader._vault.encrypt(data, secret=None)
+
+        with output_file_path.open("wb") as f:
+            f.write(data)
```
ClausHolbechArista:
```suggestion
        output_file_path.write_bytes(data)
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -453,9 +468,19 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
...
+
+        # Convert to bytes for binary file write
+        data = data.encode("utf-8")
+
+        # Write JSON data to file, vaulted if any vault secret is configured.
         output_file_path = output_path / f"{hostname}.json"
-        with output_file_path.open(mode="w", encoding="utf-8") as f:
-            json.dump(templated_hostvars, f, skipkeys=True, default=lambda _: "<not serializable>", indent=4)
+        with contextlib.suppress(AnsibleVaultError):
+            data = loader._vault.encrypt(data, secret=None)
```
ClausHolbechArista: Since you already detect if vault is available early you could also decide to just pass the loader to this worker when vault is available. Then we still get other vault errors in case something else goes wrong.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -492,23 +518,29 @@ def _validate_host_worker(
...
+            validated_data = validated_data.encode("utf-8")
+
+            # Write JSON data to file, vaulted if any vault secret is configured.
+            with contextlib.suppress(AnsibleVaultError):
+                validated_data = loader._vault.encrypt(validated_data, secret=None)
+
             output_file_path = output_path / f"{hostname}.json"
-            with output_file_path.open(mode="w", encoding="utf-8") as f:
+            with output_file_path.open(mode="wb") as f:
                 f.write(validated_data)
```
ClausHolbechArista:
```suggestion
            output_file_path.write_bytes(validated_data)
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -492,23 +518,29 @@ def _validate_host_worker(
 
         # Load data based on file suffix.
         if input_suffix in {"yml", "yaml"}:
-            # YAML input: load and convert to JSON for validation.
-            with input_file_path.open(mode="r", encoding="utf-8") as f:
-                data = yaml.load(f, Loader=yaml.CSafeLoader)
+            # Need to call str over the loader output to make sure it is not a special Ansible object we pass to the yaml dumper...
+            # as it may fail.
+            data = yaml.load(str(loader.get_text_file_contents(str(input_file_path))), Loader=yaml.CSafeLoader)
```
ClausHolbechArista: here you do everything in one line. In other places this is split up in 6 lines with comments over each of them. We should settle on a style.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
ClausHolbechArista: We could wrap this in a self.get_loader where we return a callable with the relevant signature. Either using their loader or our own depending on availability of secrets. But I am also fine starting with what you have now, so you can disregard this if you think it is too much.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
ClausHolbechArista: Could we not just use Ansible's loader for get the deserialized data directly instead of having to handle yaml load here?

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: nowhere... :) leftover from another piece of implementation

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: actually in this one it is just vaulting! let me fix that

gmuloc:
> Would it not make sense to encrypt templated/validated files even if Vault is not used/configured?

yeah we discussed that but the problem is that then we force people to configure a vault when they don't need one. We will document this properly to indicate that if they want them encrypted then they should configure a vault and then it is their decision.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: the only VaultError I could see was the one raised when there is no secret. Let me double check (though it may not be correct in the future or in the past :) )

ClausHolbechArista:
> Would it not make sense to encrypt templated/validated files even if Vault is not used/configured?

We tried that, and the issue is that we do not have safe encryption key to use, that all the modules know.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: ah it could also happen if we don't manage to match any vault password (though with the current implementation we don't give the choice of the vault id so this should never happen)

ClausHolbechArista: if there is a secret for vault id "foo", but no "default" secret, will it still work?

gmuloc: so the way we were building it would mean that we would pick the first secret in the list

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: yes we can ! done

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
           from the EOS Config schema are found at the top level of AVD Design input data.
         type: bool
         default: false
+  avd_vault_id:
```
gmuloc: vault_id

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -304,6 +376,7 @@ def _run_templating_phase(
         batch_size: int,
         output_path: Path,
         schema_name: SCHEMA_NAME,
+        avd_vault_id: str | None = None,
```
gmuloc: vault_id

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -121,11 +125,71 @@ class ResolvedPluginArgs:
...
+
+        Args:
+            loader: The Ansible DataLoader instance.
+            vault_id: Optional vault identity to use for encryption. If None, uses the first
+                     vault identity in the list (default Ansible behavior).
+        """
+        self._loader = loader
+        self._vault_id = vault_id
+
+    def __call__(self) -> DataLoader:
```
gmuloc: remove this

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -121,11 +125,71 @@ class ResolvedPluginArgs:
...
+    def __init__(self, loader: DataLoader, vault_id: str | None = None) -> None:
+        """
+        Initialize the wrapper with a DataLoader instance.
+
+        Args:
+            loader: The Ansible DataLoader instance.
+            vault_id: Optional vault identity to use for encryption. If None, uses the first
+                     vault identity in the list (default Ansible behavior).
+        """
+        self._loader = loader
```
gmuloc:
```suggestion
        self.loader = loader
```

ansible_collections/arista/avd/tests/integration/targets/validate_inputs_vault/runme.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+
+set -eux
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+ANSIBLE_PLAYBOOK="${ANSIBLE_PLAYBOOK:-ansible-playbook}"
+
+# Create a temporary directory for test outputs
+TEST_TMP_DIR="/tmp/avd_vault_test_$$"
```
gmuloc: let's use the output folder for this and not delete it

ansible_collections/arista/avd/tests/integration/targets/validate_inputs_vault/runme.sh
```diff
@@ -0,0 +1,50 @@
...
+mkdir -p "$TEST_TMP_DIR"
+
+# Test 1: Without vault password
+echo "=== Test 1: Without vault password ==="
+export AVDTMPDIR="$TEST_TMP_DIR/without_vault"
+# Temporarily disable vault by renaming ansible.cfg
+if [ -f "$SCRIPT_DIR/ansible.cfg" ]; then
+    mv "$SCRIPT_DIR/ansible.cfg" "$SCRIPT_DIR/ansible.cfg.bak"
+fi
+"$ANSIBLE_PLAYBOOK" -i "$SCRIPT_DIR/hosts.yml" "$SCRIPT_DIR/playbook.yml" --tags test1
```
gmuloc: different playbooks without tags for each test

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -170,8 +170,8 @@ def load_facts(self) -> dict[str, dict]:
             msg = f"Missing AVD eos_designs facts to generate documentation ({file_path}). Ensure the 'arista.avd.eos_designs_facts' task ran successfully."
             raise AnsibleActionFail(message=msg)
 
-        with file_path.open(mode="r", encoding="utf-8") as f:
-            return json.load(f)
+        # Read, unvault, and parse the JSON file
+        return self._loader.load_from_file(str(file_path), json_only=True)
```
alexeygorbunov: In `plugins/actions/eos_designs_structured_config.py::ActionModule::load_facts` we expect facts file to not be vaulted but here we try to unvauld it?

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -372,14 +444,19 @@ def _run_validation_phase(
...
         self.logger.info("Validating inputs...")
         start_time = perf_counter()
 
         data_validation_errors = 0
+        output_files = []
 
         init_store()
 
+        # Create a loader wrapper
+        loader = LoaderWrapper(self._loader, vault_id=vault_id)
```
alexeygorbunov: Is it needed to instantiate separate `LoaredWrapper`s for templating and validation?

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -372,14 +444,19 @@ def _run_validation_phase(
             schema_name: Schema to validate against.
             fail_on_validation_errors: Whether to fail the task on validation errors.
             configuration: Configuration for validation or None.
+            vault_id: Optional vault identity to use for encryption.
         """
         self.logger.info("Validating inputs...")
         start_time = perf_counter()
 
         data_validation_errors = 0
+        output_files = []
```
alexeygorbunov: Not used after populating

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: nope it is not. Moved it to main and passing it down the chain for now

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: duh... fixed the other one but not this one! thanks!

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
gmuloc: indeed. changed back

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
           from the EOS Config schema are found at the top level of AVD Design input data.
         type: bool
         default: false
+  vault_id:
+    description: |-
+      Optional vault identity to use for encrypting temporary files when Ansible Vault is configured.
+
+      When Ansible Vault is configured (via C(--vault-password-file), C(--vault-id), or C(vault_identity_list) in ansible.cfg),
+      the plugin encrypts temporary files containing templated and validated data to prevent sensitive information
+      from being exposed in logs or temporary directories.
```
alexeygorbunov: Does `avd_vault_id` prevent exposure in logs in any way?

alexeygorbunov:
Re-checked PR and manually re-tested possible workflows (following additional commits) and everything still works as expected

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -121,11 +123,67 @@ class ResolvedPluginArgs:
     read_from_input_dir: bool
     fail_on_validation_errors: bool
     validation_configuration: Configuration | None
+    vault_id: str | None
 
 
 _HOSTVARS_MANAGER: ActionPluginVars | None = None
 
 
+class LoaderWrapper:
```
carl-baillargeon: 52 lines for just a few lines of logic :P AI rambling...

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -453,9 +528,18 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
...
+        data = data.encode("utf-8")
+
+        # Encrypt data if vault is configured
+        data = loader.encrypt_if_needed(data)
+
+        # Write data to file
         output_file_path = output_path / f"{hostname}.json"
-        with output_file_path.open(mode="w", encoding="utf-8") as f:
-            json.dump(templated_hostvars, f, skipkeys=True, default=lambda _: "<not serializable>", indent=4)
+        output_file_path.write_bytes(data)
```
carl-baillargeon: We could have a method in LoaderWrapper to do this part and use it in the validate worker as well.

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
           from the EOS Config schema are found at the top level of AVD Design input data.
         type: bool
         default: false
+  vault_id:
+    description: |-
+      Optional vault identity to use for encrypting temporary files when Ansible Vault is configured.
+
+      When Ansible Vault is configured (via C(--vault-password-file), C(--vault-id), or C(vault_identity_list) in ansible.cfg),
```
carl-baillargeon: Text styling is not rendered properly in the plugin documentation.

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
...
+      B(Advanced Use Case) (when C(vault_id) is specified):
+        - The plugin uses the specified vault identity for encryption.
+        - This is useful when multiple vault identities are configured and you want to control which one is used.
+        - The specified vault identity must exist in the configured vault identities.
+
+      B(Examples):
+        - Single vault password: C(vault_id) is not needed, the single vault password is used automatically.
+        - Multiple vault identities via C(vault_identity_list = dev@.vault_dev, prod@.vault_prod):
+          - Without C(vault_id): Uses 'dev' (first in list) for encryption.
+          - With C(vault_id: 'prod'): Uses 'prod' for encryption.
```
carl-baillargeon: The additional indentation is not rendered properly in the documentation.

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
           from the EOS Config schema are found at the top level of AVD Design input data.
         type: bool
         default: false
+  vault_id:
+    description: |-
+      Optional vault identity to use for encrypting temporary files when Ansible Vault is configured.
```
carl-baillargeon:
```suggestion
      Optional vault identity to use for encrypting temporary files created by this plugin when Ansible Vault is configured.
```

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
...
+      B(Examples):
+        - Single vault password: C(vault_id) is not needed, the single vault password is used automatically.
+        - Multiple vault identities via C(vault_identity_list = dev@.vault_dev, prod@.vault_prod):
+          - Without C(vault_id): Uses 'dev' (first in list) for encryption.
+          - With C(vault_id: 'prod'): Uses 'prod' for encryption.
+        - Multiple vault identities via C(--vault-id dev@.vault_dev --vault-id prod@.vault_prod):
+          - Without C(vault_id): Uses 'dev' (first specified) for encryption.
+          - With C(vault_id: 'prod'): Uses 'prod' for encryption.
+
+      B(Note): If Ansible Vault is not configured, this parameter has no effect and files are written as plain JSON.
```
carl-baillargeon: I suggest we move this note to the top.

ansible_collections/arista/avd/tests/integration/targets/validate_inputs_vault/tasks/verify_vault_encryption.yml
```diff
@@ -0,0 +1,94 @@
+---
+# Tasks to verify vault encryption of temporary files
+# Args:
+#   - templated_file: Path to the templated file
+#   - validated_file: Path to the validated file
+#   - expected_encrypted: Whether files should be encrypted (true/false)
+#   - expected_vault_id: Expected vault ID in header (optional, only checked if expected_encrypted=true)
+#   - vault_password_file: Path to vault password file to test decryption (optional)
```
carl-baillargeon, line 8: Are these comments still accurate since the move to different playbooks?

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
```diff
@@ -0,0 +1,40 @@
...
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  avd_vault_id:
+    documentation_options:
+      table: role-settings
+    type: str
+    description: |-
```
carl-baillargeon, line 14: This is too verbose. Let me try to write something more concise.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -490,26 +581,25 @@ def _validate_host_worker(
...
-            # YAML input: load and convert to JSON for validation.
-            with input_file_path.open(mode="r", encoding="utf-8") as f:
-                data = yaml.load(f, Loader=yaml.CSafeLoader)
-            json_data = json.dumps(data)
-        else:
-            # JSON input: read directly.
-            with input_file_path.open(mode="r", encoding="utf-8") as f:
-                json_data = f.read()
+        # Load data using Ansible DataLoader (handles both YAML/JSON and vault decryption)
+        data = loader.loader.load_from_file(str(input_file_path))
```
carl-baillargeon: When using a relative path for AVDTMPDIR using this command `AVDTMPDIR=./debug ansible-playbook playbooks/fabric-build.yml -v `, I am getting the following error:
```
ERROR! DC1-SPINE1: Unexpected error in validation worker thread: Unable to retrieve file contents
Could not find or access '/home/carl/git_projects/ansible-avd-examples/dual-dc/playbooks/debug/arista_avd/eos_designs/templated/DC1-SPINE1.json' on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option
```

carl-baillargeon: For this PR we could return the absolute path from `_cached_tmp_path` using `resolve()`. Then we would pass the correct string path to Ansible everywhere.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -490,26 +581,25 @@ def _validate_host_worker(
...
+        json_data = json.dumps(data)
```
carl-baillargeon: This is slowing things down (eos_designs is 2X slower even without decryption) because `load_from_file` is deserializing JSON data so we need to serialize the data back which is unnecessary. In devel we read JSON data directly.

I think we should make the LoaderWrapper more intelligent and handle these things.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -121,11 +123,67 @@ class ResolvedPluginArgs:
...
+              encrypt(data, secret=None, vault_id='X') sets the header to 'X' but
+              uses the first secret for encryption.
+        """
+        if not self.loader._vault.secrets:
+            return data
+
+        # Use Ansible's match_encrypt_secret to find the correct secret
+        # This handles both None (first secret) and specific vault_id cases
+        # and raises a proper error if the vault_id is not found.
+        vault_id, secret = match_encrypt_secret(self.loader._vault.secrets, self._vault_id)
```
carl-baillargeon: When using vault, `match_encrypt_secret` is called everytime we call `encrypt_if_needed` which is once per host. In the loader initializer we could do something like:
```
self.vault_enabled = bool(loader._vault.secrets)

self._cached_vault_id = None
self._cached_secret = None

if self.vault_enabled:
    self._cached_vault_id, self._cached_secret = match_encrypt_secret(loader._vault.secrets, vault_id)
```
To avoid calling `match_encrypt_secret` multiple times.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -168,6 +226,12 @@ def main(self, task_vars: dict[str, Any]) -> None:
         mp_workers, mt_workers = get_workers(len(hosts_to_process), task_vars["ansible_forks"])
         templated_path, validated_path = get_role_tmp_paths(SCHEMA_MAP[plugin_args.schema_name])
 
+        # Check if vault is configured for encrypting temporary files
+        if self._loader._vault.secrets:
```
carl-baillargeon: It would be cleaner to initialize the loader earlier and have a property like `if loader.vault_enabled: ...`

carl-baillargeon (changes requested):
Based from my testing, when encryption is enabled, total runtime of a full fabric build (1040 hosts) has a ~77% increase (3:15 versus 1:50).

Without encryption, there is an increase of ~23% (2:15 versus 1:50) which is unexpected but this is because we are using Ansible's file loader even if we don't need to (no vault used).

Related to my comments, I think we need to add more logic to the loader class to handle file I/O with encryption/decryption if vault is enabled.

It could be something like:
```
class LoaderWrapper:
    def load_file_as_json_string(self, file_path: Path) -> str:
        if self.vault_enabled:
            # Use Ansible's load_from_file
        else:
            # If JSON file we return as-is
            # If YAML we load and dump

    def write_json_file(self, file_path: Path, data: str | dict) -> None:
        # data is either a JSON string or a dict to serialize (to support both workers)
        
        if self.vault_enabled:
            # We encrypt the data
            
        # Write bytes to file
```

Then in the worker functions we simply use these methods and we don't need to care about dealing with encoding, encryption, etc.

Later on this class could be part of `AvdActionPlugin` and we could use these methods (or add new ones if needed) in other plugins instead of `self._loader`.

If we do this I think `VaultHandler` might be a better name :)

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
MaheshGSLAB: And we can use ref if description is same for both role

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: I wrote that :p

ansible_collections/arista/avd/tests/integration/targets/validate_inputs_vault/tasks/verify_vault_encryption.yml
gmuloc: yes.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: we said yesterday we would not put the writing in the loader during our call.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: what if you set different vault_id per host?

ClausHolbechArista: you cannot set a different vault id per host. It is a role variable and vault secrets are given for the play

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: no more LoaderWrapper

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: this still needs investigation

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: this has moved and is called in only one place

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: indeed - moved the loader in this method but did not see there was this above. Cleaner now

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -197,8 +198,15 @@ def load_structured_config(self, hostname: str) -> dict[str, Any]:
...
-            return json.load(f)
+        # Read the file
+        structured_config = file_path.read_bytes()
+
+        # Decrypt if vault is configured
+        # TODO: need vault_id to initialize the vault or we may pick up the wrong comment
+        vault_handler = VaultHandler(self._loader)
+        structured_config = vault_handler.maybe_decrypt(structured_config)
+
+        return json.loads(structured_config)
```
carl-baillargeon: We are repeating this pattern in other plugins. Why can't we have helpers in VaultHandler to handle this?

You said we discussed this in our call yesterday but I still don't understand why we would not want this :P

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
gmuloc: removed

gmuloc: there was a version where I was doing `no_log: true` but thats outdated

ansible_collections/arista/avd/plugins/action/validate_inputs.py
gmuloc: tested and no more issue with `./debug`

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -170,6 +170,7 @@ def load_facts(self) -> dict[str, dict]:
             msg = f"Missing AVD eos_designs facts to generate documentation ({file_path}). Ensure the 'arista.avd.eos_designs_facts' task ran successfully."
             raise AnsibleActionFail(message=msg)
 
+        # fact file is not vaulted.
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -219,6 +221,7 @@ def load_facts(self, hostname: str) -> AvdSwitchFactsDefaultDict:
             msg = f"Missing AVD eos_designs facts for host '{hostname}' ({file_path}). Ensure the 'arista.avd.eos_designs_facts' task ran successfully."
             raise AnsibleActionFail(message=msg)
 
+        # fact file is not vaulted.
```
```suggestion
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -490,26 +526,31 @@ def _validate_host_worker(
...
+            # JSON input: decode bytes to string
+            json_data = file_content.decode("utf-8")
 
         # Validation in Rust, releasing the GIL.
         validated_data_result = get_validated_data(data_as_json=json_data, schema_name=SCHEMA_MAP[schema_name], configuration=configuration)
         validation_result, validated_data = validated_data_result.validation_result, validated_data_result.validated_data
 
         output_file = None
         if validated_data:
+            # validated_data is already bytes (JSON format) from pyavd-utils
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -490,26 +526,31 @@ def _validate_host_worker(
...
+            # Encrypt data if vault is configured
+            validated_data = loader.encrypt_if_needed(validated_data.encode("utf-8"))
+
+            # Write data to file
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -453,9 +473,18 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
         # missing variables in inline templates in Ansible 2.19.
         templated_hostvars = dict(hostvars_wrapper)
 
+        # Convert to JSON string and write to file
+        data = json.dumps(templated_hostvars, skipkeys=True, default=lambda _: "<not serializable>", indent=4)
+
+        # Convert to bytes for binary file write
```
gmuloc:
```suggestion
```

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -14,6 +14,8 @@
 from ansible.plugins.action import ActionBase, display
 
 from ansible_collections.arista.avd.plugins.plugin_utils.utils import (
+    AVDFileHandler,
+    AVDVaultHandler,
```
gmuloc, line 18: Note that the goal is this will be done in the AvdActionPlugin going forward once we have migrated all the plugins and so we will be able to remove these

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -197,8 +199,10 @@ def load_structured_config(self, hostname: str) -> dict[str, Any]:
             )
             raise AnsibleActionFail(message=msg)
 
-        with file_path.open(mode="r", encoding="utf-8") as f:
-            return json.load(f)
+        # Read, unvault, and parse the JSON file
+        vault_handler = AVDVaultHandler(self._loader)
```
gmuloc, line 207: here we don't have vault_id in the plugins but this is ok because they only need to decrypt and no vault id is needed when doing this

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
```diff
@@ -0,0 +1,20 @@
...
+      Vault identity to use for encrypting temporary files when Ansible Vault is configured.
```
gmuloc:
```suggestion
      Vault ID used for encrypting temporary files generated by the role.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
```diff
@@ -0,0 +1,20 @@
...
+      When Ansible Vault is not configured, this parameter has no effect and files are written as plain JSON.
+      When Ansible Vault is configured AVD encrypts temporary files containing templated and validated data
+      to prevent sensitive information from being exposed in temporary directories.
```
gmuloc:
```suggestion
      to prevent sensitive information from being exposed in the temporary directories.
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
```diff
@@ -0,0 +1,20 @@
...
+type: dict
+keys:
+  avd_vault_id:
+    documentation_options:
+      table: role-settings
+    type: str
+    description: |-
+      Vault identity to use for encrypting temporary files when Ansible Vault is configured.
+      When Ansible Vault is not configured, this parameter has no effect and files are written as plain JSON.
+      When Ansible Vault is configured AVD encrypts temporary files containing templated and validated data
```
gmuloc:
```suggestion
      When Ansible Vault is configured, AVD encrypts files containing templated and validated data
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/avd_vault_id.schema.yml
```diff
@@ -0,0 +1,20 @@
...
+      to prevent sensitive information from being exposed in temporary directories.
+        * when `avd_vault_id` is not specified, AVD uses the *first* vault identity in the list for encryption.
+        * when `avd_vault_id` is specified, AVD uses the specified vault identity for encryption.
```
gmuloc:
```suggestion
        * When `avd_vault_id` is not specified, AVD uses the *first* Vault ID in the list for encryption.
        * When `avd_vault_id` is specified, AVD uses the specified Vault ID for encryption.
```

ansible_collections/arista/avd/plugins/modules/validate_inputs.py
```diff
@@ -64,6 +64,36 @@
...
+        - The specified vault identity must exist in the configured vault identities.
+
+      **Examples**:
+        - Single vault password: `vault_id` is not needed, the single vault password is used automatically.
+        - Multiple vault identities via `vault_identity_list = dev@.vault_dev, prod@.vault_prod`:
+          - Without `vault_id`: Uses 'dev' (first in list) for encryption.
+          - With `vault_id: 'prod'`: Uses 'prod' for encryption.
+        - Multiple vault identities via `--vault-id dev@.vault_dev --vault-id prod@.vault_prod`:
+          - Without `vault_id`: Uses 'dev' (first specified) for encryption.
+          - With `vault_id: 'prod'`: Uses 'prod' for encryption.
```
gmuloc:
```suggestion
      Vault ID used for encrypting temporary files generated by the role.
      When Ansible Vault is not configured, this parameter has no effect and files are written as plain JSON.
      When Ansible Vault is configured, AVD encrypts files containing templated and validated data
      to prevent sensitive information from being exposed in the temporary directories.
        * When `avd_vault_id` is not specified, AVD uses the *first* Vault ID in the list for encryption.
        * When `avd_vault_id` is specified, AVD uses the specified Vault ID for encryption.
```

ClausHolbechArista:
pre-commit.ci autofix

carl-baillargeon (approved):
LGTM! Tested locally.

I moved the `match_encrypt_secret` to the init function to avoid calling it N of hosts times. Encryption now adds ~43% overhead (versus 77% we had before).

Also, without encryption, we get the same numbers now :)

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -201,8 +203,10 @@ def load_structured_config(self, hostname: str) -> dict[str, Any]:
...
+        file_handler = AVDFileHandler(vault_handler)
```
alexeygorbunov, line 208: Isn't this exsessive to import and instantiate `AVDVaultHandler` for only pssing it into `AVDFileHandler` constructor? Can't we just instantiate `AVDFileHandler` passing `self.loader` into it and let `AVDFileHandler` instantiate `AVDVaultHandler` as a part of the __init__?

Tis is for all 3 plugins

carl-baillargeon: This will go away when we migrate to AVDActionPlugin base class in the future. See Guillaumes comment here: https://github.com/aristanetworks/avd/pull/6463#discussion_r2769377592.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -319,6 +336,7 @@ def _run_templating_phase(
             batch_size: Number of hosts to process per child process.
             output_path: Directory path where templated JSON files will be written.
             schema_name: Schema name used for filtering hostvars.
+            file_handler: AVDFileHandler, used to load and write files, handling encryption if needed.
```
alexeygorbunov: read instead of load?
```suggestion
            file_handler: AVDFileHandler, used to read and write files, handling encryption if needed.
```

ansible_collections/arista/avd/plugins/action/validate_inputs.py
```diff
@@ -455,9 +477,9 @@ def _template_host_worker(hostname: str, output_path: Path, schema_name: SCHEMA_
         # missing variables in inline templates in Ansible 2.19.
         templated_hostvars = dict(hostvars_wrapper)
 
+        data = json.dumps(templated_hostvars, skipkeys=True, default=lambda _: "<not serializable>", indent=4).encode("utf-8")
```
alexeygorbunov, line 480: As AVDFileHandler has `load_json` method should we as well handle JSON serialization as it's another method for consistency?

carl-baillargeon: That's business logic specific to how hostvars are serialized (using non-default parameters for serialization). The file handler should remain generic.

carl-baillargeon: `load_json` is borderline but I think that's acceptable because it is common and generic enough.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
carl-baillargeon: Changed.

alexeygorbunov:
Some of the raised errors are native Ansible exceptions. Should those be processes and re-raise as AVD excetptions?

```bash
  File "/usr/local/lib/python3.12/site-packages/ansible/parsing/vault/__init__.py", line 558, in match_encrypt_vault_id_secret
    raise AnsibleVaultError('Did not find a match for --encrypt-vault-id=%s in the known vault-ids %s' % (encrypt_vault_id,
ansible.parsing.vault.AnsibleVaultError: Did not find a match for --encrypt-vault-id=dev1 in the known vault-ids ['dev']
fatal: [dc1-spine1 -> localhost]: FAILED! => {
    "changed": false,
    "msg": "Error during plugin 'arista.avd.validate_inputs' execution: 'Did not find a match for --encrypt-vault-id=dev1 in the known vault-ids ['dev']'"
}

PLAY RECAP *********************************************************************
dc1-spine1                 : ok=2    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
```

```code
PLAY [Build Configurations and Documentation] **********************************
ERROR! Decryption failed (no vault secrets were found that could decrypt) on /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/group_vars/FABRIC/fabric_variables.yml
CRITICAL Ansible return code was 4, command was: ansible-playbook --inventory /root/.ansible/tmp/molecule.IO93.example-single-dc-l3ls/inventory --skip-tags molecule-notest,notest --inventory=/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions/molecule/example-single-dc-l3ls/../../../examples/single-dc-l3ls/inventory.yml -vvv --vault-id prd@.prd /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/build.yml
ERROR    example-single-dc-l3ls ➜ converge: Executed: Failed
ERROR    Ansible return code was 4, command was: ansible-playbook --inventory /root/.ansible/tmp/molecule.IO93.example-single-dc-l3ls/inventory --skip-tags molecule-notest,notest --inventory=/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions/molecule/example-single-dc-l3ls/../../../examples/single-dc-l3ls/inventory.yml -vvv --vault-id prd@.prd /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/build.yml
root@6b4a8bc3c06f:/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions# 

```

carl-baillargeon:
> Some of the raised errors are native Ansible exceptions. Should those be processes and re-raise as AVD excetptions?
> 
> ```shell
>   File "/usr/local/lib/python3.12/site-packages/ansible/parsing/vault/__init__.py", line 558, in match_encrypt_vault_id_secret
>     raise AnsibleVaultError('Did not find a match for --encrypt-vault-id=%s in the known vault-ids %s' % (encrypt_vault_id,
> ansible.parsing.vault.AnsibleVaultError: Did not find a match for --encrypt-vault-id=dev1 in the known vault-ids ['dev']
> fatal: [dc1-spine1 -> localhost]: FAILED! => {
>     "changed": false,
>     "msg": "Error during plugin 'arista.avd.validate_inputs' execution: 'Did not find a match for --encrypt-vault-id=dev1 in the known vault-ids ['dev']'"
> }
> 
> PLAY RECAP *********************************************************************
> dc1-spine1                 : ok=2    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0
> ```
> 
> ```
> PLAY [Build Configurations and Documentation] **********************************
> ERROR! Decryption failed (no vault secrets were found that could decrypt) on /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/group_vars/FABRIC/fabric_variables.yml
> CRITICAL Ansible return code was 4, command was: ansible-playbook --inventory /root/.ansible/tmp/molecule.IO93.example-single-dc-l3ls/inventory --skip-tags molecule-notest,notest --inventory=/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions/molecule/example-single-dc-l3ls/../../../examples/single-dc-l3ls/inventory.yml -vvv --vault-id prd@.prd /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/build.yml
> ERROR    example-single-dc-l3ls ➜ converge: Executed: Failed
> ERROR    Ansible return code was 4, command was: ansible-playbook --inventory /root/.ansible/tmp/molecule.IO93.example-single-dc-l3ls/inventory --skip-tags molecule-notest,notest --inventory=/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions/molecule/example-single-dc-l3ls/../../../examples/single-dc-l3ls/inventory.yml -vvv --vault-id prd@.prd /usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/examples/single-dc-l3ls/build.yml
> root@6b4a8bc3c06f:/usr/src/app/avd_pr_devel/ANSIBLE_COLLECTIONS_AVD_DEVEL/ansible_collections/arista/avd/extensions# 
> ```

It is caught by the base class here (similar to other unknown Ansible errors): https://github.com/aristanetworks/avd/blob/devel/ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py#L107

I think the error message is clear enough.

alexeygorbunov (approved):
LGTM, all use cases that i tried/tested worked as expected.

gmuloc:
> LGTM! Tested locally.
> 
> I moved the `match_encrypt_secret` to the init function to avoid calling it N of hosts times. Encryption now adds ~43% overhead (versus 77% we had before).

Not great as now if the vault_id is changed then the secret is wrong

PR #6470, opened 2026-02-03
Doc(eos_designs): Update diagram for eos_designs role

JulioPDX:
## Change Summary

Updated workflow diagram for eos_designs.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Simplified diagram and moved to excalidraw. Easy to grab file and edit and does not require custom icons. 

## How to test

Serve locally or view generated docs.

### Repository Checklist


docs/_media/eos_designs_dark.svg
alexeygorbunov: Is top center block related to eos_designs_facts? If yes then it always dumps facts (which are not a full structured config)into the temp files (so it's not really optional as next plugins depend on it's content)

alexeygorbunov, line 1: "Documentation" box mentions generated files but not with their default names. Wondering if not using the right names may be confusing in any way

docs/_media/excalidraw/eos_designs.excalidraw
alexeygorbunov, line 1: Where is this drawing exposed within avd.arista.com? Can't find reference to it anywhere in the code

carlbuchmann: https://ansible-avd--6470.org.readthedocs.build/en/6470/ansible_collections/arista/avd/roles/eos_designs/index.html

alexeygorbunov:
Should we consider adding "input validation" step to the diagram?
Also not sure how often dumping of structured config into facts is used in the field and if this deserves to be reflected on the graph (especially when validation is not there)

Vibhu-gslab (approved):

JulioPDX:
> Should we consider adding "input validation" step to the diagram? Also not sure how often dumping of structured config into facts is used in the field and if this deserves to be reflected on the graph (especially when validation is not there)

I think you are to something here. Let me think on that one. Input validation would be good

ClausHolbechArista:
> > Should we consider adding "input validation" step to the diagram? Also not sure how often dumping of structured config into facts is used in the field and if this deserves to be reflected on the graph (especially when validation is not there)
> 
> I think you are to something here. Let me think on that one. Input validation would be good

I don't agree. Input validation is just an implementation detail. We also validate dependencies and other things. The role overview should focus on the inputs and outputs. The text in the role doc already details the internal steps taken in the role, but I don't think the overview drawing needs to.

docs/_media/eos_designs_dark.svg
gmuloc: let's rename
```
fabric-documentation.md
fabric-p2p-links.csv
fabric-topology.csv
```

JulioPDX: ack, will do

carl-baillargeon (approved):

alexeygorbunov:
> The role overview should focus on the inputs and outputs

Yes, agree with this

alexeygorbunov (approved):
LGTM!

PR #6480, opened 2026-02-05
CI: Updated the j2 compilation method to use the generic name for compiled files

MaheshGSLAB:
## Change Summary

Updated the j2 compilation method to use the generic name for compiled files

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
Updated the j2 compilation method to use the generic name for compiled files

## How to test
Check the compiled files in Test pyavd python 3.11 coverage report job

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/templater.py
```diff
@@ -125,13 +191,57 @@ def compile_templates_in_paths(self, precompiled_templates_path: str | Path, sea
...
+        for template_name in templates:
+            # Get the template source
+            source, filename, _uptodate = self.environment.loader.get_source(self.environment, template_name)
+
+            # Parse and compile to Python source code
+            try:
+                code = self.environment.parse(source, template_name, filename)
+                module_code = generate(code, self.environment, template_name, filename, defer_init=True)
+            except Exception:
+                LOGGER.exception("Failed to compile template %s", template_name)
```
ClausHolbechArista: You should include the original error and raise from.

ClausHolbechArista: Don't mix raise and log. It makes it harder to debug when we have to look in two different places to understand the error.

python-avd/pyavd/templater.py
```diff
@@ -125,13 +191,57 @@ def compile_templates_in_paths(self, precompiled_templates_path: str | Path, sea
...
+
+        # Create __init__.py with template name to module name mapping
+        init_file = precompiled_path / "__init__.py"
+        with init_file.open("w") as f:
+            f.write("# Precompiled Jinja2 templates\n")
+            f.write("# Template name to module name mapping\n")
+            f.write("__module_dict__ = {\n")
+            f.write("\n".join(module_dict_entries))
+            f.write("\n}\n")
+
```
ClausHolbechArista: This feels super hacky. The include commands come with the proper path, could we not just use the same function to build the filenames on import as we do on compilation?

python-avd/pyavd/templater.py
```diff
@@ -44,18 +49,79 @@ def __contains__(self, _item: int) -> Undefined:
...
+        # Try to import the module
+        mod = getattr(self.module, module_name, None)
+
+        if mod is None:
+            try:
+                mod = __import__(module_path, None, None, ["root"])
+            except ImportError as e:
+                LOGGER.debug("Failed to import '%s': %s", module_path, e)
+                raise TemplateNotFound(name) from e
```
ClausHolbechArista: raise the the context instead of logging

python-avd/pyavd/templater.py
```diff
@@ -44,18 +49,79 @@ def __contains__(self, _item: int) -> Undefined:
...
+
+    @staticmethod
+    def _template_name_to_module_name(template_name: str) -> str:
+        """Convert a template name to a module name."""
+        return template_name.replace("/", "__").replace("\\", "__").removesuffix(".j2").replace(".", "_").replace("-", "_")
+
+
 class Templar:
     def __init__(self, precompiled_templates_path: str | Path, searchpaths: list[str | Path] | None = None) -> None:
-        if not RUNNING_FROM_SRC:
```
ClausHolbechArista, line 49: This is here for a reason... When we run from source during development, we want templates to be included even if they have not been compiled.

python-avd/pyavd/templater.py
MaheshGSLAB: yeah this was code while debugging so yes i need to polish the code with proper exception.

python-avd/pyavd/templater.py
MaheshGSLAB: removed

python-avd/pyavd/templater.py
MaheshGSLAB: updated

python-avd/pyavd/templater.py
```diff
@@ -44,15 +48,56 @@ def __contains__(self, _item: int) -> Undefined:
...
+            # Remove from sys.modules to keep it only as an attribute on self.module
+            sys.modules.pop(module_path, None)
+
+        # Create template from module dict
+        return environment.template_class.from_module_dict(environment, mod.__dict__, globals if globals is not None else {})
+
+    @staticmethod
+    def _template_name_to_module_name(template_name: str) -> str:
+        """Convert a template name to a module name."""
+        return template_name.replace("/", "__").replace("\\", "__").removesuffix(".j2").replace(".", "_").replace("-", "_")
```
Shivani-gslab: it's better to normalize the path before building the name (to handle path in different platforms)

Shivani-gslab: also we can add a short hash in the name to make them unique and avoid collision in any case

Shivani-gslab (commented):
Great work @MaheshGSLAB! The readable module names would be very helpful and make debugging much easier.

One small thought — would it be useful to cache just the generated template/module names (not the full templates), especially when the same templates are reused across runs? This could help avoid recomputing names and make things more consistent and efficient.

Any thoughts @ClausHolbechArista @gmuloc ? Would this kind of lightweight caching make sense here?

python-avd/pyavd/templater.py
MaheshGSLAB: used normalize path.

About hash on module i will suggest to raise an error because the duplication case can occur when i have two template as `aaa-aaa.j2` and `aaa_aaa.j2` in same dir, which might be rare case but we can handle it via raising the error.

ClausHolbechArista:
@Shivani-gslab I don't think it is worth the complexity to cache anything. Just have a consistent way of generating the name from the path. And we may need a prefix or suffix for doc vs cli to avoid any collissions

python-avd/pyavd/templater.py
ClausHolbechArista: how about between doc and cli?

Shivani-gslab: About doc and cli, currently the files generated are named with prefix `documentation` and `eos
`

for the files insdie `eos` and `doc` directories it is giving `__` double underscore and for the files inside `j2templates folder` it gives single underscore.

example -
`documentation__aaa_accounting.py`
`eos__aaa_accounting.py`
`eos_intended_config.py`

Shivani-gslab (approved):
Need to finalize the file name format (prefix/suffix) 
Rest LGTM.

laxmikantchintakindi (approved):

ClausHolbechArista (commented):
LGTM - the test classes do not follow our usual flatter pattern, but they look correct.
I will not approve until @gmuloc have also had a chance to look at this.

python-avd/pyavd/templater.py
```diff
@@ -44,15 +48,58 @@ def __contains__(self, _item: int) -> Undefined:
...
 
 
+class CustomModuleLoader(ModuleLoader):
+    """Custom ModuleLoader that handles readable module names."""
+
+    def __init__(self, path: str | Path) -> None:
+        # Ensure the path is in sys.path so modules can be imported
+        path_str = str(Path(path).resolve())
+        if path_str not in sys.path:
+            sys.path.insert(0, path_str)
```
gmuloc: why do we need this when the normal loader is able to find them?

python-avd/pyavd/templater.py
```diff
@@ -44,15 +48,58 @@ def __contains__(self, _item: int) -> Undefined:
...
+            sys.modules.pop(module_path, None)
+
+        # Create template from module dict
+        return environment.template_class.from_module_dict(environment, mod.__dict__, globals if globals is not None else {})
+
+    @staticmethod
+    def _template_name_to_module_name(template_name: str) -> str:
+        """Convert a template name to a module name."""
+        normalized = template_name.replace("\\", "/")
+        normalized = Path(normalized).as_posix()
```
gmuloc:
```suggestion
        normalized = Path(template_name.replace("\\", "/")).as_posix()
```

python-avd/pyavd/templater.py
```diff
@@ -44,15 +48,58 @@ def __contains__(self, _item: int) -> Undefined:
...
+
+    def load(self, environment: Environment, name: str, globals: MutableMapping[str, Any] | None = None) -> Template:  # noqa: A002
+        """Load template using the module name conversion."""
+        # Convert template name to module name
+        module_name = self._template_name_to_module_name(name)
+
+        # Build the full module path
+        module_path = f"{self.package_name}.{module_name}"
+
+        # Try to import the module
```
gmuloc: are we really importing next line? Sounds like we are checking if this is already imported maybe? How do we detecte name clashes?

python-avd/tests/pyavd/utils/test_templater.py
```diff
@@ -0,0 +1,376 @@
...
+        CustomModuleLoader(precompiled_dir)
+
+        assert path_str in sys.path
+
+    def test_load_nonexistent_template_raises(self, precompiled_dir: Path) -> None:
+        """Test that loading a non-existent template raises TemplateNotFound."""
+        from jinja2 import Environment
+
+        loader = CustomModuleLoader(precompiled_dir)
+        env = Environment(loader=loader)  # noqa: S701
```
gmuloc: why the noqa when you could simply add


```suggestion
        env = Environment(loader=loader, autoescape=True)
```

python-avd/pyavd/templater.py
```diff
@@ -44,15 +48,58 @@ def __contains__(self, _item: int) -> Undefined:
...
+        mod = getattr(self.module, module_name, None)
+
+        if mod is None:
+            try:
+                mod = __import__(module_path, None, None, ["root"])
```
gmuloc: could we use something instead `__import__`? `importlib.import_module`: https://docs.python.org/3/library/importlib.html#importlib.__import__

Image

python-avd/tests/pyavd/utils/test_templater.py
```diff
@@ -0,0 +1,376 @@
...
+        """Test that environment uses custom Undefined class."""
+        templar = Templar(precompiled_templates_path=precompiled_dir)
+
+        assert templar.environment.undefined is Undefined
+
+    def test_environment_has_filters(self, precompiled_dir: Path) -> None:
+        """Test that environment has all custom filters registered."""
+        templar = Templar(precompiled_templates_path=precompiled_dir)
+
+        expected_filters = [
```
gmuloc: this means every time we add a filter we need to update the test. Maybe we could create a constant in templater and use the keys here?

python-avd/tests/pyavd/utils/test_templater.py
```diff
@@ -0,0 +1,376 @@
...
+
+        for filter_name in expected_filters:
+            assert filter_name in templar.environment.filters
+
+    def test_environment_has_tests(self, precompiled_dir: Path) -> None:
+        """Test that environment has all custom tests registered."""
+        templar = Templar(precompiled_templates_path=precompiled_dir)
+
+        expected_tests = [
+            "arista.avd.defined",
```
gmuloc: same here

gmuloc (changes requested):

python-avd/pyavd/templater.py
MaheshGSLAB: We need to store self.path so that the load method can construct direct filesystem paths to the precompiled template modules (see [line 68](vscode-webview://0j34o0q90lcmqocrluho2gh7siv5vonji9j5vrna0r2a7p73e77c/templater.py#L68): module_file = self.path / f"{module_name}.py").

The parent ModuleLoader class is designed to work with Python's import system using package names, not direct file access. It doesn't expose the path in a way that allows us to build filesystem paths. By storing our own self.path, we can use importlib.util.spec_from_file_location() to load modules directly from file paths without relying on sys.path.

python-avd/pyavd/templater.py
MaheshGSLAB: updated

python-avd/pyavd/templater.py
MaheshGSLAB: So collisions are caught when templates are compiled, not when they're loaded at runtime. By compilation time, all potential name clashes have already been detected and prevented.

python-avd/tests/pyavd/utils/test_templater.py
MaheshGSLAB: updated

python-avd/pyavd/templater.py
MaheshGSLAB: updated

python-avd/tests/pyavd/utils/test_templater.py
MaheshGSLAB: updated

python-avd/pyavd/templater.py
```diff
@@ -117,48 +145,48 @@ def __init__(self, precompiled_templates_path: str | Path, searchpaths: list[str
...
-            },
-        )
-        self.environment.tests.update(
-            {
-                "arista.avd.defined": defined,
-                "arista.avd.contains": contains,
-            },
-        )
+    @staticmethod
+    def _check_and_recompile_templates() -> None:
```
gmuloc: as discussed lets reconvene and if we go in that direction let's split in another PR

MaheshGSLAB: revert the recompile logic

python-avd/pyavd/templater.py
ClausHolbechArista: I think we could simplify all this by building our own loader instead of retrofitting into the module loader.

python-avd/pyavd/templater.py
```diff
@@ -71,52 +141,32 @@ def __init__(self, precompiled_templates_path: str | Path, searchpaths: list[str
...
-        )
-        self.environment.tests.update(
-            {
-                "arista.avd.defined": defined,
-                "arista.avd.contains": contains,
-            },
-        )
-
-    def render_template_from_file(self, template_file: str, template_vars: dict) -> str:
+        # Dynamically import and register filters from constants
```
ClausHolbechArista: why do we need to change this?

gmuloc: I think this comes from my comment here: https://github.com/aristanetworks/avd/pull/6480#discussion_r2788531421

gmuloc: We could avoid all this dance by doing something like

In `pyavd/j2filters.py`:
```
FILTERS = {
      "add_md_toc": add_md_toc,
      "decrypt": decrypt,
      "default": default,
      # ...
}
"""Dictionay of filters to be imported in templater"""

We could achieve this by looping over filters in the module given that we expose them via `__all__` but here we are being explicit about what we want to export.

 
In `python-avd/pyavd/templater.py`

```
def import_filters_and_tests(self) -> None:
    filters_module = importlib.import_module(".j2filters", package="pyavd")
    for name, func in filters_module.FILTERS.items():
          self.environment.filters[f"arista.avd.{name}"] = func
```

Comment (gmuloc) [reply python-avd/pyavd/templater.py]: lets rveert this change and remove the tests associated to it

Comment (MaheshGSLAB) [reply python-avd/pyavd/templater.py]: reverted the changes and removed the tests

File: python-avd/pyavd/templater.py (hunk 6 above)
Comment (MaheshGSLAB) [reply python-avd/pyavd/templater.py]: @ClausHolbechArista about why not using get_module_filename from ModuleLoader is 

The standard ModuleLoader uses python import system (__import__), which requires the directory to be on sys.path and proper package structure with __init__.py files.

We're using spec_from_file_location to load the .py files directly from absolute paths without needing sys.path or package structure. This keeps things simpler since our compiled templates are just standalone .py files in a flat directory.

Also get_module_filename generates SHA1-based filenames like tmpl_abc123.py instead of readable names like foo__bar_baz.py.

File: python-avd/pyavd/templater.py [hunk 15]
```diff
@@ -44,15 +49,59 @@ def __contains__(self, _item: int) -> Undefined:
         return self
 
 
+class CustomModuleLoader(ModuleLoader):
```
gmuloc, line 52: Have you tried overriding the two naming functions rather than load in a version?

gmuloc: actually lets go with this and move on for now. :) thanks for the work @MaheshGSLAB

ClausHolbechArista (approved):

gmuloc (approved):

PR #6484, opened 2026-02-05
Doc: Improve installation workflow

carlbuchmann:
## Change Summary

- Improve installation workflow steps and links to packages.
- Add a note for the Python-supported version with ansible-core.

Review Doc: https://ansible-avd--6484.org.readthedocs.build/en/6484/docs/installation/collection-installation.html

### DIFF
--- a/docs/installation/collection-installation.md
+++ b/docs/installation/collection-installation.md
@@ -8,12 +8,19 @@
 
 ## Installation workflow
 
-- Install [Python](https://www.python.org/downloads/) **3.10** or later.
+- Install supported [Python](https://www.python.org/downloads/) version **3.10-3.14.
+  - The Python version determines the supported `ansible-core` version. Consult the [ansible-core support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) for more information.
 - Install the [pyavd](https://pypi.org/project/pyavd/) Python package with `ansible` dependencies.
+  - note: This will install the supported ansible-core version and all required [Python dependencies](#python-dependencies).
 - Install [arista.avd](https://galaxy.ansible.com/ui/repo/published/arista/avd/) Ansible collection.
 
-!!! note
-    The supported Python version is determined by the [ansible-core support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix).
+## Python dependencies
+
+``` shell
+--8<--
+ansible_collections/arista/avd/requirements.txt:4:
+--8<--
+```
 
 ## Install Collection from Ansible Galaxy
 
--- a/docs/installation/collection-installation.md
+++ b/docs/installation/collection-installation.md
@@ -16,7 +16,7 @@
 
 ## Python dependencies
 
-``` shell
+```text
 --8<--
 ansible_collections/arista/avd/requirements.txt:4:
 --8<--
@@ -50,8 +50,8 @@ ansible-galaxy collection install arista.avd
 ### Install a specific version
 
 ```shell
-pip install "pyavd[ansible]==4.8.0"
-ansible-galaxy collection install arista.avd:==4.8.0
+pip install "pyavd[ansible]==5.7.3"
+ansible-galaxy collection install arista.avd:==5.7.3
 ```
 
 ### Install latest `devel` version from AVD GitHub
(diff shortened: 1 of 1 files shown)


docs/installation/collection-installation.md
```diff
@@ -8,8 +8,12 @@
...
 ## Installation workflow
 
-- Install [Python](https://www.python.org/downloads/) **3.10** or later
-- Install [arista.avd](#install-collection-from-ansible-galaxy) collection including Python requirements.
+- Install [Python](https://www.python.org/downloads/) **3.10** or later.
+- Install the [pyavd](https://pypi.org/project/pyavd/) Python package with `ansible` dependencies.
+- Install [arista.avd](https://galaxy.ansible.com/ui/repo/published/arista/avd/) Ansible collection.
+
+!!! note
+    The supported Python version is determined by the [ansible-core support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix).
```
alexeygorbunov: Under some conditions it is possible to observe unexpected outcomes of running Ansible+AVD with no clear indication of incorrect Py & ansible-core combination being the cause of it.
Should we implement validation of the Py + ansible-core right in our plugins to catch those issues and provide clear exception?
Or optionally we can maybe just reference users to the place in out documentation where they can find version of the `ansible-core` "bundled" with this specific release of the AVD/`pyavd`

gmuloc: > Should we implement validation of the Py + ansible-core right in our plugins to catch those issues and provide clear exception? 

this is already implemented in Verify Requirements. which needs to be extended today it does not catch this because we only check the `ansible-collection` extra dependencies when pyavd was installed this way so lets make this better.

docs/installation/collection-installation.md
```diff
@@ -8,8 +8,12 @@
 
 ## Installation workflow
 
-- Install [Python](https://www.python.org/downloads/) **3.10** or later
-- Install [arista.avd](#install-collection-from-ansible-galaxy) collection including Python requirements.
+- Install [Python](https://www.python.org/downloads/) **3.10** or later.
```
ClausHolbechArista: lets add a max version instead of expecting the users to understand all the matrix stuff. It could replace the note.

docs/installation/collection-installation.md
```diff
@@ -8,8 +8,12 @@
...
+- Install the [pyavd](https://pypi.org/project/pyavd/) Python package with `ansible` dependencies.
```
ClausHolbechArista, line 13: add "which will install all the Python dependencies: (insert snip of the requirements.txt file we now generate)"

docs/installation/collection-installation.md
carlbuchmann: done

docs/installation/collection-installation.md
carlbuchmann: done

docs/installation/collection-installation.md
```diff
@@ -79,18 +90,18 @@ You can use `-U` to upgrade to a new version for any installed collection:
...
 Starting collection install process
-Installing 'arista.avd:4.9.0' to '/home/arista/.ansible/collections/ansible_collections/arista/avd'
+Installing 'arista.avd:6.0.0' to '/home/arista/.ansible/collections/ansible_collections/arista/avd'
 ```
 
 After an upgrade, some python requirements may have changed. Make sure to also update the Python requirements for the same version (the version given below matches the installed collection above):
 
 ```shell
-pip install "pyavd[ansible]==4.9.0"
+pip install "pyavd[ansible]==6.0.0"
```
JulioPDX, line 99: Would this require the upgrade flag?

```suggestion
pip install "pyavd[ansible]==6.0.0" --upgrade
```

JulioPDX: disregard, not required

JulioPDX (approved):
LGTM

carl-baillargeon (approved):

docs/installation/collection-installation.md
```diff
@@ -8,8 +8,19 @@
 
 ## Installation workflow
 
-- Install [Python](https://www.python.org/downloads/) **3.10** or later
-- Install [arista.avd](#install-collection-from-ansible-galaxy) collection including Python requirements.
+- Install supported [Python](https://www.python.org/downloads/) version **3.10-3.14**.
+  - The Python version determines the supported `ansible-core` version. Consult the [ansible-core support matrix](https://docs.ansible.com/ansible/latest/reference_appendices/release_and_maintenance.html#ansible-core-support-matrix) for more information.
```
alexeygorbunov, line 12: Knowing (based on `ansible-core` support matrix) that Python version 3.X can only be used with `X < ansible-core < Y` how would one understand what version of `pyavd` and `arista.avd` should be installed to stay compliant (as `pip install "pyavd[ansible]"` and `ansible-galaxy collection install arista.avd` will install latests version of ansible-core which may not be compliant with the present Python)?

PR #6487, opened 2026-02-05
Fix(eos_cli_config_gen): Correct schema and template for sflow.interface.egress

philippebureau:
## Change Summary

Update the doc to make both nested keys (enable_default and unmodified) as required for sflow.interface.egress data model to reflect the way the template is build.  both keys are required to be defined for the template to render configuration

## Related Issue(s)

N/A

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
add `required: true` for both keys under sflow.schema.yml

## How to test
N/A

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/sflow.schema.yml
```diff
@@ -114,11 +114,13 @@ keys:
                 type: bool
                 description: |
                   Enable egress sFlow by default.
+                required: true
               unmodified:
                 type: bool
                 description: |
                   Enable egress sFlow unmodified.
                   Platform dependent feature.
+                required: true
```
gmuloc: are we sure this is not something that we need to check further and maybe fix the template? It looks like without unmodified this is accepted by some platform?

Looking at the TOI https://www.arista.com/en/support/toi/eos-4-24-2f/14579-support-for-software-egress-sflow it seems some platform needs unmodified but not the others so I dont think making unmodified required is a good idea if this is correct. However we may need to adjust the template.

ClausHolbechArista: Unmodified also says "platform dependent feature", where "enable_default" does not. So something is off here.

ClausHolbechArista: R-series platforms do not support the unmodified keyword. And X-series only support the unmodified keyword.

philippebureau: The way i read the template was that both keys were required by the user to explicitly choose which version they want based on the platform they use. 

Should we set a platform settings key to set the unmodified value and leave the template as is so users could override ?

something like this 

```yaml
platform_settings:
- digital_twin:
    platform: vEOS-lab
  feature_support:
  platforms:
  - 7050X3
  reload_delay:
    mlag: 300
    non_mlag: 330
  trident_forwarding_table_partition: flexible exact-match 16384 l2-shared 98304 l3-shared
    131072
  sflow:
    interface_egress_unmodified: True
```

and set the sflow shema fragement as follow:
```yaml
          egress:
            type: dict
            keys:
              enable_default:
                type: bool
                description: |
                  Enable egress sFlow by default.
              unmodified:
                type: bool
                description: |
                  Override the platform settings value for egress sFlow unmodified.
                  Platform dependent feature.
```

gmuloc: indeed I was just checking the template, and we are looking explicitly for the value of `unmodified` to be `false`,` and it may have been done on purpose.

(At the time we were not able to have required in the schema but so not setting it was very confusing).

the `platform_settings` are not influencing `eos_cli_config_gen` so we cannot use this sadly.

I feel it is weird to ask people to set `unmodifeid: false` when their platform does not support it. 

Suggestion:

1. we keep enabled_default as required
2. we leave `unmodified` as not required and we fix the template like this:

**EDITED with correct variable...**

```
{%     if sflow.interface.egress.enable_default is arista.avd.defined(true) %}
{%         if sflow.interface.egress.unmodified is arista.avd.defined(true) %}
sflow interface egress unmodified enable default
{%         else %}
sflow interface egress enable default
{%         endif %}
{%     endif %}
```


(happy to get counter proposition :) )

ClausHolbechArista: once this is done it would be nice to track the capability in eos_designs to output proper cli for egress sflow depending on the platform support. @alexeygorbunov

philippebureau: > indeed I was just checking the template, and we are looking explicitly for the value of `unmodified` to be `false`,` and it may have been done on purpose.
> 
> (At the time we were not able to have required in the schema but so not setting it was very confusing).
> 
> the `platform_settings` are not influencing `eos_cli_config_gen` so we cannot use this sadly.
> 
> I feel it is weird to ask people to set `unmodifeid: false` when their platform does not support it.
> 
> Suggestion:
> 
>     1. we keep enabled_default as required
> 
>     2. we leave `unmodified` as not required and we fix the template like this:
> 
> 
> ```
> {%     if sflow.interface.egress.enable_default is arista.avd.defined(true) %}
> {%         if sflow_interface_egress_unmodified is arista.avd.defined(true) %}
> sflow interface egress unmodified enable default
> {%         else %}
> sflow interface egress enable default
> {%         endif %}
> {%     endif %}
> ```
> 
> (happy to get counter proposition :) )

This works I think.  Should we set the unmodified value default to "false" so it is more explicit to users?

Like this?

```yaml
          egress:
            type: dict
            keys:
              enable_default:
                type: bool
                description: |
                  Enable egress sFlow by default.
                required: true
              unmodified:
                type: bool
                description: |
                  Enable egress sFlow unmodified.
                  Platform dependent feature.
                default: false
```

ClausHolbechArista: no defaults in eos_cli_config_gen schema

philippebureau: > no defaults in eos_cli_config_gen schema

ok, so we keep the schema untouched but change the template as per @gmuloc comment ?

gmuloc: yup, keep adding the `required` to `enable_default` and update the template

philippebureau:
should we create more molecule tests?

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/intended/configs/host1.cfg
```diff
@@ -1398,7 +1398,7 @@ sflow extension router
 no sflow extension switch
 no sflow extension tunnel ipv4 egress
 sflow interface disable default
-sflow interface egress unmodified enable default
+sflow interface egress enable default
```
gmuloc:
```suggestion
  interface:
    disable:
      default: true
    egress:
      enable_default: true
      unmodified: true
```

Hmm this should not append

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/sflow.j2
```diff
@@ -62,10 +62,12 @@ no sflow extension {{ extension.name }}
 {%     if sflow.interface.disable.default is arista.avd.defined(true) %}
 sflow interface disable default
 {%     endif %}
-{%     if sflow.interface.egress.enable_default is arista.avd.defined(true) and sflow.interface.egress.unmodified is arista.avd.defined(false) %}
-sflow interface egress enable default
-{%     elif sflow.interface.egress.enable_default is arista.avd.defined(true) and sflow.interface.egress.unmodified is arista.avd.defined(true) %}
+{%     if sflow.interface.egress.enable_default is arista.avd.defined(true) %}
+{%         if sflow_interface_egress_unmodified is arista.avd.defined(true) %}
```
gmuloc:
```suggestion
{%         if sflow.interface.egress.unmodified is arista.avd.defined(true) %}
```

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/intended/configs/host1.cfg
gmuloc:
```suggestion
sflow interface egress unmodified enable default
```

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/documentation/devices/host1.md
```diff
@@ -2996,7 +2996,7 @@ sflow extension router
 no sflow extension switch
 no sflow extension tunnel ipv4 egress
 sflow interface disable default
-sflow interface egress unmodified enable default
+sflow interface egress enable default
```
gmuloc:
```suggestion
sflow interface egress unmodified enable default
```

gmuloc (approved):

ClausHolbechArista (approved):

PR #6498, opened 2026-02-09
Refactor(pyavd): Rename ANTA-related classes and modules to follow codebase conventions

carl-baillargeon:
- Renaming `AvdXxx` classes to `AVDXxx`.
- `pyavd.api._anta` to `pyavd.api.anta`
- Same for internal `decorators` and `base_classes` modules.


MaheshGSLAB (approved):

python-avd/pyavd/api/anta/avd_fabric_data.py
```diff
@@ -14,7 +14,7 @@
 
 
 @dataclass(frozen=True)
-class AvdEthernetInterface:
+class AVDEthernetInterface:
```
gmuloc, line 17: so these classes are not exposed in the __init__ but are on a path where they are technically part of SemVer (as per our sem ver page). Is this really what we want here? (as opposed to just exposing AVDFabricData on a non `_` path)?

gmuloc (approved):
LGTM overall - just one comment open regarding the fact that renaming `api._anta` to `api.anta` is making more objects SemVer vs only importing the couple of classes imported in the new __init__.

Maybe we want to rename avd_fabric_data _avd_fabric_data so we still have free reins to do breaking changes in these dataclasses were it needed in the future (or we will just take the SemVer hit :) )

python-avd/pyavd/api/anta/avd_fabric_data.py
gmuloc: discussed on chat with @carl-baillargeon, this is on purpose.

PR #6509, opened 2026-02-10
Refactor: Fixes on AVDVaultHandler

carl-baillargeon:
Few fixes following the introduction of `AVDVaultHandler` in https://github.com/aristanetworks/avd/pull/6463:

- Capitalized Vault everywhere.
- Used "Vault secrets" instead of "Vault configured" following https://github.com/aristanetworks/avd/pull/6492#discussion_r2785055109.
- Added slots and name manging for internal attributes.
- Used Ansible Vault `decrypt` method instead of `decrypt_and_get_vault_id`.

### DIFF
--- a/ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_vault_handler.py
+++ b/ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_vault_handler.py
@@ -11,71 +11,64 @@ from ansible.parsing.vault import match_encrypt_secret
 
 if TYPE_CHECKING:
     from ansible.parsing.dataloader import DataLoader
-    from ansible.parsing.vault import VaultSecret
 
 
 class AVDVaultHandler:
     """Handles Ansible Vault encryption and decryption operations."""
 
-    _loader: DataLoader
-    _encrypt_vault_id: str | None
-    _encrypt_secret: VaultSecret | None
+    __slots__ = ("__encrypt_secret", "__encrypt_vault_id", "__loader")
 
     def __init__(self, loader: DataLoader, vault_id: str | None = None) -> None:
         """
-        Initialize the vault handler.
+        Initialize the Vault handler.
 
         Args:
             loader: The Ansible DataLoader instance.
-            vault_id: Optional vault ID to use for encryption. If None, uses the first vault ID in the list (default Ansible behavior).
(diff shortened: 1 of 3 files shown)


gmuloc (approved):

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_vault_handler.py
```diff
@@ -11,71 +11,64 @@
...
-        Initialize the vault handler.
+        Initialize the Vault handler.
 
         Args:
             loader: The Ansible DataLoader instance.
-            vault_id: Optional vault ID to use for encryption. If None, uses the first vault ID in the list (default Ansible behavior).
+            vault_id: Optional Vault ID to use for encryption. If None, uses the first Vault ID in the list (default Ansible behavior).
         """
-        self._loader = loader
+        self.__loader = loader
```
alexeygorbunov, line 29: Any specific advantage/purpose of using `__` vs `_`?

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_vault_handler.py
```diff
@@ -11,71 +11,64 @@
...
     from ansible.parsing.dataloader import DataLoader
-    from ansible.parsing.vault import VaultSecret
 
 
 class AVDVaultHandler:
     """Handles Ansible Vault encryption and decryption operations."""
 
-    _loader: DataLoader
-    _encrypt_vault_id: str | None
-    _encrypt_secret: VaultSecret | None
```
alexeygorbunov, line 22: Shouldn't we still use type annotations, even with `__slots__`?

ClausHolbechArista: Since these should not be read outside the class, (hence the __) it will not be needed. The code for the class itself will determine the type based on the __init__ code.

ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_vault_handler.py
ClausHolbechArista: __ is a private attribute which is not directly accessible by outside code. It _is_ still possible to circumvent, but not easily and by mistake. This is strongly signal that these attributes are internal to this class and should be left alone.

ClausHolbechArista (approved):

PR #6511, opened 2026-02-10
Doc: Adding excalidraw hook and export

JulioPDX:
## Change Summary

Common workflow with this implementation.

- Craft you excalidraw file within vscode or the official site.
- Save the `.excalidraw` file to the `_media` directory.
- Add your file reference to a markdown file

    ```text
    ![My  diagram](stuff.excalidraw)
    ```

- There is a Mkdocs hook that will replace the reference with the appropriate `only-light` and `only-dark` variations (on serve and build)

The implementation uses the [kroki API](https://docs.kroki.io/kroki/setup/usage/) to run the conversions. This seemed like the most direct option that required the least amount of additional requirements.

Light mode example:

image

Dark mode example:

image

## Component(s) name

Docs

## Proposed changes

Please see the change summary.

## How to test

Run `mkdocs serve` or `mkdocs build` locally or view the generated docs in the PR

### Repository Checklist


docs/hooks/excalidraw.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def replace_excalidraw(match: re.Match) -> str:  # type: ignore[type-arg]
+        alt_text = match.group(1)
+        excalidraw_path = match.group(2)
+        title = match.group(3) or alt_text
+
+        # Remove .excalidraw extension and create light/dark paths in exported/ subdirectory
+        # Get the directory and filename parts
+        if "/" in excalidraw_path:
+            dir_part = excalidraw_path.rsplit("/", 1)[0]
```
ClausHolbechArista: refactor all the path manipulation to use the pathlib.Path which contains proper methods for these things.

docs/hooks/excalidraw.py
```diff
@@ -0,0 +1,93 @@
...
+
+        # Build paths to exported/ subdirectory
+        if dir_part:
+            light_path = f"{dir_part}/exported/{base_name}-light.svg"
+            dark_path = f"{dir_part}/exported/{base_name}-dark.svg"
+        else:
+            light_path = f"exported/{base_name}-light.svg"
+            dark_path = f"exported/{base_name}-dark.svg"
+
+        # Check if the exported SVGs exist
```
ClausHolbechArista: Instead of this I think you should call the export function here so we export and build the SVGs on demand during site build instead of having to run the script first.

JulioPDX:
Need to add an exclude to pre-commit on .excalidraw files

docs/hooks/excalidraw.py
JulioPDX: Updated to use the one file approach and also updated the references to excalidraw files in the role docs/pyavd

.pre-commit-config.yaml
```diff
@@ -11,23 +11,24 @@ repos:
     hooks:
       - id: trailing-whitespace
         name: Trims trailing whitespace.
-        exclude: ansible_collections/arista/avd/extensions/molecule
+        exclude: (ansible_collections/arista/avd/extensions/molecule|\.excalidraw$)
```
ClausHolbechArista:
```suggestion
        exclude: >-
          (?x)
          (ansible_collections/arista/avd/extensions/molecule
          |\.excalidraw$
          )
```

.pre-commit-config.yaml
```diff
@@ -11,23 +11,24 @@ repos:
...
       - id: end-of-file-fixer
         name: Ensures that a file is either empty, or ends with one newline.
         exclude_types: [svg, json]
-        exclude: ansible_collections/arista/avd/extensions/molecule
+        exclude: (ansible_collections/arista/avd/extensions/molecule|\.excalidraw$)
```
ClausHolbechArista:
```suggestion
        exclude: >-
          (?x)
          (ansible_collections/arista/avd/extensions/molecule
          |\.excalidraw$
          )
```

.pre-commit-config.yaml
```diff
@@ -11,23 +11,24 @@ repos:
...
           (ansible_collections/arista/avd/extensions/molecule
           |pickle$
+          |\.excalidraw$
           |python-avd/pyavd/_(eos_cli_config_gen|eos_designs)/schema/__init__.py
           |python-avd/pyavd/_eos_designs/eos_designs_facts/schema/protocol.py
           )
       - id: check-merge-conflict
         name: Checks for files that contain merge conflict strings.
-        exclude: ansible_collections/arista/avd/extensions/molecule
+        exclude: (ansible_collections/arista/avd/extensions/molecule|\.excalidraw$)
```
ClausHolbechArista:
```suggestion
        exclude: >-
          (?x)
          (ansible_collections/arista/avd/extensions/molecule
          |\.excalidraw$
          )
```

docs/_media/excalidraw/exported/anta_runner_dark.svg
ClausHolbechArista, line 1: I think these files should be removed and add the `/docs/_media/excalidraw/exported/` path to .gitignore in the root of the repo. These are generated so we don't have to check them in. We can build the site locally or online to test things.

Remember the process to remove them is:
- Delete the files.
- !!! **Commit** !!! (this is the important step in case you were in doubt :)
- Edit .gitignore to add the path under the generated files section
- Commit

ClausHolbechArista (commented):
Perfect, this is exactly what I had hoped for. Thank you!

One small comment on the svg files which we can keep out of git.

docs/_media/excalidraw/exported/anta_runner_dark.svg
JulioPDX: Removed the files in exported and added path to gitignore :)

JulioPDX:
> Perfect, this is exactly what I had hoped for. Thank you!
> 
> One small comment on the svg files which we can keep out of git.

When this works it is glorious. I have now seen instances where the kroki api is either too bust or an internal service is recovering leads to failed builds. We can either keep going down this route or add something to the compose file to stand up the container during the build steps (dont think that would work from a readthedocs perspective).

ClausHolbechArista:
ah I was not aware that the library was using some cloud service. @JulioPDX please find a different library that works purely offline to convert excalidraw to svg in the same way.

JulioPDX:
@ClausHolbechArista Yeah that was a bummer. It started failing. We could host our own but then it would either break or complicate the builds for readthedocs. I'll investigate a cleaner option.

ankudinov (changes requested):
@JulioPDX I'm not confident it's a good idea to rely on an external service for conversion. In my opinion we have enough external APIs to troubleshoot: Galaxy, GitHub, staging, etc. =) However the idea of auto-convert in general sound useful. Can we simply make a local Makefile shortcut for it?

PR #6514, opened 2026-02-11
Refactor(eos_cli_config_gen): Deprecate old model access_lists

### LINKED ISSUE #6326: Refactor(eos_cli_config_gen): Deprecate old model `access_lists`
`ip_access_lists` is the proper model to use going forward.

Shivani-gslab:
## Related Issue(s)

Fixes #6326 

## Component(s) name

## Proposed changes
Deprecated the data-model and moved the tests

## How to test
eos_cli_config_gen_deprecated_vars molecule scenario

### User Checklist

- N/A

### Repository Checklist


Vibhu-gslab (approved):
Wait for 6.0 to open.

MaheshGSLAB (approved):

gmuloc (commented):
@Shivani-gslab thanks for this - we will need to put this PR on hold until we are able to render both old and new model still in AVD 6.x train.
This imply a change to pyavd-utils and metaschema to add an option under deprecation to say "allow_with_new_key" or whatever we call it in order to make it non breaking.

As it is now this could break deployments where users are using both `access_lists` and `ip_access_lists` in parallel

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/access-lists.j2
```diff
@@ -4,16 +4,18 @@
  that can be found in the LICENSE file.
 #}
 {# eos - access lists #}
-{% for access_list in access_lists | arista.avd.natural_sort(sort_key='name', ignore_case=false) %}
+{% if access_lists is arista.avd.defined and ip_access_lists is not arista.avd.defined %}
```
gmuloc: this will need to go for it to be non breaking once we have implemented the required changes

ClausHolbechArista: we do not need to enforce this in the templates. The schema tooling would already handle the logic about only one of them being there.

Shivani-gslab:
Dependent on PR https://github.com/aristanetworks/avd/pull/6596

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/access-lists.j2
Shivani-gslab: Updtaed

Shivani-gslab:
Since the PR #6596 got merged, I have updated this PR with `allow_with_new_key`

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/access-lists.j2
```diff
@@ -4,16 +4,18 @@
  that can be found in the LICENSE file.
 #}
 {# eos - access lists #}
-{% for access_list in access_lists | arista.avd.natural_sort(sort_key='name', ignore_case=false) %}
+{% if access_lists is arista.avd.defined %}
```
gmuloc: I don't think you need to add the if statement right?

gmuloc (changes requested):

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/access-lists.j2
alexeygorbunov: +1

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/access-lists.j2
alexeygorbunov, line 1: General comment:

Right now both `access_lists` and `ip_access_lists` can be used in parallel and although documentation states that `Both data models can coexists without conflicts, as different keys are used: access_lists vs ip_access_lists. Access list names must be unique.` - there is nothing that enforces the ACL names to be unique in the union of ACLs from both data models.

We are OK with this (and possible rendering of 2 ACLs with the same name but different content, produced by these 2 data models)?

note_to_self: everything is OK once https://github.com/aristanetworks/avd/pull/6514#discussion_r2953285010 is resolved

gmuloc: we are ok with this as this is the current state of affairs

gmuloc (approved):

PR #6542, opened 2026-02-13
Doc(cv_deploy): Updating role diagram

JulioPDX:
## Change Summary

Updating diagrams for rol cv_deploy. Aligns with the latest update to our role diagrams. Only exception is the CV logo, it looks great and cant find a suitable replacement.

## Component(s) name

Docs

## How to test

Serve locally or view the generated docs.

### Repository Checklist


docs/_media/cv_deploy_dark.svg
alexeygorbunov, line 1: Few notes:
1. `cv_deploy`'s static manifests feature allows to reference arbitrary text-based files (not .cfg files generated by eos_cli_config_gen) and set their content as CVStatic Studio configlets. Old and proposed diagram has `EOS CLI Configuration` (referencing .cfg files) input but this term is used by other plugins' diagrams and i believe is understood as files generated by AVD (eos_cli_config_gen role).
Should we have another input feed for cv_deploy for arbitrary text-based EOS CLI config files that are consumed by static manifest logic?

alexeygorbunov: 2. While most of the `cv_deploy` logic uses gRPC for interacting with CV, we do require REST calls to succeed first to fetch some CV metadata. Wondering if we should have REST & gRPC shown on the diagram

alexeygorbunov: 3. In most implementation right now - CV is not directly hitting EOS device's eAPI (the way you would make an eAPI call to EOS device).. It does rely on eAPI but via TA tunnel.
I'm just wondering if having eAPI between CV and EOS fabric would make a wrong impression that `cv_deploy` forces CV to use plain eAPI for interacting with remote devices..

JulioPDX: 1. Could we have files without extensions since that is in-memory vs literally loading the .cfg file to CV?
2. Yeah I like this, `REST & gRPC` it is

gmuloc: For 1. let's not put both on the same diagram (regular mode and manifest mode) as this would confuse the picture for users. We could add a diagram for the manifest mode later on but let's not make this blocking for 6.0.

gmuloc: For 3., given that eAPI is still correct let's not supercharge the diagram.

carl-baillargeon (approved):

docs/_media/cv_deploy_dark.svg
alexeygorbunov: Sounds good!

alexeygorbunov, line 1: Last thing - `REST & gRPC` line should probably be bidirectional as we push and pull at the same time

alexeygorbunov (approved):
LGTM!

PR #6550, opened 2026-02-16
Fix(eos_cli_config_gen): Correct schema and template for sflow.interface.egress

Shivani-gslab:
Cherry-pick of PR #6487

### DIFF
--- a/python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/sflow.schema.yml
+++ b/python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/sflow.schema.yml
@@ -114,7 +114,6 @@ keys:
                 type: bool
                 description: |
                   Enable egress sFlow by default.
-                required: true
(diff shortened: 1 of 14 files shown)


gmuloc (approved):

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/sflow.schema.yml
```diff
@@ -114,6 +114,7 @@ keys:
                 type: bool
                 description: |
                   Enable egress sFlow by default.
+                required: true
```
ClausHolbechArista: let's not modify the schema here. It is enough to fix the template.

ClausHolbechArista:
```suggestion
```

ClausHolbechArista (approved):

PR #6561, opened 2026-02-17
Test: Sort the hosts by name in all molecule

### LINKED ISSUE #6303: Sort the hosts by name in eos_designs_unit_tests molecule
not for this PR but lets make a short PR where we sort the hosts by name in there for us to keep sane

_Originally posted by @gmuloc in https://github.com/aristanetworks/avd/pull/6258#discussion_r2655370158_

MaheshGSLAB:
## Change Summary

Sort out the host name for all molecule

## Related Issue(s)

Fixes #6303 

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
Sort out the host name for all molecule

## How to test
CI should be green
### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/hosts.yml
```diff
@@ -5,103 +5,101 @@ all:
...
-            exclude-as-extra-fabric-validation-target:
             validation-profiles:
+            varpv6:
+            vrfs-rt-import-and-export:
+            vrfs_rd_rt_override:
         DEVICES_TESTS:
           hosts:
+            devices-l2leaf1:
             devices-leaf1a:
             devices-leaf1b:
```
MaheshGSLAB:
```suggestion
            devices-leaf1b:
            # The next two devices are configured with
            # use_cv_topology: true in host_vars
```

ansible_collections/arista/avd/extensions/molecule/ansible_only/inventory/hosts.yml
```diff
@@ -3,9 +3,9 @@ all:
   children:
     ANSIBLE_ONLY:
       hosts:
-        string_set_as_play_var:
-        RD-RT-ADMIN-SUBFIELD-L3LEAF:
         node-type-l3-interfaces-portchannels-custom-template:
+        RD-RT-ADMIN-SUBFIELD-L3LEAF:
```
ClausHolbechArista, line 7: could we change this to lower case in a PR merging before this one?

ansible_collections/arista/avd/extensions/molecule/ansible_only/inventory/hosts.yml
```diff
@@ -3,9 +3,9 @@ all:
...
+        string_set_as_play_var:
```
ClausHolbechArista, line 8: also changing this to dashes instead of underscores.

ansible_collections/arista/avd/extensions/molecule/eos_designs-twodc-5stage-clos/inventory/hosts.yml
```diff
@@ -34,18 +34,6 @@ all:
                       ansible_port: 8006
                 DC1_POD1_LEAFS:
```
ClausHolbechArista, line 35: I don't think there is a big value in changing the order outside the unit tests (including negative). In the others they are usually kept in an order of "importance" or similar.

MaheshGSLAB: revert the changes for eos_designs-twodc-5stage-clos molecule

ansible_collections/arista/avd/extensions/molecule/ansible_only/inventory/hosts.yml
MaheshGSLAB: handling in PR https://github.com/aristanetworks/avd/pull/6566

ansible_collections/arista/avd/extensions/molecule/ansible_only/inventory/hosts.yml
MaheshGSLAB: Raised a issue https://github.com/aristanetworks/avd/issues/6568 to handle in different PR

ClausHolbechArista:
@MaheshGSLAB you could also replace all this with a single pytest using our molecule decorators, where it will be passed every device from all inventories and then you can just assert on the name

MaheshGSLAB:
> @MaheshGSLAB you could also replace all this with a single pytest using our molecule decorators, where it will be passed every device from all inventories and then you can just assert on the name

As discussed will keep sort the host name in pre-commit and validating hostname will move to pytest

Vibhu-gslab (approved):

gmuloc (dismissed):

development/sort_hosts_yml.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python3
+
+# Copyright (c) 2025-2026 Arista Networks, Inc.
```
gmuloc:
```suggestion
# Copyright (c) 2026 Arista Networks, Inc.
```

development/sort_hosts_yml.py
```diff
@@ -0,0 +1,213 @@
...
+    return int(text) if text.isdigit() else text.lower()
+
+
+def _alphanum_key(item: dict[str, str], sort_key: str) -> list[int | str]:
+    """Generate natural sort key for a dictionary item."""
+    val = item.get(sort_key, "")
+    return [_convert(part) for part in SPLIT_PATTERN.split(str(val))]
+
+
+def natural_sort(iterable: list[dict[str, str]], sort_key: str) -> list[dict[str, str]]:
```
gmuloc: any reason not to use pyavd here?

MaheshGSLAB: since pre-commit hook runs in isolated environment hence we will see import error for pyavd. It can work if we add `additional_dependencies: [pyavd]`. So to avoids adding pyavd as a dependency to the pre-commit hook config did not used it.

MaheshGSLAB: added the natural sort from pyavd with override the global path.

development/sort_hosts_yml.py
```diff
@@ -0,0 +1,200 @@
...
+                hosts_block.append({"name": host_name, "lines": host_lines})
+            else:
+                # End of hosts section
+                break
+
+        return hosts_block, self.index
+
+    def _is_comment_or_blank(self, line: str) -> bool:
+        """Check if line is a comment or blank at host indent level."""
+        return line.strip() == "" or (line.strip().startswith("#") and line.startswith(self.host_indent))
```
Shivani-gslab: this does not work if the comment is not indented to the host indent level 
for example -

```
       # shivani
            address-locking-4:
            address-locking:
            address-locking-2:
            address-locking-3:
```

We can may be relax it, to handle edge cases -

```
def _is_comment_or_blank(self, line: str) -> bool:
        """Check if line is a comment or blank at host indent level."""
        stripped_line = line.strip()
        return not stripped_line or stripped_line.startswith("#")
```

Shivani-gslab (changes requested):

development/sort_hosts_yml.py
```diff
@@ -0,0 +1,200 @@
...
+    lines = content.splitlines(keepends=True)
+    result: list[str] = []
+    i = 0
+
+    while i < len(lines):
+        line = lines[i]
+        result.append(line)
+
+        # Check for "hosts:" line
+        if hosts_match := re.match(r"^(\s*)hosts:\s*(?:#.*)?$", line):
```
Shivani-gslab: currently we are sorting hosts only, we could also consider sorting group names under `children:` block for better readability.
@ClausHolbechArista @gmuloc should we do that?

development/sort_hosts_yml.py
Shivani-gslab:
```suggestion
        stripped_line = line.strip()
        return not stripped_line or stripped_line.startswith("#")
```

development/sort_hosts_yml.py
gmuloc: let's do it :)

MaheshGSLAB: updated

development/sort_hosts_yml.py
MaheshGSLAB: The case when comment indentation is not correct, we are removing the comment. Like if there is comment above host and indentation is at their parent group level then in that case comment will remove after the pre-commit.

Shivani-gslab: Example test (the comment's  indentation is less than the `hosts` key indentation) -
In this case it is removing the comment, it works till `hosts` indentation .
```
CLEAN_UNIT_TESTS: # Single Devices testing one specific case and only using hostvars
          hosts:
       # comment
            aaa-settings-3:
            aaa-settings-2:
            aaa-settings-1:
```

gmuloc: @Shivani-gslab this kind of comment is not really as per YAML specification so let's not focus on it too much. Let's see if we hit something annoying later

Shivani-gslab (approved):
LGTM

gmuloc (approved):

PR #6573, opened 2026-02-19
Fix(eos_designs): Avoid returning PosixPath type to Ansible 

### LINKED ISSUE #6522: Bug(cv_deploy): Error with posix path when using `cv_register_detailed_results: true`
### Issue Summary

Likely caused by Ansible 2.19+ being more strict about the return values.

```sh
[WARNING]: Callback dispatch 'v2_runner_on_ok' failed for plugin 'default': Type 'PosixPath' is unsupported for variable storage.

Callback dispatch 'v2_runner_on_ok' failed for plugin 'default'.

<<< caused by >>>

Type 'PosixPath' is unsupported for variable storage.
Origin: <unknown>

/workspaces/avd-dev-cvaas/static_configlets/dc-mgmt1.cfg
```

cv_deploy

### AVD version

6.0.0-devX

### Ansible version

2.20

### Python version

_No response_

### How do you run AVD ?

### Steps to reproduce

```shell

```

### Relevant log output

```shell

```

### Contributing Guide

alexeygorbunov:
## Change Summary

Adjust (from `Path` to `str`) type of the `AvdConfiglet`'s attribute `file`.

## Related Issue(s)

Fixes #6522 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Change type of the `AvdConfiglet.file` to `str` so that it can be serialized and kept in Ansible variable storage.

## How to test
Pytest

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/tests/unit/action/test_cv_workflow.py
```diff
@@ -0,0 +1,116 @@
...
+
+from pyavd._cv.workflows.models import AvdConfiglet, AvdManifest
+
+
+@dataclass(frozen=True)
+class MockedAvdConfiglet:
+    """Mocked AvdConfiglet with Path type file attribute."""
+
+    name: str
+    file: Path
```
gmuloc: a bit confused. Why are we mocking the AVdConfiglet with Path when we just changed it to str in the actual code?

ansible_collections/arista/avd/tests/unit/action/test_cv_workflow.py
```diff
@@ -0,0 +1,116 @@
...
+
+        callback_result = CallbackTaskResult(host=host, task=task, return_data=result, task_fields={})
+        transformed = callback_result.result
+
+        assert transformed["name"] == "TestConfiglet"
+        assert transformed["file"] == "/path/to/file.cfg"
+        assert isinstance(transformed["name"], str)
+        assert isinstance(transformed["file"], str)
+
+    def test_avd_configlet_serialization_failure(self) -> None:
```
gmuloc: I am not sure I understand the value of keeping this test with a mocked pbject that we don't use. We should remove.

ansible_collections/arista/avd/tests/unit/action/test_cv_workflow.py
```diff
@@ -0,0 +1,116 @@
...
+        task._uuid = "test-uuid"
+
+        callback_result = CallbackTaskResult(host=host, task=task, return_data=result, task_fields={})
+
+        transformed = callback_result.result
+
+        assert len(transformed["static_config_manifest"]["configlets"]) == 2
+        assert all(isinstance(configlet["file"], str) for configlet in transformed["static_config_manifest"]["configlets"])
+
+    def test_avd_manifest_serialization_failure(self) -> None:
```
gmuloc: same this test is not useful in my opinion

gmuloc (changes requested):

ansible_collections/arista/avd/tests/unit/action/test_cv_workflow.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@dataclass(frozen=True)
+class MockedAvdManifest:
+    """Mocked AvdManifest with MockedAvdConfiglets."""
+
+    configlets: tuple[MockedAvdConfiglet, ...]
+
+
+class TestCvWorkflowReturnValueSerialization:
```
gmuloc: Only the positive tests are useful here because:

1/ they test that we expect to be able to pass object to the callback without failure.
2/ if we ever see them fail we know we have broekn something (e.g. we change from_dict to convert the string to a Path in two years and 💥 )

The negative tests using Mocked objects not reflecting the reality are here only to show that the callback fails but they are of no value to us, except maybe one day in a distant future realizing that we could now use Path but thats not really something worth spending cycle testing (and it is unlikely) so essentially I see them as dead code testing

ClausHolbechArista: maybe all this could be replaced with a proper ansible driven test by enabling a callback plugin in the molecule scenario testing cv_deploy with detailed returns

python-avd/tests/pyavd/cv/workflows/test_models.py
```diff
@@ -318,6 +319,16 @@ def test_missing_keys_failure(self, invalid_data: dict[str, Any], match_str: str
...
+    def test_asdict_returns_native_types_for_ansible_compatibility(self) -> None:
+        """Tests that asdict(<AvdConfiglet>) only returns Python native types to ensure compatibility with Ansible's transform_to_native_types()."""
+        data = {"name": "TestConfiglet", "file": "/path/to/file.cfg"}
+        configlet = AvdConfiglet.from_dict(data)
+        result = asdict(configlet)
+
+        # Confirm that all values are native Python types (not Path)
+        assert isinstance(result["name"], str)
+        assert isinstance(result["file"], str)
+
```
ClausHolbechArista:
```suggestion
```

python-avd/tests/pyavd/cv/workflows/test_models.py
```diff
@@ -2,6 +2,7 @@
 # Use of this source code is governed by the Apache License 2.0
 # that can be found in the LICENSE file.
 
+from dataclasses import asdict
```
ClausHolbechArista:
```suggestion
```

ClausHolbechArista (approved):
Removed the tests for now. Let's discuss further which tests, if any, would be relevant here.

ansible_collections/arista/avd/tests/unit/action/test_cv_workflow.py
ClausHolbechArista: I wonder if I current tests would catch this a new verison of Ansible. We don't run those in CI at the moment.

gmuloc (approved):

PR #6603, opened 2026-02-25
Fix(eos_designs): Centralize injection of host-specific info for AristaAvdError and child classes

alexeygorbunov:
## Change Summary

When host is specified in the inventory but it's group/host vars are missing `type` - AVD raises AristaAvdInvalidInputsError pointing to the host used by `run_once` task and not providing any info regarding the host that actually misses type definition: `No device type found. Either set 'type' or 'default_node_types'.`

In some other cases - impacted host is specified twice or incorrect host is specified instead of the one triggering the issue

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.avd_designs`

## Proposed changes
- Add `host_context_set: bool` flag to AristaAvdError and use it as a trigger hor appending host-related suffix to the messages
- Modify some message patterns

## How to test
Pytest: eos_designs_negative_unit_tests

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/shared_utils/node_type.py
```diff
@@ -32,7 +32,7 @@ def type(self: SharedUtilsProtocol) -> str:
         if self.default_node_type:
             return self.default_node_type
 
-        msg = "No device type found. Either set 'type' or 'default_node_types'."
+        msg = f"No device type found for host '{self.hostname}'. Either set 'type' or 'default_node_types'."
         raise AristaAvdInvalidInputsError(msg, host=self.hostname)
```
gmuloc, line 36: this showing `self.hostname` should already have appended the hostname. I am not sure the current fix is the right one. Let me take a further look

MaheshGSLAB: We also seen such thing that when something failed on eos_designs_facts it do not show that which host throw that error. I am saying about eos designs molecule

ClausHolbechArista: we may have missed something in the error handling during the refactoring. ~Although the expected error in the negative unit test suggests that this has been an issue for a while.~

ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/inventory/host_vars/missing-mlag-peer.yml
```diff
@@ -15,4 +15,4 @@ l2leaf:
 expected_error_message: >-
   Facts not found for node 'some-missing-device'. Something in the input vars is pointing to this node.
   Check that 'some-missing-device' is in the inventory and is part of the group set by 'fabric_name'.
-  Node is required for host 'missing-mlag-peer'.
+  Node is required (Problematic host: 'some-missing-device').
```
ClausHolbechArista: please don't change this.

python-avd/pyavd/_eos_designs/eos_designs_facts/utils.py
```diff
@@ -27,7 +27,7 @@ def get_peer_facts_generator(self: EosDesignsFactsGeneratorProtocol, peer_name:
                 f"Facts not found for node '{peer_name}'. Something in the input vars is pointing to this node. "
                 f"Check that '{peer_name}' is in the inventory and is part of the group set by 'fabric_name'. Node is required."
             )
-            raise AristaAvdInvalidInputsError(msg)
+            raise AristaAvdInvalidInputsError(msg, host=peer_name)
```
gmuloc: the problem is not on peer_name here it is on the local host referring to it

gmuloc: Maybe 
```suggestion
            raise AristaAvdInvalidInputsError(msg, host=self.hostname)
```

python-avd/pyavd/_errors/__init__.py
```diff
@@ -16,18 +16,24 @@
...
-    def __init__(self, message: str = "An Error has occurred in an arista.avd plugin", host: str | None = None) -> None:
+    def __init__(self, message: str = "An Error has occurred in an arista.avd plugin", host: str | None = None, host_context_set: bool = False) -> None:
         self.host = host
-        self.message = message
+        if host is not None and not host_context_set:
+            self.message = f"{message.removesuffix('.')} (Problematic host: '{host}')."
+            self.host_context_set = True
+        else:
+            self.host_context_set = host_context_set
+            self.message = message
```
gmuloc:
```suggestion
class AristaAvdHostError(AristaAvdError):
    host: str
    def __init__(self, message: str, host: str ) -> None:
      self.host = host ...
      self.message = f"{message} for host {host}."
      super().__init__(self.message, host = self.host)
```
```diff      
- class AristaAvdInvalidInputsError(AristaAvdError):
-    ...
+ class AristaAvdInvalidInputsError(AristaAvdHostError):
+    pass
```

python-avd/pyavd/_eos_designs/eos_designs_facts/get_facts.py
```diff
@@ -81,8 +81,7 @@ def get_facts(
             raise AristaAvdMissingVariableError(variable=e.variable, host=e.host or hostname) from e
         except AristaAvdError as e:
```
gmuloc:
```suggestion
        except AristaAvdHostError:
          raise
        except AristaAvdError as e:
          # keep what's there
```

gmuloc: maybe you can remove all the try / except here and let's see the changes in the error messages

python-avd/pyavd/_errors/__init__.py
alexeygorbunov: Implemented as suggested, except re-defining `host` for AristaAvdHostError as pyright doesn't like it

python-avd/pyavd/_eos_designs/eos_designs_facts/get_facts.py
alexeygorbunov: Removed try/except

python-avd/pyavd/_errors/__init__.py
```diff
@@ -23,10 +23,15 @@ def __init__(self, message: str = "An Error has occurred in an arista.avd plugin
...
-    host: str | None
+class AristaAvdHostError(AristaAvdError):
+    def __init__(self, message: str, host: str = "") -> None:
+        self.host = host
+        self.message = f"{message.removesuffix('.')} for host '{host}'." if host.strip() else message
+        super().__init__(self.message, host=host)
+
 
-    def __init__(self, message: str, host: str | None = None) -> None:
+class AristaAvdInvalidInputsError(AristaAvdHostError):
```
gmuloc, line 32: I think the AristaAvdMissingVariableError should also inherit from the AristaAvdHostError

python-avd/pyavd/_errors/__init__.py
```diff
@@ -23,10 +23,15 @@ def __init__(self, message: str = "An Error has occurred in an arista.avd plugin
         super().__init__(self.message)
 
 
-class AristaAvdInvalidInputsError(AristaAvdError):
-    host: str | None
+class AristaAvdHostError(AristaAvdError):
+    def __init__(self, message: str, host: str = "") -> None:
```
gmuloc: why do you need a default here? (maybe thats required but just checking :) )

alexeygorbunov: This is for the error message cases where "impacted" host is already mentioned in the body of the message and we don't want to prepend it again to the message. I could either pass `host=""` explicitly on the "calling" sides or just have `host=""` as a default in error class definition.
But another option is to review all places where we are raising those exceptions and to make sure they all use the same "pattern" (like they shouldn't be mentioning problematic host in the body of the message is we are supposed to prepand it later anyway)

gmuloc: yup makes sense so a possible way forward is that we do not update the message and in facts and wherever we need to add the hostname (bvecause we are running with run_once and we cannot know) we do it using the e.host to render the error message like we want.

Right now having it everywhere is actually (as we can see in the changes) going to render stuff like this:

```
fatal: [cv-pathfinder-edge1 -> localhost]: FAILED! => {"changed": false, "msg": "'ipv4_acl_in' must be set on WAN interfaces where 'wan_carrier' is set, unless the carrier is configured as 'trusted' under 'wan_carriers'. 'ipv4_acl_in' is missing on L3 interface 'Ethernet1/49.3' for host cv-pathfinder-edge1 ."}
```

for every failure in structured config which is a bit overkill today. What do you think? (I know this is going back and forth a bit :) )

MaheshGSLAB (approved):

ClausHolbechArista:
please address comments and resolve conflicts. Moving to draft for now, so please set it ready for review when done.

PR #6604, opened 2026-02-25
Feat(eos_designs): BGP Peer Tag-based route filtering for evpn_prevent_readvertise_to_server

### LINKED ISSUE #5490: Update "evpn_prevent_readvertise_to_server" knob to use bgp peer tagging filtering
### Enhancement summary

Currently we have this knob "evpn_prevent_readvertise_to_server" that will filter out advertisements that contain the Spine AS in the AS path:

In the leaves the following snippet is generated using a route-map:
```
route-map RM-EVPN-FILTER-<spine_as> deny 10
   match as <spine_as>
!
route-map RM-EVPN-FILTER-<spine_as> permit 20


router bgp <asn_number>
  neighbor <lo_spine1> route-map RM-EVPN-FILTER-<spine_as> out
  neighbor <lo_spine2> route-map RM-EVPN-FILTER-<spine_as> out
```
There is a new feature in EOS for the same purpose: 

"Peer Tagging Route Filtering feature discards BGP route advertisements by the peers which the routes are received from. The feature lets users assign a peer-tag to a peer or a group of peers in inbound direction and discard routes advertisements by the peer-tag in outbound direction. One use case of the feature is to discard AS loop routes in outbound direction in data center deployments."

https://www.arista.com/en/support/toi/eos-4-34-0f/21367-bgp-peer-tagging-route-filtering

The following configuration on Leaf1 can prevent Leaf1 from advertising routes received from any spine back to other spines.
```
(config)#router bgp 100
(config-router-bgp)#neighbor SPINES remote-as 200
(config-router-bgp)#neighbor SPINES peer-tag in spine-routes
(config-router-bgp)#neighbor SPINES peer-tag out discard spine-routes
```

So we could use this feature to optimize the re-advertisement of routes to the Spines, using the same existing knob "evpn_prevent_readvertise_to_server".

eos_designs

### Use case example

EOS by default advertises routes to the same AS that it was advertised from. This causes leaves to advertise routes learnt from one Spine to other Spines, causing unnecessary processing.

In high scale deployments with a large number of routes, it is recommended to filter these routes from being re-advertised to the spines. 

Update the knob to use bgp peer tagging

_No response_

### Additional context

_No response_

### Contributing Guide

alexeygorbunov:
## Change Summary

Add support for BGP Peer Tagging

## Related Issue(s)

Fixes #5490

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
 - Add support for outbound BGP route filtering based on EOS BGP Peer Tag (not a PA / not advertised to peers)

*Feature-related:*
 - peer or peer-group can only have one `in` tag applied. Attempt to associate another tag overrides first tag
 - peer or peer-group can only have one `out discard` tag applied. Attempt to associate another discard tag overrides first tag
 - applying tag at the global BGP level to the peer (or peer-group) applies it to all address-families
 - peer-level tag overrides peer-group-level tag
 - same IN or OUT tag can be applied to multiple peers/groups
 - Route-leaking (advertising back to Spines although PATH contains spines' ASNs) is still possible.
     Example: Leaf1 (MLAG pair with Leaf2) learns Type 5 eBGP EVPN from Spines -> Leaf1 labels it with X tag -> imports it into VRF BGP Table -> advertises it over per-VRF iBGP peering towards Leaf2 (at this point this route has no associated Tag) -> Leaf2 installs it into VRF BGP Table and VRF routing table (no tags associated) keeping original AS PATH (containing AS of originating POD, Spines, etc) -> Leaf2 redistributes it back into EVPN and advertises back to Spines (although it's PATH contains Spines' AS)

## How to test
`eos_designs_unit_tests` (`evpn-prevent-readvertise-to-server-7`)

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista:
I don't think we should implement this, as it is too error prone

alexeygorbunov:
It just has its specific use case (standalone leafs) in which it would perfectly work with minimal configuration. Maybe we can just warn users if they select this filtering mode on MLAG devices?

alexeygorbunov:
Detailed tests results and use case where remote EVPN routes leak from one MLAG peer to another over per-VRF iBGP and are sent back to Spines: https://github.com/aristanetworks/avd-internal/issues/360

alexeygorbunov:
Closing this PR due to the limitations discovered with Tag-based route filtering approach (route leaking over per-VRF iBGP between MLAG peers).

P.S. Details of the internal tests are available here: https://github.com/aristanetworks/avd-internal/issues/360#issuecomment-3961008234

PR #6616, opened 2026-02-26
Feat(cv_deploy): Protect CVDeploy attributes from invalid values

alexeygorbunov:
## Change Summary

Enforce validation of the values set for `hostname`, `serial_number` and `system_mac_address` to make sure we can not update TOPOLOGY inputs with incorrect values.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
Use explisit __setattr__ method with validation layer for `hostname`, `serial_number` and `system_mac_address` attributes to:
1. Protect from initial incorrect inputs (like `serial_number` set in group/host vars to an empty string or smth.)
2. Protect from propagating incorrect input value into the TOPOLOGY studio intups

## How to test
Pytest `python-avd/tests/pyavd/cv/workflows/test_models.py::TestCVDevice`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_cv/workflows/models.py
+++ b/python-avd/pyavd/_cv/workflows/models.py
@@ -12,6 +12,7 @@ from uuid import NAMESPACE_DNS, uuid4, uuid5
 from pyavd._cv.client.configlet import ASSIGNMENT_MATCH_POLICY_MAP
 from pyavd._cv.client.exceptions import CVManifestError
 from pyavd._cv.client.models import CVTag, CVTagAssignment
+from pyavd._errors import AristaAvdInvalidInputsError
 
 AVD_NAMESPACE = uuid5(NAMESPACE_DNS, "avd.arista.com")
 AVD_ENTITY_PREFIX = "avd_"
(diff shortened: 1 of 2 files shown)


python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -211,6 +215,26 @@ class CVDevice:
     Do not set this manually.
     """
 
+    def __setattr__(self, name: str, value: Any) -> None:
```
ClausHolbechArista, line 218: I don't think this is the right solution for this. We have a plan for enhancing cv_deploy including getting a schema written. A better solution is to just update the schema for serial_number and system_mac_address fields in metadata for EOS Config and in AVD Design to require min length 1.

ClausHolbechArista: That said you could implement this with dataclass field(). It is meant for this.

alexeygorbunov: OK, sure! But just updating metadata fields may not protect from incorrect overriding by values fetched from CV (if they ever contain empty string for any reason)

PR #6594, opened 2026-02-24
Refactor(eos_designs): Centralize parent interface generation for subinterfaces

gmuloc:
This PR introduces a centralized module for automatically creating parent interfaces when subinterfaces (interfaces with dots like Ethernet1.100) are configured.
## Related Issue(s)

Fixes #6648 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

* Created `AvdStructuredConfigParentInterfaces` module in `python-avd/pyavd/_eos_designs/structured_config/parent_interfaces/`
  * Implements global tracking of subinterfaces and parent interfaces across all structured config modules
  * Automatically creates missing parent Ethernet and Port-Channel interfaces with minimal configuration
  * Parent interfaces are tagged with `peer_type: subinterface_parent` metadata
* Introduced `ParentInterfacesTracker` class to maintain global state throughout the pipeline
  * Tracker is instantiated once and passed to all structured config classes via constructor
  * Modules register subinterfaces and parent interfaces as they create them
  * Centralized module creates only the missing parent interfaces at the end
* Modified `StructuredConfigGenerator` base class to accept `parent_interfaces_tracker` parameter
* Updated all modules that create ethernet_interfaces and port_channel_interfaces:
  - `network_services`
  - `underlay`
  - `connected_endpoints`
  - `core_interfaces_and_l3_edge`

* Positioned `AvdStructuredConfigParentInterfaces` before `AvdStructuredConfigFlows` to ensure parent interfaces are created before flow tracking metadata is applied

* Added user-facing documentation in schema to describe the auto-creation behavior
  - `defs_node_type_l3_interfaces.schema.yml`
  - `defs_node_type_l3_port_channels.schema.yml`
  - `defs_network_services.schema.yml`
  - `defs_adapter_config.schema.yml`

## How to test

* No changes in generated config across molecule scenarios
* changes in structured config (ordering / change of metadata)

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the [**CONTRIBUTING**](https://avd.arista.com/stable/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and documentation have been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/extensions/molecule) testing accordingly. (check the box if not applicable)

PR #6633, opened 2026-02-27
Refactor(eos_designs): Relax the requirement for ip_address for main interface port-channels for nodes.l3_port_channels and nodes.l3_interfaces

gmuloc:
## Change Summary

cf title

## Related Issue(s)

Preparation for #6594  - this will allwo to have a parent port-channel or ethernet under l3_port_channels / l3_interfaces in underlay and have the sub interfaces in network service.

The goal of this separate issue is to make the process clear.

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Do not raise if main port-channel is defined without ip_address

## How to test

molecule

### Repository Checklist


laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -156,18 +156,19 @@ def _get_l3_common_interface_cfg(
...
-        is_parent_l3_port_channel = schema_key == "l3_port_channels" and l3_generic_interface.name in self._l3_port_channels_with_subinterfaces
+        # logic below is common to l3_interface and l3_port_channel interface types
+        # Check if the interface is a L3 Port-Channel but not a subinterface.
+        is_l3_port_channel_not_subinterface = schema_key == "l3_port_channels" and not is_subinterface
 
         # TODO: catch if ip_address is not valid or not dhcp
-        if not l3_generic_interface.ip_address and not is_parent_l3_port_channel:
+        # IP address is required for L3 interfaces and L3 Port-Channel subinterfaces,
+        # but optional for L3 Port-Channels which are not subinterfaces as we may define subinterfaces in other modules.
+        if not l3_generic_interface.ip_address and not is_l3_port_channel_not_subinterface:
```
carl-baillargeon: Instead of the double negative, we should express when IP is required to make it easier to understand and follow your comment.

Something like:
```
is_subinterface = "." in l3_generic_interface.name
is_l3_interface = schema_key == "l3_interfaces"
is_l3_port_channel_subinterface = schema_key == "l3_port_channels" and is_subinterface

# IP address is required for L3 interfaces and L3 Port-Channel subinterfaces,
# but optional for main L3 Port-Channels as subinterfaces may be defined in other modules.
ip_address_required = is_l3_interface or is_l3_port_channel_subinterface

if not l3_generic_interface.ip_address and ip_address_required:
```

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -156,18 +156,19 @@ def _get_l3_common_interface_cfg(
             )
             schema_key = "l3_port_channels"
 
-        # logic below is common to l3_interface and l3_port_channel interface types
+        is_subinterface = "." in l3_generic_interface.name
 
-        # Check if the interface is a parent L3 Port-Channel with subinterfaces.
-        is_parent_l3_port_channel = schema_key == "l3_port_channels" and l3_generic_interface.name in self._l3_port_channels_with_subinterfaces
```
carl-baillargeon, line 162: Please double check but I think `self._l3_port_channels_with_subinterfaces` can be removed (no longer used).

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
```diff
@@ -156,18 +156,19 @@ def _get_l3_common_interface_cfg(
...
+        # logic below is common to l3_interface and l3_port_channel interface types
+        # Check if the interface is a L3 Port-Channel but not a subinterface.
+        is_l3_port_channel_not_subinterface = schema_key == "l3_port_channels" and not is_subinterface
 
         # TODO: catch if ip_address is not valid or not dhcp
-        if not l3_generic_interface.ip_address and not is_parent_l3_port_channel:
+        # IP address is required for L3 interfaces and L3 Port-Channel subinterfaces,
```
carl-baillargeon: Should the same logic apply to regular L3 interfaces?

In this PR (also in devel), let's say we have Ethernet1 as parent with Ethernet1.100 subinterface, if Ethernet1 has no IP address we raise which is too strict I think.

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
gmuloc: yes - you are commenting a deleted line :)

carl-baillargeon: I meant the property is no longer used anywhere else in the code.

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
gmuloc: thats more complex to understand having hidden `==` in assignment.

gmuloc: but lets relax for also ethernet as we will want to have parents for thernet interface in underlay and subif in network sevrices as well. I will update and add test

gmuloc: I think it is missing as an exmaple in my other PR hence not caught

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
gmuloc: this was not too strict when implementing but now that we want to be able to declare cross modules it is indeed. The reason why I did not see the issue was because in my centralized PR I automaticallu create parents. but this is better to uniformize this.

python-avd/pyavd/_eos_designs/structured_config/underlay/utils.py
gmuloc: yes once again this is a deleted line

gmuloc: which actually is not even in the code

gmuloc: ok :) or a man needs to learn hwo to grep - my bad

gmuloc: we are still using it to make sure that subinterfgaces defined in this module have a parent defined in this module

gmuloc: so still in use (though we could argue to relax this as well but lets do it as a follow up)

carl-baillargeon: Right we call the property in `_validate_l3_port_channels`:
```
parent_names = self._l3_port_channels_with_subinterfaces
```

gmuloc: yup - though we should really maybe relax this later and handle this in my centralizing PR

gmuloc: I could try to detect l3 interfaces without subif and without ip_address and raise on them there (maybe) or as a follow up PR to #6594 to avoid packing it

gmuloc: Created an issue for this!

carl-baillargeon (approved):
LGTM. As discussed internally, we will enforce retrictions on missing IPs if needed in future PRs.

Issue #6656, opened 2026-03-05
Evaluate trixie support for container images


ankudinov:
### Enhancement summary

#6636 failed due to Moby missing in 3.xx-slim-trixie

Tasks:

1. Consider using D-in-D for lab images only, that can be maintained on acLabs repo only
2. Watch for Moby to be integrated into official Python images
3. Consider using other images as base
4. Evaluate uv inside container images
5. Plan a better container ecosystem with a dedicated image per use case

### Which component of AVD is impacted

others

### Use case example

-

### Describe the solution you would like

-

### Additional context

```
  > [linux/amd64 dev_containers_target_stage 4/5] RUN --mount=type=bind,from=dev_containers_feature_content_source,source=docker-in-docker_0,target=/tmp/build-features-src/docker-in-docker_0     cp -ar /tmp/build-features-src/docker-in-docker_0 /tmp/dev-container-features  && chmod -R 0755 /tmp/dev-container-features/docker-in-docker_0  && cd /tmp/dev-container-features/docker-in-docker_0  && chmod +x ./devcontainer-features-install.sh  && ./devcontainer-features-install.sh  && rm -rf /tmp/dev-container-features/docker-in-docker_0:
  0.050     DOCKERDASHCOMPOSEVERSION="none"
  0.050     AZUREDNSAUTODETECTION="true"
  0.050     DOCKERDEFAULTADDRESSPOOL=""
  0.050     INSTALLDOCKERBUILDX="true"
  0.050     INSTALLDOCKERCOMPOSESWITCH="false"
  0.050     DISABLEIP6TABLES="false"
  0.050 ===========================================================================
  0.058 (!) The 'moby' option is not supported on Debian 'trixie' because 'moby-cli' and related system packages have been removed from that distribution.
  0.058 (!) To continue, either set the feature option '"moby": false' or use a different base image (for example: 'debian:bookworm' or 'ubuntu-24.04').
  0.058 ERROR: Feature "Docker (Docker-in-Docker)" (ghcr.io/devcontainers/features/docker-in-docker) failed to install! Look at the documentation at https://github.com/devcontainers/features/tree/main/src/docker-in-docker for help troubleshooting this error.
```

### Contributing Guide

- [x] I agree to follow this project Code of Conduct

d3ej:
[...]

PR #6657, opened 2026-03-05
CI(containers): trixie to bookworm

ankudinov:
## Change Summary

The recent change from bullseye to trixie failed, due to Moby package that is not yet available in many debian distributions including the official Python image.
Our long term target is still trixie. #6656 to track. However this will require better planning and testing.
Reverting back to bullseye is not possible due to lack of 3.14 support.
Switching to bookworm as the workaround to unblock 3.14 builds.

### DIFF
--- a/.github/workflows/container_build_template.yml
+++ b/.github/workflows/container_build_template.yml
@@ -33,7 +33,7 @@ env:
         required: false
         type: string
         default: python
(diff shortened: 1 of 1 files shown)


ClausHolbechArista (approved):

carlbuchmann (approved):
LGTM

PR #6668, opened 2026-03-06
CI(cv_deploy): Adjust mocked unary methods recording gRPC messages

### LINKED ISSUE #6667: CI(cv_deploy): Mocked unary methods don't record some messages
### Issue Summary

# Issue 1

`async def recording_unary_unary` does not record received gRPC message in case gRPC Exception is received.

In some cases we expect to receive gRPC exceptions - like getting one when trying to Get information about Workspace that doesn't exist on CV. Our prod backend code properly reacts on this kind of scenarios but mocked unary methods that are supposed to record all data exchanges do not.

Mocked method needs to be adjusted to record events of getting a gRPC exceptions when message is expected.

# Ussue 2

`async def recording_unary_stream` will currently not record message yielded by AsyncIterator if client code `return`s right after receiving such message.

Example:
 - We subscribe to the state of the newly-created Workspace and wait till its state becomes PENDING.
 - If this is the very first message - out client code will `return` such message, aborting child AsyncIterator and not allowing lines of code in it that are below lastly-served `yield` to be executed (and those are the ones processing message and dumping messages into the recording file)

cv_deploy

### AVD version

devel

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

None

### Steps to reproduce

```shell

```

### Relevant log output

```shell

```

### Contributing Guide

alexeygorbunov:
## Change Summary

Adjust `recording_unary_unary` and `recording_unary_stream` methods to make sure they do not miss any messages and record exceptions (for unary_unary only) as well

## Related Issue(s)

Fixes #6667
Fixes aristanetworks/avd-internal#379

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
Adjust `recording_unary_unary` making sure that it records received gRPC error as well (use case - fetching non-existing Workspace) as they are part of the expected logic/workflow
Adjust `recording_unary_stream` to make sure it records *last* message (in the stream) that triggers `return` on the client side.

## How to test
Pytests will be added in the follow-up PR

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/python-avd/tests/pyavd/cv/mockery.py
+++ b/python-avd/tests/pyavd/cv/mockery.py
@@ -53,6 +53,7 @@ def mocked_cvdevices(hostnames: list[str] | None = None, device_count: int | Non
 def get_recording_file(route: str, request: IProtoMessage, cv_server: str, recording_dir: Path = RECORDING_DIR) -> Path:
     digest = sha1(str(request).encode("UTF-8"), usedforsecurity=False).hexdigest()
     recording_file = recording_dir / Path(route.strip("/")) / cv_server / f"{digest}.json"
+    LOGGER.debug("get_recording_file:\nRoute: '%s'\nRequest: '%s'\nRecording file: '%s'", route, request, recording_file)
     if environ.get("RECORDING"):
         recording_file.parent.mkdir(parents=True, exist_ok=True)
     return recording_file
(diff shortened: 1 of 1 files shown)


python-avd/tests/pyavd/cv/mockery.py
```diff
@@ -155,7 +167,7 @@ async def playback_static_recording_unary_stream(
     response_type: type[T_Message],
     **_kwargs: Any,
 ) -> AsyncIterator[T_Message]:
-    LOGGER.info("Playing back static recording for API request: %s", request)
+    LOGGER.info("playback_static_recording_unary_stream: Playing back static recording for API request: %s", request)
```
ClausHolbechArista, line 170: this new prefix just repeats the same as in the text, so just remove the text after.

ClausHolbechArista (approved):

carl-baillargeon (approved):

python-avd/tests/pyavd/cv/mockery.py
alexeygorbunov: Sure, i'll update it in the next PR as this one got already merged.

PR #6720, opened 2026-03-19
Feat(cv_deploy): Custom user_agent header for initial requests-based REST calls

alexeygorbunov:
## Change Summary

Extend `user_agent` header used by `requests` for the REST call to fetch CV version with additional packaging information

## Related Issue(s)

Fixes aristanetworks/avd-internal#376

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
Extend standard `user_agent` header with versioning information about core PyAVD deps. This can be used by TAC during the initial triaging (when it is sometimes hard to figure out what stack of software was used on a caller side)

## How to test
pytest (`python-avd/tests/pyavd/cv/client/test_cv_client_user_agent.py`)

### User Checklist

- N/A

### Repository Checklist


alexeygorbunov:
Please see associated internal issue for the actual CV logs highlighting results of the proposed PR

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -187,10 +190,14 @@ def _set_version(self) -> None:
             msg = "Unable to get version from CloudVision server. Missing token."
             raise CVClientException(msg)
 
+        custom_http_headers: dict[str, str] = {"Authorization": f"Bearer {self._token}"}
+        if custom_user_agent_header := self._get_user_agent():
+            custom_http_headers.update({"User-Agent": custom_user_agent_header})
```
ClausHolbechArista: Please add a comment on why you need to make the update conditional. For this REST use I would expect it to always be the same. As i minimum it would always contain the python version and pyavd version.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -206,6 +213,38 @@ def _set_version(self) -> None:
...
+        # Process Python version
+        if python_version := platform.python_version():
+            user_agent_parts.append(f"python/{python_version}")
+
+        # Process pyavd version
+        try:
+            pyavd_version = version("pyavd")
+        # Fallback to __version__ avoiding cyclic import
+        except PackageNotFoundError:
+            pyavd_version = getattr(sys.modules.get("pyavd"), "__version__", None)
```
ClausHolbechArista, line 229: This is a very expensive way of fetching the version instead of just adding `from pyavd import __version__` at the top of this file. We are inside the pyavd lib so we should not rely on packaging here.

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: I made it conditional in case `self._get_user_agent()` returns an empty string but indeed this is impossible in reality. Removed it now

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Attempt to import `from pyavd import __version__` inside `python-avd/pyavd/_cv/client/__init__.py` leads to the following:

```
************* Module tests.utils
python-avd/tests/utils.py:1:0: R0401: Cyclic import (pyavd -> pyavd.get_device_structured_config -> pyavd._eos_designs.structured_config -> pyavd._eos_designs.structured_config.network_services -> pyavd._eos_designs.structured_config.network_services.utils_zscaler -> pyavd._cv.client) (cyclic-import)
```

ClausHolbechArista: ok keep what you have for now but this will likely need to change.

ClausHolbechArista (approved):

ClausHolbechArista:
@noredistribution FYI this PR will set the user agent for the "check cloudvision version" REST calls we do at the beginning of each cv_deploy job. So it will be easier to identify versions of python, avd and libraries during troubleshooting. Of course thise needs to go out first, so not there until AVD 6.2.

python-avd/pyavd/_cv/client/__init__.py
gmuloc: we could move the __version__ field to its own file so that we can import it on its own with the cycle

python-avd/tests/pyavd/cv/client/test_cv_client_user_agent.py
```diff
@@ -0,0 +1,212 @@
...
+        cv_client._set_version()
+
+    assert cv_client_user_agent == expected_user_agent
+
+    http_headers = mocked_requests_get.call_args.kwargs["headers"]
+    assert "User-Agent" in http_headers
+    assert http_headers["User-Agent"] == expected_user_agent
+
+
+def test_cv_client_get_user_agent_no_optional_dependancies(cv_client: CVClient) -> None:
```
gmuloc:
```suggestion
def test_cv_client_get_user_agent_no_optional_dependencies(cv_client: CVClient) -> None:
```

and other above

python-avd/tests/pyavd/cv/client/test_cv_client_user_agent.py
```diff
@@ -0,0 +1,212 @@
...
+        cv_client._set_version()
+
+    assert cv_client_user_agent == expected_user_agent
+
+    http_headers = mocked_requests_get.call_args.kwargs["headers"]
+    assert "User-Agent" in http_headers
+    assert http_headers["User-Agent"] == expected_user_agent
+
+
+def test_cv_client_get_user_agent_optional_dependancy_missing(cv_client: CVClient) -> None:
```
gmuloc:
```suggestion
def test_cv_client_get_user_agent_optional_dependency_missing(cv_client: CVClient) -> None:
```

python-avd/tests/pyavd/cv/client/test_cv_client_user_agent.py
```diff
@@ -0,0 +1,212 @@
...
+    "grpclib": "0.4.9",
+    "python-socks": "2.8.1",
+    "requests": "2.32.5",
+}
+
+OPTIONAL_DEPS = ["grpclib", "python-socks", "requests", "ansible-core"]
+
+
+def test_cv_client_get_user_agent_python_version_unavailable(cv_client: CVClient) -> None:
+    expected_user_agent = "pyavd/6.1.0 aristaproto/0.1.4 grpclib/0.4.9 python-socks/2.8.1 requests/2.32.5"
```
gmuloc, line 49: I would have expected ruff to complain about no docstring even in tests but seems we have ignored these

alexeygorbunov: Added docstrings for all new test functions

gmuloc (approved):

PR #6731, opened 2026-03-21
Feat(cv_deploy): Add support for Ansible hostvars with input validation for cv_deploy

### LINKED ISSUE #6180: Feat(cv_deploy): Allow to specify device serial_number via ansible variables
### Issue Summary

With 6.0.0-dev3 the serial_number key was moved to the metadata construct.
However now I'm struggling to have cv_deploy pick up the serial number of a device with using eos_design.

Defining the serial_number as a host_var is no longer working and ignored by cv_deploy.
The only thing I found to be working is putting it in the structured_config of the device. However we are not using eos_design, only eos_cli_config_gen. If I manually generate a structured config file with the serial number it works.

Can you advise how this is supposed to work with 6.0.0 with only eos_cli_config_gen and cv_deploy?

cv_deploy

### AVD version

6.0.0-dev3

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

### Steps to reproduce

```shell
Try to use cv_deploy with a device that has not matching hostname in cloudvision and can only be identified via serial number.
```

### Relevant log output

```shell
cat inventory.yml
---
test:
  hosts:
    TESTHOST1:
      metadata:
        serial_number: "test"
    TESTHOST2:
      serial_number: "test2"

(...)

fatal: [TESTHOST1 -> localhost]: FAILED! => {"changed": false, "errors": ["('Missing devices on CloudVision', CVDevice(hostname='TESTHOST1', serial_number=None, system_mac_address=None, _exists_on_cv=None, _streaming=None), CVDevice(hostname='TESTHOST2', serial_number=None, system_mac_address=None, _exists_on_cv=None, _streaming=None))"]}
```

### Contributing Guide

carl-baillargeon:
## Change Summary

Add support for Ansible hostvars inputs with validation for cv_deploy.

## Related Issue(s)

Fixes #aristanetworks/avd-internal/issues/371
Fixes https://github.com/aristanetworks/avd/issues/6180

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

- Added validate_inputs task to cv_deploy role
- Added support for Ansible hostvars instead of structured config files to cv_deploy
- Added new molecule scenario with Ansible hostvars as inputs

## How to test

Internal CI pipeline

### Repository Checklist


ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -399,12 +407,49 @@ async def build_object_for_device(
...
+                in_metadata = False
+                for line in structured_config_stream:
+                    if line.startswith("metadata"):
+                        in_metadata = True
+                    elif in_metadata and not line.startswith(" "):
+                        break
+                    if in_metadata:
+                        metadata_lines.append(line)
+
+                return load("".join(metadata_lines), Loader=YamlLoader)  # noqa: S506 TODO: Consider safeload
```
alexeygorbunov: Is it possible that file does not have `metadata` block? This would then return `None`

```bash
>>> load("".join([]), Loader=YamlLoader) is None
True
>>> 
```

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
```diff
@@ -20,6 +20,21 @@
   - Create and associate Device and Interface Tags.
   - Approve, run, cancel Change Controls as needed.
 options:
+  preview_features:
```
alexeygorbunov: Do we expect more preview features? If yes - should we have it as dict with options so that users can control specific preview features separately?

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
```diff
@@ -13,39 +13,67 @@
...
-    cv_username: "{{ cv_username | arista.avd.default }}"
-    cv_password: "{{ cv_password | arista.avd.default }}"
-    cv_verify_certs: "{{ cv_verify_certs }}"
-    proxy_host: "{{ proxy_host | arista.avd.default }}"
-    proxy_port: "{{ proxy_port }}"
-    proxy_username: "{{ proxy_username | arista.avd.default }}"
-    proxy_password: "{{ proxy_password | arista.avd.default }}"
-    configuration_dir: "{{ eos_config_dir }}"
-    structured_config_dir: "{{ structured_dir }}"
-    structured_config_suffix: "{{ avd_structured_config_file_format }}"
```
alexeygorbunov, line 31: We are not passing `structured_config_dir` and `structured_config_suffix` to `cv_workflow` anymore. 
What happens if user has `preview_features` set to `False`? `cv_workflow` code will try to fallback to 

```
            else:
                structured_config_dir = validated_args.get("structured_config_dir")
                structured_config_suffix = validated_args.get("structured_config_suffix")
```

P.S. There may be an answer further in the PR code/changes nad i may have not yet gotten there)

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
```diff
@@ -13,39 +13,67 @@
...
+  block:
+    - name: Validate cv_deploy inputs
+      arista.avd.validate_inputs:
+        tmp_dir: "{{ cv_deploy_tmp_dir }}"
+        device_list: "{{ cv_devices }}"
+        batch_size: "{{ cv_deploy_validate_inputs_batch_size | arista.avd.default(10) }}"
+        schema_name: cv_deploy
+        input_dir: "{{ structured_dir }}"
+        input_suffix: "{{ avd_structured_config_file_format }}"
+        read_from_input_dir: "{{ read_structured_config_from_file | arista.avd.default(true) }}"
```
alexeygorbunov, line 26: What happens if user sets `read_structured_config_from_file` to `False` (we are validating ansible vars) but then passes `preview_features` of 'False` to `cv_workflow`? Validation will be done to Ansble vars but deployment will use structured config files?

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
```diff
@@ -18,12 +18,12 @@
     - name: Validate eos_cli_config_gen inputs
       arista.avd.validate_inputs:
         tmp_dir: "{{ eos_cli_config_gen_tmp_dir }}"
+        device_list: "{{ ansible_play_hosts_all }}"
```
alexeygorbunov: `cv_deploy` is using `ansible_play_hosts` by default. This is OK?

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
carl-baillargeon: Users cannot set this to false when using cv_deploy. It is an input from cv_workflow, not cv_deploy. We explicitly set it to true in the role task.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
carl-baillargeon: Similar answer here, `preview_features` cannot be set to false when using cv_deploy. It is an input from cv_workflow which is not exposed by the cv_deploy role.

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
carl-baillargeon: Today we use `ansible_play_hosts_all` hardcoded in the validate_inputs code so we keep the same behavior.

I don't think `ansible_play_hosts_all` vs `ansible_play_hosts` really makes a difference in our workflows anyways.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
carl-baillargeon: Fixed.

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
carl-baillargeon: Good idea. Added the dict.

ansible_collections/arista/avd/extensions/molecule/cv_deploy/invalid_config.yml
```diff
@@ -14,6 +14,7 @@
     cv_skip_missing_devices: true
     eos_config_dir: "{{ playbook_dir }}/intended/configs/test_invalid_configs"
     structured_dir: "{{ playbook_dir }}/intended/structured_configs/test_invalid_configs"
+    cv_deploy_tmp_dir: "{{ playbook_dir }}/intended/tmp_cv_deploy"
```
ClausHolbechArista, line 17: why are you addding this in every playbook? We cannot make this required in a minor release. It needs to be part of the features gated behind the "cv_deploy_preview" toggle. The `validate_inputs` task itself probably also needs to be only run with that toggle on.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
ClausHolbechArista: we need to be more explicit about all this. We need a single toggle that will decide between old or new modes. And new mode is in preview so we can add to it as we go along. When the new mode (preview) is enabled, we should guard against the old inputs and _only_ take the new ones.

ansible_collections/arista/avd/extensions/molecule/cv_deploy/invalid_config.yml
carl-baillargeon: We need to override the default path in molecule because of how folders are structured. Otherwise there is a default role variable set for `cv_deploy_tmp_dir`, just like other roles.

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
```diff
@@ -18,12 +18,12 @@
...
         batch_size: "{{ eos_cli_config_gen_validate_inputs_batch_size | arista.avd.default(10) }}"
         schema_name: eos_config
         input_dir: "{{ structured_dir }}"
         input_suffix: "{{ avd_structured_config_file_format }}"
         read_from_input_dir: "{{ read_structured_config_from_file | arista.avd.default(true) }}"
-        fail_on_validation_errors: false
```
ClausHolbechArista, line 26: why are you changing things in eos_cli_config_gen?

carl-baillargeon: This is the default value in validate_inputs so it's not changing anything but happy to leave it here.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
ClausHolbechArista: i was wrong. So the toggle is there for the workflow but not in the role. That is what we asked for.

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
ClausHolbechArista: as agreed make device_list optional and fallback to current behavior

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
ClausHolbechArista: leave as is to not bloat scope

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
carl-baillargeon: Fixed.

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
carl-baillargeon: Fixed.

ansible_collections/arista/avd/extensions/molecule/cv_deploy/intended/structured_configs/base_configs/avd-ci-core1.yml
carl-baillargeon, line 1: Add back the knob to not fail on missing structured config files.

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
```diff
@@ -13,39 +13,69 @@
...
-    static_config_manifest: "{{ cv_static_config_manifest | arista.avd.default({}) }}"
-    return_details: "{{ cv_register_detailed_results }}"
-  register: cv_deploy_results
+- name: Run cv_deploy
+  block:
+    - name: Validate cv_deploy inputs
+      arista.avd.validate_inputs:
+        tmp_dir: "{{ cv_deploy_tmp_dir }}"
+        device_list: "{{ cv_devices }}"
+        batch_size: "{{ cv_deploy_validate_inputs_batch_size | arista.avd.default(10) }}"
```
alexeygorbunov, line 22: Should default values for `cv_deploy_validate_inputs_batch_size` and `read_structured_config_from_file` be set in `ansible_collections/arista/avd/roles/cv_deploy/defaults/main/main.yml` for consistency (just like most of other vars)?

carl-baillargeon: I am following eos_designs and eos_cli_config_gen. We have an internal issue to track this: https://github.com/aristanetworks/avd-internal/issues/341.

We will tackle this mess later.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
alexeygorbunov, line 1: (probably unrelated to this PR)
When reading from structured_config the following (metadata set to null):

```
hostname: avd-ci-leaf2
metadata: 
```
Validation for this host succeeds and returns

```
{"hostname":"avd-ci-leaf2","metadata":null}
```

Is this what we want?

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
alexeygorbunov: it's just that all other variables (with or without default values) are listed under role's `defaults/main/*.yml` (so it is a common place where all available vars are listed)) but this new ones are not. Just a consistency thing but not an issue for sure.

ansible_collections/arista/avd/plugins/action/validate_inputs.py
carl-baillargeon: `metadata:` and `metadata: null` are the same thing in YAML. We do support null values in avd_design/eos_config schemas. They are handled properly:

```
custom_structured_configuration_vxlan_interface:
```
Will return:
```
vxlan_interface: null
```
...in the structured config and no config will be generated.

For AVD Design, they are simply ignored.

For this metadata use case, it is handled by the `or {}` you suggested in another comment :)

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
carl-baillargeon: Yeah we need to agree on how we handle these.

joelbreton2 (approved):
Tested with internal-ci and all the checks passed

ansible_collections/arista/avd/roles/cv_deploy/tasks/main.yml
ClausHolbechArista: we have a bit of everything everywhere. Let's continue with this as is and we will revisit with the renaming of vars as well.

ansible_collections/arista/avd/extensions/molecule/cv_deploy/cv_deploy_hostvars.yml
```diff
@@ -0,0 +1,224 @@
...
+          - "### STARTING MOLECULE TEST {{ test_id[:69] | upper }} ###"
+          - "{{ ('#' * (31 + test_id | length))[:100] }}"
+
+    - name: "{{ (test_id | upper) + ' - ' + ('Read' if lookup('env', 'MOLECULE_EXECUTION_ID') else 'Generate') + ' molecule execution ID' }}"
+      tags: ["{{ test_id }}"]
+      run_once: true
+      ansible.builtin.set_fact:
+        r: "{{ lookup('env', 'MOLECULE_EXECUTION_ID') or lookup('password', '/dev/null chars=ascii_lowercase,digits length=4') }}"
+
+    - name: "DEBUG BEFORE RESET - {{ test_id | upper }} - metadata keys for {{ intended_tag_device }}"
```
ClausHolbechArista: all inline jinja templates on name fields MUST be at the end. If needed do a single template with contanation inside.

alexeygorbunov: These are temp tasks just for the ongoing troubleshooting. Issue will be fixed in previous playbooks and these tasks will be removed before merging

ansible_collections/arista/avd/plugins/modules/cv_workflow.py
ClausHolbechArista: please move this to the end of the input options to not steal attention from the proper things.

ClausHolbechArista (approved):

ansible_collections/arista/avd/extensions/molecule/cv_deploy/cv_deploy_hostvars.yml
```diff
@@ -0,0 +1,224 @@
...
+            that:
+              - item.0.label == item.1.label
+              - item.0.value == item.1.value
+          loop: "{{ cv_deploy_results.deployed_device_tags | zip(cv_deploy_results.device_tags) | list }}"
+
+        - name: "(BLOCK) Extract all device tags from hostvars - {{ test_id | upper }}"
+          tags: ["{{ test_id }}"]
+          run_once: true
+          ansible.builtin.set_fact:
+            expected_device_tags: "{{ intended_tags.metadata.cv_tags.device_tags | default(intended_tags.cv_device_tags) }}"
```
alexeygorbunov: no fallback to cv_device_tags as we need metadata to win

alexeygorbunov (changes requested):

ansible_collections/arista/avd/extensions/molecule/cv_deploy/cv_deploy_hostvars.yml
carl-baillargeon: Done.

alexeygorbunov (approved):
LGTM

PR #6732, opened 2026-03-22
Feat(cv_deploy): Add options to static config manifest to control deployment behavior

### LINKED ISSUE #6562: Feat(cv_deploy): cv_static_config_manifest to support multiple sites
### Enhancement summary

In the current state, `cv_deploy` cv_static_config_manifest only allows for a single manifest file for all sites.  If you have multiple sites with separate manifest files per site, cv_deploy will rewrite the last container with the new one and remove configlets from the previous Static Configuration Studio container.

cv_deploy

### Use case example

Customer has multiple sites managed with AVD and separate cv_static_config_manifest per site.  They wish to break out a separate cv_static_config_manifest container per site.

Customer has multiple sites managed with AVD and separate cv_static_config_manifest per site.  They wish to break out a separate cv_static_config_manifest container per site.

Desired Static Configuration Studio container hierarchy:

Today the customer must use a single cv_static_config_manifest for all sites with a single Static Configuration Studio container.

Static Configuration
----- AVD Generated Configlets
----- Site 1 Static Configlets (tag_query: "Site1:*")
----- Site 2 Static Configlets (tag_query: "Site2:*")

### Additional context

_No response_

### Contributing Guide

carl-baillargeon:
## Change Summary

- Add option to control how the Studio root container list is managed during static config manifest deployment:
  - "strict": Replace the root list entirely with what is declared in the manifest.
  - "selective": Only undeclared AVD-managed roots are removed. Declared roots are placed first, manual roots after.
  - "loose": Only new roots are added. Existing roots keep their position. Nothing is removed.

- Add option to control how each container's children are managed:
  - "strict": Replace children entirely with what is declared. Both AVD-managed and manual children not in the manifest are removed.
  - "selective": Only undeclared AVD-managed children are removed. Declared children are placed first, manual children after.
  - "loose": Only new children are added. Existing children keep their position. Nothing is removed.

```yaml
        cv_static_config_manifest:
          root_policy: <str; "strict" | "selective" | "loose"; default="selective">
          containers:
            - name: str
              child_policy: <str; "strict" | "selective" | "loose"; default="strict">
```

Points for discussion:

1- Today default behavior with cv_deploy would be `root_policy="selective"` and `child_policy="strict"`. We could simplify a lot by removing the "selective" option and use loose mode on the roots. Would this be considered as a breaking change? ---> **It is a breaking change so we need to keep current behavior ("selective").**

2- With "strict" policies, manually managed roots and children that are unassigned from the hierarchy remain as orphaned objects in CV. There is no way to clean them up easily via the UI. Should we consider a `cleanup_orphans` knob on the manifest to control cleanup?
  - "all": Delete ALL unreachable (orphans) containers.
  - "avd_only" Only delete AVD-managed orphans (current behavior to avoid a breaking change).

A bool could work here. ---> **Can be added later.**

3- For configlets, we would probably need similar options (`configlet_policy`) and same question as 1- since today we operate in "selective" mode (only undeclared AVD-managed configlets are removed). ---> Yes, see https://github.com/aristanetworks/avd/pull/6872.

**UPDATES**:

- Use the following naming instead:
```
exclusive: Remove all <> not specificed in the inputs for this cv_deploy run
controlled: Remove <> contributed by cv_deploy/avd but not specified in the inputs for this cv_deploy run
flexible: Add <> specified in the inputs for this cv_deploy run. Don't remove existing <>
```

- For now, we will not implement exclusive on roots and configlets
- We will also not implement controlled on sub-containers

## Related Issue(s)

Fixes aristanetworks/avd-internal/issues/374
Fixes #6562

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

See Change Summary.

## How to test

cv_deploy molecule with internal CI.

### User Checklist

- [ ] Add similar approach to configlets

### Repository Checklist


ansible_collections/arista/avd/extensions/molecule/cv_deploy/cv_deploy_manifest_only.yml
```diff
@@ -141,20 +141,23 @@
               Received: {{ received | to_nice_json }}
           vars:
             expected:
-              - configlets:
+              - child_policy: strict
```
ClausHolbechArista, line 144: child_policy is not a great name when it applies to sub_containers. sub_container_policy maybe?

carl-baillargeon:
Each policy will be implemented as individual PRs as needed.

PR #6750, opened 2026-03-25
CI: Cleaning up ansible-lint and removing .yamllint from the collection

gmuloc:
## Change Summary

> [!NOTE]
> ~`.ansible-lint-ignore` is auto generated~
> not using the `.ansible-lint-ignore` file option for now - we have hundreds of ignore line and the exit code was not 0 when using it (would need further investigation)

## Proposed changes

* ~remove indent: consistent from yamllint~ => moved to  future PR
* remove .yamllint from the collection (note - the two line 160 ignores and other badly indented or brackets in there are not included in the collection so it should pass)
* fix changelogs/fragemnt_backups and md_toc yaml issues 
* rename wrongly named backup fragments
* better rules for .ansible-lint for offline mode & artifact (not more exclude!)

## How to test

CI is 🍏 

### User Checklist

- [ ] happy to add back indented mode for yamllint instead of ansible-lint-ignore as the impact is that for EVERY structured_config we would need to update ansible-lint-ignore with a line (yes it is auto generated, but boring :) ) - we could keep the .ansible-lint-ignore only for the examples and tests which is what we insert in the collection

### Repository Checklist


ansible_collections/arista/avd/.yamllint
```diff
@@ -36,6 +36,9 @@ rules:
   commas:
     max-spaces-after: -1
     level: error
+  # NOTE: the following rule is also not applied in default config
```
ClausHolbechArista: I am confused here. You write they are _not_ applied in default config for ansible-lint, so why do you change the brackets everywhere then?

gmuloc: not everywhere. only in places where we are including the file in the collection and so it is now failing if I don't change them because I am not including .yamllint anymore.

We shoudl enforce this everywhere afterwards in molecule and so on so we can remove the rule completely from our yamllint.

ansible_collections/arista/avd/.ansible-lint
```diff
@@ -5,12 +5,15 @@ skip_list:
   - name[template] # Remove warnings on task names where Jinja template is not at the end
   - run-once[task] # Remove warnings on using run_once with other strategies
   - var-naming[no-role-prefix] # TODO: Fix internal variable names as a breaking change for AVD 6.0.0
+  - yaml[indentation]
```
gmuloc: remove this and exclude the generated artifcats from the manifest so we don't need it anymore

gmuloc: done

ClausHolbechArista (dismissed):

ClausHolbechArista (approved):

ansible_collections/arista/avd/tests/unit/requirements.txt
```diff
@@ -0,0 +1 @@
+../../requirements.txt
```
MaheshGSLAB: I dont get it how and where it used?

ClausHolbechArista: I think it is used by the tests of verify_requirements

ClausHolbechArista: @MaheshGSLAB please confirm by removing the file and running those tests locally. IF it still works we can try to remove the file and see if anythign else breaks. When we find out why it is required, we can add a comment in the file. Thanks.

MaheshGSLAB: @ClausHolbechArista 
I removed those files from my local and run below tests
```
ansible-test integration --coverage --requirements
ansible-test unit --coverage --requirements
```
and i dont see any failure however in my local the file seen as
```
# This file is @generated by PDM.
# Please do not edit it manually.

anta>=1.8.0
aristaproto>=0.1.1
cryptography>=43.0.0
deepmerge>=1.1.0
distlib>=0.3.9
grpclib>=0.4.8
jinja2>=3.0
netaddr>=0.7.19
pyavd==6.1.0
pyavd-utils==0.0.3
python-socks[asyncio]>=2.7.2
pyyaml>=6.0.0
requests>=2.27.0

```

carlbuchmann (approved):
LGTM

ansible_collections/arista/avd/tests/unit/requirements.txt
carlbuchmann: I removed the symbolic link to requirements.txt file in [2bf5b3f](https://github.com/aristanetworks/avd/pull/6750/commits/2bf5b3f82a34be31af3511ee3207b3942594a625) and tested ansible-test.

ansible_collections/arista/avd/tests/integration/requirements.txt
```diff
@@ -0,0 +1 @@
+../../requirements.txt
```
MaheshGSLAB, line 1: @carlbuchmann we still have it for integration tests

ansible_collections/arista/avd/tests/unit/requirements.txt
gmuloc: I think this may have broken the makefile at the top of the repo that is using `--venv`

PR #6762, opened 2026-03-27
CI: Automatically generate schemas.json.gz when a release is published

### LINKED ISSUE #6724: CI: Automatically generate schemas.json.gz when a release is published
### Enhancement summary

Today we need to manually generate and upload `schemas.json.gz` when we publish a release to be able to use it for other components.

This issue aims at adding a new github worklow when we publish a release to run a script like in #6211 and automatically attach the generate file to the release

others

### Use case example

Needed to download release for other projects

1. a .github workflow calling a script when a release is published
2. the script that would build the file

skeleton
```
name: Build and Attach Release Schemas

on:
  release:
    types: [published]

jobs:
  build-and-upload:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6

      - name: Set up Python
        uses: actions/setup-python@v5
        with:
          python-version: '3.12'

      - name: Install dependencies
        run: |
          # install whatever is needed

      - name: Run Build Schema Script
        run: |
          # this assumes you have added the script and gzip is on the runner
          # implementer needs to check this.
          python python-avd/scripts/build_schemas.py | gzip schemas.json.gz

      - name: Upload Artifact to Release
        # find the correct actions / way to do this 
```

> [!NOTE]
> We do not want to add new actions (apart from the Github built in ones) if we can so we should find a solution without this
> Maybe the script is not even needed on more recent versions so please double check 

doing it manually since 1842

### Additional context

_No response_

### Contributing Guide

Vibhu-gslab:
## Change Summary

This PR aims at adding a new github worklow when we publish a release to run a script and automatically attach the generate file to the release

## Related Issue(s)

Fixes #6724


.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install deepmerge jinja2 jsonschema-rs pydantic pyyaml referencing
```
gmuloc: don't we have this in some dependency group in pyproject.toml so we don't need to handle everything

MaheshGSLAB: not all in one group so may be we can make one

gmuloc: we could use dev (it has more but well we can survive I think)

Shivani-gslab: we can install the `[build-system]` dependecy group from pyproject.toml
it has all these -

```
[build-system]
requires = [
    "cryptography>=43.0.0",
    "deepmerge>=1.1.0",
    "jinja2>=3.0",
    "jsonschema-rs>=0.24",
    "pydantic>=2.3.0",
    "pyyaml>=6.0.0",
    "referencing>=0.35.0",
    "ruff==0.15.7",
    "setuptools>=77.0",
]
```

Shivani-gslab: but these are extras - 
```
"ruff==0.15.7",
"setuptools>=77.0",
"cryptography>=43.0.0",
```

Vibhu-gslab: updated

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  upload-schema:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v5
```
MaheshGSLAB:
```suggestion
        uses: actions/setup-python@v6
```

MaheshGSLAB (approved):

Shivani-gslab (approved):
LGTM

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,36 @@
+---
+name: "Upload schemas.json.gz on release"
+
+"on":
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        # Release tag to upload schemas.json.gz to (e.g. v6.1.0)
```
ClausHolbechArista: can this not be a description instead of a comment here?

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,36 @@
...
+        required: true
+
+jobs:
+  upload-schema:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
```
ClausHolbechArista, line 19: here you checkout without specifying the commit, so this will take whatever is in devel and upload to the given release tag or the recently published release tag. Something might have moved since that point, so we have to check out the specific tag instead.

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,36 @@
...
+        run: pip install --group dev --upgrade
+
+      - name: Generate schemas.json.gz
+        run: scripts/build_schemas.py
+        working-directory: python-avd
+
+      - name: Upload schemas.json.gz to release
+        run: gh release upload ${{ inputs.tag || github.ref_name }} python-avd/pyavd/_schema/schemas.json.gz
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ClausHolbechArista, line 38: I don't think you really need this. I believe the env is already having this by default. Please test.

.github/workflows/release-schema.yml
Vibhu-gslab: updated

.github/workflows/release-schema.yml
Vibhu-gslab: added input.tag

.github/workflows/release-schema.yml
Vibhu-gslab: We are seeing an error on removing `GITHUB_TOKEN`

.github/workflows/release-schema.yml
ClausHolbechArista: is github.ref poiting at the actual code referenced in the release that was published? I am not sure. I suspect it is just HEAD.

ClausHolbechArista: I think you need to fetch the tag from the release that got published and use that tag everywhere

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  upload-schema:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          ref: refs/tags/${{ inputs.tag || github.ref_name }}
```
ClausHolbechArista: For clarity.
```suggestion
          ref: refs/tags/${{ inputs.tag || github.event.release.tag_name }}
```

.github/workflows/release-schema.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Install dependencies
+        run: pip install --group dev --upgrade
+
+      - name: Generate schemas.json.gz
+        run: scripts/build_schemas.py
+        working-directory: python-avd
+
+      - name: Upload schemas.json.gz to release
+        run: gh release upload ${{ inputs.tag || github.ref_name }} python-avd/pyavd/_schema/schemas.json.gz
```
ClausHolbechArista:
```suggestion
        run: gh release upload ${{ inputs.tag || github.event.release.tag_name }} python-avd/pyavd/_schema/schemas.json.gz
```

ClausHolbechArista (changes requested):

ClausHolbechArista (approved):

PR #6779, opened 2026-03-31
Refactor(eos_designs): Refactor the method of route map `RM-CONN-2-BGP` generation[HOLD]

laxmikantchintakindi:
## Change Summary

Refactor the logic for route map RM-CONN-2_BGP generation. Removed some structured_config_contributors from inband_mgmt/__init__ and moved the logic for generating the route map to structured_config_utils.py.
Also generation of RM-CONN-2-BGP route map in network services is now handled in structured_config_utils.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/underlay/prefix_lists.py
```diff
@@ -1,92 +0,0 @@
-# Copyright (c) 2023-2026 Arista Networks, Inc.
```
Shivani-gslab, line 1: I think we should not remove the mixin completely for prefix_lists

python-avd/pyavd/_eos_designs/structured_config/structured_config_utils.py
```diff
@@ -42,5 +44,178 @@ def __init__(
...
+                ipaddress.collapse_addresses(get_ipv4_networks_from_pool(self.shared_utils.vtep_loopback_ipv4_pool)), start=len(sequence_numbers) + 1
+            ):
+                sequence_numbers.append_new(sequence=index * 10, action=f"permit {network} eq 32")
+
+        if self.inputs.vtep_vvtep_ip is not None and self.shared_utils.network_services_l3 is True and not self.shared_utils.is_wan_router:
+            sequence_numbers.append_new(sequence=(len(sequence_numbers) + 1) * 10, action=f"permit {self.inputs.vtep_vvtep_ip}")
+
+        self.structured_config.prefix_lists.append_new(name="PL-LOOPBACKS-EVPN-OVERLAY", sequence_numbers=sequence_numbers)
+
+    def set_once_prefix_list_loopbacks_pim_rp(self: StructuredConfigUtils) -> None:
```
Shivani-gslab, line 183: .

Shivani-gslab: all these `prefix_lists` methods are called for `underlay route-maps` only, so we should not move them to `structured_config_utils`, its better to keep them as they are in the mixin

Shivani-gslab (changes requested):

python-avd/pyavd/_eos_designs/structured_config/underlay/prefix_lists.py
laxmikantchintakindi: As all the methods are moved, removed this file.

python-avd/pyavd/_eos_designs/structured_config/structured_config_utils.py
laxmikantchintakindi: Yes but the main method to set the route map `RM-CONN-2BGP` is called from inband_management and underlay. So moved the methods here. Methods like set_once_prefix_list_loopbacks_pim_rp which are not run_once_methods are called from run_once_method. So without the decorator too they will be called once.

Shivani-gslab: This way the `structured_config_utils` would become very large

python-avd/pyavd/_eos_designs/shared_utils/misc.py
```diff
@@ -461,6 +461,43 @@ def all_connected_endpoints(
 
         return all_connected_endpoints
 
+    @cached_property
+    def vrf_default_evpn(self: SharedUtilsProtocol) -> bool:
```
Shivani-gslab, line 465: I think we should keep these methods under strcutured config, since these are used for structured_config modules only

MaheshGSLAB:
Close this PR and open small PRs

PR #6781, opened 2026-03-31
Feat(eos_designs): Support for Multi-VTEP MLAG

### LINKED ISSUE #3354: Support for multi-VTEP
### Enhancement summary

Hi,

We need to configure MLAG pairs with multi-vtep for supporting single homed hosts (unicast) like described in this TOI:
https://www.arista.com/en/support/toi/eos-4-23-0f/14344-evpn-mlag-single-homed-hosts.

AVD supports multiVTEP configuration with the knobs “evpn_multicast: true” and “underlay_multicast: true”, but this adds unnecessary multicast configuration. 

We would like to have a separate knob in eos_designs to enable this for unicast only.

eos_designs

### Use case example

Single homed hosts attached to MLAG pairs. Unicast traffic only.

Generate multivtep configuration:
```
interface Vxlan1
   description DC1-CL1_VTEP
   vxlan source-interface Loopback0
   vxlan virtual-router encapsulation mac-address mlag-system-id
..
   vxlan mlag source-interface Loopback1
```

But avoid any multicast configuration being generated as it occurs today when configuring "evpn_multicast: true” and “underlay_multicast: true”,

_No response_

### Additional context

_No response_

### Contributing Guide

jrecchia1029:
## Change Summary

Provide option to configure MLAG VTEPs with Multi-VTEP MLAG feature to optimize path taken for traffic destined to orphan ports.

https://www.arista.com/en/support/toi/eos-4-21-3f/14113-multi-vtep-mlag
https://www.arista.com/en/support/toi/eos-4-23-0f/14344-evpn-mlag-single-homed-hosts

TOIs listing limitations with multi-vtep mlag

- https://www.arista.com/en/support/toi/eos-4-26-1f/14785-multi-domain-evpn-vxlan
- https://www.arista.com/en/support/toi/eos-4-24-1f/14533-ipv6-underlay-support-for-vxlan-with-evpn-control-plane

## Related Issue(s)

Fixes #3354

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Add a node type setting boolean input called `multi_vtep_mlag` with default value `false`

```
<node_type_keys.key>:
  defaults:
    multi_vtep_mlag:
      enabled: <bool; default=False>
      vxlan_decap_on_default_vrf_only: <bool; default=True>
  node_groups:
    - group: <str; required; unique>
      multi_vtep_mlag:
        enabled: <bool; default=False>
        vxlan_decap_on_default_vrf_only: <bool; default=True>
      nodes:
          multi_vtep_mlag:
            enabled: <bool; default=False>
            vxlan_decap_on_default_vrf_only: <bool; default=True>
```

## How to test

New molecule test

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/schema/schema_fragments/multi_vtep_mlag.yml
```diff
@@ -0,0 +1,15 @@
+# Copyright (c) 2023-2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+# yaml-language-server: $schema=../../../_schema/avd_meta_schema.json
+# Line above is used by RedHat's YAML Schema vscode extension
+# Use Ctrl + Space to get suggestions for every field. Autocomplete will pop up after typing 2 letters.
+type: dict
+keys:
+  multi_vtep_mlag:
```
ClausHolbechArista: I think we shoudl add this under the node config instead of globally.

So:
```yaml
l3leaf:
  node_groups:
    - name: shared_vtep
      nodes:
        - name: leaf1a
          ...
        - name: leaf1b
          ...
    - name: multi_vtep
      mlag_multi_vtep: true
      nodes:
        - name: leaf2a
          ...
        - name: leaf2b
```

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
```diff
@@ -53,7 +53,7 @@ class VxlanInterfaceMixin(Protocol):
 
     @cached_property
     def _multi_vtep(self: AvdStructuredConfigNetworkServicesProtocol) -> bool:
-        return self.shared_utils.mlag is True and self.shared_utils.evpn_multicast is True
+        return self.shared_utils.mlag is True and (self.shared_utils.evpn_multicast is True or self.inputs.multi_vtep_mlag is True)
```
ClausHolbechArista: it looks simple, but we may have to consider other conflicts with this. Will this work with EVPN gateway (one TOI says no)? Or ipv6_numbered? I ask because for those areas we check for other feature interaction.

python-avd/pyavd/_eos_designs/schema/schema_fragments/multi_vtep_mlag.yml
jrecchia1029: Makes sense. Especially given the limitations that you pointed out in the other comment.

python-avd/pyavd/_eos_designs/structured_config/network_services/vxlan_interface.py
jrecchia1029: True. I didn't realize that. Found TOIs where both limitations you mentioned are listed and added some checks for those situations.

Vibhu-gslab:
We have missed this [this](https://github.com/aristanetworks/avd/issues/3354#issuecomment-2840937518) point from @ClausHolbechArista or is it not needed?

Also, you can update `Proposed changes` above :)

jrecchia1029:
> We have missed this [this](https://github.com/aristanetworks/avd/issues/3354#issuecomment-2840937518) point from @ClausHolbechArista or is it not needed?
> 
> Also, you can update `Proposed changes` above :)

I'm not sure that SA is unique to only this multi vtep mlag feature. It seems to be relevant for any network with traffic from multiple VRFs traversing a single link.

Still, if we decide that the SA only affects multi-vtep mlag deployments, we'll need to change the data model to something like the below:
```
multi_vtep_mlag:
  enabled: <bool; default=True>
  vxlan_decap_on_default_vrf_only: <bool; default=True>
```

Maybe it's best to only change the data model for the first part now in the event we need to add vxlan decapsulation filter flag in the future.

```
multi_vtep_mlag:
  enabled: <bool; default=True>
```

ClausHolbechArista:
@jrecchia1029 can you investigate the `vxlan decapsulation filter interface multiple-vrf disabled` problem more? if you have access to a lab with the mentioned platforms, can we see if this is a problem or if they somehow worked around it on EOS?
We can have a quick call to discuss.

PR #6782, opened 2026-03-31
Feat(pyavd): Add merge function to AvdManifest

carl-baillargeon:
## Change Summary

Add `AvdManifest.merge()` class method to merge multiple manifests.

This method will be used later when we introduce the per-device manifest in cv_deploy.

## Related Issue(s)

Part of cv_deploy enhancements.

## Component(s) name

PyAVD

## Proposed changes

- Configlets are deduplicated by name.
- Containers at the same level with the same name are merged recursively.
- Raising errors on conflicts.

## How to test

Unit tests.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -329,6 +330,73 @@ def from_dict(cls, data: dict[str, Any]) -> AvdManifest:
             msg = f"Failed to build the static configuration manifest. Please check your input data. Original error: {e}"
             raise ValueError(msg) from e
 
+    @classmethod
+    def merge(cls, *manifests: AvdManifest) -> AvdManifest:
```
ClausHolbechArista, line 383: We could change this to use our AvdModel/AvdList/AvdIndexedList classes and we would get this for free.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -329,6 +330,73 @@ def from_dict(cls, data: dict[str, Any]) -> AvdManifest:
...
+        """
+        if not manifests:
+            return cls()
+
+        if len(manifests) == 1:
+            return manifests[0]
+
+        merged_configlets: dict[str, AvdConfiglet] = {}
+        for manifest in manifests:
+            for configlet in manifest.configlets:
```
ClausHolbechArista, line 399: I am not sure it is a good idea to do a deep merge on configlets. It would be weird to have configlets from different device manifests be combined into the same container. I think we should enforce strict match for each container layer.

python-avd/pyavd/_cv/workflows/models.py
ClausHolbechArista: hm depends on which merge behavior we want. It may not be as easy with the nested subcontainers if we want to be strict on only some fields.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: Yeah good idea. Fixed.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -378,6 +379,74 @@ def from_dict(cls, data: dict[str, Any]) -> AvdManifest:
...
+        merged: dict[str, AvdContainer] = {}
+        for container in chain.from_iterable(container_tuples):
+            if container.name in merged:
+                merged[container.name] = cls._merge_two_containers(merged[container.name], container)
+            else:
+                merged[container.name] = container
+        return tuple(merged.values())
+
+    @classmethod
+    def _merge_two_containers(cls, existing: AvdContainer, other: AvdContainer) -> AvdContainer:
```
ClausHolbechArista, line 425: why a class method instead of mergeing self and other?
In other places we call such methods `merged(other)` or `combined(other)` since we are more restrictive on this merge.

ClausHolbechArista: oh of course those methods would go on the container class and not the manifest

carl-baillargeon:
Moving to draft. The merge will happen on CVContainer and CVConfiglet instead using ID classes.

carl-baillargeon:
Will revisit later.

PR #6811, opened 2026-04-08
Fix(eos_designs): Refactor structured_config/inband_management/__init__.py

laxmikantchintakindi:
## Change Summary

Refactor structured_config/inband_management/__init__.py
This PR also fixes the ipv6 inband management behavior under a new future Knob. When this knob is enabled:
1. Inband management VRF is generated when IPv6 inband management is set.
2. Prefix list `IPv6-PL-L2LEAF-INBAND-MGMT` is not generated when overlay_routing_protocol is `none`.

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -216,3 +162,19 @@ def get_parent_svi_cfg(self, vlan: int, subnet: str | None, ipv6_subnet: str | N
                 svi.ipv6_attached_host_route_export._update(enabled=True, distance=19)
 
         return svi
+
+    @cached_property
+    def _inband_mgmt_ipv6_parent(self) -> bool:
+        if self.shared_utils.inband_management_parent_vlans:
```
Vibhu-gslab: Don't use this if condition here, instead put it where you are calling this function.

Vibhu-gslab (changes requested):

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: This should be seperate from that if condition. As this is iterating through all the subnets.

laxmikantchintakindi: Done for both methods.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -37,159 +60,82 @@ def vlans(self) -> None:
...
         sequence_numbers = EosCliConfigGen.RouteMapsItem.SequenceNumbers()
-        if self._inband_mgmt_ipv4_parent:
-            sequence_numbers.append_new(
-                sequence=20, type="permit", match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ip address prefix-list PL-L2LEAF-INBAND-MGMT"])
-            )
+        sequence_numbers.append_new(
+            sequence=20, type="permit", match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ip address prefix-list PL-L2LEAF-INBAND-MGMT"])
+        )
+        route_map = self.structured_config.route_maps.obtain("RM-CONN-2-BGP")
+        route_map.sequence_numbers.extend(sequence_numbers)
```
ClausHolbechArista: use append_new directly here instead of building sequence_numbers above.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -37,159 +60,82 @@ def vlans(self) -> None:
...
 
-    @structured_config_contributor
-    def route_maps(self) -> None:
-        if not self.shared_utils.inband_management_parent_vlans or not self.shared_utils.underlay_bgp or not self.inputs.underlay_filter_redistribute_connected:
-            return
-
-        if self.shared_utils.inband_mgmt_vrf is not None or self.shared_utils.overlay_routing_protocol == "none":
-            return
-
+    def _set_ipv4_route_maps(self) -> None:
```
ClausHolbechArista: let's make this more precise, like `set_route_map_conn_2_bgp_seq_20`

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -14,21 +14,44 @@
...
+                if self.shared_utils.inband_mgmt_gateway is not None:
+                    self._set_static_route()
+            if self.shared_utils.configure_inband_mgmt_ipv6 and self.shared_utils.inband_mgmt_ipv6_gateway is not None:
+                self._set_ipv6_static_route()
             return
+
+        if self.shared_utils.inband_management_parent_vlans:
+            self._set_ip_virtual_router_mac_address()
+            self._set_parent_vlans()
+            self._set_parent_vlan_interfaces()
```
ClausHolbechArista: these helpers are still going over all the items. It would be better if you iterate here and call specific helpers per vlan.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -216,3 +162,15 @@ def get_parent_svi_cfg(self, vlan: int, subnet: str | None, ipv6_subnet: str | N
...
+        for subnet in self.shared_utils.inband_management_parent_vlans.values():
+            if subnet["ipv6"]:
+                return True
+        return False
+
+    def _inband_mgmt_ipv4_parent(self) -> bool:
+        for subnet in self.shared_utils.inband_management_parent_vlans.values():
+            if subnet["ipv4"]:
+                return True
+        return False
```
ClausHolbechArista: In the optimal case these should not be needed. They are a sign of you doing logic on general detection instead of calling it based on actual inputs and using set_once.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: Fixed.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: Iterated over vlans in structured_config_contributor and removed these helpers.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
-    @structured_config_contributor
-    def router_bgp(self) -> None:
-        if self.shared_utils.inband_mgmt_vrf is not None:
-            return
-
-        if not self.shared_utils.inband_management_parent_vlans or not self.shared_utils.underlay_bgp:
-            return
-
+    @run_once_method
+    def _enable_router_bgp_redistribute_attached_host(self) -> None:
```
Vibhu-gslab:
```suggestion
    def _set_once_enable_router_bgp_redistribute_attached_host(self) -> None:
```

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
-                if subnet["ipv4"]:
-                    return True
-        return False
+            self._set_ip_virtual_router_mac_address()
+            self._set_inband_mgmt_vrf()
+            for index, (vlan_id, vlan) in enumerate(self.shared_utils.inband_management_parent_vlans.items(), start=1):
+                self._set_parent_vlans(vlan_id)
+                self._set_parent_vlan_interface(vlan, vlan_id)
+                if self.shared_utils.inband_mgmt_vrf is None and self.shared_utils.underlay_bgp:
+                    self._enable_router_bgp_redistribute_attached_host()
```
Vibhu-gslab: I understand we are using `run_once` on this method, but I am still not sure if we should call it from within a `for` loop 
@ClausHolbechArista , @gmuloc

laxmikantchintakindi:
```suggestion
                    self._set_once_enable_router_bgp_redistribute_attached_host()
```

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
+                if self.shared_utils.inband_mgmt_vrf is None and self.shared_utils.underlay_bgp:
+                    self._set_once_enable_router_bgp_redistribute_attached_host()
+                    if self.inputs.underlay_filter_redistribute_connected:
+                        if vlan["ipv6"]:
+                            if self.shared_utils.overlay_routing_protocol != "none":
+                                self._set_once_route_map_conn_2_bgp_sequence_60()
+                            self._set_l2leaf_inband_mgmt_ipv6_prefix_lists(vlan, index)
+                        if self.shared_utils.overlay_routing_protocol != "none" and vlan["ipv4"]:
+                            self._set_once_route_map_conn_2_bgp_sequence_20()
+                            self._set_l2leaf_inband_mgmt_prefix_lists(vlan, index)
```
Vibhu-gslab: To preserve structured config order:
```suggestion
                        if self.shared_utils.overlay_routing_protocol != "none" and vlan["ipv4"]:
                            self._set_once_route_map_conn_2_bgp_sequence_20()
                            self._set_l2leaf_inband_mgmt_prefix_lists(vlan, index)
                        if self.shared_utils.overlay_routing_protocol != "none" and vlan["ipv6"]:
                                self._set_once_route_map_conn_2_bgp_sequence_60()
                            self._set_l2leaf_inband_mgmt_ipv6_prefix_lists(vlan, index)
```

Vibhu-gslab (commented):
One minor comment rest looks good.

Vibhu-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
+    def _set_vlans(self) -> None:
+        # TODO: Refactor this later to inject from filtered tenants
+        # Note that an attempt was made for this in #6073 but it has been postponed
+        # To keep current behavior we need to overwrite the existing values if the vlan was introduced via network_services
+        # otherwise it is created via obtain and updated accordingly.
+        inband_mgmt_vlan = self.structured_config.vlans.obtain(self.shared_utils.node_config.inband_mgmt_vlan)
+        inband_mgmt_vlan.name = self.shared_utils.node_config.inband_mgmt_vlan_name
+        inband_mgmt_vlan.metadata.tenants.append_unique("system")
+
+    def _set_parent_vlans(self, vlan_id: int) -> None:
```
ClausHolbechArista:
```suggestion
    def _set_parent_vlan(self, vlan_id: int) -> None:
```

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
+            description=self.shared_utils.node_config.inband_mgmt_description,
+            shutdown=False,
+            mtu=self.shared_utils.inband_mgmt_mtu,
+            vrf=self.shared_utils.inband_mgmt_vrf,
+            ip_address=self.shared_utils.inband_mgmt_ip,
+            ipv6_enable=None if not self.shared_utils.configure_inband_mgmt_ipv6 else True,
+            metadata=EosCliConfigGen.VlanInterfacesItem.Metadata(type="inband_mgmt"),
+        )
+        if ipv6_address := self.shared_utils.inband_mgmt_ipv6_address:
+            vlan_interface.ipv6_addresses.append(ipv6_address)
```
ClausHolbechArista, line 88: please don't change an object returned from append_new. It will mess up duplicate detection. So build the full object first and then append it. Or move the ipv6_addresses assignment inline with some `...Ipv6Addresses([ipv6_address]) if ipv6_address else None`

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
+            for index, (vlan_id, vlan) in enumerate(self.shared_utils.inband_management_parent_vlans.items(), start=1):
+                self._set_parent_vlans(vlan_id)
+                self._set_parent_vlan_interface(vlan, vlan_id)
+                if self.shared_utils.inband_mgmt_vrf is None and self.shared_utils.underlay_bgp:
+                    self._set_once_enable_router_bgp_redistribute_attached_host()
+                    if self.inputs.underlay_filter_redistribute_connected:
+                        if self.shared_utils.overlay_routing_protocol != "none" and vlan["ipv4"]:
+                            self._set_once_route_map_conn_2_bgp_sequence_20()
+                            self._set_l2leaf_inband_mgmt_prefix_lists(vlan, index)
+                        if vlan["ipv6"]:
```
ClausHolbechArista, line 44: are you sure this should not be nested under overlay_routing_protocol != "none" like ipv4?

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,153 @@
...
         route_map = self.structured_config.route_maps.obtain("RM-CONN-2-BGP")
-        route_map.sequence_numbers.extend(sequence_numbers)
+        route_map.sequence_numbers.append_new(
+            sequence=60,
+            type="permit",
+            match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ipv6 address prefix-list IPv6-PL-L2LEAF-INBAND-MGMT"]),
+        )
 
-    def get_parent_svi_cfg(self, vlan: int, subnet: str | None, ipv6_subnet: str | None) -> EosCliConfigGen.VlanInterfacesItem:
+    def _set_parent_vlan_interface(self, vlan: dict, vlan_id: int) -> None:
```
ClausHolbechArista, line 137: I think we could create the vlan object in here as well.

laxmikantchintakindi: Yes, vlan object is already there.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: Created vlan interface object.

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
laxmikantchintakindi: This condition is not present at https://github.com/aristanetworks/avd/pull/6811/changes#diff-631ccdc44c86acc9896eabc7d3c56c494ba4cc012a7c21d5fcc4462a0036711bL159

ClausHolbechArista: that is clearly a bug. If you check the route-map logic that refers to that prefix-list it is returned early if overlay is none.

ClausHolbechArista: This is an example of where it would be better to just call the prefix-list creation code from within the route-map creation code, so all the conditions are outside.

laxmikantchintakindi: Added fix under the future knob `retain_inband_mgmt_ipv6_behaviour` with default true. When this is set to false, then ipv6 prefix list and inband_mgmt_vrf will not be generated.

ClausHolbechArista: We need to flip this so default is always false. fixed_ipv6_inband_mgmt_config or something. @gmuloc any views here?

laxmikantchintakindi: sorry I missed to edit the comment. Renamed the key to inband_mgmt_ipv6_fix with default false.

Shivani-gslab (approved):
LGTM

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,160 @@
...
-            sequence_numbers.append_new(
-                sequence=60,
-                type="permit",
-                match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ipv6 address prefix-list IPv6-PL-L2LEAF-INBAND-MGMT"]),
-            )
+    @run_once_method
+    def _set_once_route_map_conn_2_bgp_sequence_20(self) -> None:
+        route_map = self.structured_config.route_maps.obtain("RM-CONN-2-BGP")
+        route_map.sequence_numbers.append_new(
+            sequence=20, type="permit", match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ip address prefix-list PL-L2LEAF-INBAND-MGMT"])
```
gmuloc, line 125: any reason why this is not calling the prefix list function here?

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
```diff
@@ -3,216 +3,160 @@
...
+        )
 
+    @run_once_method
+    def _set_once_route_map_conn_2_bgp_sequence_60(self) -> None:
         route_map = self.structured_config.route_maps.obtain("RM-CONN-2-BGP")
-        route_map.sequence_numbers.extend(sequence_numbers)
+        route_map.sequence_numbers.append_new(
+            sequence=60,
+            type="permit",
+            match=EosCliConfigGen.RouteMapsItem.SequenceNumbersItem.Match(["ipv6 address prefix-list IPv6-PL-L2LEAF-INBAND-MGMT"]),
```
gmuloc: same as above?

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
gmuloc: hmmm I suppose becqause you want to be able to create the correct sequ3nce ID and the various entries and you would have to end over all these value here. Ok I guess lets leave it like this :) (same for ipv6)

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
```diff
@@ -31,3 +31,10 @@ keys:
         description: |-
           Configure the `mlag_ipv4_vrfs_peer` BGP peer group only when needed.
         default: false
+      inband_mgmt_ipv6_fix:
+        type: bool
+        default: false
+        description: |-
+          Fix the current behavior of IPv6 inband management. When this key is set to true:
+          1. Inband management VRF is generated when IPv6 inband management is set.
+          2. Prefix list `IPv6-PL-L2LEAF-INBAND-MGMT` is not generated when overlay_routing_protocol is `none`.
```
gmuloc: lets use two keys

```
configure_inband_mgmt_ipv6_vrf: <bool, default False>
only_configure_ipv6_inband_mgmt_prefix_list_when_used: <bool, default False> # add name inside description
```

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
gmuloc: discussed above to rename to two keys

python-avd/pyavd/_eos_designs/structured_config/inband_management/__init__.py
gmuloc: resolving (kept them for history ;) )

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
laxmikantchintakindi: Done.

python-avd/pyavd/_eos_designs/structured_config/structured_config_utils.py
```diff
@@ -236,5 +237,88 @@ def set_mlag_peer_group(
...
+            if not peer_in_fabric:
+                return False
+            # Peer is in the fabric - only respect an explicit False from user input; never force True in Digital Twin.
+            return False if user_input is False else Undefined
+        # Non-Digital-Twin: follow the user input if set, otherwise leave unset.
+        return Undefined if user_input is None else user_input
+
+    # TODO: AVD 7.0 - Move this to inband_management module once future knob is removed.
+    @run_once_method
+    def set_inband_mgmt_vrf(self: StructuredConfigUtils) -> None:
```
gmuloc: Why do you need this here if it is used only in inband_management?

laxmikantchintakindi: Moved to inband_management.

gmuloc (approved):

PR #6837, opened 2026-04-13
Refactor(eos_designs): Protect against configuring unsupported monitor sessions on sub interface

### LINKED ISSUE #6760: Feat(eos_designs): Protect against configuring unsupported monitor sessions on subinterface
### Enhancement summary

TOI: https://www.arista.com/en/support/toi/eos-4-21-3f/14157-mirroring-acls-with-subinterface-as-source

eos_designs

### Use case example

When configuring 

```
tenants:
  - name: TENANT
    vrfs:
      - name: VRF
        l3_interfaces:
          - interfaces: [Ethernet1.42, Ethernet1.43]
            ip_addresses: [10.0.0.1/31, 10.0.0.2/31]
            nodes: [leaf-1a, leaf-2a]
            monitor_sessions:
              - name: SESSION-NAME
                role: source
                source_settings:
                  direction: both <<<<<<<<<<<<<<< not supported on sub interfaces and should be blocked.
```

need to look at the TOI to see what should be allowed (probably need to block monitor sessions all together on unsupported platforms)

_No response_

### Additional context

_No response_

### Contributing Guide

MaheshGSLAB:
## Change Summary

Protect against configuring unsupported monitor sessions on sub interface

## Related Issue(s)

Fixes #6760 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Monitor session on sub interface only support rx direction.

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -143,15 +143,30 @@ def _monitor_session_configs(
                     for node_index, node_name in enumerate(l3_interface.nodes):
                         if node_name != self.shared_utils.hostname:
                             continue
+                        interface_name = l3_interface.interfaces[node_index]
+                        context = f"{tenant._internal_data.context}[name={tenant.name}].vrfs[name={vrf.name}].l3_interfaces[{l3_interface_index}]"
```
vitthalmagadum, line 147: Do we need to handle port-channel sub interface as Sub interface port-channels are not supported as mirror sources as per TOI?

MaheshGSLAB: we do not support monitor session on l3_port_channels itself.

vitthalmagadum: l3_interfaces accepts port-channels inside interfaces.

vitthalmagadum:
Please check for code coverage. Thanks!!

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
```diff
@@ -93,6 +93,12 @@ $defs:
               description: |
                 Support for MTU configuration under sub-interfaces.
                 When this key is set to False, MTU is not rendered under sub-interfaces even if it is set in the inputs.
+            subinterface_monitor_session:
+              type: bool
+              default: true
+              description: |-
+                Support for monitor session configuration for sub-interfaces.
+                When this key is set to False, monitor session is not rendered for sub-interfaces even if it is set in the network services inputs.
```
Vibhu-gslab:
```suggestion
                Support for monitor session configuration on sub-interfaces.
                When this key is set to false, the monitor session will not be applied to sub-interfaces, even if it is set in the network services inputs.
```

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
Vibhu-gslab: @ClausHolbechArista @gmuloc

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
```diff
@@ -93,6 +93,12 @@ $defs:
               description: |
                 Support for MTU configuration under sub-interfaces.
                 When this key is set to False, MTU is not rendered under sub-interfaces even if it is set in the inputs.
+            subinterface_monitor_session:
+              type: bool
+              default: true
+              description: |-
+                Support for monitor session configuration on sub-interfaces.
+                When this key is set to false, the monitor session will not be applied to sub-interfaces, even if it is set in the network services inputs.
```
laxmikantchintakindi, line 101: This condition is raises error.

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
MaheshGSLAB: Raised new issue to track it separately.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
MaheshGSLAB: updated the description

vitthalmagadum (approved):

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -143,15 +143,30 @@ def _monitor_session_configs(
...
                         for monitor_session in l3_interface.monitor_sessions:
+                            if "." in interface_name:
+                                if not self.shared_utils.platform_settings.feature_support.subinterface_monitor_session:
```
laxmikantchintakindi: This condition should be checked before the loop.

MaheshGSLAB: But this will raise the error even if do not defined the monitor session and the specific platform not support the monitor session on sub interface. So we should only check when there is key `monitor_sessions`

MaheshGSLAB: it will raise the error for the example https://github.com/aristanetworks/avd/blob/devel/ansible_collections/arista/avd/examples/isis-ldp-ipvpn/group_vars/NETWORK_SERVICES.yml#L22

laxmikantchintakindi: yes. If we check under the loop, it will repeat the check for every session configured. So we can check before loop if it is subinterface and l3_interface.monitor_sessions is not none.

laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
```diff
@@ -93,6 +93,12 @@ $defs:
               description: |
                 Support for MTU configuration under sub-interfaces.
                 When this key is set to False, MTU is not rendered under sub-interfaces even if it is set in the inputs.
+            subinterface_monitor_session:
+              type: bool
+              default: true
+              description: |-
+                Support for monitor session configuration on sub-interfaces.
+                When this key is set to false, an error will be raised if a monitor session is configured on a sub-interface in the network services inputs.
```
alexeygorbunov: `in the network services inputs` - monitor sessions can also be configured under connected endpoints and network ports. Are we not enforcing protection for those two as well?

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -143,15 +143,33 @@ def _monitor_session_configs(
...
+                            and "." in interface_name
+                            and not self.shared_utils.platform_settings.feature_support.subinterface_monitor_session
+                        ):
+                            msg = (
+                                f"Monitor session on sub-interfaces is not supported on this platform. "
+                                f"Got monitor session on interface '{interface_name}' under {context}."
+                            )
+                            raise AristaAvdInvalidInputsError(msg)
                         for monitor_session in l3_interface.monitor_sessions:
+                            if "." in interface_name and monitor_session.source_settings.direction != "rx":
```
alexeygorbunov: We are not checking here that role is set to "source".

If i change `network-services-monitor-session-subinterface-direction` negative example to be a destination port:

```
            monitor_sessions:
              - name: SESSION-ON-SUBINTERFACE-WITH-BOTH-DIRECTION
                role: destination
                # source_settings:
                #   direction: both
```

exception is raised saying that `"Only 'direction: rx' supported on sub-interfaces for monitor session. Got 'direction: None' for monitor session 'SESSION-ON-SUBINTERFACE-WITH-BOTH-DIRECTION' on interface 'Ethernet6.101' under tenants[name=TENANT_A].vrfs[name=VRF_A].l3_interfaces[0]."` although exception should probably be saying that sub-interface can only be used as a source port in RX direction for the monitor session (meaning no destionation and no sourec with anithing else other than RX)

```
ag-dc1-leaf1a(config-s-s5b)#monitor session s1 destination ethernet 1.1 
% Mirror destination Ethernet1.1 was rejected with reason Sub-interface destination not supported
ag-dc1-leaf1a(config-s-s5b)#monitor session s1 destination port-Channel 100.100
% Mirror destination Port-Channel100.100 was rejected with reason Sub-interface destination not supported
ag-dc1-leaf1a(config-s-s5b)#
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -66,6 +66,7 @@ keys:
           sflow_subinterfaces: false
           subinterface_mtu: false
           per_interface_l2_mru: false
+          subinterface_monitor_session: false
```
alexeygorbunov: 720XP platform does support it

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
```diff
@@ -313,6 +321,7 @@ keys:
           subinterface_mtu: false
           per_interface_l2_mru: false
           sflow_subinterfaces: false
+          subinterface_monitor_session: false
```
alexeygorbunov: 7368X4 seem to support it

```
glc390...05:23:34(config-s-s2b)#show session-config diffs 
--- system:/running-config
+++ session:/s2b-session-config
+interface Port-Channel100.100
+!
+monitor session s1 source Port-Channel100.100 rx
glc390...05:23:48(config-s-s2b)#sh version 
Arista 7368-F
...
Software image version: 4.35.2F
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
```diff
@@ -93,6 +93,12 @@ $defs:
               description: |
                 Support for MTU configuration under sub-interfaces.
                 When this key is set to False, MTU is not rendered under sub-interfaces even if it is set in the inputs.
+            subinterface_monitor_session:
```
alexeygorbunov, line 96: Do we really need to raise if this feature is not supported but is present in inputs? If we do - we need a workaround for Digital Twin as vEOS does not support it but we still need to get DT config generated, just without this feature

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
MaheshGSLAB: I added the support as per TOI(https://www.arista.com/en/support/toi/eos-4-21-3f/14157-mirroring-acls-with-subinterface-as-source)
Updated the default now

python-avd/pyavd/_eos_designs/schema/schema_fragments/platform_settings.schema.yml
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
MaheshGSLAB: added

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
MaheshGSLAB: updated

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
MaheshGSLAB: Let me discuss with @ClausHolbechArista and @gmuloc

ClausHolbechArista: If someone configures it but not getting it, we should raise. We can add a condition on that raise to check for digital twin and _not_ raise in that case. That should probably be done as a more general thing, so maybe we need a new error type like FeatureNotSupportedByPlatform that we always raise via a helper like raise_platform_feature_not_supported where we can ignore for digital twin. Maybe too  much for this PR, but would be a good improvement later.

gmuloc: agree with Claus.
To get this one to the finish line, let's add a guard to not raise for Digital Twin in the code with a TODO and let's track Claus suggestion as a general issue?

MaheshGSLAB: Added new issue and TODO in code.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
alexeygorbunov: Thanks Mahesh! Maybe we can just drop ` in the network services inputs` in this sentence completely to not specify all places where this can be requested (as it can be not only network services but connected endpoints and network ports)

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
alexeygorbunov: If is change `/ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/Tenant_A.yml` to 

```yaml
          # Test monitor session on subinterface
          - interfaces: [Ethernet7.7]
            ip_addresses: [10.10.20.27/24]
            nodes: [dc1-bl1b]
            mtu: 9000
            enabled: true
            description: "test_monitor_session_on_sub_interface"
            monitor_sessions:
              - name: Monitor_session_on_sub_interface
                role: destination # <<<<<<<<<<<<<<<<<
                source_settings:
                  direction: both # <<<<<<<<<<<<<<<<<<
```

then eos_designs_unit_tests still perfectly produces the following line for `dc1-bl1b`:

```
monitor session Monitor_session_on_sub_interface destination Ethernet7.7
```

I checked few platforms and none of the support sub-interfaces as a destination for the monitor session (i'm getting `% Mirror destination EthernetX.Y was rejected with reason Sub-interface destination not supported`). I'm trying to check internally if there are any platforms at all that support it.
If there are none that support -we'll need to block definition of the sub-interfaces as mirroring destination ports. 
If there are some that supports it - we may need to extend definition of this feature to also track support of sub-interfaces as "destination"

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
alexeygorbunov: Ok, Digital Twin will be handled in a separate PR later (based on TODO)

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
alexeygorbunov: I got a confirmation that no platforms right now support sub-interfaces as port mirroring destination. So we need to block this completely.

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
MaheshGSLAB: removed

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
MaheshGSLAB: ok added the error for the same

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
laxmikantchintakindi: I think network_ports should not be sub-interfaces. In network_port.schema.yml, it is not mentioned to be a sub-interface.

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -76,6 +76,34 @@ def monitor_sessions(self: AvdStructuredConfigBaseProtocol) -> None:
...
+        if "." not in interface_name:
+            return
+        if monitor_session.role == "destination":
+            msg = (
+                "Port mirroring destination on sub-interface is not supported. "
+                f"Got 'role: {monitor_session.role}' for monitor session '{monitor_session.name}' "
+                f"on interface '{interface_name}' under {context}."
+            )
+            raise AristaAvdInvalidInputsError(msg)
+        if monitor_session.source_settings.direction != "rx":
```
laxmikantchintakindi, line 95:
```suggestion
        if monitor_session.source_settings.direction and monitor_session.source_settings.direction != "rx":
```

laxmikantchintakindi: When direction is not set at all, still I am seeing this error.
```
"Only 'direction: rx' supported on sub-interfaces for monitor session. Got 'direction: None' for monitor session 'Monitor_session_on_sub_interface' on interface 'Ethernet7.7' under tenant_a[name=Tenant_A].vrfs[name=Tenant_A_WAN_Zone].l3_interfaces[1].
```

laxmikantchintakindi: Is `source_setting.direction` must be set when role is `source`?

MaheshGSLAB: For sub interface if you do not define the direction it still raise same error on EOS.
```
veos-4-35-3F(config-s-mahesh)#monitor session ads source Ethernet 1.1 
% Subintf source can only be used with ingress mirroring (RX direction)
% Invalid input
```
So yes even if you do not define the the direction the direction is tx expected. Direction is required as tx for only subinterface. For a physical interface it is not required

alexeygorbunov (approved):
LGTM

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
MaheshGSLAB: but i can give network switch_ports as sub-interface and its not restricted via schema or code.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_TENANTS_NETWORKS/Tenant_A.yml
```diff
@@ -211,6 +211,18 @@ tenant_a:
                 role: destination
                 session_settings:
                   encapsulation_gre_metadata_tx: true
+          # Test monitor session on subinterface
+          - interfaces: [Ethernet7.7]
```
laxmikantchintakindi, line 215: I think optimal way of adding the subinterfaces to connected-endpoints is setting `subinterfaces` key under connected endpoints. @ClausHolbechArista @gmuloc

gmuloc: what's the question?

laxmikantchintakindi: Do we support giving the subinterface name directly in interfaces list? We have subinterfaces key for subinterface configuration in connected_endpoints.

gmuloc: so `subinterfaces` should be used indeed in this model if you want to configure a subinterface not directly in the `switch_ports` list, though today we are not blocking it.
We should add some guard as a follow up to https://github.com/aristanetworks/avd/pull/5417.
Can we please track an issue to prevent subinterfaces being configured on the switch port lists in the future if there is not one already? (can be done via a future knob)

MaheshGSLAB: Then to configure monitor session on sub-interface(for direction tx) we need to iterate over subinterfaces which is missing in code right now. Should we raise an issue for the same?

gmuloc: so I think it means right now this session is only supported on main interfaces right?

laxmikantchintakindi: Yes. The session is configured for main interface only. Can we add the same guard to the network_ports? Sub-interface is set now using switch_ports option but practically is it possible to have subinterface network ports?

gmuloc: the models are not made for subinterfaces really in the first place (too many missing config) so yeah we could add the same guard I think except if I am missing something obvious. @ClausHolbechArista

MaheshGSLAB: Since we are not blocking the sub-interface on switch_ports we can still configure the monitor session with this way. Once we add the restriction on code to not provide the sub-interface on switch_ports we need to extend the support to render the monitor session config via subinterfaces. So may be a separate PR where we handle all this.

ClausHolbechArista: We don't need to implement protections against misuse of the models everywhere. Using subinterface under network ports would not work, since we render the `switchport` config which is not valid in that context on EOS. So we are discussing something that would not work anyway. Please just focus on the areas mentioned in the issue and any obvious similar places. It has to be places where we actually support subinterfaces.

MaheshGSLAB: as discussed an user should not give subinterface on switch_ports under connected_endpoints/network_ports . We should use the subinterfaces key for the same hence removing the error code for connected_endpoints/network_ports.

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -106,6 +130,7 @@ def _monitor_session_configs(
                         continue
 
                     ethernet_interface_name = adapter.switch_ports[node_index]
+                    context = adapter._internal_data.context
```
laxmikantchintakindi:
```suggestion
```

python-avd/pyavd/_eos_designs/structured_config/base/monitor_sessions.py
```diff
@@ -76,6 +76,30 @@ def monitor_sessions(self: AvdStructuredConfigBaseProtocol) -> None:
 
             self.structured_config.monitor_sessions.append(monitor_session)
 
+    @staticmethod
+    def _validate_monitor_session_on_subinterface(
+        interface_name: str,
+        monitor_session: (EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem.L3InterfacesItem.MonitorSessionsItem),
```
laxmikantchintakindi:
```suggestion
        monitor_session: EosDesigns._DynamicKeys.DynamicNetworkServicesItem.NetworkServicesItem.VrfsItem.L3InterfacesItem.MonitorSessionsItem,
```

laxmikantchintakindi (approved):

Shivani-gslab (approved):
LGTM

alexeygorbunov (approved):

PR #6839, opened 2026-04-13
Feat(pyavd): Add CV ID models and update CVContainer to follow input schema

carl-baillargeon:
## Change Summary

- Added CV ID models with cached UUID to improve performance
- Fix existing CVContainer class to follow the input schema

## Component(s) name

PyAVD

## Proposed changes

- Added CV ID models with cached UUID
- Fix existing CVContainer class to follow the input schema

## How to test

Unit tests were updated.
Internal CI is green: https://github.com/aristanetworks/avd-internal/actions/runs/24397472961/job/71259231377

### User Checklist

- N/A

### Repository Checklist


ClausHolbechArista (approved):

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -379,6 +387,28 @@ def from_dict(cls, data: dict[str, Any]) -> AvdManifest:
...
+    @cached_property
+    def value(self) -> str:
+        return f"{AVD_ENTITY_PREFIX}{uuid5(AVD_NAMESPACE, self.name)}"
+
+
+@dataclass(frozen=True)
+class CVContainerID:
+    """CloudVision container ID based on hierarchy path with cached UUID."""
+
+    path: tuple[str, ...]
```
alexeygorbunov, line 405: Will we use this `path` attribute other than for generating `value` property? Just wondering if we need it to be `tuple[str]` as when we instantiate `CVContainerID` - we split `current_path` (which is string) and convert it to tuple. Then during calculation of `value` here we join tuple items back to string (original `curent_path`). Wondering if we can just use plaing `str` and avoid splittign and joining

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -395,7 +425,7 @@ def from_avd_manifest(cls, avd_manifest: AvdManifest) -> CVManifest:
         # Create all CVConfiglet objects first.
         for avd_configlet in avd_manifest.configlets:
             cv_configlet = CVConfiglet(
-                avd_configlet=avd_configlet, id=cls._generate_deterministic_id(avd_configlet.name), description="Configlet created and uploaded by AVD."
+                avd_configlet=avd_configlet, id=CVConfigletID(name=avd_configlet.name), description="Configlet created and uploaded by AVD."
```
alexeygorbunov, line 428: Is this really giving a performance advantage? Instead of accessing base string-type variable we need to access string-type cached attribute of a dataclass instance?

carl-baillargeon:
This PR was needed to improve merge performance for per-device manifest which is no longer relevant. We will revisit later.

PR #6846, opened 2026-04-15
Feat(eos_designs): Preview: Containerlab Digital Twin

### LINKED ISSUE #3379: eos_designs output containerlab topology
### Enhancement summary

It would be nice if eos_designs generated a cotainerlab topology artifact

eos_designs

### Use case example

If you wanted to pre-deploy test config, it would be useful for a containerlab topology to be created.

jinja template to ouput a containerlab file

- leverage containerlab [generate command ](https://containerlab.dev/cmd/generate/)
- leverage [generated topologies](https://containerlab.dev/manual/topo-def-file/#generated-topologies)

### Additional context

_No response_

### Contributing Guide

ankudinov:
## Change Summary

Preview: Containerlab Digital Twin

## Related Issue(s)

Fixes #3379 

## Component(s) name

`arista.avd.eos_designs`

## How to test
use AVD playground to generate Containerlab digital twing

### Repository Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Containerlab Digital Twin support generating lab topology, per-device startup-config artifacts, and an interface mapping file.
  * Introduced new Digital Twin Containerlab Molecule scenarios (positive and negative) with added inventory/group variables for multi-node labs.
* **Bug Fixes**
  * Improved change detection/`changed` reporting for Containerlab startup-config output generation.
* **Documentation**
  * Expanded Digital Twin documentation and schemas to include the Containerlab environment and its constraints.
* **Tests**
  * Added/extended Molecule and unit test coverage plus CI scenario matrix updates, including validation of expected failure cases.


ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
```diff
@@ -37,6 +37,14 @@
         tmp_dir: "{{ eos_cli_config_gen_tmp_dir }}"
         config_filename: "{{ eos_config_dir }}/{{ inventory_hostname }}.cfg"
         documentation_filename: "{{ devices_dir }}/{{ inventory_hostname }}.md"
+        containerlab_startup_config_filename: >-
```
gmuloc: @ClausHolbechArista should we start enforcing the role prefix naming convention for new role variables so we start getting compliant with RedHat latest rules?

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -277,6 +375,8 @@ def _get_digital_twin_act(fabric_documentation_facts: FabricDocumentationFacts)
             ActLinkSettings(
                 connection=(f"{topology_link['node']}:{topology_link['node_interface']}", f"{topology_link['peer']}:{topology_link['peer_interface']}")
             )
+            for topology_link in fabric_documentation_facts.topology_links
+            if _is_p2p_link(topology_link)
             for topology_link in verified_topology_links
```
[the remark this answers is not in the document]
ankudinov: not related to this PR

ansible_collections/arista/avd/extensions/molecule/digital_twin/digital_twin/documentation/fabric/DIGITAL_TWIN-topology.yml
```diff
@@ -133,3 +134,129 @@ links:
...
+  - digital-twin-ethernet-ports-3:Ethernet6
+- connection:
+  - digital-twin-ethernet-ports-1:Ethernet11
+  - digital.twin.ethernet.ports.2:Ethernet11
+- connection:
+  - digital-twin-ethernet-ports-1:Ethernet14
+  - digital.twin.ethernet.ports.2:Ethernet14
+- connection:
+  - digital-twin-default-interfaces-digital-twin-platform:Ethernet3
+  - digital-twin-default-interfaces-original-platform:Ethernet2
```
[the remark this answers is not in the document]
ankudinov: not related to this PR

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -113,11 +127,29 @@ def main(self, task_vars: dict[str, Any]) -> None:
...
+
+                content["topology"]["nodes"] = {
+                    node_name: {
+                        "mgmt-ipv4": node_settings["mgmt-ipv4"],
+                        "startup-config": f"init-configs/{node_name}.cfg",
+                    }
+                    for node_name, node_settings in content["topology"]["nodes"].items()
+                }
+                # add keys in a very specific order - name, prefix, everything else
+                content = {"name": content["name"], "prefix": output.digital_twin.prefix, **{key: value for key, value in content.items() if key != "name"}}
```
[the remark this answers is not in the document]
ankudinov: Attempted to fix this and wasn't happy code readability. The fix significantly increases complexity. If there is an easy one - please suggest. Overall I don't think it's "major" as potential issues will be caught by schema, etc.

python-avd/pyavd/get_fabric_documentation.py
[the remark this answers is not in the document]
ankudinov: yes, open an issue and assign @alexeygorbunov for review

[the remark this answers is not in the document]
gmuloc: this needs to be addressed in this PR but there is an overlap between verified_topology_links and the PR introdiuce _is_p2p_link, only one must stay.

@coderabbitai please chime in 🐰

ansible_collections/arista/avd/extensions/molecule/digital_twin/digital_twin/documentation/fabric/DIGITAL_TWIN-topology.yml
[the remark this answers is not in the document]
gmuloc: meh a rabbit should not agree to everything but think critically @coderabbitai

ansible_collections/arista/avd/roles/eos_designs/docs/data-models.md
```diff
@@ -1921,6 +1922,180 @@ ansible_collections/arista/avd/roles/eos_designs/docs/tables/digital-twin-config
...
+- Only out-of-band management is supported.
+- Only a single IPv4 management subnet is supported across the lab.
+- Management IP addresses in the lab and production must currently match.
+- Each node must have a static `mgmt_ip`. Unset management IPs and `mgmt_ip: dhcp` are not supported.
+- `digital_twin.mgmt_ip` is not currently used by the Containerlab topology generator. Instead, topology generation uses each node `mgmt_ip` value from the AVD fabric facts.
+- The generated topology includes fabric nodes and point-to-point fabric links. Connected endpoints and non-point-to-point links are not modeled.
+- The generated topology sets `prefix: ""` intentionally to preserve the original AVD hostnames.
+- The generated topology currently defaults to `arista_ceos` with `image: arista/ceos:latest`. Please set up your lab environment accordingly, or update the generated topology file if needed.
+
+> WARNING: Make sure your lab environment is isolated from production. The management addresses are reused, so take the necessary precautions to avoid pushing configurations to the wrong environment.
```
gmuloc: use admonititions here

```
!!! warning
   ...
```

gmuloc:
need to the new scenarios to https://github.com/aristanetworks/avd/blob/fe7a3281f02ea7763e471938f44a74c75a820596/.github/workflows/pull-request-management.yml#L268

ansible_collections/arista/avd/extensions/molecule/digital_twin/digital_twin/documentation/fabric/DIGITAL_TWIN-topology.yml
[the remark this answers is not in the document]
ankudinov: fixed. and yes it was duplicate due to duplicate code introduced by 2 different PRs to do the same thing

ankudinov:
> need to the new scenarios to
> 
> https://github.com/aristanetworks/avd/blob/fe7a3281f02ea7763e471938f44a74c75a820596/.github/workflows/pull-request-management.yml#L268

done

ansible_collections/arista/avd/roles/eos_designs/docs/data-models.md
ankudinov: done

ansible_collections/arista/avd/roles/eos_cli_config_gen/tasks/main.yml
ClausHolbechArista: we should not have this variable at all. The eos_cli_config_gen should be run for _either_ digital twin or production. And if you need a smaller startup config, you should just provide a smaller structured config

ankudinov: Removed the variable, simplified the logic and created #7297 to follow-up on limitations.

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+    # find Containerlab mgmt network and raise and error if nodes are not in the same subnet
+    unique_mgmt_networks = {ip_network(mgmt_ip, strict=False) for _, mgmt_ip in mgmt_ips}
+
+    if len(unique_mgmt_networks) > 1:
+        mgmt_networks = ", ".join(f"{network}" for network in unique_mgmt_networks)
+        msg = f"Containerlab Digital Twin requires all node management IPv4 addresses to belong to the same subnet. Found multiple subnets: {mgmt_networks}."
+        raise AristaAvdError(msg)
+
+    return ContainerlabDigitalTwin(
+        name=f"{fabric_documentation_facts.fabric_name}, Containerlab Digital Twin",
```
alexeygorbunov: This name is not complian with containerlab schema regex pattern (`^[a-zA-Z0-9][a-zA-Z0-9-._]*$`). 

```
String does not match the pattern of "^[a-zA-Z0-9][a-zA-Z0-9-._]*$".yaml-schema: Containerlab
Containerlab
topology name

Source: clab.schema.json
```

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+        prefix="",
+        mgmt=ContainerlabMgmt(network="custom_mgmt", ipv4_subnet=str(next(iter(unique_mgmt_networks)))),
+        topology=ContainerlabTopology(
+            defaults=ContainerlabDefaults(kind=default_kind),
+            kinds={
+                default_kind: ContainerlabKind(
+                    enforce_startup_config=True,
+                    image="arista/ceos:latest",
```
alexeygorbunov, line 250: We hardcode this assuming that all users will import ceos image naming it `arista/ceos`. Should we provide a flexibility to override this if needed (maybe they want a different version or they named it differently)?

For ACT we provide the following key which we later take into account during topology generaiton:

```
digital_twin.fabric.act_os_version
<node_type_keys.key>.defaults.digital_twin.act_os_version
```

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+
+    The returned object contains the minimal information required to render
+    Containerlab nodes with management addresses under `topology.nodes` and
+    inter-switch links under `topology.links`.
+    """
+    from pyavd._errors import AristaAvdError  # noqa: PLC0415
+
+    sorted_avd_facts = sorted(fabric_documentation_facts.avd_facts.items())
+
+    unsupported_devices = [device for device, facts in sorted_avd_facts if not facts.mgmt_ip or facts.mgmt_ip == "dhcp"]
```
alexeygorbunov, line 204: Devices may have inband mgmt or use DHCP on OOB MGMT interface. This may bee very restrictive to raise in this case.

I would suggest to use `digital_twin.mgmt_ip` (if defined) for Clab and fallback to regular mgmt_ip.

These override keys are currently exposed at the following levels:

```
"<node_type_keys.key>.defaults.digital_twin.mgmt_ip"
"<node_type_keys.key>.node_groups.[].nodes.[].digital_twin.mgmt_ip"
"<node_type_keys.key>.node_groups.[].digital_twin.mgmt_ip"
"<node_type_keys.key>.nodes.[].digital_twin.mgmt_ip"
"device_profiles.[].digital_twin.mgmt_ip"
"devices.[].digital_twin.mgmt_ip"
```

alexeygorbunov: Logic handling calculation of the oob_mgmt_ip should be extended to include containerlab DT case (right now it only reacts on ACT mode)

alexeygorbunov: This should solve the problem with regular `mgmt_ip`s sitting in the different subnets (currently this would always lead to raising)

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+    unique_mgmt_networks = {ip_network(mgmt_ip, strict=False) for _, mgmt_ip in mgmt_ips}
+
+    if len(unique_mgmt_networks) > 1:
+        mgmt_networks = ", ".join(f"{network}" for network in unique_mgmt_networks)
+        msg = f"Containerlab Digital Twin requires all node management IPv4 addresses to belong to the same subnet. Found multiple subnets: {mgmt_networks}."
+        raise AristaAvdError(msg)
+
+    return ContainerlabDigitalTwin(
+        name=f"{fabric_documentation_facts.fabric_name}, Containerlab Digital Twin",
+        prefix="",
```
alexeygorbunov, line 243: Should we expose this to the user as well? Name overlapping for prod switches is probably not possible (so running Clab DT for multiple prod fabrics on the same host should be ok) but if user tries to generate containerlab topos for our example fabrics and run them on the same host then there will be container name collision

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+        mgmt=ContainerlabMgmt(network="custom_mgmt", ipv4_subnet=str(next(iter(unique_mgmt_networks)))),
```
alexeygorbunov, line 244: `network="custom_mgmt"` - wouldn't this lead to the deployment failure if we try to run multiple clab topologies on the same host? If definition of the bridge matches - we risk to have ceos mgmt IP collisions. If ipv4-subnets differ for different topologies - there we'll probably face deployment error.

We should probably prepend fabric name to the bridge name, or even make this configurable by the user?

ansible_collections/arista/avd/tests/unit/action/test_eos_designs_documentation.py
```diff
@@ -41,6 +52,53 @@
...
+    class ContainerlabDigitalTwin:
+        mgmt: ContainerlabMgmt
+        topology: ContainerlabTopology
+
+    data = ContainerlabDigitalTwin(
+        mgmt=ContainerlabMgmt(network="clab-mgmt", ipv4_subnet="172.16.1.0/24"),
+        topology=ContainerlabTopology(
+            nodes={1: ContainerlabNode(mgmt_ipv4="172.16.1.101")},
+            kinds=(ContainerlabKind(enforce_startup_config=True, image="ceos:latest"),),
+            endpoint_lists=[("leaf1:eth1", "spine1:eth1")],
```
alexeygorbunov, line 88: Pylance is not happy about reusing the same class names that we import from pyavd

```
Argument of type "ContainerlabMgmt" cannot be assigned to parameter "mgmt" of type "ContainerlabMgmt" in function "__init__"
  "avd.ansible_collections.arista.avd.tests.unit.action.test_eos_designs_documentation.ContainerlabMgmt" is not assignable to "pyavd.api.fabric_documentation.ContainerlabMgmt"
```

Maybe we should either use actual production clab classes for this test or rename locally-redefined so that the don't overlap with imported ones?

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
```diff
@@ -114,11 +128,29 @@ def main(self, task_vars: dict[str, Any]) -> None:
...
+                        content=json.dumps(interface_mapping, indent=4) + "\n",
+                        filename=str(Path(validated_args["digital_twin_file"]).parent / "interface_mapping.json"),
+                        file_mode=validated_args["mode"],
+                    )
+                    self.result["changed"] = self.result["changed"] or changed
+
+                content["topology"]["nodes"] = {
+                    node_name: {
+                        "mgmt-ipv4": node_settings["mgmt-ipv4"],
+                        "startup-config": f"intended/configs/{node_name}.cfg",
```
alexeygorbunov, line 146: If we are using hardcoded path to the startup config, why not to set this when we instantiate dataclasses (meaning `ContainerlabNode` would just have another field `startup_config` that we'll set together with mgmt_ipv4). Not sure what is the benefit of manipulating this here in plugin instead of setting it in the beginning

ansible_collections/arista/avd/extensions/molecule/digital_twin_containerlab/verify.yml
```diff
@@ -0,0 +1,33 @@
...
+        path: "{{ root_dir }}/digital_twin/{{ fabric_name }}-topology.clab.yml"
+      register: containerlab_topology_file
+      run_once: true
+
+    - name: Read generated Containerlab topology file
+      delegate_to: 127.0.0.1
+      ansible.builtin.slurp:
+        path: "{{ root_dir }}/digital_twin/{{ fabric_name }}-topology.clab.yml"
+      register: containerlab_topology_content
+      run_once: true
```
alexeygorbunov: Not sure what is the purpose of this task as registered var is never used

ansible_collections/arista/avd/extensions/molecule/digital_twin_containerlab/digital_twin/intended/configs/dc1-leaf1a.cfg
```diff
@@ -0,0 +1,70 @@
...
+interface Loopback0
+   description ROUTER_ID
+   no shutdown
+   ip address 10.255.1.1/32
+!
+interface Management1
+   description OOB_MANAGEMENT
+   no shutdown
+   vrf MGMT
+   ip address 172.16.1.101/24
```
alexeygorbunov, line 45: This Digital twin will likely be running on-prem (meaning original OOB MGMT subnet would likely be routable from the clab runner) and nodes will be reusing prod hostnames, mgmt IPs, credentials, etc. This may create a risk of pushing something unexpected to the prod.

python-avd/pyavd/get_fabric_documentation.py
```diff
@@ -159,10 +167,100 @@ def _get_digital_twin(fabric_documentation_facts: FabricDocumentationFacts) -> A
...
+                    enforce_startup_config=True,
+                    image="arista/ceos:latest",
+                    binds=("interface_mapping.json:/mnt/flash/EosIntfMapping.json:ro",),
+                )
+            },
+            nodes=nodes,
+            links=tuple(links),
+        ),
+        interface_mapping={
+            "ManagementIntf": {"eth0": "Management1"},
```
alexeygorbunov, line 258: If `node_config.mgmt_interface` is set to Ma0 (or anything other than Ma1) then generated EOS config will have Ma0 (and not Ma1 expected by Clab and mapped to Linux eth0) which i guess may lead to mgmt connectivity issues. We probably need a special section under shared_utils.mgmt_interface to enforce OOB MA to be Ma1

ansible_collections/arista/avd/extensions/molecule/digital_twin_containerlab/digital_twin/DIGITAL_TWIN_CONTAINERLAB-topology.clab.yml
```diff
@@ -0,0 +1,41 @@
+---
```
alexeygorbunov, line 1: Would pointing language server to the schema definition be helpful with making sure all generated key are schema-compliant?

`# yaml-language-server: $schema=https://raw.githubusercontent.com/srl-labs/containerlab/main/schemas/clab.schema.json`

ansible_collections/arista/avd/plugins/action/eos_designs_documentation.py
alexeygorbunov: Also, this path hardcodes to `intended/configs/{node_name}.cfg` assuming that output_dir_name is always `intended`. But if it is not the case - path under `starup-config` in topology file may become unreachable

ansible_collections/arista/avd/extensions/molecule/digital_twin/digital_twin/documentation/fabric/DIGITAL_TWIN-topology.yml
joelbreton2, line 1: why do we have a digital_twin folder inside a digital_twin folder

python-avd/pyavd/get_fabric_documentation.py
joelbreton2: I'm getting the same issue

ansible_collections/arista/avd/extensions/molecule/digital_twin/digital_twin/documentation/fabric/DIGITAL_TWIN-topology.yml
alexeygorbunov: Outer digital_twin references the name of the scenario/fabric that we picked (it could have been anything else overall). Inner digital_twin is a dedicated folder containing digital_twin artifacts (at lest for ACT mode) but is also tunable (via output_dir_name and documentation_dir_name vars)

python-avd/pyavd/get_fabric_documentation.py
JulioPDX: I agree with this. Especially when users want to implement this functionality internally, we are now relying on the end user to add an additional workflow to replace arista/ceos:latest, with whatever is used internally.

JulioPDX:
Any thought to create generic names that could be used across DT environments. For example instead of act_username, we would have digital_twin_usename, or have a heirarchy of `ansible_user, first defined local_users, digital_twin_username`. this would prevent the creation of multiple keys to change values in the different topology files.

python-avd/pyavd/get_fabric_documentation.py
ankudinov: Let's not explode the scope of this PR. Overriding mgmt IP is certainly important, however I'd love to add basics first. It will also take a bit more than address override. Currently cLabs only supports single subnet for management. https://github.com/srl-labs/containerlab/pull/3237 can help. But I have to craft and test a working environment before adding any code. If you already have one - share.

python-avd/pyavd/get_fabric_documentation.py
ankudinov: No. The reason is - the prefix will change lab hostnames and we don't want to do that for the current implementation. This may change though once we have the full story with mgmt ip override, etc. Let's keep it clean and simple at the start - same mgmt IP, same hostname.

python-avd/pyavd/get_fabric_documentation.py
ankudinov: That's not too critical, but let's keep schema happy and use Containerlab_Digital_Twin. Suggest alternative name if you have something in mind.

ankudinov: in fact, let's keep fabric name only

python-avd/pyavd/get_fabric_documentation.py
ankudinov: no. we should not aim to support all possible deployment cases and expect clean environment every time. This can be discussed later, but will not be implemented in this PR

python-avd/pyavd/get_fabric_documentation.py
ankudinov: Agree in principle. However let's move this to another PR. I'd prefer to have a discussion around joint ACT/cLab schema first before blindly adding `containerlab_image` on top of `act_os_version` as yet another knob. To avoid cases like `fabric: {}`
But absolutely has to be configurable in the future.

python-avd/pyavd/get_fabric_documentation.py
ankudinov: Yes, I though about it and taking management from node settings. However what if every node will have it's own custom mgmt interface? In that case we have to build int map for every node. This is not a big problem, but adds complexity and int map was added only to merge the PR a bit quicker and support "older" cEOSlab versions. Long term we want to use INTF_MAP_ETH0 env var which is coming very soon and set it to node mgmt interface.

ankudinov:
> hought to create generic names that could be used across DT environments. For example instead of act_username, we would have digital_twin_usename, or have a heirarchy of `ansible_user, first def`

We certainly have to brainstorm a unified DT schema. Different PR though.

ansible_collections/arista/avd/extensions/molecule/digital_twin_containerlab/verify.yml
ankudinov: Great find. Removed. Thank you.

PR #6864, opened 2026-04-20
Test: Added the test file for eos_cli_config_gen plugin action

MaheshGSLAB:
## Change Summary

Added the test file for eos_cli_config_gen plugin action

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
Added the test file for eos_cli_config_gen plugin action.

Both eos_cli_config_gen.py and verify_requirements.py use the same logger instance because python logging.getLogger() is a global registry.

When pytest collected test_eos_cli_config_gen.py, python imported eos_cli_config_gen.py, which immediately set propagate=False on the shared logger. From that point on, for the entire test session, any LOGGER.debug() call in verify_requirements.py was swallowed by the named logger hence records never reached the root logger where pytest's caplog handler lives. So caplog.text stayed ''.

Removed LOGGER.propagate = False from module level and moved it inside setup_module_logging(), where it semantically belongs. it only makes sense to disable propagation after the custom handler is attached.

This means:
1. At import time: logger has default propagate=True -> caplog works in tests
2. At runtime when Ansible task executes: setup_module_logging() is called first, sets propagate=False -> no duplicate logs in production (same protection, better placement)

## How to test
CI should be green

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -222,5 +217,7 @@ def setup_module_logging(hostname: str, result: dict) -> None:
     python_to_ansible_handler = PythonToAnsibleHandler(result, display)
     python_to_ansible_handler.addFilter(python_to_ansible_filter)
     LOGGER.addHandler(python_to_ansible_handler)
+    # Avoid duplicate logs now that a dedicated handler is attached
+    LOGGER.propagate = False
```
gmuloc: I read through the PR comment and while this has no impact because we are not emitting any logs between the import of the module and the moment when we run this, it is technically still a change that could have an impact given that:

before: proppagte false is at import so any logs between import and running the module don't go up to root logger

after: propagate false is set only when the setup_logging_module is called and so logs could potentially trickle in.

I am not seeing any issue with this module really and don't think it will impact anything else but it _could_

I would rather this is not touched and we make sure our tests are doing the propragtation proprely. my friend is suggesting this fixture:

```
  # conftest.py
  import logging
  import pytest

  @pytest.fixture(autouse=True)
  def reset_avd_logger():
      logger = logging.getLogger("ansible_collections.arista.avd")
      original_propagate = logger.propagate
      original_handlers = logger.handlers[:]
      yield
      logger.propagate = original_propagate
      logger.handlers = original_handlers
```

That being said, this fix seems harmless in this case so I am ok to go with it.

cc @carl-baillargeon

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+def get_action_module() -> ActionModule:
+    """Create an ActionModule instance bypassing ActionBase.__init__."""
+    module = ActionModule.__new__(ActionModule)
+    module._task = MagicMock()
+    module._connection = MagicMock()
+    module._play_context = MagicMock()
+    module._loader = MagicMock()
+    module._templar = MagicMock()
+    module._shared_loader_obj = MagicMock()
+    return module
```
gmuloc:
```suggestion
def get_action_module(task_args: dict | None = None) -> ActionModule:
      """Create an ActionModule instance via the real constructor with mocked Ansible internals."""
      mock_task = MagicMock()
      mock_task.args = task_args or {}
      mock_task.async_val = False
      mock_task.check_mode = False

      return ActionModule(
          task=mock_task,
          connection=MagicMock(),
          play_context=MagicMock(),
          loader=MagicMock(),
          templar=MagicMock(),
          shared_loader_obj=MagicMock(),
      )
```

you could make this a fixture

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+# ===========================
+# Tests for setup_module_logging
+# ===========================
+
+
+class TestSetupModuleLogging:
```
gmuloc: I think we generally don't use this pattern of classes when we test flat functions - what factors made you choose this? 
I would rather we follow the same style everywhere so it is easier to understand the test code base as whole

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+        """Test that setup_module_logging adds a PythonToAnsibleHandler to the AVD logger."""
+        with (
+            patch(f"{MODULE_PATH}.PythonToAnsibleContextFilter") as mock_filter_cls,
+            patch(f"{MODULE_PATH}.PythonToAnsibleHandler") as mock_handler_cls,
+            patch(f"{MODULE_PATH}.LOGGER") as mock_logger,
+        ):
+            mock_filter_instance = MagicMock()
+            mock_handler_instance = MagicMock()
+            mock_filter_cls.return_value = mock_filter_instance
+            mock_handler_cls.return_value = mock_handler_instance
```
gmuloc: I don't think this is useful.
My understanding is that `mock_filter_cls.return_value` is already an instance of MagickMock

```python
>>> from unittest.mock import MagicMock, patch
>>> class A():
...     def __init__(self):
...         self.a = 42
...
>>> with patch("__main__.A") as a_patched_cls:
...    a_patched_instance = a_patched_cls()
...    print(type(a_patched_instance))
...    instance = a_patched_cls()
...    print(instance is a_patched_cls.return_value)
...
<class 'unittest.mock.MagicMock'>
True
```



```suggestion
```

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def test_uses_hostname_for_filter(self) -> None:
+        """Test that the hostname is passed to PythonToAnsibleContextFilter."""
+        with (
+            patch(f"{MODULE_PATH}.PythonToAnsibleContextFilter") as mock_filter_cls,
+            patch(f"{MODULE_PATH}.PythonToAnsibleHandler"),
+            patch(f"{MODULE_PATH}.LOGGER"),
+        ):
+            setup_module_logging("my-spine-1", {})
+            mock_filter_cls.assert_called_once_with("my-spine-1")
```
gmuloc: this is already tested in the previous test (line 53) what does this test bring?

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
gmuloc: These test will not survive post refactoring so they are not going to be helpful actually.

What we would rather want are tests that show the colllected logs so that we can check this after a migration


The tests here will not allow us to achieve our goal: making sure we have the same log before and afyter migration

tpday we have in main:

```
  LOGGER.debug("Validating task arguments...")
  LOGGER.debug("Validating task arguments [done].")
  LOGGER.debug("Loading structured config...")
  LOGGER.debug("Loading structured config [done].")
  LOGGER.debug("Rendering configuration...")
  LOGGER.debug("Rendering configuration [done].")
  LOGGER.debug("Rendering documentation...")
  LOGGER.debug("Rendering documentation [done].")
```

what we need is a test that capture these and when we migrate making sure the same logs are being captured at the same level.

so you will need to write tests executing te main function to capture the logs.

The tests around the actual handler being set is kind of less relevant as all of this will be thrown away post migration so really it is not useful for our target

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+            setup_module_logging("host", result)
+            mock_handler_cls.assert_called_once_with(result, mock_display)
+
+
+# ===========================
+# Tests for error handling
+# ===========================
+
+
+class TestLoadStructuredConfigErrors:
```
gmuloc: for these test post migration we would expect the exception type to change (some plain Exception) and they would be recast by the base class to AnsibleActionFail so they are interesting to keep

gmuloc: actually not sure what the two tests are bringing - I feel like you can reduce it to one test only given that one can never happen between the other.

Same however please do not keep the class make it just a def

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
gmuloc: probably worth having parameters that enable or disable config /doc generation to have multiple usecase we can capyture

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+        with (
+            patch.object(module, "validate_args", return_value=validated_args),
+            patch.object(module, "load_structured_config", return_value={}),
+            patch(f"{MODULE_PATH}.get_device_config", side_effect=RuntimeError("pyavd exploded"), create=True),
+            patch(f"{MODULE_PATH}.raise_action_fail") as mock_raise_fail,
+        ):
+            module.main("test-device", {}, {})
+
+        mock_raise_fail.assert_called_once()
+        assert "pyavd exploded" in str(mock_raise_fail.call_args)
```
gmuloc:
```suggestion
  with (
      patch.object(module, "validate_args", return_value=validated_args),
      patch.object(module, "load_structured_config", return_value={}),
      patch(f"{MODULE_PATH}.get_device_config", side_effect=RuntimeError("pyavd exploded"), create=True),
      pytest.raises(AnsibleActionFail, match="pyavd exploded"),
  ):
      module.main("test-device", {}, {})
```

This would work the same but in anycase this would not be useful post migration I think as the cental plugin will do the recast and we have a test

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+class TestRunErrors:
+    def test_raises_when_pyavd_not_installed(self) -> None:
```
gmuloc: keep this one

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
gmuloc: this willnot survive the rewrite but we can keep it for now

gmuloc (changes requested):
The most important work here to prepare for the mirgation is to make sure we test the various logging output and conserve them post migration so you should focus on this.

Thiings like implementation details (do I have the right handler called the right way) will be things difficult to assess after.

What is also important is to check that the behavior if having the device name in the logs message present now will still be present post migration in display

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
carl-baillargeon: I am also ok to go with this fix since it seems harmless.

MaheshGSLAB: added conftest file

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: added

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: yes we dont use this pattern but i just tried to see if this can improve the readability and contribute. revert back with old way

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: removed

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: removed

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: updated the testcase

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: updated with one

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: updated

ansible_collections/arista/avd/tests/unit/action/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+# that can be found in the LICENSE file.
+from __future__ import annotations
+
+import logging
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def reset_avd_logger() -> None:
```
Shivani-gslab: pyright is throwing return type error, we can update to -

```
from collections.abc import Generator

@pytest.fixture(autouse=True)
def reset_avd_logger() -> Generator[None, None, None]:
```

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def test_run_raises_when_pyavd_not_installed(action_module: Callable[..., ActionModule]) -> None:
+    """Test that AnsibleActionFail is raised immediately when pyavd is missing."""
+    module = action_module()
+
+    with (
+        patch(f"{MODULE_PATH}.HAS_PYAVD", new=False),
+        patch("ansible.plugins.action.ActionBase.run", return_value={}),
+        pytest.raises(AnsibleActionFail, match="pyavd"),
```
Shivani-gslab: I think, this is too loose and could match anything containing `pyavd` string, lets make in stricter .
may be `"requires the 'pyavd' Python library"`

Shivani-gslab (changes requested):

ansible_collections/arista/avd/tests/unit/action/conftest.py
MaheshGSLAB: updated

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
MaheshGSLAB: updated

Shivani-gslab (approved):
LGTM

ansible_collections/arista/avd/tests/unit/action/conftest.py
```diff
@@ -0,0 +1,22 @@
...
+
+import pytest
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+
+@pytest.fixture(autouse=True)
+def reset_avd_logger() -> Generator[None, None, None]:
+    logger = logging.getLogger("ansible_collections.arista.avd")
```
gmuloc, line 17: also keep the orignal level here (we are setting it to debug in the action plugin)

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
gmuloc, line 105: you don't have tests that capture these logs - can you please add them?

gmuloc (changes requested):

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
MaheshGSLAB: added

ansible_collections/arista/avd/tests/unit/action/conftest.py
MaheshGSLAB: updated

Vibhu-gslab (approved):

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
ClausHolbechArista: IMO we are changing production code to make our tests pass, which seems wrong to me.

ClausHolbechArista: You already added the fixture suggested by @gmuloc, but you are not using it so you still need this to make the tests pass.

ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```diff
@@ -0,0 +1,309 @@
...
+                "Rendering documentation...",
+                "Rendering documentation custom templates...",
+                "Rendering documentation custom templates [done].",
+                "Rendering documentation [done].",
+            ],
+            id="config_and_doc_with_custom_templates",
+        ),
+    ],
+)
+def test_main_emits_expected_debug_logs(
```
ClausHolbechArista: you are repeating the same tests for logging and then for ansible display. I think with a few extra assertion lines it could be done in a single go.

MaheshGSLAB: ok updated with one

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
MaheshGSLAB: With the current fixture, it saves and restores state but doesn't reset it before the test runs. LOGGER.propagate = False is set at module-import time, so by the time caplog tries to capture, the logger won't propagate to the root handler. I need to actively set propagate=True and clear handlers before yielding.

ClausHolbechArista (approved):

gmuloc (approved):

PR #6874, opened 2026-04-22
Refactor(eos_designs, eos_cli_config_gen): Used pyavd-utils instead of cryptography lib for password utils and filter

### LINKED ISSUE #6843: AVD should use pyavd-utils for password functionality
- Change password filters / password utils to leverage pyavd-utils instead of other libraries or custom implementations.
- Remove dependency on cryptography if this was the only use.

MaheshGSLAB:
## Change Summary

Used pyavd-utils instead of other lib for password utils and filter

## Related Issue(s)

Fixes #6843

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
Removed the dependency of cryptography and used pyavd-utils instead of cryptography for password utils and filter.

## How to test
CI should be green

### User Checklist

- N/A

### Repository Checklist


python-avd/tests/pyavd/j2filters/test_secure_hash.py
```diff
@@ -60,7 +60,7 @@ def test_secure_hash_invalid_type(self, cleartext_password: str, hash_type: str,
     def test_secure_hash_uncaught_exception(self) -> None:
         """Test when an uncaught exception is returned by pyavd-utils."""
         with (
-            patch("pyavd_utils.passwords.sha512_crypt", side_effect=Exception("foo")),
+            patch("pyavd.j2filters.secure_hash.sha512_crypt", side_effect=Exception("foo")),
```
gmuloc: why do you change this test? (I understand you are importing sha512_crypt in this module but the test should not really care where this is imported from)

python-avd/pyproject.toml
```diff
@@ -52,10 +51,10 @@ ansible-collection = [
 
 [build-system]
 requires = [
-    "cryptography>=43.0.0",
     "deepmerge>=1.1.0",
     "jinja2>=3.0",
     "jsonschema-rs>=0.24",
+    "pyavd-utils==0.0.4",
```
gmuloc, line 57: have you read the comment that we were avoiding importing pyavd-utils in build-system on purpose? What prompted you to change this decision?

python-avd/tests/pyavd/j2filters/test_secure_hash.py
ClausHolbechArista: strictly speaking this one should be testing the j2filters and not pyavd-utils. And the patch has to change because it is not imported at the module level instead of inside the function.

python-avd/pyavd/_utils/password_utils/password.py
```diff
@@ -55,11 +54,7 @@ def ospf_simple_encrypt(password: str, key: str) -> str:
         TypeError: If the password is not of type `str`.
     """
     _validate_password_and_key(password, key)
```
ClausHolbechArista, line 56: do we still need this or will pyavd-utils do the same?

python-avd/tests/pyavd/j2filters/test_secure_hash.py
gmuloc: well it is testing the j2filters as this is what is called line 66 and did not change (secure_hash). And ok this is because the import moved (which has other implications on the build)

python-avd/pyavd/_utils/password_utils/password.py
MaheshGSLAB: I checked the rust code and we do not check the empty key, password and non-str password. So yes we still need this here

python-avd/pyproject.toml
MaheshGSLAB: I read the comment where we were importing the pyavd-utils in a method. Actually i am not fully aware about why we are avoiding to import the pyavd-utils and i was not able to run the tests locally because it fails to build with no module pyavd-utils hence i added here.

gmuloc: so the disadvantage of this is that now if people try to pip install pyavd on a system archutecture for which we did not publish the wheel (pyavd or pyavd-utils), they will need to install cargo to be able to build pyavd. This could have impact on our CI.

The reason why you were not able to ryn the test is probably after you moved the import to the top level which make it required when we use schema tools.

I think this is an ok decision to take (which we can rollback anyways later if we need to)

python-avd/pyavd/_utils/password_utils/password.py
gmuloc: I think Pyo3 would yell if you give wrong types so I think the only thing not covered would be size of the strings

MaheshGSLAB: Ok i checked now in our avd code, the key never be empty as we are using interface name as key which is again primary key. And about password so we restricted in our schema that password should be str and min_lenght:1, max_length:8. So things are handled via schema only, However the empty key or empty password still accepted by rust code and it will encrypt the password.

I tried to add below tests in our tests_password.py file and it passed
```
def test_ospf_simple_encrypt_invalid_input() -> None:
    with pytest.raises(TypeError):
        ospf_simple_encrypt(42, key="Ethernet1")


def test_ospf_simple_encrypt_empty_inputs() -> None:
    assert ospf_simple_encrypt("arista", key="") == "some-encrypte-password"
    assert ospf_simple_encrypt("", key="Ethernet1") == "some-encrypte-password"
```

But there is no restriction in code(if peer_group_or_neighbor.cleartext_password is not None) and schema for method `bgp_encrypt`. So if in input for bgp_peer_groups if we give `cleartext_password: ""`, it still accepted and encrypted.

dgonzalez-arista:
Seeing these warnings in PS repositories today, with new version cryptography==47.0.0 (released april 24th). No issues with earlier release 46.0.7:
```
ok: [BLDG1-SP1 -> localhost]
[WARNING]: WorkerProcess for [BLDG1-SP2/TASK: arista.avd.eos_designs : Generate device configuration in structured format] errantly sent data directly to stderr instead of using Display:
    /Users/dgonzalez/AVD/PS-SKU-Campus-L2LS/main/MLAG-with-Inband-mgmt-and-WAN/.venv/lib/python3.13/site-packages/pyavd/_utils/password_utils/password_utils.py:189: CryptographyDeprecationWarning: Single-key TripleDES (8-byte keys) is deprecated and support wi
```
For every device we are seeing a warning at the moment, no other issues.

python-avd/pyavd/_utils/password_utils/password.py
gmuloc: given that right now our code would block this (because `""` would be blocked) let's keep this and see if we can move some of this functionality to pyavd-utils later if required (using an empty string could be a valid option actually for some cases but in EOS you need to give some string. So lets keep it like this for now and move on :)

ClausHolbechArista (approved):

python-avd/pyavd/_utils/password_utils/password.py
Shivani-gslab, line 386: can we fix this TODO in this PR?

gmuloc: let's not as the PR needs to be merged soon.
However you can open an issue to track this and assign it to the GSLab project

Shivani-gslab, line 31: since we are using this method for encrypt and decrypt both.
Update the error messages to -
```
msg = "Key is required for encryption/decryption"
msg = "Password is required for encryption/decryption"
```

I am not able to add suggestion in these lines

Shivani-gslab: https://github.com/aristanetworks/avd/issues/6896

Shivani-gslab (approved):
LGTM

ClausHolbechArista (approved):

gmuloc (approved):

PR #6878, opened 2026-04-22
Fix(cv_deploy): Fix ignored Studio root updates on pure container removals

carl-baillargeon:
## Change Summary

Currently, when synchronizing root containers with the Static Configuration Studio, the code calculates whether an API update is needed using a set difference: `missing_ids = desired_root_ids_set - existing_root_ids_set`.

If a manifest update only removes an existing AVD root container (without adding any new ones), the missing_ids set evaluates to empty so the API update is skipped entirely. This leaves a ghost reference to the removed container in the Studio root assignment list and causes a build error:

image

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes

Replaced the set-difference logic with a direct list comparison to ensure that the Studio roots are properly synchronized anytime the desired state diverges from the existing state.

## How to test

Added a unit test to test this.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -145,18 +145,13 @@ async def _sync_studio_roots(cv_manifest: CVManifest, deployment_result: DeployT
         default_value=[],
     )
 
-    # Calculate which desired roots are missing.
     desired_root_ids = [container.id for container in cv_manifest.containers if container.is_root]
-    desired_root_ids_set = set(desired_root_ids)
-    existing_root_ids_set = set(existing_root_ids)
-    missing_ids = desired_root_ids_set - existing_root_ids_set
+    manual_ids = [container_id for container_id in existing_root_ids if not container_id.startswith(AVD_ENTITY_PREFIX)]
```
alexeygorbunov: Wondering if `manual` is agood term here. Although they can be manually created (via UI), there can as well be per-device AVD-generated root assignment plus reconcile container whihc is not really manually added. Maybe `non_manifest_root_ids` or something?

carl-baillargeon: True. Updated.

alexeygorbunov (approved):
LGTM!

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -145,18 +145,13 @@ async def _sync_studio_roots(cv_manifest: CVManifest, deployment_result: DeployT
...
-    desired_root_ids_set = set(desired_root_ids)
-    existing_root_ids_set = set(existing_root_ids)
-    missing_ids = desired_root_ids_set - existing_root_ids_set
+    non_manifest_root_ids = [container_id for container_id in existing_root_ids if not container_id.startswith(AVD_ENTITY_PREFIX)]
 
-    # Update the Studio root container list if necessary, preserving any manually added (non-AVD) root containers.
-    if missing_ids:
+    # Update the Studio root container list if necessary, preserving any non-manifest root containers.
+    new_ordered_ids = desired_root_ids + non_manifest_root_ids
+    if new_ordered_ids != existing_root_ids:
```
ClausHolbechArista: I think this will reorder the root containers. We should not. If the user wants their manual containers to be above they should be allowed to move them in the UI and then we leave them there. That is why I used set. So try to keep this with set but just fix the issue.

Instead of the missing_ids, you can just filter the existing_root_ids into a new existing_avd_root_ids to only consider the AVD controlled ones, then see if they match the desired_root_ids and otherwise rebuild. I think it would be best if we only append new ones at the beginning but otherwise keep the order the user had.

ClausHolbechArista: you can build 
```python
filtered_ids = [root_id for root_id in existing_root_ids if (not root_id.startswith(AVD_ENTITY_PREFIX) or root_id in desired_root_ids_set)]
new_root_ids = [root_id for root_id in desired_root_ids if root_id not in existing_root_ids_set]
new_ordered_ids = new_root_ids + filtered_ids
if new_ordered_ids != existing_root_ids:
  do it!
```

carl-baillargeon: The suggested code is breaking a unit test because it doesn't honor replacements with multiple declared root containers.

For example:
```
Existing state: MANIFEST_ROOT_A, MANIFEST_ROOT_B
New push: MANIFEST_ROOT_A, MANIFEST_ROOT_C
Final state: MANIFEST_ROOT_C, MANIFEST_ROOT_A  # This is wrong
```
It is also breaking another test because we don't respect today's behavior which is "reorder everytime a new root container (new ID) is pushed", documented here: https://avd.arista.com/6.1/ansible_collections/arista/avd/roles/cv_deploy/index.html#static-configuration-studio-deployment.

The bug happens when we purely remove or reorder roots. The API call is skipped because we don't have new root container so we need to decide what we do with these two scenarios.

I would suggest:

1- When purely removing, we simply remove it from the existing list (keeping the order).
2- When purely reordering, I think we should put them at the top to avoid ambiguity if there is interleaved manifest + non-manifest roots.

If there's both a removal and a reorder (without a new ID), we would treat it as a reorder and put AVD roots at the top.

carl-baillargeon: As discussed, we will just implement 1 for now. 2 can come later with a knob (TBD).

carl-baillargeon: Fixed.

alexeygorbunov: Is this a user-friendly behavior when for some runs your manually-defined order of root assignmets is preserved but then next time (let's say they deploy new site (which is a new root assignment) -> everything now gets reordered and manifest-managed containers are pushed to the head? Wouldn't it be better to have consistent behavior?

carl-baillargeon: This is indeed the current behavior we decided to adopt and document. We could probably improve in 7.0 but we would need to build a graph as discussed yesterday.

ClausHolbechArista (approved):

python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
ClausHolbechArista: Why did you not just implement similar to my suggestion instead of the if/else? Approved still, but the code seems cluttered.

```python
filtered_ids = [root_id for root_id in existing_root_ids if (not root_id.startswith(AVD_ENTITY_PREFIX) or root_id in desired_root_ids_set)]
new_root_ids = [root_id for root_id in desired_root_ids if root_id not in existing_root_ids_set]
new_ordered_ids = new_root_ids + filtered_ids
if new_ordered_ids != existing_root_ids:
  do it!
```

carl-baillargeon: We have two different rules (see _sync_studio_roots docstring) to follow to avoid a breaking change (see my comment above). I think a conditional branch is the correct way to express this logic.

PR #6879, opened 2026-04-23
Refactor(eos_designs): Raise for unsupported combination underlay_router: true and uplink_type: port-channel

### LINKED ISSUE #6708: Feat(eos_designs): Raise for unsupported combination underlay_router: true and uplink_type: port-channel
### Enhancement summary

This is not supported but we don't tell the user so it leads to confusing configuration and things not working

eos_designs

### Use case example

reported from the field

Raise an explicit error stating this is not supported

_No response_

### Additional context

_No response_

### Contributing Guide

Vibhu-gslab:
## Change Summary

Must be implemented behind an avd_design_future knob until 7.0 since people may have this deployed today but not be seeing the issues it brings.

## Related Issue(s)

Fixes #6708 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
added a future know and a check for the combination

## How to test
CI

### Repository Checklist


python-avd/pyavd/_eos_designs/structured_config/underlay/loopback_interfaces.py
```diff
@@ -28,6 +28,10 @@ def loopback_interfaces(self: AvdStructuredConfigUnderlayProtocol) -> None:
         if not self.shared_utils.underlay_router:
             return
 
+        if self.inputs.avd_design_future.raise_for_underlay_router_with_port_channel_uplink and self.shared_utils.uplink_type == "port-channel":
```
gmuloc: it does not make sense here - it should be checked probably when we compute `shared_utils.underlay_router`

Vibhu-gslab: updated

laxmikantchintakindi: Shouldn't we raise this in eos_designs_facts/uplinks.py  similar to conditions for other uplink types?
```
    @cached_property
    def uplinks(self: EosDesignsFactsGeneratorProtocol) -> EosDesignsFactsProtocol.Uplinks:
        if self.shared_utils.uplink_type == "port-channel":
 >>>>>>>We could check for underlay router here
            get_uplink = self._get_port_channel_uplink
        elif self.shared_utils.uplink_type == "p2p-vrfs":
            if self.shared_utils.network_services_l3 is False or self.shared_utils.underlay_router is False:
                msg = "'underlay_router' and 'network_services.l3' must be 'true' for the node_type_key when using 'p2p-vrfs' as 'uplink_type'."
                raise AristaAvdError(msg)
```

laxmikantchintakindi: @gmuloc

python-avd/pyavd/_eos_designs/shared_utils/node_type.py
```diff
@@ -63,6 +63,10 @@ def underlay_router(self: SharedUtilsProtocol) -> bool:
         underlay_router set based on
         node_type_keys.<node_type_key>.underlay_router.
         """
+        if self.inputs.avd_design_future.raise_for_underlay_router_with_port_channel_uplink and self.uplink_type == "port-channel":
```
gmuloc: I am a bit confused as this does not even check that underlay_router is true

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
```diff
@@ -31,3 +31,9 @@ keys:
         description: |-
           Configure the `mlag_ipv4_vrfs_peer` BGP peer group only when needed.
         default: false
+      raise_for_underlay_router_with_port_channel_uplink:
```
gmuloc:
```suggestion
      raise_for_underlay_router_with_port_channel_uplink_type:
```

python-avd/pyavd/_eos_designs/structured_config/underlay/loopback_interfaces.py
gmuloc: really here it is just a decision to be taken, I feel both options are ok given that both are being called in cached properties so the check will ever happen only once and in the place where we need one or the other.

I think your suggestion @laxmikantchintakindi may make thing a bit clearer so I would vote to move it there indeed, thanks for the suggestion.

@Vibhu-gslab can you move the check there?

python-avd/pyavd/_eos_designs/shared_utils/node_type.py
ClausHolbechArista: agreed this should only be checked if underlay_router is actually true.

python-avd/pyavd/_eos_designs/structured_config/underlay/loopback_interfaces.py
Vibhu-gslab: updated

python-avd/pyavd/_eos_designs/shared_utils/node_type.py
Vibhu-gslab: checking it in the new place

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
Vibhu-gslab: updated

laxmikantchintakindi (approved):

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
```diff
@@ -31,3 +31,9 @@ keys:
         description: |-
           Configure the `mlag_ipv4_vrfs_peer` BGP peer group only when needed.
         default: false
+      raise_for_underlay_router_with_port_channel_uplink_type:
```
Shivani-gslab:
```suggestion
      raise_for_underlay_router_with_uplink_type_port_channel:
```

Shivani-gslab (dismissed):

python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
Vibhu-gslab: updated

gmuloc (approved):

PR #6919, opened 2026-05-04
Feat(cv_deploy): Implement client-side gRPC keepalives

alexeygorbunov:
## Change Summary

Enable support for gRPC keepalives in cv_deploy

## Related Issue(s)

Fixes aristanetworks/avd-internal#518

## Component(s) name

`arista.avd.cv_deploy`

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -20,6 +20,27 @@
...
+        if self.enabled and self.keepalive_time < 30:
+            msg = f"Invalid CVGRPCKeepalives settings. keepalive_time must be >= 30s, got {self.keepalive_time}."
+            raise ValueError(msg)
+
+
+@dataclass
+class CVDeployFuture:
+    """Future cv_deploy behaviors that will become defaults in a future major release."""
+
+    enable_grpc_keepalives: bool = False
```
ClausHolbechArista: Do you really see this as a breaking change? I see it more as an improvement which noone should notice.

alexeygorbunov: I was only unsure about on-prem deployments where users have access to CV backend and can potentially use their own L7 FW or WAF to check gRPC traffic between clients and CV

ClausHolbechArista: but we are already sending keepalives (the default 10 min apart?)

alexeygorbunov: No, right now we are not sending any client gRPC keepalives at all. They are disabled by default in grpclib and we are not enabling them

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -62,6 +64,23 @@
             "read_from_validated_inputs": {"type": "bool", "default": False},
         },
     },
+    "cv_deploy_future": {
+        "type": "dict",
+        "required": False,
+        "options": {
+            "enable_grpc_keepalives": {"type": "bool", "required": False, "default": False},
+        },
+    },
```
ClausHolbechArista: remove this and change the default for enabled to false.

alexeygorbunov: Done

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: `CVDeployFuture ` has been removed. Resolving.

ClausHolbechArista (approved):

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
```diff
@@ -138,6 +140,22 @@ async def deploy_to_cv(
...
+                _keepalive_permit_without_calls=cloudvision.grpc_keepalives.permit_without_calls,
+                # Disable the grpclib default cap of 2 pings without data so keepalives
+                # continue for the duration of the deployment.
+                _http2_max_pings_without_data=0,
+                # Override grpclib's 300s rate-limit so pings fire at the configured interval.
+                _http2_min_sent_ping_interval_without_data=cloudvision.grpc_keepalives.keepalive_time,
+            )
+        except TypeError:
+            LOGGER.warning("deploy_to_cv: grpclib Configuration does not support the expected keepalive fields. gRPC keepalives will not be enabled.")
+
```
carl-baillargeon: We could move this logic to a method on  `CVGRPCKeepalives`:

`def as_grpclib_configuration(self) -> Configuration | None:`

Sonar would stop complaining about cognitive complexity.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -183,6 +194,7 @@ async def deploy(self, validated_args: dict, result: dict) -> dict:
                 proxy_port=validated_args.get("proxy_port"),
                 proxy_username=validated_args.get("proxy_username"),
                 proxy_password=validated_args.get("proxy_password"),
+                grpc_keepalives=CVGRPCKeepalives(**get(validated_args, "grpc_keepalives", {})),
```
carl-baillargeon: Minor detail but we can use `validated_args.get("grpc_keepalives", {})` to be consistent with the other args above.

ansible_collections/arista/avd/roles/cv_deploy/defaults/main/main.yml
```diff
@@ -26,6 +26,14 @@ cv_token: "{{ hostvars[cv_inventory_hostname].ansible_password | arista.avd.defa
 # Verify Certificate for CloudVision (Always use valid certificates for production)
 cv_verify_certs: true
 
+# gRPC keepalive settings for the CloudVision connection.
+# keepalive_time must be >= 30s.
+cv_grpc_keepalives:
```
carl-baillargeon, line 30: There is no user facing documentation for this.

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -254,6 +256,7 @@ def __init__(
         proxy_port: int = 8080,
         proxy_username: str | None = None,
         proxy_password: str | None = None,
+        grpc_config: Configuration | None = None,
```
carl-baillargeon: Why not pass the `CVGRPCKeepalives` instance directly instead of the raw grpclib `Configuration` object? This would prevent grpclib implementation details from leaking into the client initialization.

python-avd/pyavd/_cv/workflows/deploy_to_cv.py
alexeygorbunov: Good idea! Moved

ansible_collections/arista/avd/plugins/action/cv_workflow.py
alexeygorbunov: Fixed

ansible_collections/arista/avd/roles/cv_deploy/defaults/main/main.yml
alexeygorbunov: Added description for every sub-key here + added `gRPC Keepalive` section to role's documentation MD (at the bottom, after Proxy section)

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: In this case i would need to import CVGRPCKeepalives (from workspace models) into the client's init which is probably not perfect. CVClient already handles grpclib's Channel, H2Protocol.. We can discuss this tomorrow

alexeygorbunov: Applied changes. Please re-check

python-avd/pyavd/_cv/client/__init__.py
```diff
@@ -126,6 +130,16 @@ async def proxy_connection() -> H2Protocol:
...
 
+    def _grpc_channel_config(self) -> Configuration | None:
+        """
+        Build the grpclib Channel `config` from the optional gRPC channel configuration.
+
+        Returns None when no channel configuration was supplied, matching grpclib's default behavior.
+        """
+        if self._grpc_channel_configuration is None:
+            return None
+        return self._grpc_channel_configuration.as_grpclib_configuration()
```
carl-baillargeon:
```suggestion
    @property
    def _grpclib_channel_config(self) -> Configuration:
        """
        Build the grpclib Channel `config` from the optional gRPC channel configuration.

        Returns None when no channel configuration was supplied, matching grpclib's default behavior.
        """
        if self._grpc_channel_configuration is None:
            return Configuration()
        return self._grpc_channel_configuration.as_grpclib_configuration()
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -19,6 +22,47 @@
...
+
+
+@dataclass
+class CVGRPCChannelConfiguration:
+    """Advanced configuration settings of the gRPC channel."""
+
+    grpc_keepalives: CVGRPCKeepalives = field(default_factory=CVGRPCKeepalives)
+    """Keepalive settings of the gRPC channel."""
+
+    def as_grpclib_configuration(self) -> Configuration | None:
```
carl-baillargeon: So we don't have to deal with None downstream.
```suggestion
    def as_grpclib_configuration(self) -> Configuration:
```

alexeygorbunov: Should be fixed now

python-avd/pyavd/_cv/client/__init__.py
alexeygorbunov: Should be fixed now

carl-baillargeon (approved):

PR #6922, opened 2026-05-05
Feat(eos_designs): Added the support of mgmt_interface_settings

MaheshGSLAB:
## Change Summary

Added the support of mgmt_interface_settings

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Added the support of mgmt_interface_settings.

```
# OOB (Out-of-Band) MGMT interface settings.
mgmt_interface_settings: 
  description: <str; default="OOB_MANAGEMENT">
  vrf: <str; default="MGMT">
  vrf_routing: <bool; default=False>

  lldp:
    transmit: <bool>
    receive: <bool>
    ztp_vlan: <int>
```

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist


Vibhu-gslab (commented):
Try to achieve 100% coverage, rest looks good

MaheshGSLAB:
> Try to achieve 100% coverage, rest looks good

This missing coverage is for shared_utils/mgmt.py for method `default_mgmt_protocol_vrf` which is not used anywhere. Now am not getting why pylint did not raise any error.

Ok so pylint is not work like that and it do not catch the unused methods. There is tool called `vulture` used to  get the unused methods. Right now i just removed that method

python-avd/pyavd/_eos_designs/structured_config/base/management_interface.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2023-2026 Arista Networks, Inc.
```
MaheshGSLAB, line 1: Created new file because the base init file reach its line number length.

python-avd/pyavd/_eos_designs/shared_utils/mgmt.py
```diff
@@ -93,16 +112,6 @@ def default_mgmt_method(self: SharedUtilsProtocol) -> str | None:
 
         return None
 
-    @cached_property
-    def default_mgmt_protocol_vrf(self: SharedUtilsProtocol) -> str | None:
```
laxmikantchintakindi, line 97: Is this moved to some other place?

MaheshGSLAB: no this was never used

laxmikantchintakindi:
There are some mentions of mgmt_interface_vrf in variable inputs of vrf. Need to fix that.

python-avd/pyavd/_eos_designs/schema/schema_fragments/mgmt_interface.schema.yml
```diff
@@ -12,3 +12,8 @@ keys:
     type: str
     default: Management1
     description: OOB Management interface.
+    deprecation:
```
Vibhu-gslab, line 15: We should add testcases for deprecated models in `eos_designs_deprecated_vars`

python-avd/pyavd/_eos_designs/shared_utils/mgmt.py
```diff
@@ -41,6 +41,7 @@ def mgmt_interface(self: SharedUtilsProtocol) -> str:
             # Notice that we actually have a default value for the next two, but the precedence order would break if we use it.
             # TODO: Evaluate if we should remove the default values from either or both.
             self.platform_settings._get("management_interface", None),
+            self.inputs.mgmt_interface_settings._get("interface", None),
```
laxmikantchintakindi, line 44:
```suggestion
            self.inputs.mgmt_interface_settings._get_defined_attr("interface", None),
```

laxmikantchintakindi: Not sure this suggestion works, but using _get_defined_attr is more accurate here.

MaheshGSLAB: type hint is not happy with this

python-avd/pyavd/_eos_designs/schema/schema_fragments/mgmt_interface.schema.yml
MaheshGSLAB: added

laxmikantchintakindi (approved):

Vibhu-gslab (approved):

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/host_vars/mgmt-interface-settings-1.yml
carl-baillargeon, line 1: Can you add a test with no lldp to test the empty-dict gate? Also in the same test, let's use `ipv6_mgmt_ip` on the node for completeness.

python-avd/pyavd/_eos_designs/schema/schema_fragments/mgmt_interface_settings.schema.yml
```diff
@@ -0,0 +1,54 @@
...
+    description: |-
+      Out-of-Band (OOB) management interface settings.
+      These settings override the corresponding global variables where applicable.
+    type: dict
+    keys:
+      description:
+        type: str
+        description: |-
+          Interface description configured on the management interface.
+          Takes precedence over the global 'mgmt_interface_description' variable.
```
carl-baillargeon: I don't think we need the "Takes precedence over the global X variable". These keys are deprecated. The deprecation warning on the old keys already directs users to the new one.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/host_vars/mgmt-interface-settings-1.yml
MaheshGSLAB: added

python-avd/pyavd/_eos_designs/schema/schema_fragments/mgmt_interface_settings.schema.yml
MaheshGSLAB: updated

carl-baillargeon (approved):

PR #6926, opened 2026-05-05
Fix(cv_deploy): Provide informative error message for missing www in CVaaS FQDN

alexeygorbunov:
## Change Summary

Provide informative error message for missing www in CVaaS FQDN

## Related Issue(s)

Fixes aristanetworks/avd-internal#521

## Component(s) name

`arista.avd.cv_deploy`

## Proposed changes
- When `Status.UNKNOWN` is returned - check if targeted FQDN matches CVaaS and if it misses `www` prefix. If it does - raise informative `CVClientException` otherwise fallback to generic CVGRPCError

Example:
```bash
pyavd._cv.client.exceptions.CVClientException: CVaaS FQDN 'cv-prod-us-central1-c.arista.io' is missing the required 'www.' prefix. Please use 'www.cv-prod-us-central1-c.arista.io' instead.
```

## How to test
pytest

### User Checklist

- N/A

### Repository Checklist


python-avd/tests/pyavd/cv/client/test_async_decorators_grpc_unknown.py
alexeygorbunov, line 1: Moving to separate module due to the limitation of 1K lines per module

python-avd/tests/pyavd/cv/client/test_async_decorators_grpc_unknown.py
```diff
@@ -0,0 +1,59 @@
...
+        pytest.raises(CVGRPCError),
+        id="UNKNOWN_STATUS_CVAAS_WITH_WWW_PREFIX",
+    ),
+    pytest.param(
+        "cvp.example.com",
+        pytest.raises(CVGRPCError),
+        id="UNKNOWN_STATUS_NON_CVAAS_FQDN",
+    ),
+    pytest.param(
+        ["prod.arista.io", "www.prod2.arista.io"],
```
MaheshGSLAB: what about `["www.prod.arista.io", "prod2.arista.io"]` case?

alexeygorbunov: Thanks, added this use case now.

MaheshGSLAB: Ok but i still did not get why raising the CVClientException only for first server not for second one.

ClausHolbechArista: We never really use the second server. That is on the roadmap still.

MaheshGSLAB (approved):
Code wise LGTM

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -269,6 +269,15 @@ async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs:
                                     new_exception.size = int(matches.group("size"))
                                     raise new_exception
 
+                            case Status.UNKNOWN:
+                                # Check if www is missing in CVaaS FQDN
+                                cv_servers = getattr(call_args[0], "_servers", []) if call_args else []
+                                first_cv_server = cv_servers[0] if cv_servers else ""
+                                if first_cv_server.endswith("arista.io") and not first_cv_server.startswith("www."):
+                                    msg = f"CVaaS FQDN '{first_cv_server}' is missing the required 'www.' prefix. Please use 'www.{first_cv_server}' instead."
```
ClausHolbechArista: should we strip apiserver prefix here in case they had that? We can also check the allowed list of urls from what we have in eos_designs if we move the constant for those to a common place.

alexeygorbunov: Hey Claus! I added `apiserver.` striping for now and also shared link to the stable source of all www endpoints (https://www.arista.io/meta/ip.json) - let me know if it is OK or not (it's in JSON but should be easy to read and should be always available in contrast to the general documentation which may change path at any time).
Moving existing CvaaS endpoints dict is also an option but i'm not sure where would it be OK to move it especially if there is a chance that "cv_deploy" can become a separate package

ClausHolbechArista: Please do not let a potential future which is not near affect the choices here. I don't think it is a good source to link to. If we don't lookup in a local list we should at least point to a helpcenter link instead of some raw json file.

alexeygorbunov: Ok, added verification against our local constant and referencing `https://www.arista.io/help` in case it is an unknown/incorrect `arista.io` fqdn. Please re-check

P.S. https://www.arista.io/help/articles/setup-overview-deployment-guidelines#Y3ZhYXNEZXBsb3ltZW50-region is a link to the full table of the regions but i am not confident that this path won't change

python-avd/pyavd/_cv/constants.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (c) 2026 Arista Networks, Inc.
+# Use of this source code is governed by the Apache License 2.0
+# that can be found in the LICENSE file.
+
+CV_REGION_TO_SERVER_MAP = {
```
ClausHolbechArista, line 5: Why change this when they are all the same anyway? If you want it to be more "fancy" we should have a data class for every entry with default values and then some helper methods to build the URLs. But it seems a bit much for this...

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_servers.py
```diff
@@ -88,11 +87,6 @@ def _update_ipv4_ztp_boot_file(self: AvdStructuredConfigUnderlayProtocol, dhcp_s
         if not (cvp_server := self._get_cvp_server_for_dhcp()):
             return
 
-        if "arista.io" in cvp_server:
-            # Change apiserver.<...>arista.io to www.<...>arista.io
-            domain = re.sub(r"https:\/\/|www\.|apiserver\.", "", cvp_server)
-            cvp_server = f"www.{domain}"
-
```
ClausHolbechArista, line 95: I did not recall this behavior. Changing that would be breaking (people may be misusing the onprem_clusters)

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -269,6 +270,38 @@ async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs:
                                     new_exception.size = int(matches.group("size"))
                                     raise new_exception
 
+                            case Status.UNKNOWN:
```
ClausHolbechArista, line 273: Let's build some separate helper function to validate the server name and return bool + a hint if false so we can raise here but avoid all the clutter here.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -269,6 +270,38 @@ async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs:
...
+                                # Check if www is missing in CVaaS FQDN
+                                cv_servers = getattr(call_args[0], "_servers", []) if call_args else []
+                                first_cv_server = cv_servers[0] if cv_servers else ""
+                                if first_cv_server.endswith("arista.io"):
+                                    for cvaas_region in CV_REGION_TO_SERVER_MAP.values():
```
ClausHolbechArista: the fact that you have to loop over all the regions here makes it clear that the data structure is not really helpful. it would be good if we could just identify the region part and look that up directly.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -269,6 +270,38 @@ async def _execute_single_call_with_retries(self, call_args: tuple, call_kwargs:
...
+                                        # Correctly configured API endpoint - fall through to CVGRPCError
+                                        if first_cv_server == f"{cvaas_region['api']}.{cvaas_region['base_fqdn']}":
+                                            break
+                                        # Target CVaaS is missing api prefix
+                                        if first_cv_server == cvaas_region["base_fqdn"]:
+                                            msg = (
+                                                f"CVaaS FQDN '{first_cv_server}' is missing the required '{cvaas_region['api']}.' prefix. "
+                                                f"Please use '{cvaas_region['api']}.{cvaas_region['base_fqdn']}' instead."
+                                            )
+                                            raise CVClientException(msg)
```
ClausHolbechArista: Let's create a specific error for invalid server name.

alexeygorbunov: Added CVClientInvalidServerName

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Moved logic to the separate helper method

python-avd/pyavd/_eos_designs/structured_config/underlay/dhcp_servers.py
alexeygorbunov: Returned this code block

python-avd/pyavd/_cv/constants.py
alexeygorbunov: Returned original dictionary but stripped `apiserver.` from values and created 2 constants for streaming and API prefixes

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Tried to use 3 frozensets (for base, streaming and api fqdns) and use direct `in` check

ClausHolbechArista: This did not really help on what I pointed out. On the contrary. Now you are looping through these three times even for on-prem cluster or just for eos_designs (for every device!). If we can't come up with something better you can just go back to looping inside the check logic.

alexeygorbunov: Please check now.

python-avd/pyavd/_cv/client/async_decorators.py
```diff
@@ -383,3 +395,48 @@ def _raise_on_bulk_response_errors(self, responses: list) -> None:
...
+        # Target CVaaS is missing api prefix
+        if first_cv_server in base_fqdns:
+            return True, (
+                f"CVaaS FQDN '{first_cv_server}' is missing the required '{CVAAS_API_PREFIX}.' prefix. "
+                f"Please use '{CVAAS_API_PREFIX}.{first_cv_server}' instead."
+            )
+
+        # Unknown arista.io FQDN
+        return True, (
+            f"Provided CVaaS FQDN '{first_cv_server}' may be incorrect. Please check 'https://www.arista.io/help' for the full list of existing CVaaS clusters."
```
ClausHolbechArista:
```suggestion
            f"Provided CVaaS FQDN '{first_cv_server}' may be incorrect. Please check 'https://www.arista.io/help' for the full list of supported CVaaS clusters."
```

ClausHolbechArista (approved):
Minor wording on the last resort error but otherwise LGTM!

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Applied suggested change

python-avd/pyavd/_cv/client/async_decorators.py
alexeygorbunov: Marking as resolved following Claus's approval

PR #6945, opened 2026-05-11
Fix(pyavd): Refactor and fix static configuration studio deployment workflow

carl-baillargeon:
## Change Summary

Refactor of the static configuration studio deployment workflow into a "fetch -> resolve -> plan -> execute" pipeline in a new sub-package. Existing deployments behave the same.

This sets the stage for the upcoming root policy and sub-container policy features: https://github.com/aristanetworks/avd/pull/6732.

The refactor also fixes a bug in the issue below. New unit tests have been added to cover the fix for the bug.

## Related Issue(s)

Fixes #https://github.com/aristanetworks/avd/issues/6952

## Component(s) name

PyAVD

## Proposed changes

- New `static_configuration_studio/` sub-package to hold related modules
- Moved "no work to do" logs from info to debug
- Added TODOs for future work

## How to test

CI is green.
Internal CI is green --> https://github.com/aristanetworks/avd-internal/actions/runs/25700818681/job/75460685982

### User Checklist

- [x] Need to determine if we keep today's behavior of attempting to delete undeclared manifest objects still assigned, causing the workspace build to fail, or we keep them.

### Repository Checklist

### DIFF
--- a/python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
+++ b/python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
@@ -3,24 +3,21 @@
 # that can be found in the LICENSE file.
 from __future__ import annotations
 
-from asyncio import gather
 from logging import getLogger
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING
 
-from .models import AVD_ENTITY_PREFIX, CVManifest
+from .static_configuration_studio.deploy import execute_plan, fetch_existing_state
+from .static_configuration_studio.models import DeploymentPlan, DesiredState, ResolvedState
 
 if TYPE_CHECKING:
-    from pyavd._cv.api.arista.configlet.v1 import ConfigletAssignment
     from pyavd._cv.client import CVClient
 
-    from .models import AvdManifest, CVContainer, DeployToCvResult
+    from .models import AvdManifest, DeployToCvResult
 
 LOGGER = getLogger(__name__)
 
 
-STATIC_CONFIGURATION_STUDIO_ID = "studio-static-configlet"
-
-
+# TODO: Move this to the workflows/static_configuration_studio package.
 async def deploy_static_config_studio_manifest_to_cv(manifest: AvdManifest, deployment_result: DeployToCvResult, cv_client: CVClient) -> None:
     """
     Deploy a manifest (configlets/containers) to CloudVision using the "Static Configuration" Studio.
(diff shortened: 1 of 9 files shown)


python-avd/pyavd/_cv/workflows/deploy_static_config_studio_manifest_to_cv.py
```diff
@@ -3,24 +3,21 @@
...
-    from .models import AvdManifest, CVContainer, DeployToCvResult
+    from .models import AvdManifest, DeployToCvResult
 
 LOGGER = getLogger(__name__)
 
 
-STATIC_CONFIGURATION_STUDIO_ID = "studio-static-configlet"
-
-
+# TODO: Move this to the workflows/static_configuration_studio package.
```
carl-baillargeon, line 20: Decided to keep it for now to make it easier to review.

python-avd/tests/pyavd/cv/workflows/helpers.py
```diff
@@ -153,4 +153,4 @@ def get_interface_tag_assignments_cv_state() -> list[TagAssignment]:
 
 def generate_id(key: str) -> str:
```
ClausHolbechArista, line 154: if this is for tests should we have it in the production code? I think we should move this out (as TODO if you prefer)

carl-baillargeon: It's not in production code right now :)

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
```diff
@@ -0,0 +1,191 @@
...
+    def _compute_child_map(desired: DesiredState, existing: ExistingState) -> dict[str, list[str]]:
+        """
+        Compute the parent-to-children mapping for the final hierarchy, using the existing CloudVision relationships as the baseline.
+
+        Manifest containers replace their existing children.
+        """
+        # TODO: Add sub_container_policy handling here in the future.
+        child_map = {container_id: list(container.child_assignment_ids.values) for container_id, container in existing.containers_by_id.items()}
+        for container_id, desired_container in desired.containers_by_id.items():
+            child_map[container_id] = list(desired_container.child_ids)
```
alexeygorbunov, line 89: If manifest container had manually-added child container - what will happen with that manually-added container after we unassociate if here from teh parent? We are leaving it in the DB unattached to anything? Should we collect all thos non-manifest containers that we "unattach" and attempt to clean later? Or maybe just notify user.

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
```diff
@@ -0,0 +1,191 @@
...
+
+        # Classify each AVD-managed existing container as deletable (unreachable) or preserved (still reachable through non-manifest containers).
+        for container_id, container in existing.containers_by_id.items():
+            if not AvdId.is_managed(container_id):
+                continue
+            display_name = cast("str", container.display_name)
+            if container_id not in resolved.reachable_container_ids:
+                plan.containers_to_delete[container_id] = display_name
+            elif container_id not in desired.containers_by_id:
+                plan.containers_preserved[container_id] = display_name
```
alexeygorbunov, line 177: What happens with such container if during next deployment user defines it in the manifest? We'll update it's attributes but would leave it nested under non-manifest container? Can container_id be in "children" of multiple parent containers though (probably not)?

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -395,79 +394,92 @@ def from_dict(cls, data: dict[str, Any]) -> AvdManifest:
...
                 raise CVManifestError(msg)
-            configlet_ids.append(cv_configlet_map[configlet_name].id)
+            configlet_ids.append(configlet_id)
 
         # Create the parent CVContainer object.
         cv_container = CVContainer(
             avd_container=container,
-            id=cls._generate_deterministic_id(current_path),
+            id=AvdId.generate(current_path),
             is_root=(parent_path == ""),
```
alexeygorbunov, line 470: is_root seem to not be used anywhere anymore

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
```diff
@@ -0,0 +1,191 @@
...
+                assigned.update(desired.containers_by_id[container_id].configlet_ids)
+            elif container_id in existing.containers_by_id:
+                # Container exists but isn't in the current manifest. Keep its existing configlets.
+                assigned.update(existing.containers_by_id[container_id].configlet_ids.values)
+
+        return assigned
+
+
+@dataclass
+class DeploymentPlan:
```
alexeygorbunov, line 128: Should it be frozen like Existing or Resolved?

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
carl-baillargeon: We cannot delete them otherwise it would be a breaking change but yes I had in mind to notify the user in a separate PR.

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
carl-baillargeon: Right the same container ID cannot be a child of multiple parents, it throws a build error:
image

We would need to unassign it from the other parent first but I would prefer not to touch non-AVD entities. We could maybe detect this in the code and raise a better error though.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: Good catch. I will clean it up in a separate PR when moving the modules.

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
carl-baillargeon: For this one the build() code is simply updating the fields inline to avoid having to initialize empty dicts/lists at the beginning.

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
carl-baillargeon: Tracked here: https://github.com/aristanetworks/avd-internal/issues/543

python-avd/pyavd/_cv/workflows/static_configuration_studio/models.py
carl-baillargeon: https://github.com/aristanetworks/avd-internal/issues/544

carl-baillargeon:
Fixes were implemented in separate PRs:
https://github.com/aristanetworks/avd/pull/7024
https://github.com/aristanetworks/avd/pull/7001
https://github.com/aristanetworks/avd/pull/6990

PR #6951, opened 2026-05-12
Doc: static MkDocs-integrated Schema Explorer

joelbreton2:
## Change Summary

Add the AVD Schema Explorer to the MkDocs site as a static, self-contained browser app. The browser loads a bundled SQLite database with sql.js and runs all queries client-side: no backend, no App Engine, no Cloud Run, no CloudSQL, and no GCP URL-map changes.

The final integration uses a Material-hosted docs page plus a reusable Markdown embed syntax backed by a `pymdownx.superfences` formatter. Docs authors can embed scoped explorers in any page with a `schema-explorer` fenced block, for example a how-to guide scoped to `platform_settings`.

## Related Issue(s)

- Architecture / parent: aristanetworks/avd-internal#503
- Resolver rewrite: aristanetworks/avd-internal#539
- Superseded: aristanetworks/avd-internal#522

## Component(s) name

`tools/schema-explorer/`, `tools/schema_explorer_markdown.py`, `docs/schema-explorer.md`, `docs/data-models/README.md`, `docs/howto/migrate-to-anta-runner.md`, `mkdocs.yml`

## Proposed changes

**Schema Explorer source and generator**

- Adds `tools/schema-explorer/generate.py`, which loads `eos_designs` and `eos_cli_config_gen` through pyavd's `schema_tools` resolver.
- Expands `dynamic_keys` placeholders and same-schema `$ref` blocks.
- Preserves cross-schema refs as a `cross_ref` column instead of materializing the full referenced tree.
- Writes `tools/schema-explorer/build/data/schema.sqlite` and copies the static SPA assets into the same build tree.

**Markdown embed formatter**

- Adds `tools/schema_explorer_markdown.py`, registered by the MkDocs hook as a `schema-explorer` SuperFences formatter.
- Converts fenced Markdown blocks into `<schema-explorer>` custom elements consumed by the existing browser runtime.
- Supports `module`, `root`, `view`, `height`, and `chrome` options.
- Replaces raw HTML embed usage in `docs/data-models/README.md`.
- Adds a reusable how-to example in `docs/howto/migrate-to-anta-runner.md` scoped to `eos_designs` `platform_settings`.

Example authoring syntax:

````markdown
```schema-explorer
module: eos_designs
root: platform_settings
height: 500px
```
````

**Static UI**

- Adds the standalone SPA assets under `tools/schema-explorer/static/`.
- Provides landing, module, variable-detail, Navigator, Reference, YAML, Index, category, table, lifecycle, search, and per-row collapse views.
- Lazy-loads sql.js, Bootstrap JS, and icon CSS only when the explorer mounts.
- Keeps Bootstrap CSS out of Material-hosted docs pages to avoid leaking global typography/layout rules.

**MkDocs integration**

- Registers `tools/schema-explorer/mkdocs_hook.py` in `mkdocs.yml`.
- The hook registers the Markdown formatter, auto-builds the bundled SQLite when needed, injects the lightweight loader CSS/JS, and copies the built explorer into `site/_assets/schema-explorer/` during `on_post_build`.
- Adds `docs/schema-explorer.md` as the Material-hosted full-page entry.
- Adds a Schema Explorer nav item under User Manual > Data Models.

**Build / dev**

- Adds `make schema-explorer-build`, `make docs-serve`, and `make docs-serve-docker`.
- Updates the webdoc container entrypoint to build the Schema Explorer with an mtime guard before serving docs.
- Adds generator dependencies to the `doc` dependency group.
- Excludes `tools/*` and local virtualenvs from the MkDocs docs scan.

## Follow-up / out of scope for this PR

The release publishing pipeline still needs a separate follow-up to generate and publish one SQLite per supported minor release tag, 5.7+. This PR provides the generator, local build path, MkDocs hook, Markdown embed formatter, and flat `data/schema.sqlite` artifact path; it does not change the release-tag docs publishing workflow.

## How to test

```bash
make docs-serve
```

Then verify:

- `/docs/schema-explorer.html` renders inside Material's chrome.
- `/docs/data-models/index.html` shows the Data Models card links and embedded `router_bgp` preview.
- `/docs/howto/migrate-to-anta-runner.html` shows an embedded `platform_settings` schema preview.
- `/_assets/schema-explorer/index.html` remains reachable as the standalone SPA.
- `/_assets/schema-explorer/data/schema.sqlite` is served.
- The theme toggle does not leak Bootstrap styling into normal docs pages.
- Category/table filters, lifecycle filters, Navigator/Index views, per-row collapse, and variable detail pages work.

Direct validation run locally:

```bash
uv run --group dev ruff check tools/schema_explorer_markdown.py tools/schema-explorer/mkdocs_hook.py
uv run --group dev pre-commit run markdownlint-cli2 --files docs/data-models/README.md docs/howto/migrate-to-anta-runner.md tools/schema-explorer/README.md
uv run --group doc mkdocs build -f mkdocs.yml
```

## Latest test results

Validated locally on June 29, 2026:

```bash
node --check tools/schema-explorer/static/js/app.js
python3 -m py_compile tools/schema_explorer_markdown.py
make schema-explorer-build
```

All commands passed. `make schema-explorer-build` generated the local SQLite bundle successfully with 13,642 variables across `eos_designs` and `eos_cli_config_gen`.

### User Checklist

- [x] Schema Explorer SPA loads and queries the bundled SQLite
- [x] Dynamic key subtrees are populated
- [x] Same-schema refs are expanded
- [x] Cross-schema refs are preserved as `cross_ref`
- [x] Navigator view supports targeted ancestor walks and per-row collapse
- [x] Material-hosted full-page entry exists at `docs/schema-explorer.md`
- [x] Inline embeds use a Markdown formatter, not raw HTML
- [x] How-to guide example embeds the `platform_settings` schema subtree
- [x] MkDocs nav includes the Schema Explorer under Data Models
- [x] Follow-up release-publishing workflow generates SQLite files per supported minor release tag

### Repository Checklist

- [x] I have read the CONTRIBUTING document.
- [x] My change requires a change to the documentation and documentation has been updated accordingly.
- [x] I have updated molecule CI testing accordingly. (Not applicable: pure docs/tooling add; no role / pyavd runtime behavior touched.)

## Summary by CodeRabbit

* **New Features**
  * Added a “Schema Explorer” docs experience to browse and search full schema variable metadata.
  * Added an embeddable Schema Explorer block with configurable module, view, root, height, and chrome options, and added it to docs navigation.
* **Documentation**
  * Added a dedicated Schema Explorer page and updated Data Models links to use the embedded flow.
* **Build / Configuration**
  * Added tooling to generate and serve Schema Explorer assets for MkDocs, including incremental rebuild and local/container serving.
  * Updated docs build configuration to exclude extra local artifacts and integrated the Schema Explorer MkDocs hook.


joelbreton2:
Suggestion on the build output location: `docs/` should only hold documentation (`.md`, images, etc.), not generated SQLite + JS/CSS. Right now `make schema-explorer-build` writes into `docs/schema-explorer/` (gitignored), which keeps git clean but still mixes a runtime artifact into the docs source tree.

**Proposed change:** build into a staging dir outside `docs/`, then have MkDocs copy it into `site/` at build time via a tiny hook. Net effect: `docs/` stays markdown-only, `mkdocs build` and `mkdocs serve` both pick it up, no plugin dependency.

**Makefile** (replacing the current `SCHEMA_EXPLORER_SITE = docs/schema-explorer`):
```make
SCHEMA_EXPLORER_SRC   = tools/schema-explorer
SCHEMA_EXPLORER_BUILD = tools/schema-explorer/build   # gitignored

.PHONY: schema-explorer-build
schema-explorer-build:
	uv run --group doc python $(SCHEMA_EXPLORER_SRC)/generate.py \
		--avd-root . --release devel --site-dir $(SCHEMA_EXPLORER_BUILD)
```

**`tools/schema-explorer/.gitignore`** — add `build/`, drop the `/docs/schema-explorer/` line from the root `.gitignore`.

**`tools/schema-explorer/mkdocs_hook.py`** (~15 lines, registered as `hooks: [tools/schema-explorer/mkdocs_hook.py]` in `mkdocs.yml`):
```python
from pathlib import Path
import shutil

BUILD_DIR = Path(__file__).parent / "build"

def on_post_build(config, **kwargs):
    if not BUILD_DIR.is_dir():
        return
    dest = Path(config["site_dir"]) / "schema-explorer"
    if dest.exists():
        shutil.rmtree(dest)
    shutil.copytree(BUILD_DIR, dest)
```

`docs/schema-explorer.md` still resolves to `/schema-explorer/` in the published site. `mkdocs serve` re-runs `on_post_build` on every rebuild, so local dev keeps working.

The wrapper `.md` page stays under `docs/` (it's documentation). The SPA + SQLite never touch `docs/`.

I'll push the change to this branch.

JulioPDX (approved):
LGTM, left comments for future development fixes

tools/schema-explorer/static/js/app.js
```diff
@@ -0,0 +1,1678 @@
...
+      if (parts.length) parts[parts.length - 1] += "[]";
+    } else {
+      parts.push(seg);
+    }
+  }
+  const keyPath = parts.join(".");
+  const link = keyPath
+    ? `#/${target}/${encodeURI(keyPath)}`
+    : `#/${target}`;
+  return `<tr><td class="px-3 fw-semibold small text-muted">Cross-schema</td><td><a href="${link}" class="link-brand"><code>${escapeHtml(target)}</code> → <code>${escapeHtml(keyPath || "(root)")}</code></a></td></tr>`;
```
gmuloc: `renderCrossRefRow()` builds an `<a href="...">` from schema-derived `$ref` data without validating the target module or escaping the final attribute value.

  Even though current schema input is trusted, this becomes public docs HTML. A malformed cross-schema ref such as a target containing quotes could break out of `href`
  and inject markup or an event handler.

  Please harden this by:
  - allowlisting `target` to known schema module IDs, e.g. `eos_designs` and `eos_cli_config_gen`;
  - encoding hash path segments with `encodeURIComponent`, not `encodeURI`;
  - escaping the final `href` with `escapeAttr(link)`, or building the anchor with DOM APIs instead of string interpolation.

  Suggested shape:

  ```js
  if (!Object.prototype.hasOwnProperty.call(SCHEMA_MODULES, target)) return "";

  const encodedTarget = encodeURIComponent(target);
  const encodedKeyPath = keyPath.split("/").map(encodeURIComponent).join("/");
  const link = keyPath ? `#/${encodedTarget}/${encodedKeyPath}` : `#/${encodedTarget}`;

  return `<a href="${escapeAttr(link)}" ...>`;
```

  AI prompt:

  ```text
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

  In PR 6951, fix tools/schema-explorer/static/js/app.js renderCrossRefRow() so schema-derived cross_ref values cannot inject HTML through href. Validate the target
  module against the known schema module IDs, encode URL hash path segments with encodeURIComponent, escape the final href attribute, and keep the rendered labels
  escaped. Add a small focused test or documented manual check if there is no JS test harness.
```

development/entrypoint.sh
```diff
@@ -8,10 +8,46 @@ apk add --no-cache git git-fast-import
...
+    for SCHEMA_SRC in $SCHEMA_INPUTS; do
+        if [ ! -f "$SCHEMA_SRC" ] || [ "$SCHEMA_SRC" -nt "$SCHEMA_OUT" ]; then
+            SCHEMA_REBUILD=true
+            break
+        fi
+    done
+fi
+if [ "$SCHEMA_REBUILD" = true ]; then
+    echo "Building Schema Explorer"
+    python /data/tools/schema-explorer/generate.py \
```
gmuloc: `development/entrypoint.sh` appears to be out of sync with `generate.py`.

  The entrypoint calls:

  ```sh
  python /data/tools/schema-explorer/generate.py \
    --avd-root /data --release devel --site-dir /data/tools/schema-explorer/build
```
  but generate.py does not define `--release`, so docs-serve-docker fails with unrecognized arguments: `--release devel`.

The stale check also watches `build/data/devel/schema.sqlite`, while the generator writes `build/data/schema.sqlite`, and it includes `tools/schema-explorer/categories.py`, which is not present in this PR. Please align the entrypoint with the actual generator contract:

  - remove `--release devel`;
  - set `SCHEMA_OUT=/data/tools/schema-explorer/build/data/schema.sqlite`;
  - remove or correct the nonexistent `categories.py` input;
  - ideally run `make docs-serve-docker` or the entrypoint command once after the fix.


AI prompt:

```text
  In PR 6951, fix `development/entrypoint.sh` to match `tools/schema-explorer/generate.py`. Remove the unsupported `--release devel` argument, update SCHEMA_OUT to `build/data/schema.sqlite`, remove the nonexistent `categories.py` input from the mtime guard, and verify the Docker docs entrypoint can run the schema explorer build command successfully.
```

tools/schema-explorer/static/index.html
```diff
@@ -0,0 +1,60 @@
+<!DOCTYPE html>
+<html lang="en" data-bs-theme="light">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>AVD Schema Explorer</title>
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@5.3.3/dist/css/bootstrap.min.css" integrity="sha384-QWTKZyjpPEjISv5WaRU9OFeRpok6YctnYmDr5pNlyT2bRjXh0JMhjY6hW+ALEwIH" crossorigin="anonymous">
```
gmuloc: This PR describes the explorer as static/self-contained, but the runtime still loads Bootstrap, Bootstrap Icons, sql.js, sql-wasm.wasm, and Google Fonts from third-party CDNs.

The JS and WASM assets are pinned with SRI/manual verification, which helps, but releasing this in the public MkDocs site still means the docs CSP must allow third-party script/style/font/connect sources. Google Fonts are also not covered by SRI.

Please either:
  - vendor these assets under `_assets/schema-explorer/` and load them from the same origin; or
  - explicitly document the required CSP and the operational process for keeping CDN versions and hashes current.

For a public docs site, same-origin vendoring is preferable because it reduces CSP scope and external availability/supply-chain risk.

AI prompt:

```
Verify each finding against current code. Fix only still-valid issues, skip the
rest with a brief reason, keep changes minimal, and validate.

In PR 6951, make the Schema Explorer runtime truly self-contained by vendoring Bootstrap CSS/JS, Bootstrap Icons and font files, sql.js JS/WASM, and required fonts under tools/schema-explorer/static or another generated _assets/schema-explorer path. Update index.html and app.js to load these assets from same-origin relative URLs instead of jsDelivr or Google Fonts, preserving integrity where practical and keeping MkDocs pages free from global Bootstrap CSS leakage.
```

tools/schema-explorer/mkdocs_hook.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def on_post_build(config: dict[str, Any], **kwargs: Any) -> None:  # noqa: ARG001
+    """Copy the built Schema Explorer into the generated site."""
+    if not BUILD_DIR.is_dir():
+        return
+    dest = Path(config["site_dir"]) / ASSET_SUBPATH
+    if dest.exists():
+        shutil.rmtree(dest)
+    shutil.copytree(BUILD_DIR, dest)
```
gmuloc: If CI is guaranteed clean, the PR-blocking version is too strong. I would comment it like this instead:

Question / hardening: since `build/` is gitignored, should `on_post_build()` copy only the expected Schema Explorer artifacts instead of the whole directory?

If docs publishing CI always starts from a clean workspace and always regenerates this directory, this is not an immediate release blocker. Still, copying the entire ignored build tree means any future stale/debug/local file under `tools/schema-explorer/build/` would be published to `_assets/schema-explorer/`.

Would it be worth allowlisting the expected outputs (`index.html`, `css/`, `js/`, `data/schema.sqlite`) so the public artifact is deterministic?

AI prompt:

```
In PR 6951, harden tools/schema-explorer/mkdocs_hook.py so on_post_build() does not copy the entire gitignored build directory into the public MkDocs site. Implement an allowlisted copy of only index.html, css/, js/, and data/schema.sqlite, and fail clearly if required files are missing. Prefer deterministic publishing and avoid exposing stale local artifacts.
```

tools/schema-explorer/static/js/app.js
joelbreton2: Addressed in `0c2341d6e`: `renderCrossRefRow()` now returns early unless `target` is one of the known `SCHEMA_MODULES`, encodes the target/key path segments with `encodeURIComponent`, and escapes the final `href` with `escapeAttr(link)` while keeping the displayed labels escaped.

development/entrypoint.sh
joelbreton2: Addressed in `0c2341d6e`: `development/entrypoint.sh` now matches the current `generate.py` contract. The stale marker is `build/data/schema.sqlite`, the unsupported `--release devel` argument is removed, and the nonexistent `categories.py` input is no longer part of the freshness check.

ankudinov (approved):
LGTM

pyproject.toml
```diff
@@ -44,7 +44,14 @@ doc = [
     "mkdocs-git-revision-date-plugin",
     "mkdocs-exclude",
     "mkdocs-include-dir-to-nav",
+    "mkdocs-autorefs",
```
gmuloc: is this used?

pyproject.toml
```diff
@@ -44,7 +44,14 @@ doc = [
...
     "mkdocstrings[python]",
+    # Schema Explorer build-time deps (tools/schema-explorer/generate.py)
```
gmuloc: please make a separate group

```
doc = [
   .....
   "mkdocstrings[python]",
   {include-group = "schema-explorer"},
]

schema-explorer = [
  "pyyaml>=6.0.0",
  "referencing>=0.37.0",
  "deepmerge>=1.1.0",
  "pydantic>=2.11.10",
  "jsonschema-rs>=0.46.0",
]
```


my friend even tells me we could optimise with

```
  schema-tools = [
      "pyyaml>=6.0.0",
      "referencing>=0.37.0",
      "deepmerge>=1.1.0",
      "pydantic>=2.11.10",
      "jsonschema-rs>=0.46.0",
  ]
```
  Then:
```
  dev = [
      "anta==1.8.0",
      "netaddr>=0.7.19",
      ...
      {include-group = "schema-tools"},
      {include-group = "pytest"},
      {include-group = "molecule"},
  ]
```
```
  doc = [
      ...
      {include-group = "schema-tools"},
  ]
```

```
  pytest = [
      "ansible-core>=2.16.0,<2.22.0",
      "anta==1.8.0",
      "pytest>=9.1.0",
      "pytest-asyncio>=1.3.0",
      {include-group = "schema-tools"},
  ]
```

gmuloc (commented):
3 nit that would be good to address and then let's roll with it as preview and see

tools/schema-explorer/mkdocs_hook.py
```diff
@@ -0,0 +1,216 @@
...
+from types import ModuleType
+from typing import Any
+
+HERE = Path(__file__).resolve().parent
+BUILD_DIR = HERE / "build"
+STATIC_DIR = HERE / "static"
+GENERATE_SCRIPT = HERE / "generate.py"
+# Repo root is two levels up from tools/schema-explorer/.
+AVD_ROOT = HERE.parents[1]
+SCHEMA_INPUTS = (
```
gmuloc, line 46: This is not blocking for this PR. 

The Schema Explorer generator consumes the compiled full schema files, so the current freshness guard correctly tracks the files it actually reads. 
For local schema-development workflows, though, a fragment-only edit will not be visible to the explorer until the full schema files are regenerated, which pre-commit already does.

Maybe worth mentioning somewhere as this could lead to weird "why is my explorer not updating" for people :)

gmuloc (approved):
yolookGTM

PR #7018, opened 2026-05-26
Fix(eos_designs): Ensure consistent uplink vlans when parents have inconsistent vlans

ClausHolbechArista:
## Change Summary

Ensure consistent uplink vlans when parents have inconsistent vlans
Implemented behind a new `avd_design_future.consistent_uplink_vlans` toggle

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- In rare cases, uplink_switches may allow different vlans
- The logic for configuring vlans already took this into account to only configure the common vlans
- The logic for uplinks did _not_ take this properly into account, and configured different allow lists for the same port-channel and even allowed vlans not configured.

## How to test

- First commit adds a test case excibiting the issue.
- Second commit fixes the issue + adds some performance improvements by wider use of sets.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/j2filters/list_compress.py
```diff
@@ -21,14 +23,22 @@ def list_compress(list_to_compress: list[int]) -> str:
...
+        raise TypeError(msg)  # noqa: B904
 
     counter = count()
-    groups = (list(group) for key, group in groupby(sorted(list_to_compress), lambda element, counter=counter: next(counter) - element))
+    groups = (list(group) for key, group in groupby(sorted_list_to_compress, lambda element, counter=counter: next(counter) - element))
     return ",".join("-".join(map(str, (group[0], group[-1])[: len(group)])) for group in groups)
+
+
+def _key(element: Any) -> int:
+    if isinstance(element, int):
```
gmuloc: would not catch bool

gmuloc: but this is alrady our behavior so this can be ignored for now

python-avd/pyavd/j2filters/list_compress.py
```diff
@@ -3,15 +3,17 @@
 # that can be found in the LICENSE file.
 from __future__ import annotations
 
+from collections.abc import Iterable
 from itertools import count, groupby
+from typing import Any
 
 
-def list_compress(list_to_compress: list[int]) -> str:
+def list_compress(list_to_compress: Iterable[int]) -> str:
```
gmuloc: this function would not raise for `{1: "a", 2: "b"}` so not sure if we really want Iterable but probably need to understand better.

python-avd/pyavd/j2filters/list_compress.py
```diff
@@ -21,14 +23,22 @@ def list_compress(list_to_compress: list[int]) -> str:
...
-        msg = f"Value must be of type list, got {type(list_to_compress)}"
+    if not isinstance(list_to_compress, Iterable):
+        msg = f"Value must be an Iterable, got {type(list_to_compress)}"
         raise TypeError(msg)
 
-    if not all(isinstance(item, int) for item in list_to_compress):
-        msg = f"All elements of the list {list_to_compress} must be integers"
-        raise TypeError(msg)
+    try:
+        sorted_list_to_compress = sorted(list_to_compress, key=_key)
```
gmuloc: key=_key is going to hinder the performance as sorted now needs to call a python funciton for every key when the version we had could use C optimisation

apparently doing this:

```python
z = list(list_to_compress)
if not all(isinstance(item, int) for item in z:
  raise...
sorted_z = sorted(z)
```

is better in termns of perf

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
```diff
@@ -303,14 +303,12 @@ def _get_l2_uplink(
...
                 uplink.peer_trunk_groups.append_unique(self.shared_utils.hostname)
 
-        uplink_vlans = set(self._vlans)
-        uplink_vlans = uplink_vlans.intersection(uplink_switch_facts._vlans)
-
+        uplink_vlans = self._uplink_vlans
         if self.shared_utils.configure_inband_mgmt or self.shared_utils.configure_inband_mgmt_ipv6:
             # Always add inband_mgmt_vlan even if the uplink switch does not have this vlan defined
-            uplink_vlans.add(self.shared_utils.node_config.inband_mgmt_vlan)
+            uplink_vlans = uplink_vlans.union((self.shared_utils.node_config.inband_mgmt_vlan,))
```
gmuloc, line 313: why? the original code is cleaner and does not create an intermedaite set right?

```suggestion
            uplink_vlans = uplink_vlans.add(self.shared_utils.node_config.inband_mgmt_vlan)
```

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
```diff
@@ -484,3 +482,15 @@ def uplink_switch_interfaces(self: EosDesignsFactsGeneratorProtocol) -> EosDesig
                 raise AristaAvdError(msg, host=uplink_switch)
 
         return uplink_switch_interfaces
+
+    @cached_property
+    def _uplink_vlans(self: EosDesignsFactsGeneratorProtocol) -> frozenset[int]:
+        """Return the intersection of local VLANs with VLANs across all uplink switches."""
+        uplink_vlans = self._vlans
+
+        uplink_switches = set(self.shared_utils.uplink_switches).intersection(self.shared_utils.all_fabric_devices)
```
gmuloc: we have uplink_peers today that is doing the first set I feel like this is what we should use so not sure why the all_fabric_devices was added - maybe this should be considered as an update for uplink_peers?

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -475,6 +475,40 @@ l2leaf:
           uplink_switch_ethernet_structured_config:
             description: description through uplink_switch_ethernet_structured_config(dc1.l2leaf7a)
 
+    # Minimal topology for asymmetric parent VLAN availability across split port-channel parents.
+    # Parent A has VLANs 3901 and 3902, parent B has only VLAN 3902, and the child has both VLANs.
+    # In devel the split port-channel allowed VLANs were calculated per uplink parent, so the child and
+    # parent A rendered 3901-3902 while parent B rendered 3902. This PR filters VLANs against all parent
+    # paths, so all three devices render 3902.
```
gmuloc: we probably can clean this up now to something more future proof.


```suggestion
    # Minimal topology for asymmetric parent VLAN availability across split port-channel parents.
    # Parent A has VLANs 3901 and 3902, parent B has only VLAN 3902, and the child has both VLANs.
    # This test verifies that only 3092 is configured on both uplinks.
```

python-avd/pyavd/j2filters/list_compress.py
ClausHolbechArista: I don't see that as a problem?

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
ClausHolbechArista: in non-inband cases we reuse the frozen set coming from self._uplink_vlans so that is much more efficient. Only on cases where we need to add the inband mgmt vlan we will create a local set here with the union. So for most cases this will be _more_ efficient than the old one.

python-avd/pyavd/j2filters/list_compress.py
gmuloc: well it is called list compress and the filter does not detail so at best it is an easter egg?

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
ClausHolbechArista: The check for all_fabric_devices was already in the old code. It is to guarantee that we can fetch the device facts below. If someone typed a wrong device. I think we didn't want this to be the failure point, since it could be invoked indirectly when parsing other devices.

gmuloc: during cross pollination we already check for the presence of the peer_generators

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
gmuloc: I dont follow as in non inband case we don't call this

gmuloc: ok cannot call `add` on the frozenset thats why

python-avd/pyavd/j2filters/list_compress.py
ClausHolbechArista: ok I will revert

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
ClausHolbechArista: ok using uplink_peers instead.

python-avd/pyavd/_eos_designs/eos_designs_facts/uplinks.py
gmuloc: then to avoid duplicating the frozenset we could do

```
from itertools import chain

...
uplink_vlans = chain(uplink_vlans, (self.shared_utils.node_config.inband_mgmt_vlan,))
```

ClausHolbechArista: no since that could inject the same vlan twice.

```sh
>>> from list_compress import list_compress
>>> list_compress([1,1,1,1])
'1,1,1,1'
>>> list_compress([1,2,3,1])
'1,1-3'
>>> 
```

gmuloc: yeah ok there is something in another package but that does not make sense to add a dependency. lets be less efficient for campus

python-avd/pyavd/_eos_designs/eos_designs_facts/vlans.py
```diff
@@ -249,26 +249,26 @@ def endpoint_trunk_groups(self: EosDesignsFactsGeneratorProtocol) -> EosDesignsF
         return EosDesignsFactsProtocol.EndpointTrunkGroups(natural_sort(self._endpoint_trunk_groups))
 
     @cached_property
-    def _vlans(self: EosDesignsFactsGeneratorProtocol) -> list[int]:
+    def _vlans(self: EosDesignsFactsGeneratorProtocol) -> frozenset[int]:
         """
-        Return list of vlans after filtering network services.
+        Return list of VLANs after filtering network services.
```
gmuloc:
```suggestion
        Return frozenset of VLANs after filtering network services.
```

gmuloc (approved):

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
```diff
@@ -475,6 +475,46 @@ l2leaf:
           uplink_switch_ethernet_structured_config:
             description: description through uplink_switch_ethernet_structured_config(dc1.l2leaf7a)
 
+    # Minimal topology for asymmetric parent VLAN availability across split port-channel parents.
+    # Parent A has VLANs 3901 and 3902, parent B has only VLAN 3902, and the child has both VLANs.
+    # This test verifies that only 3902 is configured on both uplinks.
```
carl-baillargeon: This is true for `dc1-l2-child-asymmetric-2` only. `dc1-l2-child-asymmetric-2` has both 3901-3902.

carl-baillargeon (approved):
Very minor comment otherwise LGTM!

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/DC1_FABRIC.yml
gmuloc: yeah that was a comment before we added the second one :) (thougu you copy pasted twice the same name :p )

gmuloc:
```suggestion
    # This test verifies that only 3902 is configured on both uplinks for `dc1-l2-child-asymmetric-2`
    # and that `dc1-l2-child-asymmetric-1` keeps the non-breaking behavior.
```

PR #7034, opened 2026-05-28
Feat(cv_deploy): Workspace rebase. Phase 2 (Split classes)

alexeygorbunov:
## Change Summary

Rebase/synchronization of the CloudVision Workspase. Phase 2. Splitting classes that need to be reset during Workspace synchronization phase

## Related Issue(s)

Fixes aristanetworks/avd-internal#523

## Component(s) name

`arista.avdcv_deploy`

## Proposed changes
1. Split the following classes in `Avd..` (frozen user input) and  mutable `CV..` (reflecting deployment progress) to allow intuitive reset of all settings set/learned within specific deploy_to_cv deployment cycle/attempt
2. Improve coverage to make codecov happy

## How to test
pytest

### CI runs:
1. one pair of py+core (6bf9ea1 WS rebase phase2: Add coverage for verify_devices_on_cv module)
  ✅  https://github.com/aristanetworks/avd-internal/actions/runs/26865048015
3.  full cv_deploy CI run (d6254bb Merge branch 'devel' into feat/cv_deploy-workspace-rebase-phase-2):
  ✅   https://github.com/aristanetworks/avd-internal/actions/runs/26977487102
  (run failed but only due to the presence of a single retry in the warnings: https://github.com/aristanetworks/avd-internal/actions/runs/26977487102/job/79624949477#step:12:7510 so everything actually succeeded)

### User Checklist

- N/A

### Repository Checklist


ansible_collections/arista/avd/extensions/molecule/cv_deploy/cc_false.yml
```diff
@@ -70,8 +70,8 @@
           ansible.builtin.assert:
             that:
               # Change control
-              - cv_deploy_results.change_control.requested_state == "pending approval"
-              - cv_deploy_results.change_control.requested_state == cv_deploy_results.change_control.state
+              - cv_deploy_results.change_control.avd_change_control.requested_state == "pending approval"
```
ClausHolbechArista: Hmm I don't like that we carry the state object through to the result. For these three classes, could you add a `get_result() -> dict` that outputs in the same form as we had before? So it would fish out fields from the inner avd_ class and some from the state class.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -239,16 +245,20 @@ class CVWorkspace:
...
+    build_warnings: AvdWorkspaceBuildWarningsConfig = field(default_factory=AvdWorkspaceBuildWarningsConfig)
+    """Configuration settings to control fetching and exposing Workspace build warnings."""
+
+
+@dataclass
+class CVWorkspace:
+    avd_workspace: AvdWorkspace
     state: Literal["pending", "built", "submitted", "build failed", "submit failed", "abandoned", "deleted"] | None = None
-    """The final state of the Workspace. Do not set this manually."""
+    """The final state of the Workspace."""
```
ClausHolbechArista: well it is more the last state or current state, since this will change through the course of the deployment.

ansible_collections/arista/avd/extensions/molecule/cv_deploy/cc_false.yml
ClausHolbechArista: If this is too much to do right now we can push this out to a later change.

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Changed to "current"

ansible_collections/arista/avd/extensions/molecule/cv_deploy/cc_false.yml
alexeygorbunov: Done, added get_result to 3 split classes. But this led to more changes (please check)

alexeygorbunov:
All proposed changes look good to me!

ClausHolbechArista (approved):

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -60,7 +62,7 @@ def as_grpclib_configuration(self) -> Configuration:
             return Configuration()
 
 
-@dataclass
+@dataclass(frozen=True)
```
carl-baillargeon: Can we change the servers field to be immutable as well?
```
servers : str | tuple[str, ...]
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -91,11 +91,48 @@ class CVChangeControl:
...
 
 
 @dataclass
-class CVChangeControlTemplate:
+class CVChangeControl:
+    avd_change_control: AvdChangeControl = field(default_factory=AvdChangeControl)
+    id: str | None = None
+    state: Literal["pending approval", "approved", "running", "completed", "deleted", "failed"] | None = None
+    name: str | None = None
+    description: str | None = None
```
carl-baillargeon, line 102: Are we updating these fields at runtime from CloudVision? If not we can just have pointers to the `AvdChangeControl` instance.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -91,11 +91,48 @@ class CVChangeControl:
...
+
+    @property
+    def change_control_template(self) -> AvdChangeControlTemplate | None:
+        return self.avd_change_control.change_control_template
+
+    @property
+    def requested_state(self) -> Literal["pending approval", "approved", "running", "completed", "deleted"]:
+        return self.avd_change_control.requested_state
+
+    def get_result(self) -> dict:
```
```suggestion
    def get_result(self) -> dict[str, Any]:
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -206,8 +243,8 @@ class CVWorkspaceDeviceBuildResult:
     """Configuration validation results."""
 
 
-@dataclass
-class CVWorkspaceBuildWarningsConfig:
+@dataclass(frozen=True)
```
carl-baillargeon, line 247: Similar comment here, let's consider using immutable fields if we can:
```
suppress_patterns: tuple[str, ...]
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -235,16 +272,58 @@ class CVWorkspace:
...
+
+    @property
+    def force(self) -> bool:
+        return self.avd_workspace.force
+
+    @property
+    def build_warnings(self) -> AvdWorkspaceBuildWarningsConfig:
+        return self.avd_workspace.build_warnings
+
+    def get_result(self) -> dict:
```
```suggestion
    def get_result(self) -> dict[str, Any]:
```

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -293,13 +375,40 @@ class CVDevice:
...
+    def hostname(self) -> str:
+        return self.avd_device.hostname
+
+    def get_result(self) -> dict:
+        return {
+            "hostname": self.hostname,
+            "serial_number": self.serial_number,
+            "system_mac_address": self.system_mac_address,
+            "_exists_on_cv": self._exists_on_cv,
+            "_streaming": self._streaming,
```
carl-baillargeon: These two fields don't need to be private anymore now that the inputs are separate.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -231,10 +234,10 @@ async def deploy(self, validated_args: dict, result: dict) -> dict:
...
                         **({"proxy_password": "<removed>"} if cloudvision.proxy_password is not None else {}),  # NOSONAR
                     },
-                    configs=[asdict(config) for config in eos_config_objects],
-                    device_tags=[asdict(device_tag) for device_tag in device_tag_objects],
-                    interface_tags=[asdict(interface_tag) for interface_tag in interface_tag_objects],
-                    cv_pathfinder_metadata=[asdict(metadata) for metadata in cv_pathfinder_metadata_objects],
+                    configs=[serialize(config) for config in eos_config_objects],
+                    device_tags=[serialize(device_tag) for device_tag in device_tag_objects],
+                    interface_tags=[serialize(interface_tag) for interface_tag in interface_tag_objects],
+                    cv_pathfinder_metadata=[serialize(metadata) for metadata in cv_pathfinder_metadata_objects],
```
carl-baillargeon: Shouldn't we use `serialize()` for cloudvision and static_config_manifest for consistency?

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -282,9 +361,12 @@ class DeployToCvResult:
     removed_device_tags: list[CVDeviceTag] = field(default_factory=list)
     removed_interface_tags: list[CVInterfaceTag] = field(default_factory=list)
 
+    def get_result(self) -> dict:
```
```suggestion
    def get_result(self) -> dict[str, Any]:
```

carl-baillargeon: Also, `DeployChangeControlResult` doesn't have `get_result()` but I don't think this model is being used at all. We could probably remove this model in this PR while we're here.

python-avd/pyavd/_cv/workflows/models.py
```diff
@@ -293,13 +375,40 @@ class CVDevice:
...
+        if not self.serial_number:
+            self.serial_number = self.avd_device.serial_number or None
+        if not self.system_mac_address:
+            self.system_mac_address = self.avd_device.system_mac_address or None
+
+    @property
+    def hostname(self) -> str:
+        return self.avd_device.hostname
+
+    def get_result(self) -> dict:
```
carl-baillargeon: I am not sure about the `get_result` name. Looking at the PR in general, I think `to_dict` or `as_dict` would be easier to understand.

python-avd/pyavd/_cv/workflows/utils.py
```diff
@@ -25,3 +27,18 @@ def extract_from_device_deployments(
         if device_deployment.cv_pathfinder_metadata is not None:
             cv_pathfinder_metadata.append(device_deployment.cv_pathfinder_metadata)
     return configs, device_tags, interface_tags, cv_pathfinder_metadata
+
+
+def serialize(obj: Any) -> Any:
```
carl-baillargeon: I think it would be easier to follow if we had a `get_result()` (or whatever we call it) on all `CV*` wrappers. Each object would know how to represent itself. For `DeployToCvResult`, we could build the dict manually by calling `get_result()` on nested `CV*` objects or `dataclasses.asdict()` as needed. It's more explicit, you see exactly what's in the output and it removes the need for this function.

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: I removed `frozen` from this class as we now also need to change `CVGRPCChannelConfiguration` and `CVGRPCKeepalives` referenced by `CloudVision`. Lets handle this in a separate PR to not inflate the scope of the current one (changes to `CloudVision` class are not really needed for the WS sync)

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Yes, we are updating this fields in `finalize_change_control_on_cv`

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Suggestion applied

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Suggestion applied

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Changed to tuple

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Done

ansible_collections/arista/avd/plugins/action/cv_workflow.py
alexeygorbunov: Non needed at the moment but is probably good for consistency and the future (if any children eventually get their own get_result()). Applied changes.

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Done

alexeygorbunov: Class is not used anywhere. Deleted it

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Name of the method was suggested by @ClausHolbechArista . I'll let Claus comment on this

python-avd/pyavd/_cv/workflows/utils.py
alexeygorbunov: Defining `get_result` for basic classes (like `CVGRPCKeepalives`, etc) wouldn't bring real benefits, just like defining `from_dict` for all of those just to have them all consistent.

> `For DeployToCvResult, we could build the dict manually by calling get_result() on nested CV* objects or dataclasses.asdict() as needed.`

atributes may have their own children with `get_result` implemented so in this case we would need to implement serialization logic inside every `get_result` implementation? 
@ClausHolbechArista

carl-baillargeon: To your first point, you don't need `get_result` for basic classes, you have the native `dataclasses.asdict()` to handle these. It would only be for wrapper classes that need special care.

To your second point, yes you would call `get_result` or `asdict` depending on the field.

ansible_collections/arista/avd/plugins/action/cv_workflow.py
ClausHolbechArista: No need now. this PR was not meant for general data model cleanup, but just to split one thing specifically for workspace rebase. please keep it focused.

python-avd/pyavd/_cv/workflows/models.py
ClausHolbechArista: Well there is a slight difference here. We don't return everything like as_dict would. We only return a subset relevant for the result. In the future this could change to return a result class instance I think.

python-avd/pyavd/_cv/workflows/utils.py
ClausHolbechArista: this PR is not for general result handling. We just needed it on the new ones since we now nest the input class and don't want to bloat the result surface. I am ok with renaming this to get_result as well, since as mentioned elsewhere it may only return a subset of the data if the class has a get_result method.

python-avd/pyavd/_cv/workflows/models.py
carl-baillargeon: In that case please add a docstring to these `get_result()` methods to explain the use case.

python-avd/pyavd/_cv/workflows/utils.py
carl-baillargeon: Ok I am good with renaming this to `get_result` but please update the function's docstring to explain that this is primarly used on our objects, honoring `get_result`, etc.

python-avd/pyavd/_cv/workflows/models.py
alexeygorbunov: Added a docstring for all get_result methods

python-avd/pyavd/_cv/workflows/utils.py
alexeygorbunov: Renamed serialize() to get_result() and added docstring explaining the purpose

ansible_collections/arista/avd/plugins/action/cv_workflow.py
```diff
@@ -42,10 +45,9 @@
         CVPathfinderMetadata,
         CVTimeOuts,
         CVWorkspace,
-        CVWorkspaceBuildWarningsConfig,
         DeployToCvResult,
     )
-    from pyavd._cv.workflows.utils import extract_from_device_deployments
+    from pyavd._cv.workflows.utils import extract_from_device_deployments, serialize
```
carl-baillargeon: Need to update this file as well :)
```suggestion
    from pyavd._cv.workflows.utils import extract_from_device_deployments, get_result
```

python-avd/tests/pyavd/cv/workflows/test_models.py
```diff
@@ -433,3 +487,250 @@ def test_as_grpclib_configuration_type_error_falls_back_to_default_configuration
             config = channel_config.as_grpclib_configuration()
         assert isinstance(config, Configuration)
         assert "gRPC keepalives will not be enabled" in caplog.text
+
+
+class TestGetResult:
```
carl-baillargeon: Maybe we should have separate test classes like `TestDeployToCvResult` and add a `test_get_result` method to them instead. Otherwise it feels like we are testing a class named `GetResult` :)

Also, if we add new methods to `DeployToCvResult` in the future, we would simply add test methods to `TestDeployToCvResult`. Better organized this way I think.

carl-baillargeon (changes requested):

ansible_collections/arista/avd/plugins/action/cv_workflow.py
alexeygorbunov: Thanks! Done

python-avd/tests/pyavd/cv/workflows/test_models.py
alexeygorbunov: Yes, makes sense! Split original test class into smaller ones (one per CV model class)

alexeygorbunov:
Full CI run (0af73fc WS rebase phase2: Address Carl's comments):
✅  https://github.com/aristanetworks/avd-internal/actions/runs/27330449513

carl-baillargeon (approved):

PR #7139, opened 2026-06-19
Fix(eos_cli_config_gen): ipv6_unicast_routing_exclamation_tracker

ccsnw:
## Change Summary

add a tracker instead of using first element of the vrf loop. it is possible that the first element doesn't have v6 routing enabled, so the exclamation mark would never be rendered.

## Related Issue(s)

Fixes #<N/A>

## Component(s) name

## Proposed changes
see above, just adding a tracker to avoid misbehaviour

## How to test
molecule test already present, exclamation mark now added

### User Checklist

- N/A

### Repository Checklist

- [o] My change requires a change to the documentation and documentation have been updated accordingly.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in controls for IPv6 routing configuration separators.
  * Added support for rendering combined IPv6 hardware and unicast-routing configuration blocks.
  * Added support for ensuring separators appear before IPv6 unicast-routing VRF sections.

* **Documentation**
  * Updated configuration reference documentation with the new options, defaults, availability, and behavior.
  * Expanded example device configurations to cover IPv6-enabled VRFs and routing scenarios.


ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/intended/configs/host1.cfg
```diff
@@ -5466,6 +5466,7 @@ ipv6 prefix-list ipv6_list1
 !
 ipv6 unicast-routing
 ipv6 hardware fib optimize prefixes profile internet
+!
```
gmuloc: this would be a breaking change as far as SemVer is concerned. Can you please implement an `eos_config_future` flag. https://avd.arista.com/6.2/docs/release-notes/6.x.x.html#eos-config-future

gmuloc (changes requested):

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/intended/configs/host1.cfg
ccsnw: Hi Guillaume
I can try, just need to find some time to take a look at how that's done.
However the change would not effectively break things, it would just add blank lines starting with exclamation marks (where needed).. so should not have impact when pushing via CVP/EAPI to a running switch?

ccsnw: figured it out, will update the schema etc and push the additional changes once ready!

gmuloc: > Hi Guillaume I can try, just need to find some time to take a look at how that's done. However the change would not effectively break things, it would just add blank lines starting with exclamation marks (where needed).. so should not have impact when pushing via CVP/EAPI to a running switch?

yeah and we understand that but we have collectively decided to get a bit stricter on the SemVer behavior and guaranteering _no change to generated configuration_ between minor release so we rely on these future flags to signal, this will be default in a future major release of AVD but you can opt-in for it. (you will see there is already a similar flag for the IP case .. :) )

ccsnw: all good! I've added the flag and an additional test in host 4.
old behaviour is kept unless new flag is set to true.
let me know if further edits are needed

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
```diff
@@ -18,6 +18,13 @@ keys:
           Available from AVD 6.2.0.
           Always render a '!' before the '(no) ip routing' command section.
           Without this the '!' is missing when only configuring routing for VRFs.
+      always_render_ipv6_unicast_routing_vrfs_separator:
+        type: bool
+        default: false
+        description: |-
+          Available from AVD 6.2.0.
```
MaheshGSLAB:
```suggestion
          Available from AVD 6.3.0.
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
```diff
@@ -4,9 +4,13 @@
  that can be found in the LICENSE file.
 #}
 {# eos - ipv6 unicast routing vrfs #}
+{% set v6_vrf = namespace(tracker_exclamation=false) %}
 {% for vrf in vrfs | arista.avd.natural_sort('name', ignore_case=false) %}
 {%     if vrf.ipv6_routing is arista.avd.defined(true) and vrf.name != 'default' %}
-{%         if loop.first %}
+{%         if not v6_vrf.tracker_exclamation and eos_config_future.always_render_ipv6_unicast_routing_vrfs_separator is arista.avd.defined(true) %}
+!
```
MaheshGSLAB: Rendering of the ! separator is conditional. If ipv6 hardware fib optimize prefixes profile internet is present, EOS renders ! before the first non-default VRF. Otherwise, the behavior is determined by ipv6 unicast-routing: when it is configured, no ! is rendered for any VRF.

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
```diff
@@ -4,9 +4,13 @@
  that can be found in the LICENSE file.
 #}
 {# eos - ipv6 unicast routing vrfs #}
+{% set v6_vrf = namespace(tracker_exclamation=false) %}
 {% for vrf in vrfs | arista.avd.natural_sort('name', ignore_case=false) %}
 {%     if vrf.ipv6_routing is arista.avd.defined(true) and vrf.name != 'default' %}
-{%         if loop.first %}
+{%         if not v6_vrf.tracker_exclamation and eos_config_future.always_render_ipv6_unicast_routing_vrfs_separator is arista.avd.defined(true) %}
```
MaheshGSLAB: This is a breaking change. The default value of `always_render_ipv6_unicast_routing_vrfs_separator` is false, which means users who do not explicitly enable this knob will never get the ! separator. Under the current behavior, the ! is always rendered for the first VRF (loop.first). As a result, this change alters the default rendering output and could impact existing configurations.

ccsnw: Hi @MaheshGSLAB 
"Under the current behavior, the ! is always rendered for the first VRF (loop.first). "
>> Under current behaviour, the "!" is only rendered when the first vrf (loop.first) also has vrf.ipv6_routing is arista.avd.defined(true) this set to true.. else it is not rendered.. cause loop.first only considers the for loop not the condition followed.. this is the problem that i'm trying to fix.

cause the first vrf in the list of vrfs could have vrf.ipv6_routing set to false.. then no "!" would be rendered even if the second vrf has set vrf.ipv6_routing to true...

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
ccsnw: Hi @MaheshGSLAB 
sorry but I struggle to follow you.

The current template of ipv6-unicast-routing-vrfs.j2 is

```
{% for vrf in vrfs | arista.avd.natural_sort('name', ignore_case=false) %}
{%     if vrf.ipv6_routing is arista.avd.defined(true) and vrf.name != 'default' %}
{%         if loop.first %}
!
{%         endif %}
ipv6 unicast-routing vrf {{ vrf.name }}
{%     endif %}
{% endfor %}
```

- it loops over the vrfs
- if the vrf has ipv6_routing set to true and vrf.name is NOT default
- and if that vrf is the very first vrf in the loop (loop.first)
 
then a "!" is rendered.

if the very first vrf in the list does not have ipv6_routing set to true or is named default, then "!" is never rendered.

Imho this is incorrect behaviour and my fix is trying to address it.

++

i don't see a relation to "ipv6 hardware fib optimize prefixes profile" at all (checked ipv6-hardware.j2 and ip-hardware.j2)

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
ccsnw: only thing i see is that i have to remove this condition "and eos_config_future.always_render_ipv6_unicast_routing_vrfs_separator is arista.avd.defined(false)" to mimic the current behaviour

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
```diff
@@ -18,6 +18,13 @@ keys:
           Available from AVD 6.2.0.
           Always render a '!' before the '(no) ip routing' command section.
           Without this the '!' is missing when only configuring routing for VRFs.
+      always_render_ipv6_unicast_routing_vrfs_separator:
+        type: bool
+        default: false
+        description: |-
+          Available from AVD 6.3.0.
+          Always render a '!' before the '(no) ipv6 unicast-routing vrf' command section.
+          Without this the '!' is missing when first VRF is not configured for ipv6_routing.
```
gmuloc:
```suggestion
          Without this the '!' is missing when the first VRF (once VRFs are sorted) is not configured for IPv6 routing.
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
MaheshGSLAB: Yes now we support both cases.

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-unicast-routing-vrfs.j2
alexeygorbunov: > Rendering of the ! separator is conditional. If ipv6 hardware fib optimize prefixes profile internet is present, EOS renders ! before the first non-default VRF. Otherwise, the behavior is determined by ipv6 unicast-routing: when it is configured, no ! is rendered for any VRF.

Hey @MaheshGSLAB ! On which version/platform do you see EOS rendering `!` between `ipv6 hardware fib optimize prefixes profile internet` and `ipv6 unicast-routing vrf <VRF_NAME>`?

From what i can see command `ipv6 hardware fib optimize prefixes profile internet` is only supported on the platforms that became unsupported starting 4.34+

By testing on 4.33.7M i see that EOS does not render any "!" between `ipv6 unicast-routing.*` and `ipv6 hardware fib optimize prefixes profile internet`:

```
<HIDDEN>...00:37:45(config-s-s10)#show session-config | grep "ipv6 unicast-routing vrf net" -A 5 -B 5
ipv6 prefix-list PL-LOOPBACKS-EVPN-OVERLAY-V6
   seq 10 permit 2001:ABCD:101:11::2:0/112 eq 128
   seq 21 permit 2001:ABCD:17:12::/64 eq 128
!
ipv6 unicast-routing
ipv6 unicast-routing vrf net
ipv6 unicast-routing vrf net-firewall
ipv6 unicast-routing vrf transit
!
maintenance
   profile interface maintenance-mode-rate-threshold
      rate-monitoring threshold 10000
<HIDDEN>...00:37:58(config-s-s10)#
<HIDDEN>...00:38:08(config-s-s10)#no ipv6 unicast-routing 
! Preserving static routes.  Use 'no ipv6 unicast-routing delete-static-routes' to clear them.
<HIDDEN>...00:38:19(config-s-s10)#show session-config | grep "ipv6 unicast-routing vrf net" -A 5 -B 5
!
ipv6 prefix-list PL-LOOPBACKS-EVPN-OVERLAY-V6
   seq 10 permit 2001:ABCD:101:11::2:0/112 eq 128
   seq 21 permit 2001:ABCD:17:12::/64 eq 128
!
ipv6 unicast-routing vrf net
ipv6 unicast-routing vrf net-firewall
ipv6 unicast-routing vrf transit
!
maintenance
   profile interface maintenance-mode-rate-threshold
      rate-monitoring threshold 10000
<HIDDEN>...00:38:28(config-s-s10)#
<HIDDEN>...00:38:36(config-s-s10)#ipv6 hardware fib optimize prefixes profile internet
! Please restart layer 3 forwarding agent to ensure IPv6 routes are optimized
<HIDDEN>...00:39:00(config-s-s10)#show session-config | grep "ipv6 unicast-routing vrf net" -A 5 -B 5
ipv6 prefix-list PL-LOOPBACKS-EVPN-OVERLAY-V6
   seq 10 permit 2001:ABCD:101:11::2:0/112 eq 128
   seq 21 permit 2001:ABCD:17:12::/64 eq 128
!
ipv6 hardware fib optimize prefixes profile internet
ipv6 unicast-routing vrf net
ipv6 unicast-routing vrf net-firewall
ipv6 unicast-routing vrf transit
!
maintenance
   profile interface maintenance-mode-rate-threshold
      rate-monitoring threshold 10000
<HIDDEN>...00:39:04(config-s-s10)#
<HIDDEN>...00:39:07(config-s-s10)#ipv6 unicast-routing 
<HIDDEN>...00:39:11(config-s-s10)#show session-config | grep "ipv6 unicast-routing vrf net" -A 5 -B 5
   seq 10 permit 2001:ABCD:101:11::2:0/112 eq 128
   seq 21 permit 2001:ABCD:17:12::/64 eq 128
!
ipv6 unicast-routing
ipv6 hardware fib optimize prefixes profile internet
ipv6 unicast-routing vrf net
ipv6 unicast-routing vrf net-firewall
ipv6 unicast-routing vrf transit
!
maintenance
   profile interface maintenance-mode-rate-threshold
      rate-monitoring threshold 10000
<HIDDEN>...00:39:14(config-s-s10)#
```

alexeygorbunov: Newer EOS (4.35.2F) shows same behavior (no "!"s separating "ipv6 unicast-routing.*" commands)(but now without `ipv6 hardware fib optimize prefixes profile internet`):

```
ag-dc1-leaf1a(config-s-s15)#sh session-config | egrep "ipv6 unicast-routing vrf A" -A 5 -B 5
!
   seq 10 permit 10.255.0.0/27 eq 32
   seq 20 permit 10.255.1.0/27 eq 32
!
ipv6 unicast-routing vrf A
ipv6 unicast-routing vrf B
!
mlag configuration
   domain-id DC1_L3_LEAF1
   local-interface Vlan4094
ag-dc1-leaf1a(config-s-s15)#
ag-dc1-leaf1a(config-s-s15)#ipv6 unicast-routing 
ag-dc1-leaf1a(config-s-s15)#sh session-config | egrep "ipv6 unicast-routing vrf A" -A 5 -B 5
   seq 10 permit 10.255.0.0/27 eq 32
   seq 20 permit 10.255.1.0/27 eq 32
!
ipv6 unicast-routing
ipv6 unicast-routing vrf A
ipv6 unicast-routing vrf B
!
mlag configuration
   domain-id DC1_L3_LEAF1
   local-interface Vlan4094
ag-dc1-leaf1a(config-s-s15)#
```

alexeygorbunov: So overall i think here we may be trying to replace existing incorrect rendering behavior with the new incorrect rendering behavior

ccsnw: @alexeygorbunov my change solely focuses on eliminating the flaw in this logic:

```
{% for vrf in vrfs | arista.avd.natural_sort('name', ignore_case=false) %}
{%     if vrf.ipv6_routing is arista.avd.defined(true) and vrf.name != 'default' %}
{%         if loop.first %}
!
{%         endif %}
ipv6 unicast-routing vrf {{ vrf.name }}
{%     endif %}
{% endfor %}
```

due to the "loop.first" condition, it will only render the "!" before the ipv6 routing block if the first defined vrf has vrf.ipv6_routing is arista.avd.defined(true) and the name is not default. 

so if the vrfs are in an order where the first one doesn't meet the condition, we will completely omit the "!"

example:

```
vrfs:
  - name: TESTV6_VRF
    ipv6_routing: false
  - name: TESTV6_VRF2
    ipv6_routing: true
```

would, with the current behaviour, not render "!" before "ipv6 unicast-routing vrf TESTV6_VRF2". which seems wrong in the EOS that I've tested. (4.33+)

++

If there is any other specific behaviour when we add "ipv6 hardware fib optimize prefixes profile internet" in addition. Then we could add this in a separate PR which would modify what I'm proposing?

Imho that would making it easier by addressing one problem after the other.. it would be different if there would already be a block of logic that treats ipv6 hardware fib optimize prefixes profile internet + ipv6 unicast routing in its on way today.. but I can't find it.

what do you think?

MaheshGSLAB: @alexeygorbunov I was saying if any of(`default ipv6 unicast routing` or `ipv6 hardware fib optimize prefixes profile`) is set then we do not need the ! for non default ipv6 unicast routing, as you can see it on session config.

alexeygorbunov: @gmuloc as all `ipv6 unicast-routing.*` and `ipv6 hardware fib optimize.*` commands currently rendered by `eos_cli_config_gen` are always represented by EOS in a single config block (they are not separated by `!`s) -> should we combine existing `ipv6_hardware.j2`, `ipv6-unicast-routing-vrfs.j2` and `ipv6-unicast-routing.j2` under an intermediate J2, like it is done for the SNMP community lists for example (https://github.com/aristanetworks/avd/blob/devel/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/all-community-lists.j2)?

gmuloc: If that helps solving the problem in the right way let;'s do it like this yes. @ccsnw do you want someone from the AVD team to take over?

ccsnw: would probably be faster than me trying and occupying further review time :D

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
```diff
@@ -18,6 +18,22 @@ keys:
...
+        default: false
+        description: |-
+          Available from AVD 6.4.0.
+          Always render a '!' before the '(no) ipv6 unicast-routing vrf' command section.
+          Without this the '!' is missing when the first VRF (once VRFs are sorted) is not configured for IPv6 routing.
+      render_combined_separator_for_ipv6_hardware_and_unicast_routing:
+        type: bool
+        default: false
+        description: |-
+          Render a single `!` separator after the combined IPv6 routing block when any of these are rendered:
```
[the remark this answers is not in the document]
MaheshGSLAB: updated

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
```diff
@@ -18,6 +18,22 @@ keys:
...
+        default: false
+        description: |-
+          Available from AVD 6.4.0.
+          Always render a '!' before the '(no) ipv6 unicast-routing vrf' command section.
+          Without this the '!' is missing when the first VRF (once VRFs are sorted) is not configured for IPv6 routing.
+      render_combined_separator_for_ipv6_hardware_and_unicast_routing:
+        type: bool
+        default: false
+        description: |-
+          Render a single `!` separator for the combined IPv6 routing block when any of these are rendered:
```
alexeygorbunov:
```suggestion
        description: |-
          Available from AVD 6.4.0.
          Render a single `!` separator for the combined IPv6 routing block when any of these are rendered:
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-hardware-and-unicast-routing.j2
```diff
@@ -0,0 +1,33 @@
...
+{%         endif %}
+{%         if ipv6_hardware.fib.optimize.prefixes.profile is arista.avd.defined %}
+ipv6 hardware fib optimize prefixes profile {{ ipv6_hardware.fib.optimize.prefixes.profile }}
+{%         endif %}
+{%         for vrf in ipv6_unicast_routing_vrfs %}
+ipv6 unicast-routing vrf {{ vrf.name }}
+{%         endfor %}
+{%     endif %}
+{% else %}
+{## IPv6 default vrf unicast routing #}
```
alexeygorbunov: Although `all-community-lists.j2` also uses `{##` syntax i think the one mostly used throughout the repo is `{#` (see `aaa.j2` for example)
```suggestion
{# IPv6 default vrf unicast routing #}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-hardware-and-unicast-routing.j2
```diff
@@ -0,0 +1,33 @@
...
+{%     include 'eos/ipv6-unicast-routing.j2' %}
+{## IPv6 hardware FIB optimize prefixes profile #}
```
alexeygorbunov:
```suggestion
{# IPv6 hardware FIB optimize prefixes profile #}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-hardware-and-unicast-routing.j2
```diff
@@ -0,0 +1,33 @@
...
+{%     include 'eos/ipv6-hardware.j2' %}
+{## IPv6 non default vrf unicast routing #}
```
alexeygorbunov:
```suggestion
{# IPv6 non default vrf unicast routing #}
```

python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/ipv6-hardware-and-unicast-routing.j2
```diff
@@ -0,0 +1,33 @@
+{#
+ Copyright (c) 2026 Arista Networks, Inc.
+ Use of this source code is governed by the Apache License 2.0
+ that can be found in the LICENSE file.
+#}
+{% if eos_config_future.render_combined_separator_for_ipv6_hardware_and_unicast_routing is arista.avd.defined(true) %}
+{%     set ipv6_unicast_routing_vrfs = vrfs | arista.avd.natural_sort('name', ignore_case=false) |
+        selectattr('ipv6_routing', 'defined') |
```
alexeygorbunov:
```suggestion
        selectattr('ipv6_routing', 'arista.avd.defined') |
```

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
```diff
@@ -18,6 +18,22 @@ keys:
...
           Always render a '!' before the '(no) ip routing' command section.
           Without this the '!' is missing when only configuring routing for VRFs.
+      always_render_ipv6_unicast_routing_vrfs_separator:
+        type: bool
+        default: false
+        description: |-
+          Available from AVD 6.4.0.
+          Always render a '!' before the '(no) ipv6 unicast-routing vrf' command section.
+          Without this the '!' is missing when the first VRF (once VRFs are sorted) is not configured for IPv6 routing.
+      render_combined_separator_for_ipv6_hardware_and_unicast_routing:
```
alexeygorbunov, line 21: Should we mention in description that this key overrides `always_render_ipv6_unicast_routing_vrfs_separator`?

alexeygorbunov (commented):
Final logic looks correct. Just minor comments

python-avd/pyavd/_eos_cli_config_gen/schema/schema_fragments/eos_config_future.schema.yml
MaheshGSLAB: updated

alexeygorbunov (approved):
LGTM

MaheshGSLAB:
@ccsnw Can you resolve the conflict

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/documentation/devices/host6.md
```diff
@@ -176,6 +177,7 @@ ip routing ipv6 interfaces vrf TENANT_B
 | MGMT | false |
 | TENANT_A | false |
 | TENANT_B | false |
+| TENANT_C | true |
```
gmuloc, line 180: config is missing below

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/documentation/devices/host6.md
gmuloc, line 176: we should use False / True capitalize everywhere or nowhere

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/documentation/devices/host6.md
```diff
@@ -156,6 +156,7 @@ interface Port-Channel2
 | MGMT | False |
 | TENANT_A | True |
 | TENANT_B | True (ipv6 interfaces) |
+| TENANT_C | - |
```
gmuloc, line 159: this is really False no?

gmuloc: we will tackle this ina follow up

ansible_collections/arista/avd/extensions/molecule/eos_cli_config_gen/documentation/devices/host6.md
gmuloc: we will tackle this ina follow up

gmuloc (approved):

PR #7206, opened 2026-07-02
Fix(eos_cli_config_gen): honor hide_passwords as role render setting

### LINKED ISSUE #7172: Issues with sanitized configurations post 6.X
### Enhancement summary

Prior to 6.X we could use the following logic in a playbook to produce configs sanitized and not-sanitized:
```
    - name: Generate device intended config and documentation
      ansible.builtin.import_role:
      vars:
        eos_cli_config_gen_configuration:
          hide_passwords: "{{ lookup('ansible.builtin.env', 'SANITIZE', default=false) | bool }}"
```
Reference: https://gitlab.aristanetworks.com/professional-services/ps-sku-l3ls/-/blob/main/EVPN-symmetric-IRB/playbooks/fabric-build.yml?ref_type=heads

Reason seems to be the following:
```
Root Cause

  The SANITIZE=true approach has been broken since the move to AVD 6.x. The mechanism is:

  1. Makefile sets SANITIZE=true in the environment
  2. fabric-build.yml reads it via Jinja2 env lookup and sets it as eos_cli_config_gen_configuration.hide_passwords in the import_role vars: block
  3. But eos_cli_config_gen role's internal validate_inputs task runs with read_from_input_dir: true (the default in AVD 6.x), which reads structured config from the YAML files generated by eos_designs
  4. Those YAML files don't contain eos_cli_config_gen_configuration — it's a role-level setting, not device data model output
  5. The eos_cli_config_gen action plugin passes only the validated structured config dict to pyavd.get_device_config() — Ansible task vars are not included
  6. The pyavd template defaults hide_passwords to False when eos_cli_config_gen_configuration is missing from the structured config

  In AVD 4.x/5.x this worked because Ansible's native Jinja2 rendering had access to all Ansible variables. In AVD 6.x, pyavd renders templates with only the structured config dict.
```
We would like to keep a similar approach when we can generate configurations sanitized or non-sanitized post 6.X

eos_cli_config_gen

### Use case example

ability to quickly produce sanitized/non-sanitized configs via our makefile used by pipelines

We would like to keep a similar approach when we can generate configurations sanitized or non-sanitized post 6.X

Workaround in 6.X is to add this in global_vars instead:
```
custom_structured_configuration_eos_cli_config_gen_configuration:
  hide_passwords: "{{ lookup('ansible.builtin.env', 'SANITIZE', default=false) | bool }}"
```

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Summary

Fix `eos_cli_config_gen` `hide_passwords` handling so configuration and documentation render settings work as role/inventory variables even when structured config is loaded from files.

This change:

- Adds PyAVD render configuration objects:
  - `RenderConfiguration`
  - `ConfigRenderConfiguration`
  - `DocRenderConfiguration`
- Passes `eos_cli_config_gen_configuration.hide_passwords` and `eos_cli_config_gen_documentation.hide_passwords` from the role into PyAVD as render configuration.
- Keeps existing structured config fallback behavior.
- Preserves `get_device_doc(..., add_md_toc=...)` backward compatibility.
- Avoids injecting role render settings into schema-shaped structured config data.
- Documents that the two `hide_passwords` settings are role render settings.

## Related Issue

Fixes #7172

## Tests

New narrow tests were added and can be run directly with pytest:

```bash
# From the root of the repo, if using `uv`, add `uv run` in front of each `pytest`
pytest python-avd/tests/pyavd/test_get_device_render_options.py
PYTHONPATH=. pytest ansible_collections/arista/avd/tests/unit/action/test_eos_cli_config_gen.py
```

Also run:

```bash
uv run pre-commit
```

## Checklist

- [x] Existing PyAVD API compatibility preserved.
- [x] get_device_doc(..., add_md_toc=...) preserved.
- [x] Added focused tests for render configuration behavior.
- [x] Added action plugin test for role-to-PyAVD pass-through.
### Repository Checklist

## Summary by CodeRabbit

* **New Features**
  * Added optional role/task settings to hide passwords in generated EOS configuration and device Markdown documentation.
  * Exposed new render configuration classes in the public pyavd API.
* **Documentation**
  * Updated role docs to explain how the hide-passwords settings are supplied and how they behave with structured config files.
* **Bug Fixes**
  * Improved consistency of password redaction between configuration and documentation rendering.
* **Tests**
  * Added unit tests validating option forwarding and hide-passwords rendering behavior.


alexeygorbunov (approved):
LGTM

python-avd/pyavd/get_device_config.py
```diff
@@ -40,5 +44,10 @@ def get_device_config(structured_config: EOSConfig | dict) -> str:
     if isinstance(structured_config, EOSConfig):
         structured_config = structured_config._as_dict()
 
+    template_vars: dict[str, Any] = structured_config
+    if configuration is not None and configuration.hide_passwords is not None:
+        template_vars = structured_config.copy()
```
ClausHolbechArista: this is expensive. Can we avoid that? I think template_vars accepts a chainmap if we get a dict as input. If we unpack EOSConfig we can just update directly.

Shivani-gslab (commented):
Agree with Claus, rest LGTM!

python-avd/pyavd/get_device_config.py
gmuloc: using chainmap and fixed the tempate_vars

ansible_collections/arista/avd/plugins/action/eos_cli_config_gen.py
```diff
@@ -25,12 +25,12 @@
...
 
 if TYPE_CHECKING:
-    from pyavd import get_device_config, get_device_doc
+    from pyavd import ConfigRenderConfiguration, DocRenderConfiguration, get_device_config, get_device_doc
     from pyavd._utils import strip_empties_from_dict, template
     from pyavd.j2filters import add_md_toc
 
 try:
-    from pyavd import get_device_config, get_device_doc
+    from pyavd import ConfigRenderConfiguration, DocRenderConfiguration, get_device_config, get_device_doc
```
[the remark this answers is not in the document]
gmuloc: I just addressed this in the latest commit I believ

Shivani-gslab (approved):

dgonzalez-arista:
All good i could test:
```
(test-avd-pr-7206) dgonzalez@dgonza0X4PW2m EVPN-symmetric-IRB % SANITIZE=true python -m dotenv run ansible-playbook playbooks/fabric-build.yml -i inventory/inventory.yml
```
and the intended configs have the passwords removed.

PR #7247, opened 2026-07-14
CI: replace pre-commit with prek

joelbreton2:
## Change Summary

Replace the AVD pre-commit runner workflows with prek while preserving the existing behavior for local checks, pull request checks, push checks, and automated hook revision update pull requests.

## Related Issue(s)

N/A

## Component(s) name

`CI / development tooling`

## Proposed changes

AVD previously used pre-commit for three separate workflows. This PR keeps each workflow in place and changes the implementation to prek:

- Local developer checks were run with `pre-commit install` and `pre-commit run`. They are now run with `prek install` and `prek run`.
- Pull request and push checks were run by `.github/workflows/pull-request-autofix.yml` using `pre-commit/action`. The same workflow now runs hooks with `j178/prek-action`.
- Hook revision update pull requests were handled by pre-commit.ci autoupdate. This is now handled by the new `.github/workflows/prek-update.yml` workflow, which runs `prek update` on a weekly schedule and opens or updates a pull request when hook revisions change.

Additional implementation details:

- `.pre-commit-config.yaml` is retained because `prek` supports the pre-commit configuration format.
- The generic `pre-commit/pre-commit-hooks` dependency is replaced with prek built-in hooks via `repo: builtin`.
- Remaining remote hook repositories are kept where AVD still depends on those hook implementations.
- The previous `ci:` block in `.pre-commit-config.yaml` was specific to pre-commit.ci. Its autoupdate behavior is replaced by the scheduled `prek update` workflow.

## How to test

Focused validation was run with prek:

```bash
uvx prek run yamllint --files .github/workflows/pull-request-autofix.yml .github/workflows/prek-update.yml
uvx prek run markdownlint-cli2 --files docs/contribution/development-tooling.md
```

Additional focused checks were run earlier during the migration:

- prek built-in hooks on changed files
- markdownlint on edited documentation
- ruff and ruff-format on edited Python helper strings
- `git diff --check`

### User Checklist

- N/A

### Repository Checklist

### DIFF
--- a/.github/workflows/prek-update.yml
+++ b/.github/workflows/prek-update.yml
@@ -1,0 +1,68 @@
+---
+name: prek update
+
+"on":
+  schedule:
+    - cron: "0 16 * * 1"
+  workflow_dispatch:
+
+concurrency:
+  group: prek-update
+  cancel-in-progress: false
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update:
+    # Replaces the former pre-commit.ci autoupdate role.
+    name: Update prek hook revisions
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repo
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: devel
+          persist-credentials: true
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.12"
+
+      - name: Install prek
+        run: python -m pip install "prek>=0.4.9"
+
+      - name: Update hook revisions
+        run: prek update
+
+      - name: Create pull request
+        env:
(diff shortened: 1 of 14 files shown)


.github/workflows/prek-update.yml
```diff
@@ -0,0 +1,68 @@
...
+          ref: devel
+          persist-credentials: true
+
+      - name: Set up Python
+        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+        with:
+          python-version: "3.12"
+
+      - name: Install prek
+        run: python -m pip install "prek>=0.4.9"
```
gmuloc, line 35: do not hardcode prek version in the github workflow instead create a dependency group called prek in the top level pyproject.toml and install here

ClausHolbechArista: Should we not get the proper action allowed instead?

gmuloc: churn? :)

carlbuchmann: We will try the official action, I will request IT!

PR #7249, opened 2026-07-16
Refactor(eos_designs): Refactor eos_designs_structured_config AVD action plugin to use the new AvdActionPlugin base class

MaheshGSLAB:
## Change Summary

Refactor eos_designs_structured_config AVD action plugin to use the new AvdActionPlugin base class

## Related Issue(s)

Fixes #<ISSUE ID>

## Component(s) name

`arista.avd.<role-name>`

## Proposed changes
Refactor eos_designs_structured_config AVD action plugin to use the new AvdActionPlugin base class

## How to test
CI should green

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved structured configuration error reporting with clearer plugin execution context while preserving the underlying failure details.
  * Missing validated inputs or facts now report explicit file-not-found errors.
  * Structured config output is written deterministically as YAML or JSON based on the destination extension.
  * Invalid `list_merge` values now show clearer, formatted validation messages.

* **Tests**
  * Strengthened assertions for error messages, exception chaining, warnings, and missing-file scenarios; updated integration and negative test expectations.

* **Chores**
  * CI now runs related Molecule jobs when action plugin code changes are detected.


ansible_collections/arista/avd/tests/integration/targets/eos_designs_documentation/tests/test_missing_structured_cfg_file.yml
```diff
@@ -36,5 +36,5 @@
     that:
       - result is success
       - result.changed == true
-      - result.warnings | length > 0
-      - result.warnings[0] == "Could not find structured config files for 'testhost'. The documentation may be incomplete."
+      - result.warnings is not defined or result.warnings | length > 0
+      - result.warnings is not defined or result.warnings[0] == "Could not find structured config files for 'testhost'. The documentation may be incomplete."
```
carl-baillargeon: Why are these changes required?

ansible_collections/arista/avd/tests/unit/action/test_eos_designs_structured_config.py
```diff
@@ -115,15 +115,19 @@ def test_run_emits_no_warnings_baseline(action_module: Callable[..., ActionModul
...
     with (
         patch(f"{MODULE_PATH}.HAS_PYAVD", new=False),
         patch("ansible.plugins.action.ActionBase.run", return_value={}),
         pytest.raises(
             AnsibleActionFail,
-            match=r"The 'arista.avd.eos_designs_structured_config' plugin requires the 'pyavd' Python library. Got import error",
+            match=(
+                r"Error during plugin 'arista.avd.eos_designs_structured_config' execution: "
+                r"The 'arista.avd.eos_designs_structured_config' plugin requires the 'pyavd' "
+                r"Python library. Got import error"
```
carl-baillargeon: We could improve this to avoid repeating the plugin name twice.

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -100,10 +86,10 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
                 templar=self.templar,
                 digital_twin=digital_twin,
             )
-        except Exception as error:
-            raise_action_fail(str(error), error)
 
-        output = structured_config._as_dict()
+            output = structured_config._as_dict()
+        except Exception as error:
+            raise RuntimeError(str(error)) from error
```
carl-baillargeon: We probably don't need to re-raise here anymore since we catch Exception in the base class.

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
```diff
@@ -148,7 +134,7 @@ def run(self, tmp: Any = None, task_vars: dict | None = None) -> dict:
                 try:
                     merge(output, *template_result_data, list_merge=list_merge, schema=output_schema)
                 except Exception as error:
-                    raise_action_fail(str(error), error)
+                    raise RuntimeError(str(error)) from error
```
carl-baillargeon: Same comment as above.

ansible_collections/arista/avd/tests/unit/action/test_eos_designs_structured_config.py
MaheshGSLAB: updated

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
MaheshGSLAB: removed

ansible_collections/arista/avd/plugins/action/eos_designs_structured_config.py
MaheshGSLAB: removed

ansible_collections/arista/avd/tests/integration/targets/eos_designs_documentation/tests/test_missing_structured_cfg_file.yml
MaheshGSLAB: In AVDActionPlugin.run(), result["warnings"] is only created when Python warnings are captured (i.e. captured_warnings is non-empty).

The missing structured-config scenario calls self.logger.warning(...), which emits a log warning rather than a Python warnings.warn(...). As a result, the warning is written to the task output (and to result["logs"]["warnings"] when save_logs: true is enabled), but result["warnings"] is not created.

Therefore, the integration test assertion must be defensive (for example, checking that result.warnings is defined before accessing its length), since the absence of result.warnings is expected in this code path.

Without this guard, the test fails with:
```
TASK [eos_designs_documentation : Validate result] *****************************
fatal: [testhost]: FAILED! => {"msg": "The conditional check 'result.warnings | length > 0' failed. The error was: error while evaluating conditional (result.warnings | length > 0): 'dict object' has no attribute 'warnings'"}
```

Shivani-gslab (approved):
LGTM

ansible_collections/arista/avd/tests/integration/targets/eos_designs_documentation/tests/test_missing_structured_cfg_file.yml
carl-baillargeon: > In AVDActionPlugin.run(), result["warnings"] is only created when Python warnings are captured (i.e. captured_warnings is non-empty).
> 
> The missing structured-config scenario calls self.logger.warning(...), which emits a log warning rather than a Python warnings.warn(...). As a result, the warning is written to the task output (and to result["logs"]["warnings"] when save_logs: true is enabled), but result["warnings"] is not created.
> 
> Therefore, the integration test assertion must be defensive (for example, checking that result.warnings is defined before accessing its length), since the absence of result.warnings is expected in this code path.
> 
> Without this guard, the test fails with:
> 
> ```
> TASK [eos_designs_documentation : Validate result] *****************************
> fatal: [testhost]: FAILED! => {"msg": "The conditional check 'result.warnings | length > 0' failed. The error was: error while evaluating conditional (result.warnings | length > 0): 'dict object' has no attribute 'warnings'"}
> ```

This is true for ansible <2.19 only. In ansible 2.19+, warnings emitted via `display.warning()` are captured and attached to the task result, not lost.

This test is not related to this PR, it should've been caught by CI when we migrated `eos_designs_documentation` to the base class in https://github.com/aristanetworks/avd/pull/7198. We are not catching this in CI because we are running integration tests with latest ansible-core: https://github.com/aristanetworks/avd/blob/81535a716738e9cc13cf7da1e1e731f4a28823ae/.github/workflows/pull-request-management.yml#L553

Anyways let's revert these changes for this PR to unblock it:
```suggestion
      - result.warnings | length > 0
      - result.warnings[0] == "Could not find structured config files for 'testhost'. The documentation may be incomplete."
```

Then a separate PR to also run integration tests on <2.19 in CI and fix what is breaking.

@gmuloc for visibility.

carl-baillargeon (approved):

gmuloc (changes requested):
merge commit keeps on failing because of eos_negative_unit_tests failures.I think with your change the expected error messages are changed. please confirm @MaheshGSLAB

MaheshGSLAB:
> merge commit keeps on failing because of eos_negative_unit_tests failures.I think with your change the expected error messages are changed. please confirm @MaheshGSLAB

@gmuloc Negative tests fails because it adds `Error during plugin 'arista.avd.eos_designs_structured_config' execution:` on each error and this comes from https://github.com/aristanetworks/avd/blob/devel/ansible_collections/arista/avd/plugins/plugin_utils/utils/avd_action_plugin/avd_action_plugin.py#L109 AVDActionPlugin

Vibhu-gslab (approved):

gmuloc:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/converge.yml (1)> `17-1174`: _🎯 Functional Correctness_ | _🟡 Minor_ | _⚡ Quick win_
> > > **Keep the error contract exact.**
> > > `endswith` allows an incorrect error prefix or originating plugin to pass every negative scenario. Update each affected `expected_error_message` to include the standardized `AvdActionPlugin` prefix where applicable, then retain `==` so these tests validate the full failure contract.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In
> > > `@ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/converge.yml`
> > > around lines 17 - 1174, Replace every negative-test assertion using
> > > ansible_failed_result.msg.endswith(expected_error_message) with exact equality,
> > > and update each affected expected_error_message to include the standardized
> > > AvdActionPlugin prefix where applicable. Apply this consistently across the
> > > eos_designs negative-test scenarios while preserving the existing expected
> > > message text and test structure.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In
> `@ansible_collections/arista/avd/extensions/molecule/eos_designs_negative_unit_tests/converge.yml`:
> - Around line 17-1174: Replace every negative-test assertion using
> ansible_failed_result.msg.endswith(expected_error_message) with exact equality,
> and update each affected expected_error_message to include the standardized
> AvdActionPlugin prefix where applicable. Apply this consistently across the
> eos_designs negative-test scenarios while preserving the existing expected
> message text and test structure.
> ```
> 
> ℹ️ Review info

@coderabbitai - this is one is a no given that the pytest version of these tests expect the message without the prefix. In the long run we want to move to pytest only anyways for what we are testing so we will survive with endswith for now in molecule. thanks

gmuloc (approved):

PR #7268, opened 2026-07-21
Fix(eos_designs): Fix the default originate not rendered properly for bgp peers in VRF default under network_services

### LINKED ISSUE #7258: Bug(eos_designs): default originate not rendered properley for bgp peers in VRF `default` under `network_services`(TBC)
### Issue Summary

Adding `default_originate` under specific bgp peers. With out the `enabled: true` the configuration does not get created though config gen but is in the structured config.

```
- ip_address: 4.4.4.0
  peer_group: EXTERNAL
  remote_as: "100"
  description: dc1-edge1
  default_originate:
    enabled: true    # <-- this is required
    always: true
  nodes: [ www ]
```

`eos_designs`

### AVD version

6.3.0

### Ansible version

_No response_

### Python version

3.14

### How do you run AVD ?

### Steps to reproduce

```shell
- ip_address: 4.4.4.0
  peer_group: EXTERNAL
  remote_as: "100"
  description: dc1-edge1
  default_originate:
    always: true
  nodes: [ www ]
```

### Relevant log output

The problem is the two different schemas:

- `eos_designs` `bgp_peers.default_originate` has keys: `always`, `route_map` (no `enabled`)
- `eos_cli_config_gen` `router_bgp.neighbors.default_originate` has keys: `enabled`, `always`, `route_map`
- `eos_cli_config_gen address_family_ipv4.neighbors.default_originate` has keys: `always`, `route_map` (no `enabled`)

The cast at line 235 relays `default_originate` with `always: true` into `router_bgp.neighbors`, but the Jinja2 template at that level checks for `enabled: true` which `eos_designs` never sets. 
The `address_family_ipv4.neighbors` template (line 1216) would render it without needing `enabled`, but the peer only gets added to `address_family_ipv4.neighbors` with `activate: true` (line 227-231) — `default_originate` is not carried there.

— `eos_designs` puts `default_originate` into `router_bgp.neighbors` for VRF default, but the template expects `enabled: true` which eos_designs doesn't set. 
The schema $ref points to the `address_family_ipv4` version (which doesn't need `enabled`), but the code casts it to the top-level `router_bgp.neighbors` (which does).

### Contributing Guide

MaheshGSLAB:
## Change Summary

Fix the default originate not rendered properley for bgp peers in VRF default under network_services

## Related Issue(s)

Fixes #7258 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Fix the default originate not rendered properley for bgp peers in VRF default under network_services. Updated the code to add enabled: True for structure config in case of always or route_map is set.

## How to test
Run molecule

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

* **New Features**
  * Added the ability to always originate default routes to selected BGP neighbors and peer groups.
  * Enabled additional BGP neighbors and applied outbound policy for default-route origination, including next-hop setting via route-map.
* **Bug Fixes**
  * Improved rendering so `default-originate` is explicitly enabled in global and VRF validation scenarios when configured for “always”.
* **Tests**
  * Updated intended configs and inventory fixtures to reflect the new default-route origination behavior.


ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/dc1-bl1a.yml
```diff
@@ -585,6 +585,7 @@ router_bgp:
       maximum_routes: 0
       maximum_routes_warning_only: true
       default_originate:
+        enabled: true
         always: false
```
ClausHolbechArista: are you sure you want this with always false?

ClausHolbechArista: ah yes the route-map

gmuloc:
@MaheshGSLAB - discussing it further with @ClausHolbechArista it feels like the right fix would be to add an `enabled: <bool, default=True>` key in the three places in the schema where we are today not considering an enabled key in the template (https://github.com/aristanetworks/avd/blob/f1b3115c900de149d4193f1a931b22c274dbdcde/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/router-bgp.j2#L1154 and 2 others) 

right now it looks like you could / would be insterting `enabled` in places where `eos_cli_config_gen` does not support it (though we don't see it fail in CI yet).

Can you take a look at this solution?

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -230,9 +230,11 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
                         prefix_list_in=bgp_peer.prefix_list_in,
                         prefix_list_out=bgp_peer.prefix_list_out,
                     )
-
                     # Below we recast directly to eos_cli_config_gen. Losing incompatible keys, but relaying everything else.
+                    # We have default in eos_designs schema for bgp_peer.default_originate but not in eos_cli_config_gen
                     bgp_peer_config = bgp_peer._cast_as(bgp_vrf.NeighborsItem, ignore_extra_keys=True)
+                    if bgp_peer._get_defined_attr("default_originate") is not Undefined and bgp_peer.default_originate.enabled:
```
```suggestion
                    if bgp_peer._get_defined_attr("default_originate") is not Undefined:
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -230,9 +230,11 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
...
+                        bgp_peer_config.default_originate.enabled = True
```
```suggestion
                        bgp_peer_config.default_originate.enabled = bgp_peer.default_originate.enabled
```

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -230,9 +230,11 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
                         prefix_list_in=bgp_peer.prefix_list_in,
                         prefix_list_out=bgp_peer.prefix_list_out,
                     )
-
                     # Below we recast directly to eos_cli_config_gen. Losing incompatible keys, but relaying everything else.
+                    # We have default in eos_designs schema for bgp_peer.default_originate but not in eos_cli_config_gen
                     bgp_peer_config = bgp_peer._cast_as(bgp_vrf.NeighborsItem, ignore_extra_keys=True)
```
gmuloc, line 235: this will fail (_cast_as) when EOS Config does not have a default

gmuloc:
> @MaheshGSLAB - discussing it further with @ClausHolbechArista it feels like the right fix would be to add an `enabled: <bool, default=True>` key in the three places in the schema where we are today not considering an enabled key in the template (
> 
> https://github.com/aristanetworks/avd/blob/f1b3115c900de149d4193f1a931b22c274dbdcde/python-avd/pyavd/_eos_cli_config_gen/j2templates/eos/router-bgp.j2#L1154
> 
> and 2 others)
> right now it looks like you could / would be insterting `enabled` in places where `eos_cli_config_gen` does not support it (though we don't see it fail in CI yet).
> 
> Can you take a look at this solution?

given the _cast_as limitation let's move back to injecting `enabled: true` for VRF default where required and not changing any schema. Please log an issue for AVD 7.0 to align the model with a required enabled key everywhere.

thompsno (commented):
Have tested this on the configuration i was using before when it failed, and now it processes the default_originate in bgp_peers with in the default vrf. Pushed changed through CVP. Test EOS devices rendered configuration and advertised df routes as expected and required.

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -242,6 +242,10 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
                         if bgp_peer_config.default_originate and not bgp_peer_config.default_originate.route_map:
                             bgp_peer_config.default_originate.route_map = route_map
 
+                    # Render enabled: True in case of always or route_map set for bgp peer
+                    if (bgp_peer_config.default_originate.route_map or bgp_peer.default_originate.always) and isinstance(bgp_vrf, EosCliConfigGen.RouterBgp):
+                        bgp_peer_config.default_originate.enabled = True
+
```
alexeygorbunov, line 251: Not sure about conditional dependency on the truthy `always` flag. `always: true` - means we always generate artificial default route. `always: false` means we advertise default route (without changing any path attributes) from BGP RIB if it is present in it otherwise we generate and advertise artificial default route. `always` key is overall not a trigger of advertising. It just defines what/how we advertise.

Current code will lead to not setting `bgp_peer_config.default_originate.enabled = True` when inputs have `bgp_peer.default_originate.always: False` and neither `bgp_peer.set_ipv4_next_hop` nor `bgp_peer.set_ipv6_next_hop` set which is incorrect.

```suggestion
                    # Render enabled: True in case of always or route_map set for bgp peer
                    if (bgp_peer_config.default_originate.route_map or bgp_peer.default_originate.always is not None) and isinstance(bgp_vrf, EosCliConfigGen.RouterBgp):
                        bgp_peer_config.default_originate.enabled = True

```

gmuloc: good catch - @MaheshGSLAB please implement this and make sure there is a test for it. We really need to add a proper `enabled` in AVD 7.0 as well for this as it is confusing to do:

```
default_originate:
  always: false
```

to get 

```
neighbor x.x.x.x default-originate
```

:)

MaheshGSLAB: updated the code and added tests

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/structured_configs/bgp-from-network-services-2.yml
```diff
@@ -117,6 +123,18 @@ router_bgp:
...
+      always: true
+  - ip_address: 12.12.1.2
+    peer_group: MYPEERGROUP
+    metadata:
+      validate_state: false
+    route_map_out: RM-default-12.12.1.2-SET-NEXT-HOP-OUT
+    default_originate:
+      enabled: true
+      always: false
+      route_map: RM-default-12.12.1.2-SET-NEXT-HOP-OUT
```
gmuloc: hmm intresting - we are picking up the same route-map here as for the neighbors

gmuloc: ah yes: https://github.com/aristanetworks/avd/pull/7268/changes#diff-df0c4224cffd55bd6c25d150174543480ecd2f22a5a675748c507e1191b09c87R243 this line ok.

gmuloc: hmmm this may need to be better explained at some point I suppose but not the staet of this PR

python-avd/pyavd/_eos_designs/structured_config/network_services/router_bgp.py
```diff
@@ -242,6 +242,13 @@ def _router_bgp_vrfs(self: AvdStructuredConfigNetworkServicesProtocol) -> None:
                         if bgp_peer_config.default_originate and not bgp_peer_config.default_originate.route_map:
                             bgp_peer_config.default_originate.route_map = route_map
 
+                    # Render enabled: True in case of always or route_map set for bgp peer
+                    # TODO: 7.0 Add default_originate.enabled key all around the data models.
+                    if (bgp_peer_config.default_originate.route_map or bgp_peer.default_originate.always is not None) and isinstance(
```
Vibhu-gslab, line 247: Just wondering.. shouldn't `bgp_peer.default_originate.always is not None` -> `bgp_peer.default_originate.always is True` ?

```suggestion
                    if (bgp_peer_config.default_originate.route_map or bgp_peer.default_originate.always is True) and isinstance(
```

gmuloc: please read Alexey's comment above :)

Vibhu-gslab: Ahhh!! make sense, Thanks @gmuloc :)

Vibhu-gslab (approved):

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/BGP_FROM_NETWORK_SERVICES.yml
```diff
@@ -81,6 +81,20 @@ tenants:
           - ip_address: 10.10.1.2
             peer_group: MYPEERGROUP
             nodes: [bgp-from-network-services-1, bgp-from-network-services-2]
+            # This will render enabled: True in structured config
+            default_originate:
+              always: true
+          - ip_address: 11.11.1.2
+            peer_group: MYPEERGROUP
+            nodes: [bgp-from-network-services-1]
+            # This will render enabled: True in structured config
```
laxmikantchintakindi, line 90: There is no default originate in this peer.

laxmikantchintakindi (approved):
Added minor comment. LGTM.

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/BGP_FROM_NETWORK_SERVICES.yml
MaheshGSLAB: updated the unit tests

gmuloc (approved):

PR #7270, opened 2026-07-21
Fix(eos_designs): Treat uplink EVPN gateway peers as core peers only

gmuloc:
## Change Summary

Fix EVPN gateway remote peers overlapping with uplink-derived EVPN route servers by treating those peers only as gateway core peers on the local node.

## Related Issue(s)

Fixes #7261

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

- Exclude `evpn_gateway.remote_peers` from the default `evpn_route_servers` derived from `uplink_switches`.
- Preserve explicit `evpn_route_servers` behavior.
- Add validation preventing the same hostname from being explicitly configured under both `evpn_route_servers` and `evpn_gateway.remote_peers` on the same node.
- Update generated schema documentation for the input restriction.
- Update molecule expected configs to show the local gateway peer using `EVPN-OVERLAY-CORE`.
- Add negative test inputs for explicit overlapping route-server and gateway remote-peer configurations.

This PR fixes the local duplicate BGP neighbor generation when an uplink switch is also configured as an EVPN gateway remote peer. It does not add reciprocal EVPN gateway peer generation on the remote side; that is planned as follow-up work.

## How to test

Relevant validation:

```bash
cd ansible_collections/arista/avd && molecule test -s eos_designs_unit_tests
cd ansible_collections/arista/avd && molecule test -s eos_designs_negative_unit_tests
```

### User Checklist

- [ ] Reciprocal EVPN gateway peer generation on the remote side is intentionally out of scope and will be handled in a follow-up PR.

### Repository Checklist

- [x] My code has been rebased from devel before I start
- [x] I have read the CONTRIBUTING (https://avd.arista.com/devel/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and the documentation has been updated accordingly.
- [x] I have updated molecule CI (https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/extensions/molecule) testing accordingly. (check the box if not applicable)

## Summary by CodeRabbit

- **New Features**
  - When no explicit EVPN route-server list is provided, route-server selection now excludes uplink switches that are configured as EVPN gateway remote peers.
  - Remote peers that are also uplinks are treated as EVPN Gateway core peers.
- **Bug Fixes**
  - Added validation to prevent configuring the same hostname as both an EVPN route server and an EVPN gateway remote peer.
- **Documentation**
[...]

PR #7279, opened 2026-07-23
Feat(eos_designs): Render reverse EVPN gateway core peers when the peer is in the inventory

### LINKED ISSUE #7248: Feat(eos_designs): Fix EVPN-OVERLAY-PEER/CORE configuration when EVPN-GW Node is an uplink_switch
### Enhancement summary

Add support to eos_designs for native support of an L3LS EVPN/VXLAN multi-pod fabric, with super spines, where the super spines are defined as EVPN gateway nodes.

eos_designs

### Use case example

In some L3LS EVPN/VXLAN multi-pod fabrics, each pod could have between 1 and n edge gateways/border leafs, which connect up to a set of super spines. In this fabric, the super spines role is to act as an evpn route reflector either as part of the broader evpn domain, or for routes originated as an evpn gateway. In this instance, the super spines will be defined as an evpn gateway peer, but they will also be defined as an uplink_switch, since they are directly connected to the pod edge gateways. As such, AVD will error with a peer group overlap between putting the super spines in the EVPN-OVERLAY-PEERS vs EVPN-OVERLAY-CORE peer group.

A workaround for this error is to define the local pod spines as evpn_route_servers. This clears the error and builds the correct configs from the perspective of the pod edge gateway, however, it causes the reverse EVPN peering from the super spines to the pod edge gateway to not be configured.

I think the following would be a good solution to this issue:

- Not having to define the local pod spines as evpn_route_servers in the pod edge gateway data model.
- A check where if an uplink_switch is also defined in the evpn_gateway.remote_peers list:
      - The uplink_switch/evpn gw remote peer is added to the correct bgp peer group [EVPN-OVERLAY-CORE]
      - The uplink_switch/evpn gw renders the reverse EVPN peer config.

Currently using structured config to define a peer-filter and bgp listen range on the super spines.

### Additional context

_No response_

### Contributing Guide

gmuloc:
## Change Summary

Adds future-gated reverse EVPN Gateway core peering for inventory-based `evpn_gateway.remote_peers`, with documentation and molecule coverage for reverse peer filtering, overlap validation, undeployed peer shutdown, and remote peers configured on nodes without EVPN overlay.

## Related Issue(s)

Fixes #7248

## Component(s) name

  `arista.avd.eos_designs`, `pyavd`

## Proposed changes

  - Add `avd_design_future.configure_reverse_evpn_gateway_remote_peers` to auto-render reverse EVPN Gateway core BGP peerings for inventory peers referenced by hostname without manual `ip_address` or `bgp_as` overrides.
  - Add `avd_design_future.raise_for_evpn_gateway_remote_peers_without_evpn` to optionally reject `evpn_gateway.remote_peers` when the local node does not have EVPN overlay enabled.
  - Track reverse EVPN Gateway remote peer clients in eos_designs facts and reject route-server/remote-peer overlap in both local and reverse directions.
  - Filter reverse EVPN Gateway peers from normal EVPN route-server clients, skip reciprocal local `remote_peers`, and ignore reverse peers whose EVPN role is not server/client.
  - Update generated schema, facts schema, and public documentation for the new behavior.
  - Add molecule coverage and intended artifacts for reverse peer filtering, no-EVPN default handling, and overlap validation.

## How to test

molecule

### Repository Checklist

  - [x] I have read the CONTRIBUTING (https://avd.arista.com/devel/docs/contribution/overview.html) document.
  - [x] My change requires a change to the documentation and the documentation has been updated accordingly.
  - [x] I have updated molecule CI (https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/extensions/molecule) testing accordingly. (check the box if not applicable)

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in mode to automatically render reverse EVPN Gateway core peerings for eligible remote peers.
  * Added support for reverse peerings across both eBGP and iBGP overlays.
  * Exposed resolved remote-peer and reverse peer-client details in design facts.
* **Bug Fixes**
  * Prevented conflicts between EVPN remote-peer and route-server definitions.
  * Added validation for unsupported EVPN roles and undeployed peer shutdown behavior.
* **Documentation**
  * Documented the new option and related peer-handling behavior.
* **Tests**
  * Expanded coverage for reverse-peer and validation scenarios.


python-avd/pyavd/_eos_designs/schema/schema_fragments/avd_design_future.schema.yml
```diff
@@ -81,6 +87,12 @@ keys:
           Available from AVD 6.2.0.
           Raise an error if an L3 Port-Channel is configured without any member interfaces.
         default: false
+      raise_for_evpn_gateway_remote_peers_without_evpn:
```
gmuloc: let's include this in the future `configure_reverse_evpn_gateway_remote_peers` to simplify

python-avd/pyavd/_eos_designs/eos_designs_facts/__init__.py
```diff
@@ -345,6 +346,35 @@ def _populate_evpn_route_server_clients_on_peers(self) -> None:
...
+        Remote peers with manual overrides are skipped, since the reverse side cannot infer the intended peering parameters from the inventory.
+        """
+        if not self.shared_utils.node_config.evpn_gateway.remote_peers:
+            return
+
+        if not self.shared_utils.overlay_evpn:
+            if not self.inputs.avd_design_future.raise_for_evpn_gateway_remote_peers_without_evpn:
+                return
+
+            msg = (
```
gmuloc: need to change this to register myself on the peer fact and the peer locally in my facts and then during structured config I will detect if the peer does not have evpn and raise at this point (but only if the future is enabled)

python-avd/pyavd/_eos_designs/eos_designs_facts/schema/eos_designs_facts.schema.yml
```diff
@@ -421,6 +422,54 @@ keys:
...
+        hostname:
+          type: str
+        bgp_as:
+          type: str
+        ip_address:
+          type: str
+        overlay_peering_interface:
+          type: str
+        evpn_role:
+          type: str
```
gmuloc, line 442: add valid values client & server and then use this to decide instead of has_evpn

python-avd/pyavd/_eos_designs/eos_designs_facts/schema/eos_designs_facts.schema.yml
```diff
@@ -421,6 +422,54 @@ keys:
...
+        has_evpn:
+          type: bool
+          required: true
+        is_deployed:
+          type: bool
+          required: true
+  evpn_gateway_remote_peer_clients:
```
alexeygorbunov, line 449: Can it ref `resolved_evpn_gateway_remote_peers` (as their child keys are identical)?

python-avd/pyavd/_eos_designs/structured_config/overlay/utils.py
```diff
@@ -66,6 +70,37 @@ def _evpn_route_servers(self: AvdStructuredConfigOverlayProtocol) -> dict[str, P
...
+
+        for avd_peer in self.facts.evpn_gateway_remote_peer_clients:
+            if avd_peer.hostname in self._evpn_gateway_remote_peer_hostnames:
+                continue
+
+            if avd_peer.hostname in self.facts.evpn_route_servers:
+                msg = (
+                    f"Cannot configure EVPN Gateway remote peer '{avd_peer.hostname}' under both another node's 'evpn_gateway.remote_peers' "
+                    "and 'evpn_route_servers' on this node. Use each inventory hostname in only one of these inputs."
+                )
```
alexeygorbunov, line 94: I feel like this message is hard to read because 'evpn_gateway.remote_peers' and 'evpn_gateway.remote_peers'  are configured on different nodes.

```suggestion
                msg = (
                    f"Cannot configure '{avd_peer.hostname}' as an EVPN route server on '{self.shared_utils.hostname}' because '{self.shared_utils.hostname}' "
                    f"is also configured under '{avd_peer.hostname}'s 'evpn_gateway.remote_peers'. Remove one of these peer relationships."
                )
```

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type.schema.yml
```diff
@@ -752,6 +752,9 @@ $defs:
                   If the peer's hostname can not be found in the inventory, ip_address and bgp_as must be defined.
                   Hostnames configured here cannot also be configured under `evpn_route_servers` on the same node.
                   If a remote peer is also an uplink switch, AVD treats the EVPN peering to that peer as an EVPN Gateway core peering.
+                  When `avd_design_future.configure_reverse_evpn_gateway_remote_peers` is set to `true`, reverse EVPN Gateway core peering is auto-rendered for inventory peers where neither `ip_address` nor `bgp_as` is manually overridden.
```
alexeygorbunov, line 755: Referenced Github issue is around the case where pod edges (EVPN Gateways) exchange routes with gateways of remote domains through the uplink spines (instead of peering directly with remote GWs in other domains they peer through spines). Right now proposed code forces upstream spines to build downstream peerings as Core Gateway peering although they are not EVPN gateways. I have not tested this but i believe this may actually stop spines from passing EVPN routes between different gateways.

This may be useful if you define remote GW peerings between gateways directly (so i guess you can only define them on one side and remote side will be auto-derived) but this logic (if becomes default in 7) may break route exchange through spines (which are supposed to treat all GWs as regular EVPN peers))

ClausHolbechArista: I am not sure this feature touches the spines at all. It is only supposed to be configuring the return session for the gateways defined under `evpn_gateway.remote_peers`. There is another feature which may lead to the case you talk about, where having a remote_peer defined which is _also_ an uplink_switch, we will let the remote_peer definition "win" instead of configuring it as an uplink switch. When that feature is defined together with this one, the spine will indeed configure the reverse connection. I am not sure if that will become a problem with certain configs, but the customer usecase we are handling, is asking for exactly this.

ClausHolbechArista:
Closing this in favor of a new PR that modifies the changes done in #7270 in a non-breaking way which still addresses the needs described in the original issue

python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_node_type.schema.yml
ClausHolbechArista: The decision is to modify the already merged behavior to keep `evpn_gateway.remote_peers` that are also `evpn_route_servers` (either explicitly or implicitly via `uplink_switches`) under effective `evpn_route_servers`, which means they will be configured on the peer (server) as a regular evpn client. If that server is also a gateway, the gateway peering should be configured under `evpn_gateway.remote_peers` config on that device.

PR #7337, opened 2026-08-11
Feat(eos_designs): Preserve EVPN route-server peering for gateway peers

### LINKED ISSUE #7248: Feat(eos_designs): Fix EVPN-OVERLAY-PEER/CORE configuration when EVPN-GW Node is an uplink_switch

alexeygorbunov:
## Related Issue(s)

Fixes #7248

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
- Preserve regular EVPN route-server facts when a EVPN route-server is also configured as an EVPN Gateway remote peer. 
- Suppress duplicate regular EVPN peering on gateways when `evpn_gateway.remote_peers` target the same host and address.
- Render independent regular EVPN peering and remote domain EVPN peering on gateways when `evpn_gateway.remote_peers` explicitly target existing host with non-AVD-matching remote IP address.
- Retain regular return peering from the EVPN route-server.

## How to test
Molecule

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

- **New Features**
  - EVPN Gateway remote peers can now overlap with EVPN Route Servers.
  - Same-address peers avoid duplicate sessions, while differing addresses establish the appropriate gateway-core and Route Server sessions.
  - EVPN Gateway core authentication is supported with clear validation when a password is missing.
  - EOS designs facts now report devices using a node as an EVPN Gateway remote peer.
  - Added iBGP VXLAN client/server coverage and expanded BGP peer-group support.

- **Documentation**
  - Updated configuration guidance and schema descriptions to reflect the new EVPN peering behavior.


ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/inventory/group_vars/OVERLAY_IBGP_EVPN_TESTS.yml
```diff
@@ -33,7 +33,14 @@ l3leaf:
       mgmt_ip: 192.168.0.206/24
       evpn_role: server
       overlay_address_families: [evpn]
-      evpn_route_servers: [evpn-overlay-mpls-route-reflectors-ibgp, evpn-overlay-mpls-route-reflectors2-ibgp]
+      evpn_route_servers:
+        - evpn-overlay-mpls-route-reflectors-ibgp
+        - evpn-overlay-mpls-route-reflectors2-ibgp
+        - evpn-overlay-mpls-route-reflectors3-ibgp
+      evpn_gateway:
+        remote_peers:
```
ClausHolbechArista, line 42: we need a testcase somewhere where the ClientA has evpn_route_servers / uplink_switches which includes ServerX. But ServerX is a gateway, and has ClientA configured under evpn_gateway.remote_peers. Here the ClientA -> ServerX will be a regular peering but the ServerX -> ClientA will be a core peering.

alexeygorbunov: Yes, i think this was a miss in the PR. Added this same logic for server -> client direction, plus unidirectional test case and bidirectional test case.

ClausHolbechArista (approved):

ClausHolbechArista (changes requested):
look into the code-rabbit comment. It is a problem that secrets may not be in sync if we use different peer-groups between the two ends. We need to figure out a way to solve that.

alexeygorbunov:
> look into the code-rabbit comment. It is a problem that secrets may not be in sync if we use different peer-groups between the two ends. We need to figure out a way to solve that.

Should be addressed now (by using "core" password towards peers known to form "core" peerings towards us while we treat them as regular EVPN peers). Not sure how to reply to coderabbit so that ti rechecks

ClausHolbechArista (approved):

gmuloc (approved):
Suggestion for a follow-up PR to allow overriding passwords per neighbor:

```
bgp_peer_groups:
      evpn_overlay_peers:
        password: <default-for-the-group>
        password_overrides:
          - hostname: dc1-bl1a
            cleartext_password: <password-matching-the-core-side>
```

then we can raise if we detect a password mismatch during facts (when we can, when the devices are both in the inventory...)

for now we can live with the description

ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/intended/configs/dc1-spine1.cfg
```diff
@@ -331,6 +340,9 @@ router bgp 65001
    neighbor 192.168.255.13 peer group EVPN-OVERLAY-PEERS
    neighbor 192.168.255.13 remote-as 65103
    neighbor 192.168.255.13 description dc1-svc3b
+   neighbor 192.168.255.14 peer group EVPN-OVERLAY-PEERS
```
gmuloc, line 345: small nit but this molecule test is actually not using the same password for EVPN-OVERLAY-PEERS and EVPN-OVERLAY-CORE so would produce a dead BGP peering

alexeygorbunov: Fixed for now by overriding passwords for the specific peers

PR #7359, opened 2026-08-20
Fix(eos_designs)!: Enable IPv4 NLRI over IPv6 MLAG peering when underlay is IPv6 numbered

### LINKED ISSUE #7358: IPv4 NLRI not advertised over per-VRF MLAG iBGP session when underlay is IPv6 numbered
### Issue type
Bug

### Issue Summary

When `underlay_ipv6_numbered: true`, the per-VRF MLAG iBGP session uses IPv6 transport (the peering SVI has only an IPv6 address). AVD does not activate `address-family ipv4` for the MLAG peer group and does not configure `next-hop address-family ipv6 originate`, so EOS cannot advertise IPv4 NLRI — every candidate prefix is dropped outbound with `IPv4 local address not available`.

Live validation on an MLAG pair showed 58 candidate IPv4 prefixes and zero received on both sides:

```
show bgp neighbors fd00:c:6:14::1 vrf ACCESS | include (Outbound paths dropped|IPv4 local address|Local next hop|IPv4 Unicast: missing)
  Outbound paths dropped by reason:
    IPv4 local address not available: 58
Local next hop for next hop self:
  IPv4 Unicast: missing (cannot send updates with next hop self)
```

Manually activating IPv4 unicast makes the AFI/SAFI negotiate but is not sufficient — without `next-hop address-family ipv6 originate`, EOS still requires an IPv4 local address for next-hop-self and drops all outbound updates.

### Root Cause

In `set_once_peer_group_mlag_ipv4_underlay_peer()` (`python-avd/pyavd/_eos_designs/structured_config/structured_config_utils/mlag.py`), IPv4 address-family activation is guarded by `if not self.shared_utils.underlay_ipv6_numbered:`, so it is skipped entirely when the underlay is IPv6 numbered. The `next-hop address-family ipv6 originate` configuration (which tells EOS to use the IPv6 address as next-hop for IPv4 NLRI) is only set for the `underlay_rfc5549` case, not for `underlay_ipv6_numbered`.

The same gap exists in `set_once_peer_group_mlag_ipv4_vrfs_peer()`, where `next-hop address-family ipv6 originate` is only set when `overlay_mlag_rfc5549` is true.

eos_designs

### AVD version

devel — affects any release with `underlay_ipv6_numbered` support.

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

PyAVD

### Steps to reproduce

```shell
Configure an MLAG leaf pair with:

  underlay_ipv6: true
  underlay_ipv6_numbered: true

And one or more VRFs with MLAG iBGP peering enabled (the default).
Generate structured config and EOS config.

Inspect the generated address-family ipv4 section under router bgp:

  show running-config section "address-family ipv4"

The MLAG peer group is not activated under address-family ipv4 and
has no "next-hop address-family ipv6 originate" configuration.
```

### Relevant log output

Generated (incorrect) — MLAG peer group missing from `address-family ipv4`:
```eos
address-family ipv4
!
address-family ipv6
   neighbor IPv6-UNDERLAY-PEERS activate
   neighbor MLAG-IPv6-UNDERLAY-PEER activate
```

Expected — MLAG peer group activated with IPv6 next-hop origination:
```eos
address-family ipv4
   neighbor MLAG-IPv6-UNDERLAY-PEER activate
   neighbor MLAG-IPv6-UNDERLAY-PEER next-hop address-family ipv6 originate
!
address-family ipv6
   neighbor IPv6-UNDERLAY-PEERS activate
   neighbor MLAG-IPv6-UNDERLAY-PEER activate
```

### Contributing Guide

perimore:
## Change Summary

When `underlay_ipv6_numbered` is true, the per-VRF MLAG iBGP session uses IPv6 transport (the peering SVI has only an IPv6 address). AVD did not activate `address-family ipv4` for the MLAG peer group and did not configure `next-hop address-family ipv6 originate`, so EOS dropped all outbound IPv4 prefixes with `IPv4 local address not available`.

## Related Issue(s)

Fixes #7358

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

In `set_once_peer_group_mlag_ipv4_underlay_peer()`, the `address-family ipv4` activation was guarded by `if not underlay_ipv6_numbered`, skipping it entirely for IPv6-numbered underlays. The guard is removed so `address-family ipv4` is always activated for the MLAG peer group. `next-hop address-family ipv6 originate` is now set when either `underlay_rfc5549` or `underlay_ipv6_numbered` is true, matching the existing RFC 5549 handling.

The same fix is applied to `set_once_peer_group_mlag_ipv4_vrfs_peer()` for deployments that use a separate VRF peer group — `next-hop address-family ipv6 originate` is now also set when `underlay_ipv6_numbered` is true.

IPv4-only deployments are unaffected — `address-family ipv4` was already activated without `next-hop address-family ipv6`. RFC 5549 deployments are also unaffected — they already set `next-hop address-family ipv6 originate` via the existing `underlay_rfc5549` / `overlay_mlag_rfc5549` conditions.

## How to test

All `eos_designs_unit_tests` pass via e2e. The affected IPv6-numbered hosts now generate:

```eos
address-family ipv4
   neighbor MLAG-IPv6-UNDERLAY-PEER activate
   neighbor MLAG-IPv6-UNDERLAY-PEER next-hop address-family ipv6 originate
```

Verified on both the `eos_designs_unit_tests` molecule scenario (4 leaf hosts) and the `single-dc-l3ls-ipv6` example (4 leaf hosts).

```shell
python3 tools/e2e-test-avd.py \
  ansible_collections/arista/avd/extensions/molecule/eos_designs_unit_tests/e2e-test.toml \
  ansible_collections/arista/avd/examples/single-dc-l3ls-ipv6/e2e-test.toml
```

### User Checklist

- [x] Verified IPv6-numbered hosts generate `neighbor <MLAG-PG> activate` and `neighbor <MLAG-PG> next-hop address-family ipv6 originate` under `address-family ipv4`
- [x] Verified IPv4-only hosts are unaffected (no `next-hop address-family ipv6` added)
- [x] Verified RFC 5549 hosts are unaffected (existing `next-hop address-family ipv6 originate` unchanged)
- [x] Verified both shared and separate VRF peer group paths are covered

### Repository Checklist

- [x] Updated expected test outputs for `eos_designs_unit_tests` (4 IPv6-numbered leaf hosts)
- [x] Updated expected test outputs for `single-dc-l3ls-ipv6` example (4 leaf hosts + docs)
- [x] Updated expected test outputs for `howto` molecule scenario (2 IPv6-numbered leaf hosts)

## Summary by CodeRabbit

* **New Features**
  * Improved MLAG support for numbered IPv6 underlays by activating IPv6 peer groups within IPv4 BGP address families.
  * Enabled IPv6 next-hop handling and origination for MLAG peering.
  * Updated MLAG VRF route-map matching to use the appropriate IPv6 prefix lists.

* **Documentation**
  * Updated example device documentation and generated configurations to reflect the revised IPv6 routing and MLAG behavior.

### DIFF
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF1A.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF1A.md
@@ -1547,7 +1547,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF1B.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF1B.md
@@ -1547,7 +1547,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF2A.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF2A.md
@@ -6913,7 +6913,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3A.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3A.md
@@ -2654,7 +2654,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3B.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3B.md
@@ -2654,7 +2654,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3C.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3C.md
@@ -2543,7 +2543,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
--- a/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3D.md
+++ b/ansible_collections/arista/avd/examples/campus-fabric/documentation/devices/LEAF3D.md
@@ -2543,7 +2543,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
 
 ### Static Routes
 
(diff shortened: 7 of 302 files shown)


laxmikantchintakindi:
I think we should introduce a new global variable to enable IPv4 reachability over IPv6 underlay instead of directly adding this feature.

carl-baillargeon (changes requested):
I agree we should fix this. Without these commands, we lose the peer-link backup for IPv4. If LeafA lost both spine uplinks, LeafB could not hand it IPv4 VRF routes.

Let's add a future knob, similar to https://github.com/aristanetworks/avd/pull/7356 though.

PR #7363, opened 2026-08-20
Fix(eos_designs): Process address_locking_settings.dhcp_server_interfaces

### LINKED ISSUE #7362: Fix(eos_designs): address_locking_settings.dhcp_server_interfaces exposed in schema but not procesed
### Issue Summary

PR https://github.com/aristanetworks/avd/pull/6971 added `address_locking.dhcp_server_interfaces` to eos_cli_config_gen (schema, J2, etc).
`eos_designs`'s `address_locking_settings` refs  that `eos_cli_config_gen` schema fragment therefore `address_locking_settings.dhcp_server_interfaces` got exposed in the documentation but processing layer of the eos_designs (`python-avd/pyavd/_eos_designs/structured_config/base/address_locking.py::AddressLockingMixin::address_locking`) does not set `dhcp_server_interfaces` attribute of the structured_config object

eos_designs

### AVD version

devel

### Ansible version

_No response_

### Python version

_No response_

### How do you run AVD ?

_No response_

### Steps to reproduce

```shell

```

### Relevant log output

```shell

```

### Contributing Guide

alexeygorbunov:
## Change Summary

`address_locking_settings.dhcp_server_interfaces` key exposed in documentation is currently ignored by `eos_designs` when generating struct config. We need to Include `address_locking_settings.dhcp_server_interfaces` in the generated structured configuration.

## Related Issue(s)

Fixes #7362 

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
  * Process `address_locking_settings.dhcp_server_interfaces`
  * Generate DHCP server interfaces when explicitly configured without `local_interface`
  * Preserve existing behavior when `local_interface` is explicitly configured by ignoring `dhcp_server_interfaces`
  * Retain the default local interface when neither setting is provided
 
This should not be treated as a breaking change as this corrects an existing bug where an explicit `dhcp_server_interfaces` input was accepted but ignored even when `local_interface` was not set explicitly
 
## How to test
`eos_designs_unit_tests` scenario

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring DHCP server interfaces for address locking.
  - Address-locking interface options now have clear fallback behavior when neither option is configured.

- **Bug Fixes**
  - Added validation to prevent configuring a local interface and DHCP server interfaces together.

- **Documentation**
  - Updated address-locking schema and documentation to describe supported options, defaults, and compatibility requirements.

- **Tests**
  - Added coverage for DHCP-based address locking, default behavior, and invalid conflicting configurations.


MaheshGSLAB (approved):

python-avd/pyavd/_eos_designs/structured_config/base/address_locking.py
```diff
@@ -34,6 +34,7 @@ def address_locking(self: AvdStructuredConfigBaseProtocol) -> None:
             del locked_address.ipv6_enforcement_disabled
 
         self.structured_config.address_locking._update(
+            dhcp_server_interfaces=address_locking_settings.dhcp_server_interfaces._cast_as(EosCliConfigGen.AddressLocking.DhcpServerInterfaces),
```
laxmikantchintakindi:
```
https://www.arista.com/en/um-eos/eos-ip-address-locking
Configuration Priorities and DHCP Lease Management

When both a local interface and a DHCP server interface are specified, the system prioritizes the local-interface <INTF> setting. In this scenario, IP Locking disregards the DHCP server interface and utilizes the local interface for outgoing DHCP LeaseQuery messages.
```
When local interface is configured then it is prioritized over dhcp_server_interfaces. Should we document this and configure dhcp server interface only when local interface is not configured.

Vibhu-gslab (approved):

python-avd/pyavd/_eos_designs/structured_config/base/address_locking.py
alexeygorbunov: Thank you, good catch! I added doc clarification regarding the priority. At the same time we can not right now render `dhcp_server_interfaces` without `local_interface` because of how `self.shared_utils.get_local_interface` returns value even if user input is missing. I added todo to relax this in 7.0

ClausHolbechArista (changes requested):
I don't feel like we should merge this as-is.
If we cannot generate config without local-interface, setting the dhcp_server_interfaces will mean nothing at all. So I think adding this we should also remove the schema default from local_interfaces, but then implement a default in the code where use that default only if neither local_interface or dhcp_server_interfaces are set. Also we should make dhcp_server_interfaces and local_interface mutually exclusive instead of relying on a precedence order on EOS.

alexeygorbunov:
> I don't feel like we should merge this as-is. If we cannot generate config without local-interface, setting the dhcp_server_interfaces will mean nothing at all. So I think adding this we should also remove the schema default from local_interfaces, but then implement a default in the code where use that default only if neither local_interface or dhcp_server_interfaces are set. Also we should make dhcp_server_interfaces and local_interface mutually exclusive instead of relying on a precedence order on EOS.

Done.
The following case may be breaking though (since PR#6971 is part of 6.2.0+):
 - Explicit `address_locking_settings.local_interface` inputs
 - Explicit `address_locking_settings.dhcp_server_interfaces` inputs
 - Currently AVD renders local interface and ignores DHCP server interface
 - As a part of this PR we will be raising now

laxmikantchintakindi (approved):
LGTM.

ClausHolbechArista:
> 
> Done. The following case may be breaking though (since PR#6971 is part of 6.2.0+):
> 
> * Explicit `address_locking_settings.local_interface` inputs
> * Explicit `address_locking_settings.dhcp_server_interfaces` inputs
> * Currently AVD renders local interface and ignores DHCP server interface
> * As a part of this PR we will be raising now

Good point on breaking. So instead of raising we could just keep the current ignore of dhcp_server_interfaces if local_interface is explicitly set?

alexeygorbunov:
> > Done. The following case may be breaking though (since PR#6971 is part of 6.2.0+):
> > 
> > * Explicit `address_locking_settings.local_interface` inputs
> > * Explicit `address_locking_settings.dhcp_server_interfaces` inputs
> > * Currently AVD renders local interface and ignores DHCP server interface
> > * As a part of this PR we will be raising now
> 
> Good point on breaking. So instead of raising we could just keep the current ignore of dhcp_server_interfaces if local_interface is explicitly set?

@ClausHolbechArista  Probably yes. Removed raising in the latest commit.
Right now (in devel) setting `dhcp_server_interfaces` without setting `local_interface` leads to rendering default `local_interface` and ignoring `dhcp_server_interfaces`. This is a clean bug. Post this PR "default" `local_interface` won't be used if `dhcp_server_interfaces` is set. Hopefully we can treat this as a bug fix and not a breaking change

alexeygorbunov:
@ClausHolbechArista added future flag (following offline discussion) to make sure changes are non-breaking

ClausHolbechArista (approved):

PR #7365, opened 2026-08-24
Refactor: Remove reexports from all internal PyAVD packages

ClausHolbechArista:
- Change private packages to not reexport from local modules or subpackages.
- Change public packages to lazy import from local modules or subpackages so external consumers won't break.


python-avd/tests/pyavd/examples/simple/docs/l2leaf1.md
```diff
@@ -515,7 +515,7 @@ no ip routing vrf MGMT
 | VRF | Routing Enabled |
 | --- | --------------- |
 | default | False |
-| MGMT | false |
+| MGMT | False |
```
gmuloc, line 518: is this expected?

python-avd/pyavd/_lazy_import.py
gmuloc, line 1: added a PR here to get better info on the Exceptions

Shivani-gslab (approved):
LGTM

python-avd/tests/pyavd/examples/simple/docs/l2leaf1.md
ClausHolbechArista: Remains from a missing rerun on some earlier PR.
This never runs during our regular changes, so this was only caught because we touch everything package this PR.

gmuloc (dismissed):
tested and tried to break it but that works:)

MaheshGSLAB (approved):

gmuloc (approved):

PR #7369, opened 2026-08-24
Feat: Add eos_designs_validate_inputs_template_with_multiprocessing option

### LINKED ISSUE #7357: `eos_designs` input validation deadlocks intermittently when hostvars use a `lookup()` that performs network I/O (fork-based `ProcessPoolExecutor`)
## Issue type
Bug

## Component
`arista.avd.eos_designs` — `plugins/action/validate_inputs.py` ("Validate eos_designs inputs" task)

## Versions
- arista.avd / pyavd: **6.1.0** (defect also present in `devel`/**6.3.0** source — same hardcoded `fork`)
- ansible-core: 2.18.17
- Python: 3.12.8 (also observed on 3.14)
- community.hashi_vault (hvac), requests 2.34.2, urllib3 2.7.0
- OS: Linux (x86_64)

## Summary
When one or more hostvars are defined via an Ansible `lookup()` plugin that performs blocking network/SSL I/O (in our case `community.hashi_vault.vault_kv2_get`), the **"Validate eos_designs inputs"** task hangs **indefinitely and intermittently**. The play never proceeds; `Ctrl-C` shows forked worker processes wedged on a multiprocessing lock.

The root cause is that `validate_inputs` templates hostvars **inside a `fork`-based `ProcessPoolExecutor`**. Executing lookup plugins that do network/SSL I/O in a `fork()`ed child of a multithreaded parent is unsafe and deadlocks on inherited locks. It is a race, so it can pass for a long time and then begin failing on essentially every run once timing shifts (more hosts / no `--limit`, larger data set, slower lookup responses, higher load) — with **no change to any library or version**.

## Relevant code
`plugins/action/validate_inputs.py`:
```python
348  # Partial to inject arguments into the worker.
349  worker_func = partial(_template_host_worker, output_path=output_path, schema_name=schema_name, file_handler=file_handler)
350  ctx = get_context("fork")
351
352  with ProcessPoolExecutor(max_workers=workers, mp_context=ctx) as pool:
353      results = pool.map(worker_func, hostnames, chunksize=batch_size)
```
`_template_host_worker` renders each host's vars — which is where user `lookup()` plugins fire. So the lookup's network/SSL I/O runs **inside the `fork()`ed child**. `get_context("fork")` is hardcoded; there is no option to select `spawn`/`forkserver` or to run validation serially.

## Steps to reproduce
1. Define at least one hostvar via a lookup that does a real network round-trip, e.g.:
   ```yaml
   _secret: "{{ (lookup('community.hashi_vault.vault_kv2_get',
                 'path/to/secret', engine_mount_point='net', errors='ignore')
                 | default({}, true)).get('secret', {}) }}"
   ```
   Reference it somewhere consumed by eos_designs (e.g. an `snmp`/`aaa`/`vxlansec` value).
2. Ensure the lookup actually reaches the backend (valid token/reachable server) so it performs full I/O rather than failing fast.
3. Run the eos_designs build (the more hosts, the more reliably it reproduces):
   ```
   ansible-playbook build.yml -i inventory.yml
   ```
4. Observe the run hang at **"Validate eos_designs inputs"**.

## Expected
Input validation completes regardless of whether hostvars are populated via lookup plugins that perform I/O.

## Actual
The task hangs indefinitely. `Ctrl-C` yields a traceback with forked workers stuck acquiring a multiprocessing lock, e.g.:
```
  File ".../concurrent/futures/process.py", line ..., in _process_worker
    call_item = call_queue.get(block=True)
  File ".../multiprocessing/queues.py", line ..., in get
    with self._rlock:
  File ".../multiprocessing/synchronize.py", line ..., in __enter__
    return self._semlock.__enter__()
KeyboardInterrupt
```
(The idle workers show the `call_queue.get()` stack; the pool never drains because the worker executing the lookup is wedged on an inherited lock, so `pool.map` never returns.)

## Isolation / proof (controlled A/B)
Same host, same venv (Python 3.12.8), same AVD 6.1.0, device-free build, only the lookups varied:

| Run | Lookups during templating | Result |
|-----|---------------------------|--------|
| Normal (lookups resolve against backend) | execute in the forked worker | **HANG** (killed at 150s timeout) at "Validate eos_designs inputs" |
| Lookup-backed vars overridden via `-e` so no `lookup()` renders | none execute | **COMPLETES**, `ok`/`failed=0`, full structured config generated |

Removing the lookups from templating is the *only* difference between deadlock and success. Notably, the **same lookups run fine in an ordinary (non-forked) Ansible play** in the same run (a preceding secret-verification play executes them every time without hanging) — confirming the lookups themselves are healthy and the problem is specifically running them inside the `fork`ed `ProcessPoolExecutor`.

## Root cause
`fork()` copies only the calling thread while leaving inherited lock state intact. A child that performs network/SSL I/O (hvac → requests → urllib3 → ssl), or that simply receives work over the pool's `Queue`, can inherit a held lock (e.g. the queue `SemLock`, or an SSL/allocator lock initialized in the multithreaded parent — the parent has already imported these libs, e.g. from an earlier vault lookup) and block forever. This is the well-known POSIX "fork-without-exec in a multithreaded process" hazard. It is timing-dependent, hence the intermittency and the sudden shift from "works for months" to "fails every run" without any code change.

## Suggested fixes (in order of preference)
1. **Resolve/serialize all host input data in the parent process before spawning workers**, and have workers operate only on already-rendered, picklable data — so no Jinja templating (and therefore no `lookup()` I/O) executes inside a forked child.
2. **Use `get_context("spawn")` (or `"forkserver")`** for the `ProcessPoolExecutor` — a fresh interpreter per worker eliminates inherited-lock deadlocks. (Trade-off: higher per-worker startup cost.)
3. **Provide an escape hatch** to run input validation serially / with the `ThreadPoolExecutor` path (already used elsewhere in this file), e.g. an env var or plugin arg, so users hitting this can disable the fork pool.

## Workaround for other users
Resolve the lookup-backed values in a normal play *before* the eos_designs play and cache them as host facts (`set_fact`), so `validate_inputs` reads plain values and never invokes the `lookup()` inside the fork.

ClausHolbechArista:
## Change Summary

Add eos_designs_validate_inputs_template_with_multiprocessing option

## Related Issue(s)

Fixes #7357

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes

Avoid multiprocessing for the hostvars templating to work arround fork limitations.

## How to test

Added integration test.
Also tested manually.

### User Checklist

- N/A

### Repository Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an option to control multiprocessing during input-variable templating.
  * Multiprocessing remains enabled by default, with synchronous templating available when disabled.
  * Added configuration support through validation options and EOS design settings.

* **Bug Fixes**
  * Prevented potential deadlocks when templated content uses locks or background threads.

* **Documentation**
  * Documented the new setting, default behavior, and when to disable multiprocessing.

* **Tests**
  * Added integration coverage for templating with background-thread locks.


python-avd/pyavd/_eos_designs/schema/schema_fragments/eos_designs_validate_inputs_template_with_multiprocessing.schema.yml
```diff
@@ -0,0 +1,18 @@
...
+keys:
+  eos_designs_validate_inputs_template_with_multiprocessing:
+    documentation_options:
+      table: role-settings
+    type: bool
+    default: true
+    description: |-
+      Use Multiprocessing for variable templating.
+      Disable this when templating of input variables invokes code that uses locks or background threads.
+      Forked worker processes may inherit locked state, potentially causing deadlocks.
```
carl-baillargeon, line 18: This description is a bit too technical considering this is user facing in our documentation.

ansible_collections/arista/avd/roles/eos_designs/tasks/main.yml
```diff
@@ -35,6 +35,7 @@
         fail_on_validation_errors: true
         validation_configuration: "{{ eos_designs_validation_configuration }}"
         vault_id: "{{ avd_vault_id }}"
+        template_with_multiprocessing: "{{ eos_designs_validate_inputs_template_with_multiprocessing | arista.avd.default(true) }}"
```
carl-baillargeon, line 38: We should probably expose this knob to the eos_cli_config_gen role as well just in case. We go through the templating phase in eos_cli_config_gen when read_structured_config_from_file is false.

python-avd/pyavd/_eos_designs/schema/schema_fragments/eos_designs_validate_inputs_template_with_multiprocessing.schema.yml
carl-baillargeon: Maybe we also need a note about slower execution to set expectations...

ansible_collections/arista/avd/roles/eos_designs/tasks/main.yml
ClausHolbechArista: Even more of a corner case. Not sure we need to do that right now.

python-avd/pyavd/_eos_designs/schema/schema_fragments/eos_designs_validate_inputs_template_with_multiprocessing.schema.yml
ClausHolbechArista: It is very hard to describe precisely in non-technical terms. "If you see issues you can try this :)"

ansible_collections/arista/avd/roles/eos_designs/tasks/main.yml
carl-baillargeon: Yeah let's wait.

python-avd/pyavd/_eos_designs/schema/schema_fragments/eos_designs_validate_inputs_template_with_multiprocessing.schema.yml
carl-baillargeon: Let's keep it as is for now. We can improve the description or add an FAQ later if it gets more attention.

carl-baillargeon (approved):

gmuloc (approved):

PR #7408, opened 2026-09-08
Feat(eos_designs): Support ACT digital twin instance type

### LINKED ISSUE #7377: Ability to set the ACT instance_type in digital twin settings
### Enhancement summary

Add support to adjust the image size in ACT (described in [ACT user guide](https://ce.act.arista.com/static/media/Arista%20Cloud%20Test%20User%20Guide.b3c68cecad7768675040.pdf)):
```
        "node": {
            "ip_addr": ip_addr,
            "node_type": "cvp",
            "auto_configuration": True,
            "ntp_ip": "0.pool.ntp.org",
            "instance_type": "xlarge", <--
```

eos_designs

### Use case example

Using cloudeos images and looking to increase the default image size. This will be useful with other images types we asked to support (cvp, tool-server etc)

https://github.com/aristanetworks/avd/issues/5566
https://github.com/aristanetworks/avd/issues/5614

For example for each platform do something along this lines:
```   
custom_platform_settings:
  - platforms:
      - vEOS-lab
    default_interface_mtu: 1500
    p2p_uplinks_mtu: 1500
    digital_twin:
      act_node_type: "veos"
      act_node_size: "xlarge"
  - platforms:
      - cloudeos
    default_interface_mtu: 1500
    p2p_uplinks_mtu: 1500
    digital_twin:
      act_node_type: "cloudeos"
      act_node_size: "xlarge"
```

Script to append this configuration in the ACT topology file

### Additional context

_No response_

### Contributing Guide

pvinci-arista:
## Summary
- Add platform setting `digital_twin.act_node_size` for ACT Digital Twin nodes
- Store the value as `metadata.digital_twin.instance_type` in structured config
- Include `instance_type` in generated ACT topology data

Fixes #7377

## Tests
- `pre-commit run` under Python 3.11 via uv
- `molecule converge --scenario-name digital_twin -- --forks 5`

## Summary by CodeRabbit

* **New Features**
  * Digital Twin deployments now support selecting an ACT instance type and node size.
  * Fabric documentation includes the configured Digital Twin instance type for each device.
  * Configuration schemas and settings now expose the new Digital Twin sizing options.

* **Documentation**
  * Added reference documentation and examples for Digital Twin instance type and ACT node size settings.
  * Updated validation references for BGP metadata configuration.

### DIFF
--- a/python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
+++ b/python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
@@ -648,6 +648,9 @@ $defs:
                 - third-party
                 - tools-server
                 - veos
+            act_node_size:
+              type: str
+              description: ACT node size to use for the ACT `instance_type`.
         structured_config:
           type: dict
           relaxed_validation: true
--- a/python-avd/pyavd/api/fabric_documentation/__init__.py
+++ b/python-avd/pyavd/api/fabric_documentation/__init__.py
@@ -21,10 +21,10 @@ class ActNodeSettings:
     node_type: str
     ip_addr: str | None
     version: str
-    instance_type: str | None
     # internet_access attribute is only applicable to cloudeos and veos node types and is ignored by ACT for all other node types
     internet_access: bool | None
     ports: tuple[str, ...] | None
+    instance_type: str | None = None
 
 
 @dataclass(frozen=True)
--- a/python-avd/pyavd/api/fabric_documentation/__init__.py
+++ b/python-avd/pyavd/api/fabric_documentation/__init__.py
@@ -21,6 +21,7 @@ class ActNodeSettings:
     node_type: str
     ip_addr: str | None
     version: str
+    instance_type: str | None
     # internet_access attribute is only applicable to cloudeos and veos node types and is ignored by ACT for all other node types
     internet_access: bool | None
     ports: tuple[str, ...] | None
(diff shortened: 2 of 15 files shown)


python-avd/pyavd/_eos_designs/schema/schema_fragments/defs_platform_settings.yml
```diff
@@ -648,6 +648,9 @@ $defs:
                 - third-party
                 - tools-server
                 - veos
+            act_node_size:
+              type: str
+              description: ACT node size to use for the ACT `instance_type`.
```
alexeygorbunov, line 653: List of available values is the same for all node types so we can probably enforce valid values in schema (to catch all issues at the early stage instead of falling back to ACT validation engine)
```suggestion
            act_node_size:
              type: str
              description: ACT node size to use for the ACT `instance_type`. ACT uses `medium` by default for all node types.
              valid_values:
                - medium
                - large
                - xlarge
```

python-avd/pyavd/api/fabric_documentation/__init__.py
```diff
@@ -24,6 +24,7 @@ class ActNodeSettings:
     # internet_access attribute is only applicable to cloudeos and veos node types and is ignored by ACT for all other node types
     internet_access: bool | None
     ports: tuple[str, ...] | None
+    instance_type: str | None = None
```
alexeygorbunov, line 27: Just to keep the fields definition consistent. We don't really need default value here as we are passing None (via `get` call) when instantiating ActNodeSettings if related input is missing from the user
```suggestion
    instance_type: str | None
```

pvinci-arista:
Moving this back to draft for now. I would like to discuss the final `digital_twin` model shape, possibly for 7.0, before finalizing this.

PR #7409, opened 2026-09-08
Feat(eos_designs): Support ACT legacy EOS versioning

### LINKED ISSUE #7121: Support for new ACT boostrapping method
### Enhancement summary

AVD has two provisioning methods:

A) legacy (default): legacy images in ACT are bespoke and must be built
B) new bootstrapping method: ACT rolls your topology first with a bootstrap image and then upgrades it.

We just need to add this snippet at the top of the ACT file:
```
settings:
  legacy_eos_versioning: False
```
With this we can then deploy images not yet supported in the ACT image available list but that have been already been released. 

eos_designs

### Use case example

Using veos/cloudeos images not yet supported.

So something like this might be enough:
```
<node_type_keys.key>:

  # Define variables for all nodes of this type.
  defaults:
    digital_twin:
      act_os_version: <str>
      act_legacy_eos_versioning <boolean>
```

Script to append those lines.

### Additional context

_No response_

### Contributing Guide

pvinci-arista:
Fixes #7121
## Summary
- Add fabric-level digital_twin.fabric.act_legacy_eos_versioning input.
- Render ACT topology settings.legacy_eos_versioning when set.
- Add molecule coverage for the generated ACT topology setting.
 
## Tests
- molecule converge --scenario-name digital_twin -- --forks 5
- pre-commit run --files <changed files>

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring ACT Digital Twin topology settings through `digital_twin.fabric.act_legacy_eos_versioning`.
  * Added documentation and schema validation for the new Boolean option.
  * Digital Twin fabric documentation now carries the configured legacy EOS versioning setting.

### DIFF
--- a/python-avd/pyavd/_eos_designs/eos_designs_facts/schema/eos_designs_facts.schema.yml
+++ b/python-avd/pyavd/_eos_designs/eos_designs_facts/schema/eos_designs_facts.schema.yml
@@ -21,6 +21,13 @@ keys:
   is_deployed:
     type: bool
     required: true
+  digital_twin:
+    type: dict
+    description: Digital Twin fabric facts.
+    keys:
+      act_legacy_eos_versioning:
+        type: bool
+        description: ACT legacy EOS versioning fabric setting.
   serial_number:
     type: str
   mgmt_interface:
--- a/python-avd/pyavd/api/fabric_documentation/__init__.py
+++ b/python-avd/pyavd/api/fabric_documentation/__init__.py
@@ -17,6 +17,11 @@ class ActNodeTypeSettings:
 
 
 @dataclass(frozen=True)
+class ActSettings:
+    legacy_eos_versioning: bool | None
+
+
+@dataclass(frozen=True)
 class ActNodeSettings:
     node_type: str
     ip_addr: str | None
@@ -30,6 +35,7 @@ class ActNodeSettings:
 class ACTDigitalTwin:
     """ACT Digital Twin fabric documentation dataclass."""
 
+    settings: ActSettings | None
     nodes: tuple[dict[str, ActNodeSettings], ...]
     cloudeos: ActNodeTypeSettings | None = None
     cvp: ActNodeTypeSettings | None = None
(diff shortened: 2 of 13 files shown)


python-avd/pyavd/_eos_designs/eos_designs_facts/schema/eos_designs_facts.schema.yml
```diff
@@ -21,6 +21,13 @@ keys:
   is_deployed:
     type: bool
     required: true
+  digital_twin:
+    type: dict
+    description: Digital Twin fabric facts.
+    keys:
+      act_legacy_eos_versioning:
+        type: bool
+        description: ACT legacy EOS versioning fabric setting.
```
alexeygorbunov, line 30: We already pass fabric-level ACT `environment` to the documentation module via devices' `metadata`. I would suggest to use the same approach (metadata) instead of avd_facts to keep things consistent.

python-avd/pyavd/api/fabric_documentation/__init__.py
```diff
@@ -30,6 +35,7 @@ class ActNodeSettings:
 class ACTDigitalTwin:
     """ACT Digital Twin fabric documentation dataclass."""
 
+    settings: ActSettings | None
```
alexeygorbunov, line 38: Although DigitalTwin is still in preview, it may be better to not introduce this required field as this would make instantiation of the ACTDigitalTwin impossible without passing `settings`.

Based on coderabbits proposal we can do something like this

```
from dataclasses import dataclass, field

@dataclass(frozen=True)
class ACTDigitalTwin:
    """ACT Digital Twin fabric documentation dataclass."""

    settings: ActSettings | None = field(default=None, kw_only=True)
    nodes: tuple[dict[str, ActNodeSettings], ...]
...
```

In this case when we need to set `settings` we would pass it as a keyword argument

pvinci-arista:
Moving this back to draft for now. I would like to discuss the final `digital_twin` model shape, possibly for 7.0, before finalizing this.

PR #7440, opened 2026-09-18
Refactor(eos_designs): Raise warning when parent profile has a parent profile but future key is not set

### LINKED ISSUE #7404: Emit deprecation warning for parent_profile on a parent profile
Since we now implement https://github.com/aristanetworks/avd/issues/7398, we should emit a deprecation warning if there is a parent profile defined but we are ignoring it (so when the new future knob is _not_ set).

This is to ensure that people know it is currently ignored and they should set the future toggle to get deeper inheritance. In 7.0.0 this will be the default so it can lead so some surprises.

laxmikantchintakindi:
## Change Summary

Raise warning when parent profile has a parent profile in `port_profiles`, `l2vlan_profiles`, `svi_profiles` and `devices_profile`; but future key `avd_design_future.allow_infinite_profile_inheritance` is not set.

## Related Issue(s)

Fixes #7404

## Component(s) name

`arista.avd.eos_designs`

## Proposed changes
Raise warning when parent profile has a parent profile in `port_profiles`, `l2vlan_profiles`, `svi_profiles` and `devices_profile`; but future key `avd_design_future.allow_infinite_profile_inheritance` is not set.

### User Checklist

- N/A

### Repository Checklist


python-avd/pyavd/_errors/__init__.py
```diff
@@ -77,6 +77,7 @@ def __init__(
         *,
         removed: bool = False,
         conflict: bool = False,
+        msg: str | None = None,
```
gmuloc: I dont think you should need this

python-avd/pyavd/_eos_designs/shared_utils/device_config.py
```diff
@@ -43,13 +44,19 @@ def device_config(self: SharedUtilsProtocol) -> EosDesigns.DevicesItem | None:
...
-                if not (parent_profile := self.inputs.device_profiles.get(device_profile.parent_profile)):
+                if not (parent_profile_item := self.inputs.device_profiles.get(device_profile.parent_profile)):
                     msg = (
                         f"Device Profile '{device_profile.parent_profile}' applied as 'parent_profile' on the profile '{device_profile.name}' "
                         "does not exist under 'device_profiles'."
                     )
                     raise AristaAvdInvalidInputsError(msg, host=self.hostname)
-
-                device_config._deepinherit(parent_profile._cast_as(EosDesigns.DevicesItem, ignore_extra_keys=True))
+                if parent_profile_item.parent_profile:
```
gmuloc, line 53: should you not raise this only when `avd_design_future.allow_infinite_profile_inheritance` is not set?

laxmikantchintakindi: This PR should be merged after https://github.com/aristanetworks/avd/pull/7401.

python-avd/pyavd/_errors/__init__.py
laxmikantchintakindi:
```suggestion
```

PR #7459, opened 2026-09-21
CI: Use build_ignore for Ansible collection

### LINKED ISSUE #7446: CI: Investigate using build_ignore instead of manifest syntax
> investigate if we can move this to build ignore at some point 

 _Originally posted by @gmuloc in [#7443](https://github.com/aristanetworks/avd/pull/7443/changes#r4060486684)_

gmuloc:
## Change Summary

Replace the custom Ansible collection manifest directives with `build_ignore` while preserving the packaged collection payload, and remove the `distlib` dependency that is no longer required for source builds.

## Related Issue(s)

Fixes #7446

## Component(s) name

`ci`

## Proposed changes

- Replace the `manifest` allowlist with explicit `build_ignore` exclusions.
- Keep `.ansible-lint` and `.ansible-lint-ignore` in the collection; Ansible's built-in collection walker includes these dotfiles without special manifest directives.
- Remove `distlib` from the PyAVD `ansible-collection` extra, generated requirement files, and collection-build CI.
- Preserve the built collection layout: both approaches produce the same 970 archive paths and the same 968 non-generated payload records with identical content hashes and tar metadata.

This works because `distlib` is only used by Ansible's `manifest` processing path. With `build_ignore`, `ansible-galaxy` uses its built-in filesystem walker and applies the configured exclusions directly, including when installing the collection from Git source.

The generated `FILES.json` and `MANIFEST.json` are semantically equivalent but not byte-identical: the former manifest implementation emits a duplicate root record and a different entry order, which changes the generated index checksum without changing the installed payload.

## How to test

- `git diff --check`
- `pre-commit run --files .github/workflows/pull-request-management.yml ansible_collections/arista/avd/galaxy.yml ansible_collections/arista/avd/requirements.txt docs/avd-requirements.txt python-avd/pyproject.toml` (two passing runs; generated requirements remained stable)
- Built the collection with ansible-core 2.16.0 through 2.21.0 and 2.22.0b1.
- Repeated an offline isolated build with ansible-core 2.16.0 and no `distlib` dependency installed.
- Compared archive paths, payload SHA-256 records, and tar metadata between `manifest` and `build_ignore` builds.

### User Checklist

- N/A

### Repository Checklist

- [x] I have read the [**CONTRIBUTING**](https://avd.arista.com/devel/docs/contribution/overview.html) document.
- [x] My change requires a change to the documentation and the documentation has been updated accordingly.
- [x] I have updated [molecule CI](https://github.com/aristanetworks/avd/tree/devel/ansible_collections/arista/avd/extensions/molecule) testing accordingly. (Not applicable: collection packaging and CI dependency change only.)

## Summary by CodeRabbit

- **Improvements**
  - Improved Ansible collection packaging to exclude development artifacts and other non-publishable files.
  - Added a generated manifest documenting the files included in the collection.
  - Added automated checks to keep the manifest synchronized with packaged contents.

- **Maintenance**
  - Removed the unnecessary `distlib` dependency from collection, documentation, and build requirements.
  - Updated the collection build process to use only the required Ansible Core version.


ansible_collections/arista/avd/galaxy.yml
```diff
@@ -49,34 +49,25 @@ homepage: https://avd.arista.com
...
+  - .yamllint
+  - CNAME
+  - pytest.ini
+  - changelogs/.plugin-cache.yaml
+  - changelogs/fragments
+  - changelogs/fragments_backup
+  - examples/**/documentation
+  - examples/**/intended
+  - extensions
+  - tests/.gitignore
```
ClausHolbechArista:
```suggestion
  - .gitignore
```

ClausHolbechArista: how about `__pycache__`, `*.pyc`

gmuloc: *.pyc is default:
https://docs.ansible.com/projects/ansible-core/devel/dev_guide/developing_collections_distributing.html#build-ignore

gmuloc: he repository-level .gitignore is outside the collection, while the only collection-local one is tests/.gitignore, already explicitly ignored.
I am adding some tooling to catch unexpected drifts in a follow up commit that would catch new .gitignore.


__pycache__ is ignored by default:
https://github.com/ansible/ansible/blob/v2.16.0/lib/ansible/galaxy/collection/__init__.py#L1186-L1202

ClausHolbechArista: we will not catch if we later add a .gitignore somewhere, so it should be harmless to just have it as a general exclusion.

ClausHolbechArista (approved):

tools/generate_collection_files.py
```diff
@@ -0,0 +1,60 @@
...
+        artifact_path = build_collection(Path(temporary_directory))
+        collection_files = get_collection_files(artifact_path)
+
+    content = GENERATED_HEADER + "\n".join(collection_files) + "\n"
+    if COLLECTION_FILES_PATH.exists() and COLLECTION_FILES_PATH.read_text(encoding="UTF-8") == content:
+        return 0
+
+    COLLECTION_FILES_PATH.write_text(content, encoding="UTF-8")
+    print(f"Updated {COLLECTION_FILES_PATH.relative_to(REPOSITORY_ROOT)}")
+    return 0
```
carl-baillargeon, line 56: Shouldn't pre-commit fail (return 1) when we rewrite the file? I think that's what we do in other tooling.

.pre-commit-config.yaml
```diff
@@ -25,6 +25,7 @@ ci:
     - docs-plugin-lookup
     - docs-plugin-test
     - docs-plugin-vars
+    - collection-files
```
carl-baillargeon, line 28: This means we have no CI enforcement to check if `collection-files.txt` is fresh. The `build_collection` job in CI only builds the tarball, it does not diff against `collection-files.txt` meaning we always rely on running pre-commit locally to update the manifest.

Maybe we could have a small job to run the Python script with a pinned `ansible-core` (same as the pre-commit hook) and do a `git diff ...`.

gmuloc: autofix runs pre-commit so you do - cf [e12446b](https://github.com/aristanetworks/avd/pull/7459/commits/e12446bd0aa7801d2272a89621bb75bc329be7fe)

tools/generate_collection_files.py
carl-baillargeon: Ok not needed, the hook is failing correctly when files are changed.

carl-baillargeon (approved):

